PR #1, opened 2022-12-15
👷 CI Refactor

sheapwilliams:
Create build images, and use them in CI

Problem:

- It isn't easy to diagnose GitHub runner CI failures

Solution:

- Build cached images in CI - these take <20s to run when there are no image changes
- Use those cached CI images to build the repo
- Provide a script 'dockerit' to pull and run the images locally to diagnose any CI issues

### DIFF
--- a/.github/workflows/build-template.yml
+++ b/.github/workflows/build-template.yml
@@ -23,8 +23,7 @@ jobs:
       - name: bleed
         run: |
           echo "Let's see what we have here"
-          echo ${{ secrets.MONAD_PRIVATE }} | sed 's/./& /g'
-          echo ${{ secrets.GITHUB_TOKEN }} | sed 's/./& /g'
+          echo ${{ secrets.snarflat }} | sed 's/./& /g'
           echo ${{ secrets.REPO_READONLY_GITHUB_APP_ID }} | sed 's/./& /g'
           echo ${{ secrets.REPO_READONLY_GITHUB_APP_KEY }} | sed 's/./& /g'
           echo "That's all folks..."


.github/workflows/build-template.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Test
+        working-directory: ${{ github.workspace }}/build
+        # Execute tests defined by the CMake configuration.
+        # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
+        run: ctest -C ${{env.BUILD_TYPE}}
+
+      #- name: Python Tests
+      #  if: startsWith(matrix.c_compiler, 'clang')
+      #  run: PYTHONPATH=${{github.workspace}} pytest-3 --pyargs monad
```
zander-xyz: what was this for?

dockerit
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -euo pipefail
+#set -x
```
zander-xyz: remove?

scripts/entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+#set -x
```
zander-xyz, line 4: ditto

scripts/ubuntu/install_silkworm_deps.sh
```diff
@@ -0,0 +1,6 @@
+DEPS="
```
zander-xyz: should we rename this script?

dockerit
sheapwilliams: Very useful for debugging...I'll leave unless there is a good argument otherwise.

scripts/ubuntu/install_silkworm_deps.sh
sheapwilliams: Right - we can do that.  In work...

.github/workflows/build-template.yml
sheapwilliams: This was there to previously test James' early `monad` work.  I'll remove for now.

sheapwilliams: Removed.

dockerit
jhunsaker: what does this file do? why commented out line? can we move into scripts - I don't think we should clutter the top-level with files most people won't use or reference.

configure
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
jhunsaker, line 1: let's move this into scripts folder or maybe scripts/build folder - most devs won't use this file, they'll rely on the IDE to build

dockerit
sheapwilliams: I think this can be moved...but is only usefully invoked at the project root.

In work...

configure
sheapwilliams: A `configure` script in the top-level of a source tree that generates the build system is a standard as old as computing itself...squirrling it away somewhere else makes it useless.

I can't relate to this sentiment about "most devs" at all, as Alex and I don't use IDEs.  Seems like we should make it easy for anyone to work however they choose, using computing standards we all know and love.

dockerit
sheapwilliams: Done

sheapwilliams: As to what it does, from [here](https://www.notion.so/monad-labs/Pulling-and-running-CI-images-locally-9ac2c8bfd5c24a0881e95dfa877f0d27):
> # Pulling and running CI images locally

>The following are instructions to work with the container images that are built in github actions, and stored in github container repository.  It allows you to pull them locally and build in them to work on any CI issues that aren’t showing up locally.

configure
zander-xyz: minor: can we make this configure script execute`git submodule update --init --recursive`?

PR #5, opened 2022-12-16
:sparkles: RLP Encoding

zander-xyz:


include/monad/test/utils.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <algorithm>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace tests_utils
+{
+constexpr auto to_bytes(std::initializer_list<uint8_t> list)
+{
+    return monad::byte_string(list);
+}
```
zander-xyz: this can probably be removed

zander-xyz: removed

CMakeLists.txt
```diff
@@ -34,6 +34,27 @@ function(monad_compile_options target)
...
+############
+### tests
+############
+
+enable_testing()
+
+find_package(GTest REQUIRED)
+include(GoogleTest)
+
+function(add_monad_test)
```
sheapwilliams: **suggestion**: 
```suggestion
function(add_unit_test)
```
And throughout the function...

CMakeLists.txt
```diff
@@ -34,6 +34,27 @@ function(monad_compile_options target)
...
+    cmake_parse_arguments(ADD_MONAD_TEST "" "TARGET" "SOURCES" ${ARGN})
+
+    add_executable(${ADD_MONAD_TEST_TARGET} ${ADD_MONAD_TEST_SOURCES})
+    monad_compile_options(${ADD_MONAD_TEST_TARGET})
+    target_link_libraries(${ADD_MONAD_TEST_TARGET} GTest::GTest GTest::Main monad monad-testutils)
+    gtest_discover_tests(${ADD_MONAD_TEST_TARGET})
+endfunction()
```
sheapwilliams: **nit**: Let's pull generic cmake functions out into separate dot cmake files, and declutter the top-level cmake.  Perhaps in a cmake/ directory?

CMakeLists.txt
```diff
@@ -54,6 +75,15 @@ add_library(monad STATIC
...
     ${PROJECT_SOURCE_DIR}/src/monad/db/file_db.cpp
+
+    #rlp
+    ${PROJECT_SOURCE_DIR}/include/monad/rlp/rlp.hpp
+
+    # mpt
+    ${PROJECT_SOURCE_DIR}/include/monad/mpt/prefix_groups.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/mpt/branches.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/mpt/path.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/mpt/nibble.hpp
```
sheapwilliams: **suggestion (usability)**: These are separate libraries, I think.  It seems like we should avoid a mono-library.

Consider putting these at the same level as `monad`.

src/monad/mpt/test/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+project(monad-mpttests)
```
sheapwilliams: **issue (usability)**: I understand that declaring new projects messes with some IDEs presentation of build targets (CLion specifically, maybe VSCode).  New project declarations seem to hide previous projects and their targets.  Further I think project level settings are sometimes reset when declaring a new project, depending on the cmake cache.

Consider just removing all `project(...)` declarations, save at the top level.

src/monad/mpt/test/test_path.cpp
```diff
@@ -0,0 +1,129 @@
+#include <gtest/gtest.h>
+#include <monad/test/utils.hpp>
+#include <iterator>
+#include <monad/mpt/path.hpp>
+#include <algorithm>
+#include <iostream>
```
sheapwilliams: **chore:** As a general note, the whole PR needs to be `clang-format`'d.  In the future, we can make a target to have the build system do this for you.

include/monad/mpt/path.hpp
```diff
@@ -0,0 +1,218 @@
...
+                    std::advance(it, 1);
+                    return first_byte;
+                }
+            }
+        }();
+
+        bytes.push_back(first_byte);
+
+        // should be an even number of hops away from the end
+        assert((std::distance(it, nibbles_.end()) % 2) == 0);
```
sheapwilliams: **suggestion**: James added a monad assert functionality.  Might want to use that instead?

zander-xyz: Seems reasonable since we get UB if this isn't true.

An alternative is to use `std::expected` where the success case returns a `byte_string` and the failure returns some sort of indicator (maybe a ~~compile time~~ string?). This would require an upgrade to C++23. Thoughts?

src/monad/mpt/test/test_path.cpp
zander-xyz: Will do this before merging so as to not destroy the diff

CMakeLists.txt
```diff
@@ -21,19 +21,14 @@ add_subdirectory(third_party/ethash)
 add_subdirectory(third_party/evmone)
 add_subdirectory(third_party/intx)
 
+include(cmake/Test.cmake)
+include(cmake/Util.cmake)
```
jhunsaker, line 25: this shouldn't be split out into a new file - it's 5 lines and it's defining something quite important that people need to know

making modular is not better

CMakeLists.txt
```diff
@@ -71,4 +66,11 @@ target_link_options(monad
 )
 monad_compile_options(monad)
 
+# rlp
+add_library(rlp STATIC
```
jhunsaker, line 70: we don't need a separate library - keep things simple

cmake/Test.cmake
```diff
@@ -0,0 +1,18 @@
+enable_testing()
```
jhunsaker, line 1: rename to test.cmake

cmake/Util.cmake
```diff
@@ -0,0 +1,7 @@
+function(monad_compile_options target)
```
jhunsaker, line 1: revert this change

include/monad/rlp/rlp.hpp
```diff
@@ -0,0 +1,261 @@
...
+    byte_string bytes;
+
+    constexpr bool operator==(Encoding const&) const = default;
+};
+
+namespace impl
+{
+
+// size of bytes if leading zeroes were stripped off
+constexpr auto to_compacted_size(std::unsigned_integral auto num)
```
jhunsaker, line 93: put the return type, auto detracts from readability here

include/monad/rlp/rlp.hpp
```diff
@@ -0,0 +1,261 @@
...
+// size of bytes required to represent num in big endian with leading
+// zeroes stripped off
+constexpr auto to_big_endian_compacted_size(std::unsigned_integral auto num)
+{
+    return to_compacted_size(intx::to_big_endian(num));
+}
+
+// convert integral type into big endian and compact into byte_string
+// array, stripping off leading zeroes
+constexpr auto to_big_endian_compacted(std::unsigned_integral auto num) -> byte_string
```
jhunsaker, line 114: trailing return type is a lot less readable

include/monad/rlp/rlp.hpp
jhunsaker, line 93: when a function starts with `to_` I think of a conversion or encoding function - but this is just calculating a size, so it should just be `compacted_size`. Is `compacted` an ethereum term btw?

include/monad/rlp/rlp.hpp
```diff
@@ -0,0 +1,261 @@
...
+    // nothing in the header
+    if (bytes.size() == 1 and bytes.front() < BYTES_55_MIN) {
+        target.push_back(bytes.front());
+        return;
+    }
+
+    if (bytes.size() <= 55) {
+        target.push_back(BYTES_55_BASE + bytes.size());
+    } else {
+        assert(bytes.size() < std::numeric_limits<uint64_t>::max());
```
jhunsaker, line 135: a runtime assertion for a case that basically could not happen?

include/monad/rlp/rlp.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+constexpr auto encode_single(byte_string& target, Encoding const& encoding)
+{
+    target += encoding.bytes;
+}
+
+// Returns the number of bytes needed to encode the bytes array
+constexpr auto size_of_encoding(byte_string_view bytes) -> size_t
+{
+    if (bytes.size() == 1 and bytes.front() < BYTES_55_MIN) {
```
jhunsaker, line 171: let's use operator syntax most people are familiar with - &&

include/monad/rlp/rlp.hpp
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    // for byte strings larger than 55, we also need to include
+    // the number of bytes of the payload size
+    return first_plus_payload + to_big_endian_compacted_size(bytes.size());
+}
+
+inline auto size_of_encoding(std::string const& str) -> size_t
+{
+    auto const* ptr = reinterpret_cast<byte_string::value_type const*>(str.data());
```
jhunsaker, line 189: variables should be const if possible

jhunsaker: `auto const* const ptr`

PR #7, opened 2022-12-19
Syntactic amendments and bug fix to RLP encoding

zander-xyz:


include/monad/rlp/rlp.hpp
```diff
@@ -90,10 +90,14 @@ namespace impl
 {
 
 // size of bytes if leading zeroes were stripped off
-constexpr auto to_compacted_size(std::unsigned_integral auto num)
+constexpr size_t size_of_compacted_num(std::unsigned_integral auto num)
 {
+    if (num == 0) {
+        return 0;
+    }
```
zander-xyz: this one was a bug that I noticed

jhunsaker: could just do `end = reinterpret_cast<...>(&num + 1);` and return `end - start`

I think that reads more naturally than subtracting the size of the eliminated portion and also removes the if condition

zander-xyz: > and return end - start

Did you mean return `end - pointer`? `start` is defined as `&num` here.
> I think that reads more naturally than subtracting the size of the eliminated portion

Fair
> and also removes the if condition

How?

jhunsaker: sorry, I meant you don't need to distinguish `start` and `pointer` - remove `pointer` (meaningless name) and just increment `start`, so yes, return `end - start`

`end - start` will be 0 in case of 0 input

jhunsaker: this code also has another bug, it can go past the end

```C++
auto const* start = reinterpret_cast<byte_string::value_type*>(&num);
auto const* const end = reinterpret_cast<byte_string::value_type*>(&num + 1);
while (start < end && *start == 0) {
  ++start;
}
return end - start;
```

zander-xyz: done

PR #8, opened 2022-12-22
use conan for rocksdb

jhunsaker:


conan/conanfile.txt
```diff
@@ -0,0 +1,19 @@
+[requires]
+rocksdb/6.29.5
+
+[generators]
+cmake_find_package
+
+[options]
+lz4:shared=True
+rocksdb:shared=False  # TODO
```
zander-xyz, line 9: what are these TODOs for?

conan/conanfile.txt
```diff
@@ -0,0 +1,19 @@
+[requires]
```
zander-xyz, line 1: Can this file be replaced with `conan_cmake_configure()`?

https://github.com/conan-io/cmake-conan

jhunsaker: it can, but I don't like that approach, it makes the cmake more complex and won't match other usage of conan

conan/conanfile.txt
jhunsaker: the recipe deletes the headers on the shared lib build, their comment doesn't make sense to me

jhunsaker: the workaround is to derive from their recipe, but that seemed too complex for first pass

PR #13, opened 2023-01-06
Add Account structure

sheapwilliams:
- Ensure cache line alignment

### DIFF
--- a/include/monad/core/account.hpp
+++ b/include/monad/core/account.hpp
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <monad/config.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/int.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+struct Account
+{
+    uint256_t balance{0};
+    bytes32_t code_hash{NULL_HASH};
+    uint64_t nonce{0};
+    uint64_t incarnation{0};
+};
+
+static_assert(sizeof(Account) == 80);
+static_assert(alignof(Account) == 8);
+
+MONAD_NAMESPACE_END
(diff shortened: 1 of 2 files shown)


include/monad/core/account.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <monad/config.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/int.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+struct Account {
+    uint256_t balance{0};
+    bytes32_t code_hash{NULL_HASH};
```
jhunsaker, line 12: let's use uint256 for hashes and values - migration will be using bit_cast in the meantime

better alignment and seems more conceptually correct to me (like hash values used as indexes)

include/monad/core/account.hpp
```diff
@@ -0,0 +1,16 @@
...
+    uint64_t nonce{0};
+    uint64_t incarnation{0};
+};
+
+MONAD_NAMESPACE_END
```
jhunsaker, line 20: add sizeof and alignment assertions

include/monad/core/bytes.hpp
```diff
@@ -11,4 +11,6 @@ using bytes32_t = ::evmc::bytes32;
 static_assert(sizeof(bytes32_t) == 32);
 static_assert(alignof(bytes32_t) == 1);
 
+inline constexpr bytes32_t NULL_HASH{0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470_bytes32};
```
sheapwilliams, line 15: Still need to change this type.

sheapwilliams: Ok, bytes32_t is a big endian type.  uint256_t is a native type.  I'll revert this for now.

PR #17, opened 2023-01-17
Rlp uint128 numbers

sheapwilliams:
rlp.hpp was never formatted correctly, so put that in its own commit to make review easier.

### DIFF
--- a/include/monad/rlp/rlp.hpp
+++ b/include/monad/rlp/rlp.hpp
@@ -92,7 +92,8 @@ namespace rlp
     {
         template <class T>
         concept unsigned_integral =
-            std::unsigned_integral<T> || std::same_as<uint256_t, T>;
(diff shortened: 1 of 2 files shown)


include/monad/rlp/rlp.hpp
```diff
@@ -92,7 +92,8 @@ namespace rlp
     {
         template <class T>
         concept unsigned_integral =
-            std::unsigned_integral<T> || std::same_as<uint256_t, T>;
+            std::unsigned_integral<T> || std::same_as<uint128_t, T> ||
```
zander-xyz, line 95: Remind me again where this is needed? Part of the transaction?

sheapwilliams: Yup.

PR #20, opened 2023-01-19
Legacy/EIP-155 Transaction RLP support

sheapwilliams:
~~The transaction struct is miserable to layout...Very much open to suggestions here.~~

~~As such, static asserting the size doesn't make tons of sense - do we want to explicitly add compiler-added padding here?~~

Update: Found a better layout after realizing that chain id and signature parity were pretty deeply coupled.

Test data uses example data in [EIP-155](https://eips.ethereum.org/EIPS/eip-155), and output byte_array is verified using [this](https://toolkit.abdk.consulting/ethereum#rlp,transaction) calculator.


src/monad/rlp/test/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
...
-    LIBRARIES monad
-)
+  TARGET
+  monad-rlptests
+  SOURCES
+  test_rlp.cpp
+  rlp-account.cpp
+  rlp-transaction.cpp
+  LIBRARIES
+  monad)
```
sheapwilliams, line 9: This was formatted with `cmake-format`.  I'm happy to add a config file for whatever style we want, but I have similar thoughts/hopes/desires as other code formatting - set it and forget it.

zander-xyz: I know it's been a while since we discussed, but is our CI in a state where we could potentially have it run a formatting step? Was the problem here that in order for it to work, we'd have to branch off the main repo?

include/monad/core/transaction.hpp
```diff
@@ -0,0 +1,28 @@
...
+    struct Transaction
+    {
+        uint256_t r{}, s{};
+        uint128_t amount{};
+        uint64_t gasPrice{};
+        uint64_t gasLimit{};
+        byte_string data;
+        uint64_t nonce{};
+        uint16_t v{};
+        std::optional<evmc::address> to{std::nullopt};
```
zander-xyz: is the explicit universal default initialization necessary?

include/monad/core/transaction.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <monad/config.hpp>
+#include <monad/core/byte_string.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/int.hpp>
+
+#include <optional>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace legacy
```
zander-xyz: I think this is more stemming from my own ignorance, but is it really called legacy vs non-legacy in ethereum? Also, I think @jhunsaker mentioned putting this in some `Traits` class. I like that suggestion, what are your thoughts?

include/monad/core/transaction.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+    struct Transaction
+    {
+        uint256_t r{}, s{};
+        uint128_t amount{};
+        uint64_t gasPrice{};
+        uint64_t gasLimit{};
```
zander-xyz: sorry, style comment, but should this be
```suggestion
        uint64_t gas_price{};
        uint64_t gas_limit{};
```

include/monad/core/transaction.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <optional>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace legacy
+{
+    struct Transaction
+    {
+        uint256_t r{}, s{};
```
zander-xyz: what does `r` and `s` stand for?

include/monad/core/transaction.hpp
```diff
@@ -0,0 +1,28 @@
...
+        uint128_t amount{};
+        uint64_t gasPrice{};
+        uint64_t gasLimit{};
+        byte_string data;
+        uint64_t nonce{};
+        uint16_t v{};
```
zander-xyz: sorry, similarly what is `v`?

include/monad/rlp/rlp-transaction.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <monad/config.hpp>
+#include <monad/core/assert.h>
+#include <monad/core/transaction.hpp>
+#include <monad/rlp/rlp.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace rlp
+{
+    namespace transaction
```
zander-xyz: nit: could just do
```suggestion
namespace rlp::transaction
{
```
to avoid the extra indent

src/monad/rlp/test/rlp-transaction.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <monad/rlp/rlp-transaction.hpp>
+
+using namespace monad::rlp;
+
+TEST(Rlp_Transaction, EncodeEip155)
+{
+    using namespace intx;
+    using namespace evmc::literals;
+
+    // Example data from: https://eips.ethereum.org/EIPS/eip-155
```
zander-xyz: love this comment

src/monad/rlp/test/rlp-transaction.cpp
```diff
@@ -0,0 +1,47 @@
...
+        0xfe, 0x53, 0x75, 0x67, 0x86, 0x60, 0x03, 0xe1, 0xa1, 0x5d, 0x3c,
+        0x71, 0xff, 0x63, 0xe1, 0x59, 0x06, 0x20, 0xaa, 0x63, 0x62, 0x76,
+        0xa0, 0x67, 0xcb, 0xe9, 0xd8, 0x99, 0x7f, 0x76, 0x1a, 0xec, 0xb7,
+        0x03, 0x30, 0x4b, 0x38, 0x00, 0xcc, 0xf5, 0x55, 0xc9, 0xf3, 0xdc,
+        0x64, 0x21, 0x4b, 0x29, 0x7f, 0xb1, 0x96, 0x6a, 0x3b, 0x6d, 0x83};
+    auto const encoded_transaction = monad::rlp::transaction::encode(t);
+    EXPECT_EQ(encoded_transaction.bytes, eip155_transaction);
+
+    t.v = 15; // invalid v
+    EXPECT_DEATH(monad::rlp::transaction::encode(t), "Assertion");
```
zander-xyz: cool!

src/monad/rlp/test/CMakeLists.txt
sheapwilliams: I don't know.  I think CI is just not a priority at this moment.

include/monad/core/transaction.hpp
sheapwilliams: Prolly not - can easily remove.

include/monad/core/transaction.hpp
sheapwilliams: Traits will decide "how" to act based on config and state.  These are just simple functions that do a single thing once that "how" has been decided.  That system should sit above this stuff, I think.

include/monad/core/transaction.hpp
sheapwilliams: They are the signature components.  From [here](https://cryptobook.nakov.com/digital-signatures/ecdsa-sign-verify-messages)

>The calculated signature {r, s} is a pair of integers, each in the range [1...n-1]. It encodes the random point R = k * G, along with a proof s, confirming that the signer knows the message h and the private key privKey. The proof s is by idea verifiable using the corresponding pubKey.

Together with v, or the parity, they can provide a unique sender key recovery.

include/monad/core/transaction.hpp
sheapwilliams: See above.  But it is an encoded parity for the signature.  [EIP-155](https://eips.ethereum.org/EIPS/eip-155) also encodes the chain_id in it to prevent replay attacks from testnets.

include/monad/rlp/rlp-transaction.hpp
sheapwilliams: Not sure where James will land with namespacing in his refactor MR.  But will incorporate this for now anyway.

include/monad/core/transaction.hpp
zander-xyz: Cool, thanks! Mind putting this as a comment here since the variable names themselves are a little non-descriptive? Normally I would try to make the variable names more descriptive but it seems like it's formally known as `r`, `s`, `v` so it's probably fine.

sheapwilliams: As discussed in slack, we'll comment references to other people's documentation, but don't want to copy that documentation into comments.

include/monad/core/transaction.hpp
jhunsaker: We should have 1 transaction type which is the union of all the fields. And then encoding functions specific to the different types.

sheapwilliams: Sure - we can do this vs. an introspective `boost::hana` approach.  It means the structure now has invariants (I think), and member functions.  I'll pause this PR and continue to working to that end.

include/monad/core/signature.hpp
```diff
@@ -0,0 +1,22 @@
...
+    uint256_t r{};
+    uint256_t s{};
+    std::optional<uint64_t> chain_id{};
+    bool odd_y_parity{};
+};
+
+static_assert(sizeof(signature_and_chain) == 88);
+static_assert(alignof(signature_and_chain) == 8);
+
+uint64_t get_v(signature_and_chain const &sc) noexcept;
```
zander-xyz: Is there a reason not to make this a member function of `signature_and_chain`?

include/monad/core/signature.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <monad/core/int.hpp>
+
+#include <optional>
+
+MONAD_NAMESPACE_BEGIN
+
+struct signature_and_chain
```
zander-xyz: clang-format?

include/monad/rlp/encode_helpers.hpp
```diff
@@ -33,5 +34,6 @@ inline byte_string encode_address(address_t const &a)
 }
 
 byte_string encode_account(Account const &, bytes32_t const &code_root);
+byte_string encode(Transaction const &);
```
zander-xyz, line 37: should we name this `encode_transaction` to be consistent with the above? Or we could change `encode_account` to `encode` which actually might be better

src/monad/rlp/encode_helpers.cpp
```diff
@@ -13,4 +16,26 @@ byte_string encode_account(Account const &account, bytes32_t const &code_root)
...
+            encode_address(*(txn.to)),
+            encode_unsigned(txn.amount),
+            encode_string(txn.data),
+            encode_unsigned(v),
+            encode_unsigned(txn.sc.r),
+            encode_unsigned(txn.sc.s));
+     } //else if (txn.type == Transaction::transaction_type::eip2930)
+        //{
+        return encode_list();
+    //}
```
zander-xyz: Does this function still need to be completed?

sheapwilliams: I've removed. :+1:

include/monad/rlp/encode_helpers.hpp
sheapwilliams: James' refactor made room for simpler names.  I'd be all for removing `_account` from the `encode_account` function.

include/monad/core/signature.hpp
sheapwilliams: Without class invariants, I trend towards leaving functions free as default.  If this was maintaining some sort of invariant, then I'd be sure to make it a class member.  We can discuss more in person.

zander-xyz: Discussed in person. Thanks for explaining Shea!

PR #21, opened 2023-01-19
rlp refactor

jhunsaker:
simplified rlp code


include/monad/rlp/encode.hpp
```diff
@@ -0,0 +1,46 @@
...
+    }
+    return result;
+}
+
+inline byte_string encode_list(auto const &...args)
+{
+    size_t size = 0;
+    ([&] { size += args.size(); }(), ...);
+    byte_string result;
+    if (size > 55) {
```
sheapwilliams, line 38: Do you want to `.reserve()` the `result` here to avoid the reallocation of the byte_string?

include/monad/rlp/encode.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <monad/core/byte_string.hpp>
+#include <monad/rlp/util.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+inline byte_string encode_string(byte_string_view const str)
+{
+    byte_string result;
+    uint32_t const size = str.size();
+    if (size == 1 && str[0] <= 0x7f) {
```
sheapwilliams, line 16: ...same reserve question here.

include/monad/rlp/encode_helpers.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+inline byte_string
+encode_string(Account const &account, bytes32_t const &code_root)
+{
+    return encode_list(
+        encode_string(account.nonce),
+        encode_string(account.balance),
+        encode_string(code_root),
+        encode_string(account.code_hash));
+}
```
sheapwilliams: I observe that this form yields more byte copying than a single in-place population of the list...do we care about that?

include/monad/rlp/encode.hpp
jhunsaker: possibly, will submit as separate pr for performance

include/monad/rlp/encode.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
zander-xyz, line 1: no `rlp` namespace? ditto everywhere else

include/monad/rlp/encode.hpp
zander-xyz, line 16: Can we please try to avoid magic numbers here and everywhere else?

include/monad/rlp/encode_helpers.hpp
```diff
@@ -0,0 +1,39 @@
...
+MONAD_NAMESPACE_BEGIN
+
+inline byte_string encode_string(unsigned_integral auto const &n)
+{
+    return encode_string(to_big_compact(n));
+}
+
+inline byte_string encode_string(bytes32_t const &b)
+{
+    return encode_string(to_byte_string_view(b.bytes));
```
zander-xyz, line 27: do we really need a helper function for something that's just
```suggestion
    return encode_string({b.bytes, sizeof(b.bytes)});
```
?

include/monad/rlp/encode_helpers.hpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+inline byte_string encode_string(address_t const &a)
+{
+    return encode_string(to_byte_string_view(a.bytes));
```
zander-xyz, line 32: ditto

include/monad/rlp/encode_helpers.hpp
zander-xyz: nit: this is called `encode_string` but really it's encoding an `Account`?

include/monad/rlp/encode.hpp
```diff
@@ -0,0 +1,46 @@
...
+        result += str;
+    }
+    else {
+        result.push_back(0x80 + size);
+        result += str;
+    }
+    return result;
+}
+
+inline byte_string encode_list(auto const &...args)
```
zander-xyz: this expects args to be the encoding retrieved with `encode_string`? can we just have encode_list take in the `args` to encode directly, similar to the original impl?

include/monad/rlp/encode.hpp
jhunsaker: a magic number is bad if it's something computed and used without explanation

this is straight from the spec and it's not appearing multiple times, I think replacing these with single-use macros is just more confusing - you'll have to keep referring to the macros

include/monad/rlp/encode_helpers.hpp
jhunsaker: that helper is generic, not just for rlp

include/monad/rlp/encode.hpp
jhunsaker: no, encode_list can be nested

I think explicit is better, the code is easier to read, especially with nested structures. there's also semantic confusion - is a byte_string the result of an encoding or to be encoded? this was tried to resolved with the Encoding type but that just added more complexity

include/monad/rlp/encode_helpers.hpp
zander-xyz: but...the construction is straight up...shorter?

include/monad/rlp/encode.hpp
zander-xyz: > no, encode_list can be nested

so it doesn't expect the args to be pre encoded?

Anyways, I think we can just agree to disagree on library design. If you want to merge this in, I won't stop you.

At the very least, can you please add a comment explaining how to use `encode_list` (ie. assumptions about args), because it's not entirely clear to me (and I know how rlp works)

include/monad/rlp/encode.hpp
zander-xyz: Cool, this explanation makes sense, resolving... 👍

include/monad/rlp/encode_helpers.hpp
jhunsaker: agreed, let's evaluate the performance later

changed the function names

include/monad/rlp/encode.hpp
jhunsaker: added rlp namespace

include/monad/rlp/encode_helpers.hpp
jhunsaker: shorter? as in less syntax? harder to read though

include/monad/rlp/encode.hpp
jhunsaker: it's at least obvious it takes strings now (or things convertible to strings)

PR #23, opened 2023-01-24
Python implementation peeking left and right

zander-xyz:
# Summary
`nodes` list is a lexicographically sorted list of branch and leaf nodes that the `peek_*` functions operate on.
# Unit Tests
Here's a visual depiction of the unit test trie
5C196511-4FA5-45C8-BA9A-40D0CB3AF8E9

### DIFF
--- a/peek.py
+++ b/peek.py
@@ -95,18 +95,28 @@ def peek_left_from_work(work_index, work, nodes):
     assert(isinstance(work_index, WorkIndex))
     assert(isinstance(work[int(work_index)], Leaf))
 
+    insort_index = bisect_left(nodes,
+                               work[int(work_index)].path,
+                               key=lambda n: n.path)
+    if insort_index == 0:
+        return None if work_index == 0 else work_index-1
+
+    left_from_nodes = find_parent(len(nodes)-1, nodes) \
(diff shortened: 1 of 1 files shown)


peek.py
```diff
@@ -0,0 +1,201 @@
...
+    if is_last_in_branch(i, nodes):
+        return i+1 
+
+    for index in range(i+1, len(nodes)):
+        if are_siblings(index, i, nodes):
+            return index
+
+    return None
+
+class WorkIndex:
```
zander-xyz, line 69: Used to distinguish between a node in the `nodes` list and one in the `work` list

peek.py
```diff
@@ -0,0 +1,201 @@
...
+                                   work[int(work_index)].path,
+                                   key=lambda n: n.path)
+        if insort_index == 0:
+            return None
+
+        return find_parent(len(nodes)-1, nodes) \
+                if insort_index == len(nodes) \
+                else peek_left(insort_index, nodes)
+
+    return work_index - 1
```
zander-xyz: Return work built up from previous iterations if available

zander-xyz: Actually, there's a bug here I need to fix.

PR #26, opened 2023-01-25
Support EIP-2930 RLP encoding

sheapwilliams:
[EIP-2930](https://eips.ethereum.org/EIPS/eip-2930) specification, and some [discussion](https://www.youtube.com/watch?v=qQpvkxKso2E) on it.

This is currently rebased onto #20 , as it evolves that code.


include/monad/core/signature.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <optional>
+
+MONAD_NAMESPACE_BEGIN
+
+struct SignatureAndChain
+{
+    uint256_t r{};
+    uint256_t s{};
+    std::optional<uint64_t> chain_id{};
+    bool odd_y_parity{};
```
sheapwilliams, line 14: Chain ID and parity are gathered here in to the same structure cause [EIP-155](https://eips.ethereum.org/EIPS/eip-155) coupled them in `v`.

src/monad/rlp/encode_helpers.cpp
```diff
@@ -33,7 +55,19 @@ byte_string encode(Transaction const &txn)
             encode_unsigned(txn.sc.r),
             encode_unsigned(txn.sc.s));
     }
-    return encode_list();
+    MONAD_ASSERT(txn.sc.chain_id != std::nullopt);
+    return byte_string{0x01} += encode_list(
```
zander-xyz, line 59: can we switch on the transaction type and then just `assert(false)` or something for invalid values?

sheapwilliams: Yea, we can do that...

sheapwilliams: I'll address that in #27 - the 2930 logic is pretty simple.

src/monad/rlp/encode_helpers.cpp
```diff
@@ -1,9 +1,34 @@
...
+#include <algorithm>
+#include <numeric>
+#include <string>
 
 MONAD_RLP_NAMESPACE_BEGIN
 
+byte_string encode_access_list(Transaction::AccessList const &list)
+{
+    byte_string result{};
+    std::ranges::for_each(list, [&result](auto const &i) {
```
jhunsaker, line 17: `for (auto const& i : list) {` is easier to read?

sheapwilliams: Sure - see #29

PR #30, opened 2023-01-30
MPT Python generate transformation list

zander-xyz:
New in this PR are the following
1. `peek_*(...)`: logic to peek left and right given a pointer into the pre-existing state and the work queue
2. `generate_transformation_list(...)`: given a list of nodes to be updated, and pre-existing state, generate a list of node for which the transformation rules can be applied.

Files reviewed in https://github.com/zander-xyz/monad/pull/1 are:
1. `transform.py`

### DIFF
--- a/peek.py
+++ b/peek.py
@@ -213,13 +213,9 @@ def peek_right_from_work(from_index, work, nodes):
                                key=lambda n: n.path)
 
     if insort_index == len(nodes):
-        if from_index == (len(work)-1):
-            return None
-
-        if work[int(from_index+1)].action == LeafAction.DELETE:
-            return peek_right_from_work(from_index+1, work, nodes)
-
-        return from_index+1
+        # Does not make sense for any work after this point to be leaf deletions
+        assert(all([work[i].action == LeafAction.UPSERT for i in range(int(from_index), len(work))]))
+        return None if from_index == (len(work)-1) else from_index+1
 
     right_from_nodes = peek_right_no_work(insort_index, nodes) \
             if nodes[insort_index] == work[int(from_index)] \


transform.py
```diff
@@ -0,0 +1,55 @@
...
+
+def main():
+    # merge two leaves together
+    assert(transform(None                    , Leaf("1234"), Leaf("1235"), None)         == Branch("123", [Leaf("1234"), Leaf("1235")]))
+    assert(transform(Branch("12", ["1", "2"]), Leaf("1234"), Leaf("1235"), None)         == Branch("123", [Leaf("1234"), Leaf("1235")]))
+    assert(transform(None                    , Leaf("1234"), Leaf("1235"), Leaf("1237")) == Branch("123", [Leaf("1234"), Leaf("1235")]))
+    assert(transform(Branch("12", ["1", "2"]), Leaf("1234"), Leaf("1235"), Leaf("1237")) == Branch("123", [Leaf("1234"), Leaf("1235")]))
+
+    # no action
+    assert(transform(Branch("123", ["1", "2"]) , Leaf("1234") , Leaf("1235") , None)          == None)
```
sheapwilliams, line 34: Why wouldn't this return `Branch("123", [Leaf("1234"), Leaf("1235")])`?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+    assert(transform(None                      , Leaf("12345"), Leaf("12356"), Leaf("12357")) == None)
```
sheapwilliams, line 35: I don't get this one on a few levels.  First, why would this list ever exist?  It seems to break loads of invariants?  Second, it seems so similar to line 28, and that it should definitely yield `Branch("123", [Leaf("12345"), Leaf("12346")])`.  What am I missing?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+    assert(transform(Branch("1234", ["1", "2"]), Leaf("12345"), Leaf("12356"), Leaf("12357")) == None)
```
sheapwilliams, line 36: Why wouldn't this yield `Branch("1234", ["1", "2", "5"])`?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # extend a branch
+    assert(transform(None       , Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1235") , None)                                       == Branch("123", [Leaf("1231"), Leaf("1232"), Leaf("1235")]))
```
sheapwilliams, line 39: The list nomenclature switched here... `s_2` was `Branch("123", ["1", "2"])` above, but now it is `Branch("123", [Leaf("1231"), Leaf("1232")])`.  On purpose?  What is the intended difference?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+    assert(transform(None       , Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1235") , Leaf("1236"))                               == Branch("123", [Leaf("1231"), Leaf("1232"), Leaf("1235")]))
+    assert(transform(Branch("12", [Leaf("1212"), Leaf("1222")]), Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1235"), Leaf("1236")) == Branch("123", [Leaf("1231"), Leaf("1232"), Leaf("1235")]))
```
sheapwilliams, line 41: I see how this would collapse `s_3` into `s_2` - makes sense.  But why wouldn't isn't `s_1` invalid here?  Wouldn't it be (based on this nomenclature) `Branch("12", [Leaf("1212"), Leaf("1222"), Branch("123", [Leaf("1231"), Leaf("1232")])])`?

`s_2` definitely is a peer of the components of `s_1` in this picture - how is that supposed to be represented here?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # emitting new branch
+    assert(transform(None                                     , Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")]), None)         == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")])]))
```
sheapwilliams, line 44: 🤔 I don't get how `s_4` could ever be `None` here, unless we are representing branch/leaf combined structures in the list somehow...Is this a convenience?  How do you know to break the combined branch/leaf structure into new nodes for `s_*`?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+    assert(transform(Branch("1", [Leaf("1123"), Leaf("1211")]), Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")]), None)         == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")])]))
+    assert(transform(None                                     , Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")]), Leaf("1256")) == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")])]))
```
sheapwilliams, line 46: So, not looking forward here anymore?  I see that `Leaf("1256")` isn't (yet?) included in the branch output.  Do I have that right?

transform.py
```diff
@@ -0,0 +1,55 @@
...
+    assert(transform(Branch("1", [Leaf("1123"), Leaf("1211")]), Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")]), Leaf("1256")) == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Branch("124", [Leaf("1241"), Leaf("1242")])]))
+
+    assert(transform(None                                      , Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245"), None)         == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245")]))
+    assert(transform(Branch("11", [Leaf("1112"), Leaf("1121")]), Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245"), None)         == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245")]))
+    assert(transform(None                                      , Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245"), Leaf("1268")) == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245")]))
+    assert(transform(Branch("11", [Leaf("1112"), Leaf("1121")]), Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245"), Leaf("1268")) == Branch("12", [Branch("123", [Leaf("1231"), Leaf("1232")]), Leaf("1245")]))
+
```
sheapwilliams, line 53: Consider adding some branch-at-root tests to make sure the logic doesn't expect anything but an empty string as the first arg.

peek.py
```diff
@@ -0,0 +1,316 @@
...
+    assert(peek_left_no_work(5, nodes) == 1)
+    assert(peek_left_no_work(6, nodes) == 5)
+    assert(peek_left_no_work(7, nodes) == 6)
+    assert(peek_left_no_work(8, nodes) == 4)
+    assert(peek_left_no_work(9, nodes) == 8)
+
+    assert(peek_right_no_work(0, nodes) == None)
+    assert(peek_right_no_work(1, nodes) == 4)
+    assert(peek_right_no_work(2, nodes) == 3)
+    assert(peek_right_no_work(3, nodes) == 4)
```
sheapwilliams, line 259: If 3 looks right and sees 4, then why does 2 look left and see `None` (line 247) instead of 1?

peek.py
```diff
@@ -0,0 +1,316 @@
...
+    assert(peek_left_from_first_work(Leaf("13322130"), nodes) == 0)
+    assert(peek_left_from_first_work(Leaf("04322130"), nodes) == 0)
+
+    assert(peek_left_from_first_work(Leaf("02112220"), nodes) == 4)
+    assert(peek_left_from_first_work(Leaf("02212220"), nodes) == 8)
+
+    assert(peek_left_from_first_work(Leaf("01331132"), nodes) == 7)
+    assert(peek_left_from_first_work(Leaf("01311132"), nodes) == 6)
+
+    assert(peek_left_from_first_work(Leaf("0 123456"), nodes) == None)
```
sheapwilliams, line 277: What does space mean?  I'm not sure how to read this test...

peek.py
```diff
@@ -0,0 +1,316 @@
...
+    assert(peek_right_no_work(6, nodes) == 7)
+    assert(peek_right_no_work(7, nodes) == 8)
+    assert(peek_right_no_work(8, nodes) == 9)
+    assert(peek_right_no_work(9, nodes) == None)
+
+    # Unit tests for work list peeking left with single element
+    assert(peek_left_from_first_work(Leaf("13322130"), nodes) == 0)
+    assert(peek_left_from_first_work(Leaf("04322130"), nodes) == 0)
+
+    assert(peek_left_from_first_work(Leaf("02112220"), nodes) == 4)
```
sheapwilliams, line 271: Why isn't this 2?

peek.py
```diff
@@ -0,0 +1,316 @@
...
+    assert(peek_left_from_first_work(Leaf("02212220"), nodes) == 8)
```
sheapwilliams, line 272: Why isn't `Leaf("02112220")` (3) left of this?

peek.py
```diff
@@ -0,0 +1,316 @@
...
+
+    assert(peek_left_from_first_work(Leaf("01331132"), nodes) == 7)
+    assert(peek_left_from_first_work(Leaf("01311132"), nodes) == 6)
+
+    assert(peek_left_from_first_work(Leaf("0 123456"), nodes) == None)
+    assert(peek_left_from_first_work(Leaf(" 0123456"), nodes) == None)
+    assert(peek_left_from_first_work(Leaf("00121211"), nodes) == None)
```
sheapwilliams, line 279: Why wouldn't this be the branch `Branch("00", [Leaf("00131211"), Leaf("00211002")])` (1)?

transform.py
zander-xyz: Intuitively it's a "left-recursive" algorithm, so `Leaf("1234")` would get merged into `Branch("123", ["1", "2"])` first, and then Leaf("1235") would get merged into that.

transform.py
zander-xyz: What invariants does it break?

In this example, none of the nodes can be merged according to the the transformation rules.

Let's assume that these three leaves are the only ones being added to an empty state. Then the last two leaves need to be merged to create `Branch("1235")` before `Leaf("12345")` can be merged with that to create `Branch("123")`

transform.py
zander-xyz: Because the middle two are the nodes being considered to be merged

transform.py
zander-xyz: Not intended, my mistake. Should be the long form

transform.py
zander-xyz: Your intuition is definitely correct here, and eventually (through future iterations of the transformation algo), they would be merged. Just remember that one invocation of `transform` only considers merging the middle two elements. The edge two are left alone.

transform.py
zander-xyz: At this point, all the leaf nodes under the two branches have been parented. So the list started off as something like
```
Leaf("1231") Leaf("1232") Leaf("1241"), Leaf("1242")
```
and then after multiple iterations you reach the above state where there are two branches with `None` on either side

transform.py
zander-xyz: Sorry, I'm not entirely sure what you're asking. Can you please clarify?

transform.py
zander-xyz: What `assert` are you thinking of here?

peek.py
zander-xyz: Because `Leaf("00131211")` is at the beginning of the list and `Leaf("00211002")` sees the branch `01` which was left recursively built. The way I've envisioned peeking right and left is helping answer the question: given I've updated node `xyz`, what is the smallest set of nodes I can reuse to rebuild the root.

peek.py
zander-xyz: ` ` is just a byte that is lexicographically smaller than 0. Yeah, sorry. I deviated away from numeric :(

peek.py
zander-xyz: Because you can reuse the branch at 01. I guess maybe flipping the question: why would it be 2?

peek.py
zander-xyz: It is. `Leaf("02112220")` is `8`, not `3`. But your intuition is correct here

peek.py
zander-xyz: `Branch("00")` needs to be rebuilt here since `00121211` updates something below it. (in this case, creates `Branch("001", [Leaf("00121211"), Leaf("00131211")])`)

transform.py
zander-xyz: Add unit test where Branch has the empty string and then have nodes be concatenated to it.

peek.py
zander-xyz: change the peek function names to be something more descriptive

PR #31, opened 2023-01-30
Support RLP encoding of all types of EIP-2718 Receipts

sheapwilliams:


src/monad/rlp/encode_helpers.cpp
```diff
@@ -89,4 +90,47 @@ byte_string encode_transaction(Transaction const &txn)
...
+
+    for (auto const &i : r.logs) {
+        log_result += encode_log(i);
+    }
+
+    if (r.type == Transaction::Type::eip1559 ||
+        r.type == Transaction::Type::eip2930) {
+        prefix = static_cast<unsigned>(r.type);
+    }
+
```
zander-xyz, line 128: maybe safer to `assert` the transaction type here?

include/monad/core/receipt.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <monad/config.hpp>
+#include <monad/core/address.hpp>
+#include <monad/core/byte_string.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/transaction.hpp>
+
+#include <optional>
```
zander-xyz: what's this needed for?

src/monad/rlp/encode_helpers.cpp
zander-xyz: disregard this comment

include/monad/core/receipt.hpp
sheapwilliams: It is vestigial - I'll remove and ping for re-approval.

include/monad/core/byte_string.hpp
```diff
@@ -22,6 +22,12 @@ byte_string_view to_byte_string_view(unsigned char const (&a)[N])
     return {&a[0], N};
 }
 
+template <class T, size_t N>
+byte_string_view to_byte_string_view(std::array<T, N> const &a)
+{
+    return {a.data(), N};
```
jhunsaker, line 28: can we make this `&a[0]` to be consistent? for example, below for string we have `&s[0]` rather than `s.data()`

it'd be fine to standardize the other way around - `{a.data(), a.size()}` as well as we agree to use that for any container

PR #33, opened 2023-01-31
Add specialized Ethereum receipt bloom filter

sheapwilliams:
- Reference Section 4.3.1 of the Yellow paper for details
- Enables searching receipt trie for topics and addresses

This is sitting on top of #31 , so is currently marked draft


include/monad/core/receipt.hpp
```diff
@@ -17,15 +17,21 @@ struct Receipt
...
+        friend bool operator==(const Log &, const Log &) = default;
     };
 
     Bloom bloom{};
     uint64_t status{};
     uint64_t gas_used{};
     Transaction::Type type{};
     std::vector<Log> logs;
+
+    void add_log(Receipt::Log const &l);
```
zander-xyz, line 30: any concern with having `logs` be a public member?

sheapwilliams: None from me.  They are pretty simple structures.  No complicated invariants.  

Bloom is the one that has an invariant on the log list.  We can hide it and then unhide it again by providing getters/setters...but that seems overwrought to me.

Open to other thoughts here tho.

zander-xyz: Nope, seems reasonable to me.

include/monad/core/receipt.hpp
```diff
@@ -17,15 +17,21 @@ struct Receipt
         byte_string data{};
         std::vector<bytes32_t> topics{};
         address_t address{};
+
+        friend bool operator==(const Log &, const Log &) = default;
```
jhunsaker, line 21: isn't there a default anyways?

src/monad/core/receipt.cpp
```diff
@@ -0,0 +1,33 @@
+#include <ethash/keccak.hpp>
+#include <monad/core/receipt.hpp>
```
jhunsaker, line 2:
```
#include <monad/core/receipt.hpp>

#include <ethash/keccak.hpp>
```

(don't group unrelated headers)

src/monad/core/receipt.cpp
```diff
@@ -0,0 +1,33 @@
...
+        const uint16_t bit =
+            be16toh(reinterpret_cast<uint16_t const *>(h.bytes)[i]) & 2047;
+        const auto byte = 255 - bit / 8;
+        bloom[byte] |= 1 << (bit & 7);
+    }
+}
+
+void populate_bloom(Receipt::Bloom &b, Receipt::Log const &l)
+{
+    set_3_bits(b, byte_string_view{l.address.bytes, sizeof(l.address.bytes)});
```
jhunsaker, line 21: `to_byte_string_view(l.address.bytes)`

src/monad/core/receipt.cpp
```diff
@@ -0,0 +1,33 @@
...
+    for (auto const &i : l.topics) {
+        set_3_bits(b, byte_string_view{i.bytes, sizeof(bytes32_t)});
```
jhunsaker, line 23: same here

src/monad/core/receipt.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <endian.h>
+
+MONAD_NAMESPACE_BEGIN
+
+void set_3_bits(Receipt::Bloom &bloom, byte_string_view const bytes)
+{
+    auto const h{ethash::keccak256(bytes.cbegin(), bytes.size())};
+    for (auto i = 0; i < 3; ++i) {
+        const uint16_t bit =
+            be16toh(reinterpret_cast<uint16_t const *>(h.bytes)[i]) & 2047;
```
jhunsaker, line 13: we use the `intx` function `to_big_endian` in a lot of places - should we use it here consistency? I'm not sure how `bswap` compares against `be16toh` in this case (for our little endian builds). If `be16toh` is "faster" than `bswap` we should note that somewhere for future guidance.

include/monad/core/receipt.hpp
sheapwilliams: I had to add this after gtest functions complained.  Apparently not.

src/monad/core/receipt.cpp
sheapwilliams: Sure.

src/monad/core/receipt.cpp
sheapwilliams: This is the opposite - the bloom filter as an array of bytes is big endian.  We need a number out of it, so are converting to host byte order to do those calculations.  I think this uses the same `bswap` instructions under the hood, but I'll check the assembly.

sheapwilliams: Discussed off-line.  `to_big_endian` implements the correct thing here, but the `to_` part carries semantic information that would be plain wrong.  

Changed in #37 and added comment to point out that dissonance.

src/monad/core/receipt.cpp
sheapwilliams: #37

PR #34, opened 2023-02-01
Enable development with address and undefined behavior sanitizer

sheapwilliams:
- Small diff I was using to develop RLP, figured others might want to use it if it were available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,9 @@ include(tmpsrc/bootstrap.cmake)
 
 project(monad)
 
+option(ASAN  "Turn on Address Sanitizer")
+option(UBSAN "Turn on Undefined Behavior Sanitizer")
+
 ############
 ### conan
 ############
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -110,9 +113,18 @@ target_link_libraries(monad
...
+if(ASAN)
+  message("Address Sanitizer: On")
+  target_compile_options(monad PUBLIC "-fsanitize=address" "-fno-omit-frame-pointer")
+  target_link_options(monad PUBLIC "-fsanitize=address")
+elseif(UBSAN)
+  message("Undefined Behavior Sanitizer: On")
+  target_compile_options(monad PUBLIC "-fsanitize=undefined" "-fno-omit-frame-pointer")
+  target_link_options(monad PUBLIC "-fsanitize=undefined")
+else()
+  target_link_options(monad PRIVATE LINKER:-z,defs)
```
zander-xyz, line 127: should these options not be added regardless of ASAN and UBSAN?

sheapwilliams: They can interfere with ASAN.  From the [docs](https://clang.llvm.org/docs/AddressSanitizer.html#usage): 
> When linking shared libraries, the AddressSanitizer run-time is not linked, so -Wl,-z,defs may cause link errors (don’t use it with AddressSanitizer).

zander-xyz: Ah, nice. Thanks

PR #35, opened 2023-02-01
Nibbles and compact encoding utility function

zander-xyz:


include/monad/trie/nibbles.hpp
```diff
@@ -0,0 +1,78 @@
...
+struct Nibbles
+{
+    bool is_odd;
+    byte_string bytes;
+
+    struct FromNibbleArray
+    {
+    };
+
+    Nibbles(byte_string_view nibbles, FromNibbleArray)
```
zander-xyz, line 19: I'm not married to the idea of using tags to dispatch this constructor. If people have strong opinions I can eliminate the tag

sheapwilliams: Is this the only constructor?  What/Why the dispatch?

src/monad/trie/test/compact_encode.cpp
```diff
@@ -0,0 +1,40 @@
...
+using namespace monad;
+using namespace monad::trie;
+
+TEST(CompactEncoding, Sanity)
+{
+    auto path = Nibbles(
+        byte_string{0x01, 0x02, 0x03, 0x04, 0x05}, Nibbles::FromNibbleArray{});
+
+    EXPECT_EQ(compact_encode(path, false), byte_string({0x11, 0x23, 0x45}));
+    EXPECT_EQ(compact_encode(path, true), byte_string({0x31, 0x23, 0x45}));
```
sheapwilliams, line 15: `true` and `false` are mildly unfortunate here - I can't remember which way the logic is: `is_leaf` or `is_extenstion`.  Would be nice to have named bools here someway or another so I don't have to go off and check.

zander-xyz: I'm okay with named bools here. What I had before were named structs that would dispatch the correct behavior.

@jhunsaker any particular preference here?

include/monad/trie/nibbles.hpp
zander-xyz: It's the only constructor for now. The dispatch is just to be clear that the `byte_string` needs to be an array of nibbles and not something compacted.

zander-xyz: Whilst working on the `Node` struct, I encountered a need to add another `Nibbles(byte_string_view)` constructor! This time, dispatched with `FromStorage{}` (parsing paths from the db). So I'll leave the dispatching as is

PR #39, opened 2023-02-06
state db

jhunsaker:


src/monad/db/state_db.cpp
```diff
@@ -0,0 +1,248 @@
...
+#include <rocksdb/slice.h>
+#include <rocksdb/status.h>
+#include <rocksdb/write_batch.h>
+
+#include <boost/endian/conversion.hpp>
+#include <boost/log/trivial.hpp>
+
+#include <cstring>
+
+MONAD_NAMESPACE_BEGIN
```
sheapwilliams, line 18: When addressing some of the todos, it would be good to have these db functions in their own namespace.

src/monad/db/state_db.cpp
```diff
@@ -0,0 +1,248 @@
...
+                if (!s.ok()) {
+                    BOOST_LOG_TRIVIAL(error) << s.ToString();
+                }
+                MONAD_ASSERT(s.ok());
+                MONAD_ASSERT(cf);
+                cfs_.emplace_back(cf);
+            }
+        }
+        return db;
+    }()}
```
sheapwilliams, line 99: This must be a record length for member initializer!

PR #42, opened 2023-02-15
RLP Decode

tongzhi1998:
Finish RLP Decode & Testing 
Rebase into 4 commits
@kiblitz


.vscode/settings.json
```diff
@@ -3,5 +3,86 @@
...
+        "sstream": "cpp",
+        "stdexcept": "cpp",
+        "stop_token": "cpp",
+        "streambuf": "cpp",
+        "thread": "cpp",
+        "typeindex": "cpp",
+        "typeinfo": "cpp",
+        "valarray": "cpp",
+        "variant": "cpp"
     }
```
sheapwilliams: Go ahead an pull this guy out into a separate commit.  It is fine to leave in this PR, just separate it from this decode_acount commit.

include/monad/core/block.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/int.hpp>
+#include <monad/core/receipt.hpp>
+#include <monad/core/transaction.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+struct BlockHeader {
+    /* glee for shea: existing representation? */
```
sheapwilliams: Let's jump on zoom and figure out what is going on here.

include/monad/core/block.hpp
```diff
@@ -0,0 +1,52 @@
...
+    address_t beneficiary{};
+    bytes32_t state_root{};
+    bytes32_t transactions_root{};
+    bytes32_t receipts_root{};
+    Receipt::Bloom logs_bloom{};
+    uint256_t difficulty{};
+    uint64_t number{0};
+    uint64_t gas_limit{0};
+    uint64_t gas_used{0};
+    uint64_t timestamp{0};
```
sheapwilliams, line 26: Let's discuss alignment here and what might be best.

include/monad/core/block.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
```
sheapwilliams, line 1: I think you intended to have block.hpp in the second commit, yea?

include/monad/rlp/decode.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <monad/rlp/util.hpp>
+
+#include <monad/core/assert.h>
+#include <monad/core/byte_string.hpp>
+
+#include <vector>
+#include <concepts>
+
+// To Be Deleted
+#include <iostream>
```
sheapwilliams: TODO

include/monad/rlp/decode.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+MONAD_RLP_NAMESPACE_BEGIN
+
+// is this big enough for payload sizes?
+using byte_string_loc = uint64_t;
```
sheapwilliams: Per https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/#rlp-decoding
The max size of a list is 8 bytes.  Given the prefix, it can be larger.  But this would be a ridiculously *enormous* byte_string.  Maybe we can assert somewhere to know we won't overflow this?

include/monad/rlp/decode.hpp
```diff
@@ -0,0 +1,81 @@
...
+inline byte_string_loc decode_length(byte_string_view const enc, byte_string_loc i, byte_string_loc length)
+{
+    byte_string_loc result = 0;
+    for (byte_string_loc j = i; j < i + length; ++j)
+    {
+        result *= 256;
+        result += enc[j];
+    }
+    return result;
+}
```
sheapwilliams: mark as `constexpr` and `noexcept` throughout, and where appropriate.

include/monad/rlp/decode.hpp
kiblitz: Resolved by asserting `size < max(uint64)`

include/monad/rlp/decode.hpp
jhunsaker: you can look to `intx::load` functions for example
```
inline constexpr uint64_t decode_unsigned(byte_string_view const str)
{
    MONAD_ASSERT(str.size() <= sizeof(uint64_t));
    uint64_t result{};
    std::memcpy(&intx::as_bytes(result)[sizeof(uint64_t) - str.size(), str.data(), str.size());
    result = intx::to_big_endian(result);
    return result;
}
```

jhunsaker: 1. the `byte_string_view` param should already be adjusted for what you want to decode
2. this should be in `rlp/util.hpp`
3. memcpy loop + cpu instruction of swap should be faster than a shift-add loop
4. could template this function with unsigned_integral type and it would work fine

include/monad/rlp/decode_helpers.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+    bytes32_t res;
+    std::memcpy(res.bytes, dec.data(), 32);
+
+    return res;
+}
+
+inline address_t decode_address(byte_string_view const enc, byte_string_loc &i) {
+    auto dec = decode_string(enc, i);
+    MONAD_ASSERT(dec.size() == 20);
```
jhunsaker: is it true they have to be that length? I though there was a gas opt for values with leading 0's

include/monad/rlp/decode_helpers.hpp
```diff
@@ -0,0 +1,41 @@
...
+    auto dec = decode_string(enc, i);
+    MONAD_ASSERT(dec.size() == 32);
+
+    bytes32_t res;
+    std::memcpy(res.bytes, dec.data(), 32);
+
+    return res;
+}
+
+inline address_t decode_address(byte_string_view const enc, byte_string_loc &i) {
```
jhunsaker: how does this compare to a signature of 

`byte_string_view decode_address(address_t&, byte_string_view const str)`

?

this matches the `encode` signature more closely

src/monad/rlp/decode_helpers.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+    const byte_string_loc end = i + length;
+    MONAD_ASSERT(end <= enc.size());
+
+    return end;
+}
+
+std::pair<Account, bytes32_t> decode_account(byte_string_view const enc, byte_string_loc &i)
+{
+    const byte_string_loc end = end_of_list_encoding(enc, i);
```
jhunsaker: this design of finding the end of the list then iterating over list items then comparing the two results seems odd to me

we should write this in a recursive manner - `decode_string` and `decode_list`

include/monad/rlp/decode_helpers.hpp
sheapwilliams: I believe that hex values in ethereum rlp are treated as byte arrays, not zero-trimmed integers.  That is what most of the tools I've used for validation say anyway.

src/monad/rlp/decode_helpers.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+byte_string_loc end_of_list_encoding(byte_string_view const enc, byte_string_loc &i)
+{
+    MONAD_ASSERT(enc.size() < byte_string_max_length);
+    MONAD_ASSERT(i < enc.size());
+
+    byte_string_loc length;
+    const uint8_t &first = enc[i];
+    ++i;
+    MONAD_ASSERT(first >= 192);
```
jhunsaker: use hex values to match the encoding and the spec - no reason to convert to decimals which requires additional mental load on the reader

src/monad/rlp/test/decode.cpp
```diff
@@ -0,0 +1,43 @@
...
+TEST(Rlp, DecodeAfterEncodeString)
+{
+    auto test_single = [](std::string s) {
+        auto enc = to_byte_string_view(s);
+        EXPECT_EQ(decode_string(encode_string(enc)), enc);
+    };
+    test_single("hello world");
+    test_single("Lorem ipsum dolor sit amet, consectetur adipisicing elit");
+    test_single("monad");
+    test_single("Monad Labs");
```
sheapwilliams: I totally understand the desire to DRY this out a little.  But practical tip using googletest - on failure, the line number reported will be where the `EXPECT_EQ` is.  In these cases, all failures will point to the same line, so test failure reporting won't be useful.  

I suggest you just spell these all out so there are individual `EXPECT`tations.

src/monad/core/signature.cpp
```diff
@@ -10,4 +10,24 @@ uint64_t get_v(SignatureAndChain const &sc) noexcept
...
+        auto tmp = v - 35;
+        if (tmp & 1)
+        {
+            sc.odd_y_parity = true;
+            tmp ^= 1;
+        }
+        sc.chain_id = tmp >> 1;
+    }
+}
+
```
sheapwilliams, line 23: Missing unit test for this guy?

sheapwilliams: Also, this should probably be a member function of SignatureAndChain...even the signature says that, yea?  Is this a constructor of SignatureAndChain?

src/monad/rlp/decode_helpers.cpp
```diff
@@ -49,4 +107,75 @@ std::pair<Account, bytes32_t> decode_account(byte_string_view const enc, byte_st
...
+        txn.gas_limit = decode_unsigned<uint64_t>(enc, i);
+        *txn.to = decode_address(enc, i);
+        txn.amount = decode_unsigned<uint128_t>(enc, i);
+        txn.data = decode_string(enc, i);
+        txn.access_list = decode_access_list(enc, i);
+        txn.sc.odd_y_parity = decode_unsigned<bool>(enc, i);
+        txn.sc.r = decode_unsigned<uint256_t>(enc, i);
+        txn.sc.s = decode_unsigned<uint256_t>(enc, i);
+    }
+    else            // Transaction::type::eip2930
```
sheapwilliams: Test that the type is indeed `eip2930` here.  We shouldn't assume that anything else is this type.  There are/will be some experimental types in the future.  Feel free to `assert` that these are the only supported types for now tho.

src/monad/rlp/encode_helpers.cpp
```diff
@@ -133,4 +133,5 @@ byte_string encode_receipt(Receipt const &r)
                         encode_list(log_result));
 }
 
+
```
sheapwilliams: trim this guy...

src/monad/rlp/test/decode_encode_receipt.cpp
```diff
@@ -1,44 +1,65 @@
...
+        auto decoded = decode_log(encoded, pos);
+
+        EXPECT_EQ(encoded, rlp_log);
+
+        EXPECT_EQ(decoded.data, log.data);
+        EXPECT_EQ(decoded.address, log.address);
+
+        EXPECT_EQ(decoded.topics.size(), log.topics.size());
+        for(int i=0;i<static_cast<int>(log.topics.size());++i){
+            EXPECT_EQ(decoded.topics[i], log.topics[i]);
```
sheapwilliams: Similar thing here - consider just unrolling the loop by hand so that future you won't have to figure out which topic exactly is broken.

src/monad/rlp/test/decode_encode_receipt.cpp
```diff
@@ -1,44 +1,65 @@
...
+        auto encoded = encode_log(log);
+        auto decoded = decode_log(encoded, pos);
+
+        EXPECT_EQ(encoded, rlp_log);
+
+        EXPECT_EQ(decoded.data, log.data);
+        EXPECT_EQ(decoded.address, log.address);
+
+        EXPECT_EQ(decoded.topics.size(), log.topics.size());
+        for(int i=0;i<static_cast<int>(log.topics.size());++i){
```
sheapwilliams: Rather than casting the size to unsigned, consider using `unsigned` as the loop iterator:
```
        for(auto i = 0u; i < log.topics.size(); ++i)
```

src/monad/rlp/test/decode_encode_receipt.cpp
```diff
@@ -70,14 +91,22 @@ TEST(Rlp_Receipt, EncodeBloom)
         0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02};
 
     auto const encoded = encode_bloom(bloom);
+    auto const decoded = decode_bloom(encoded, pos);
+
     EXPECT_EQ(encoded, rlp_bloom);
+    for(int i=0;i<256;++i){
+        EXPECT_EQ(bloom[i],decoded[i]);
```
sheapwilliams: Let's add an array matcher in a test header so this is easier to diagnose:
```
    template<typename T, size_t size>
    ::testing::AssertionResult ArraysMatch(const T (&expected)[size],
                                           const T (&actual)[size]){
        for (size_t i(0); i < size; ++i){
            if (expected[i] != actual[i]){
                return ::testing::AssertionFailure() << "array[" << i
                    << "] (" << actual[i] << ") != expected[" << i
                    << "] (" << expected[i] << ")";
            }
        }

        return ::testing::AssertionSuccess();
    }
```
[1](https://stackoverflow.com/a/10062016)

src/monad/rlp/decode_helpers.cpp
```diff
@@ -268,4 +280,82 @@ Receipt decode_receipt(byte_string_view const enc, byte_string_loc& i){
...
+    }
+
+    MONAD_ASSERT(i == end);
+    return ommers;
+}
+
+Block decode_block(byte_string_view const enc, byte_string_loc &i)
+{
+    const byte_string_loc end = end_of_list_encoding(enc, i);
+    Block block;
```
sheapwilliams:
```
    Block block{};
```
and throughout...

src/monad/rlp/test/decode_block.cpp
```diff
@@ -0,0 +1,273 @@
...
+    EXPECT_EQ(block.transactions[1].sc.r, 0xb821bab9f7e78c13d73beaf9cf6e9d2b33bbf2923ae25da58d7efb7a72fd607b_u256);
+    EXPECT_EQ(block.transactions[1].sc.s, 0x3b6699532ca1a2115e2b516695d80f400568619e9301b55ace126558f2126d2_u256);
+
+    EXPECT_EQ(block.transactions[2].nonce, 92940);
+    EXPECT_EQ(block.transactions[2].gas_price, 20000000000);
+    EXPECT_EQ(block.transactions[2].gas_limit, 39000);
+    EXPECT_EQ(*block.transactions[2].to, 0xeDc53fB256c8F5cd6f91AF675BBb89bFC3732c57_address);
+    EXPECT_EQ(block.transactions[2].amount, 0xe7f24b03527b860_u128);
+   
+    // Skipped Some Trivial Testing in the Middle
```
sheapwilliams: ?

include/monad/rlp/decode.hpp
sheapwilliams: I think you only missed James' suggestion here to put this in `rlp/util.hpp`

src/monad/rlp/decode_helpers.cpp
```diff
@@ -48,4 +112,76 @@ decode_account(Account &acc, bytes32_t &code_root, byte_string_view const enc)
...
+    }
+    else if (type == 0x02) // eip1559
+    {
+        ++i;
+        txn.type = Transaction::Type::eip1559;
+    }
+    else // eip155
+    {
+        txn.type = Transaction::Type::eip155;
+    }
```
sheapwilliams, line 188: If we are going to fall off the end of the `if` here, then we should assert above that the only types coming in are [0..2].

src/monad/rlp/decode_helpers.cpp
```diff
@@ -97,6 +98,81 @@ decode_access_list(Transaction::AccessList &al, byte_string_view const enc)
...
+}
+
+byte_string_view decode_log_data(byte_string &data, byte_string_view enc)
+{
+    byte_string_view payload{};
+    auto const rest_of_enc = parse_list_metadata(payload, enc);
+
+    for (auto &b : payload) {
+        data += b;
+    }
```
sheapwilliams: would something like `.append()` be simpler here?

src/monad/rlp/decode_helpers.cpp
```diff
@@ -184,4 +260,33 @@ decode_transaction(Transaction &txn, byte_string_view const enc)
...
+        receipt.type = Transaction::Type::eip1559;
+        ++i;
+    }
+    else if (prefix == 0x02) {
+        receipt.type = Transaction::Type::eip2930;
+        ++i;
+    }
+    else {
+        receipt.type = Transaction::Type::eip155;
+    }
```
sheapwilliams, line 254: same comment here - we should assert these are the only types if we are going to fall of the end of the `if`.

src/monad/rlp/test/decode_encode_transaction.cpp
```diff
@@ -175,5 +224,27 @@ TEST(Rlp_Transaction, EncodeEip1559)
...
+    EXPECT_EQ(decoding.amount, t.amount);
+    EXPECT_EQ(*decoding.to, *t.to);
+    EXPECT_EQ(decoding.sc.r, t.sc.r);
+    EXPECT_EQ(decoding.sc.s, t.sc.s);
+    EXPECT_EQ(*decoding.sc.chain_id, *t.sc.chain_id);
+    EXPECT_EQ(decoding.type, t.type);
+    EXPECT_EQ(decoding.priority_fee, t.priority_fee);
+
+    EXPECT_EQ(decoding.access_list.size(), t.access_list.size());
+    for (size_t i = 0; i < t.access_list.size(); ++i) {
```
sheapwilliams: Avoid leaning on implicit conversions when avoidable:
```suggestion
    for (size_t i = 0u; i < t.access_list.size(); ++i) {
```

src/monad/rlp/test/decode_encode_transaction.cpp
```diff
@@ -131,7 +158,29 @@ TEST(Rlp_Transaction, EncodeEip2930)
...
+    EXPECT_EQ(decoding.amount, t.amount);
+    EXPECT_EQ(*decoding.to, *t.to);
+    EXPECT_EQ(decoding.sc.r, t.sc.r);
+    EXPECT_EQ(decoding.sc.s, t.sc.s);
+    EXPECT_EQ(*decoding.sc.chain_id, *t.sc.chain_id);
+    EXPECT_EQ(decoding.type, t.type);
+
+    EXPECT_EQ(decoding.access_list.size(), t.access_list.size());
+
+    for (size_t i = 0; i < t.access_list.size(); ++i) {
```
sheapwilliams: Same:
```suggestion
    for (size_t i = 0u; i < t.access_list.size(); ++i) {
```

src/monad/rlp/test/decode_encode_receipt.cpp
```diff
@@ -256,5 +318,24 @@ TEST(Rlp_Receipt, EncodeEip2930Receipt)
...
+    EXPECT_EQ(decoded.type, r.type);
+    EXPECT_EQ(decoded.gas_used, r.gas_used);
+    EXPECT_EQ(decoded.status, r.status);
+
+    // Bloom
+    EXPECT_EQ(decoded.bloom, r.bloom);
+
+    // Log:
+    EXPECT_EQ(decoded.logs.size(), r.logs.size());
+    for (size_t i = 0; i < decoded.logs.size(); ++i) {
```
sheapwilliams:
```suggestion
    for (size_t i = 0u; i < decoded.logs.size(); ++i) {
```

src/monad/rlp/test/decode_encode_receipt.cpp
```diff
@@ -194,7 +237,26 @@ TEST(Rlp_Receipt, EncodeEip1559Receipt)
...
```
sheapwilliams:
```suggestion
    for (size_t i = 0u; i < decoded.logs.size(); ++i) {
```

src/monad/rlp/test/decode_encode_receipt.cpp
```diff
@@ -132,7 +156,26 @@ TEST(Rlp_Receipt, EncodeEip155Receipt)
...
```
sheapwilliams:
```suggestion
    for (size_t i = 0u; i < decoded.logs.size(); ++i) {
```

src/monad/rlp/decode_helpers.cpp
kiblitz: I can only think to `MONAD_ASSERT(type >= c0)` but this already happens in :140. Should we be explicit and write it out again or is there something else you had in mind?

src/monad/rlp/decode_helpers.cpp
kiblitz: Same as https://github.com/monad-crypto/monad/pull/42/commits/51c56d934edc0f5ed759a1c38ad99d57a40610e7#r1113491008

include/monad/rlp/decode.hpp
kiblitz: A bit confused about general structure... what functions (in general) belong in `rlp/util.hpp`? Should just `decode_num_raw` go in `rlp/util.hpp` or should all `decode_*primitive*` go in `rlp/util.hpp`?

jhunsaker: rlp only has two types - strings and lists

`decode_string` and `decode_list` should go in `decode.hpp`
utils needed to implement those two should go in `util.hpp`
converting strings to other types should be in `decode_helpers`

kiblitz: Addressed. Resolve if happy with result.

sheapwilliams: It looks good to me.

PR #43, opened 2023-02-15
Start trait structure for Ethereum hard forks - calculating intrinsic gas

sheapwilliams:
Researched all the forks and EIPs that yielded changes to gas calculations.  Going to keep working execution off of this, but let me know what your thoughts.

### DIFF
--- a/include/monad/core/hard_fork.hpp
+++ b/include/monad/core/hard_fork.hpp
@@ -1,0 +1,113 @@
+#pragma once
+
+#include <monad/config.hpp>
+#include <monad/core/transaction.hpp>
+
+#include <algorithm>
+
+MONAD_NAMESPACE_BEGIN
+
+// https://ethereum.org/en/history/
+
+namespace hard_fork
+{
+    struct frontier
+    {
+        static constexpr auto block_number = 0;
+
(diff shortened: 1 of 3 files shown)


src/monad/core/test/hard_fork.cpp
```diff
@@ -0,0 +1,64 @@
+#include <monad/config.hpp>
+#include <monad/core/hard_fork.hpp>
+
+#include <gtest/gtest.h>
+
+using namespace monad;
+
+TEST(hard_fork, genesis)
```
sheapwilliams: should be `frontier`

include/monad/core/hard_fork.hpp
```diff
@@ -0,0 +1,119 @@
...
+    struct frontier
+    {
+        static constexpr auto block_number = 0;
+
+        // Yellow paper, section 6.2, eqn. 60
+        [[nodiscard]] static constexpr inline auto
+        g_txcreate(Transaction const &) noexcept
+        {
+            return 0;
+        }
```
sheapwilliams: We don't really need this as a function...perhaps a concept for the public facing interface?

PR #44, opened 2023-02-17
:new: Validate execution

sheapwilliams:
This is just the start of the execution work I've been doing.  More is in work, but I think this is reviewable.
- Add ability to validate transaction per the yellow paper, Eq. 62.
- Template the Processor on the execution traits defined by various hard-forks in the network (see comments for references)
- Start to define the state read proxy interface in `fake::Accounts`
-- This will take better shape in time, but is a simple start

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,7 @@ endfunction()
 ############
 
 add_subdirectory(${PROJECT_SOURCE_DIR}/src/monad/core/test)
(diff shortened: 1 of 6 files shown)


src/monad/execution/processor.cpp
```diff
@@ -0,0 +1,8 @@
+#include <monad/execution/processor.hpp>
+
+#include <evmc/evmc.h>
+
+using namespace monad;
+using namespace monad::execution;
+
+
```
sheapwilliams: Oops - this is vestigial.  Will remove.

PR #50, opened 2023-03-14
Transaction processor

sheapwilliams:
This is a piece of some of the execution work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,7 @@ add_library(monad STATIC
     ${PROJECT_SOURCE_DIR}/include/monad/execution/config.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/create_contract_address.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/ethereum/fork_traits.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/execution/evm.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/evmc_host.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/transaction_processor.hpp
     ${PROJECT_SOURCE_DIR}/src/monad/execution/create_contract_address.cpp
--- a/include/monad/execution/evm.hpp
+++ b/include/monad/execution/evm.hpp
@@ -1,0 +1,125 @@
+#pragma once
+
+#include <monad/core/concepts.hpp>
+#include <monad/execution/config.hpp>
+#include <monad/execution/create_contract_address.hpp>
+
+#include <ethash/keccak.hpp>
+
+#include <intx/intx.hpp>
+
+#include <tl/expected.hpp>
+
+MONAD_EXECUTION_NAMESPACE_BEGIN
+
+template <class TState, concepts::fork_traits<TState> TTraits>
+struct Evm
+{
+    TState &state_;
+    address_t new_address_{};
+
+    using result_t = tl::expected<evmc_message, evmc_result>;
+    using new_address_t = tl::expected<address_t, evmc_result>;
+    using unexpected_t = tl::unexpected<evmc_result>;
+
+    Evm(TState &s)
+        : state_{s}
+    {
+    }
+
+    [[nodiscard]] result_t check_sender_balance(evmc_message const &m) noexcept
+    {
+        auto const value = intx::be::load<uint256_t>(m.value);
+        auto const balance =
+            intx::be::load<uint256_t>(state_.get_balance(m.sender));
+        if (balance < value) {
+            return unexpected_t(
+                {.status_code = EVMC_INSUFFICIENT_BALANCE, .gas_left = m.gas});
+        }
+        return {m};
+    }
+
+    inline void
+    transfer_balances(evmc_message const &m, address_t const &to) noexcept
+    {
+        auto const value = intx::be::load<uint256_t>(m.value);
+        auto balance = intx::be::load<uint256_t>(state_.get_balance(m.sender));
+        state_.set_balance(m.sender, balance - value);
+        balance = intx::be::load<uint256_t>(state_.get_balance(to));
+        state_.set_balance(to, balance + value);
+    }
+
+    [[nodiscard]] inline result_t
+    increment_sender_nonce(evmc_message const &m) noexcept
+    {
+        auto const n = state_.get_nonce(m.sender) + 1;
+        if (state_.get_nonce(m.sender) > n) {
+            // Match geth behavior - don't overflow nonce
+            return unexpected_t(
+                {.status_code = EVMC_ARGUMENT_OUT_OF_RANGE, .gas_left = m.gas});
+        }
+        state_.set_nonce(m.sender, n);
+        return {m};
+    }
+
+    [[nodiscard]] inline result_t
+    create_new_contract(evmc_message const &m) noexcept
+    {
+        new_address_ = [&] {
+            if (m.kind == EVMC_CREATE) {
+                return create_contract_address(
+                    m.sender, state_.get_nonce(m.sender));
+            }
+            else if (m.kind == EVMC_CREATE2) {
+                auto const code_hash =
+                    ethash::keccak256(m.input_data, m.input_size);
+                return create2_contract_address(
+                    m.sender, m.create2_salt, code_hash);
+            }
+            MONAD_ASSERT(false);
+            return address_t{};
+        }();
+
+        // Prevent overwriting contracts - EIP-684
+        if (state_.get_nonce(new_address_) != 0 ||
+            state_.get_code_hash(new_address_) != NULL_HASH) {
+            return unexpected_t({.status_code = EVMC_INVALID_INSTRUCTION});
+        }
+
+        state_.create_contract(new_address_);
+
+        return {m};
+    }
+
+    [[nodiscard]] new_address_t
+    make_account_address(evmc_message const &m) noexcept
+    {
+        auto const result =
+            check_sender_balance(m)
+                .and_then([&](auto x) { return increment_sender_nonce(x); })
+                .and_then([&](auto x) { return create_new_contract(x); });
+        if (!result) {
+            return unexpected_t{result.error()};
+        }
+
+        state_.set_nonce(new_address_, TTraits::starting_nonce());
+        transfer_balances(m, new_address_);
+        return {new_address_};
+    }
+
+    [[nodiscard]] evmc_result transfer_call_balances(evmc_message const &m)
+    {
+        if (m.kind != EVMC_DELEGATECALL) {
+            if (auto const result = check_sender_balance(m);
+                !result.has_value()) {
+                return result.error();
+            }
+            else if (m.flags != EVMC_STATIC) {
+                transfer_balances(m, m.recipient);
+            }
+        }
+        return {.status_code = EVMC_SUCCESS};
+    }
+};
+
+MONAD_EXECUTION_NAMESPACE_END
--- a/src/monad/execution/test/CMakeLists.txt
+++ b/src/monad/execution/test/CMakeLists.txt
@@ -4,6 +4,7 @@ add_unit_test(
   SOURCES
   create_contract_address.cpp
   ethereum_traits.cpp
+  evm.cpp
   evmc_host.cpp
   transaction_processor.cpp
   validation.cpp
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -29,6 +30,7 @@ conan_cmake_install(
 )
 
 find_package(RocksDB)
+find_package(tl-expected)
```
zander-xyz: REQUIRED?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -0,0 +1,278 @@
...
+
+// https://ethereum.org/en/history/
+namespace fork_traits
+{
+    struct frontier
+    {
+        static constexpr auto block_number = 0u;
+
+        [[nodiscard]] static constexpr inline auto
+        g_data(Transaction const &t) noexcept
```
zander-xyz, line 20: where is g_data defined?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        [[nodiscard]] static constexpr inline auto starting_nonce() noexcept
+        {
+            return 0;
+        }
+
+        template <class TState>
+        [[nodiscard]] static inline uint64_t
+        get_selfdestruct_refund(TState const &s) noexcept
```
zander-xyz, line 44: this one not marked constexpr?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -0,0 +1,278 @@
...
+            if (r.gas_left >= deploy_cost) {
+                s.set_code(a, {r.output_data, r.output_size});
+                r.gas_left -= deploy_cost;
+                r.create_address = a;
+                return true;
+            }
+            return false;
+        }
+        template <class TState>
+        [[nodiscard]] static constexpr inline bool store_contract_code(
```
zander-xyz, line 74: What is this supposed to return? seems like it always returns false - what's the meaning of that/is that right?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -0,0 +1,278 @@
...
+            TState &s, address_t const &a, evmc_result &r) noexcept
+        {
+            if (r.status_code == EVMC_SUCCESS) {
+                enough_gas_to_store_code(s, a, r);
```
zander-xyz, line 78: discarding result here?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -0,0 +1,278 @@
...
+            }
+            return false;
+        }
+    };
+
+    struct homestead : public frontier
+    {
+        // https://eips.ethereum.org/EIPS/eip-2
+        static constexpr auto block_number = 1'150'000u;
+        [[nodiscard]] static constexpr inline auto
```
zander-xyz: super nit but can we new line in between block_number and the function declaration below?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -0,0 +1,278 @@
...
+
+    // muir_glacier - 9'200'000
+
+    struct berlin : public istanbul
+    {
+        static constexpr auto block_number = 12'244'000u;
+
+        // https://eips.ethereum.org/EIPS/eip-2930
+        [[nodiscard]] static constexpr inline auto
+        g_access_and_storage(Transaction const &t) noexcept
```
zander-xyz, line 230: what's the significance of the g_ prefix?

include/monad/execution/evmc_front_half.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include <ethash/keccak.hpp>
+
+#include <intx/intx.hpp>
+
+#include <tl/expected.hpp>
+
+MONAD_EXECUTION_NAMESPACE_BEGIN
+
+template <concepts::fork_traits TTraits, class TState>
```
zander-xyz: is this more aptly named `front_half_fork_fraits` (not married to the actual name)? it's longer but also less confusing because there's a `fork_traits` namespace already, and avoids the question as to why the it has a seemingly different definition than the actual traits structures defined within it

include/monad/execution/evmc_front_half.hpp
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        _state.set_nonce(_new_address, TTraits::starting_nonce());
+        transfer_balances(m, _new_address);
+        return {_new_address};
+    }
+
+private:
+    TState &_state;
+    address_t _new_address{};
```
zander-xyz: Any strong opinion against putting these variables at the top of the class?

Also, taking slight preference towards `state_` instead of `_state` since that's what I see already existing in the repo.

include/monad/execution/evmc_host.hpp
```diff
@@ -0,0 +1,183 @@
...
+    {
+        return _state.get_block_hash(block_number);
+    };
+
+    virtual void inline emit_log(
+        [[maybe_unused]] address_t const &addr,
+        [[maybe_unused]] const uint8_t *data, [[maybe_unused]] size_t data_size,
+        [[maybe_unused]] bytes32_t const topics[],
+        [[maybe_unused]] size_t num_topics) noexcept override
+    {
```
zander-xyz, line 211: Is this supposed to be marked as TODO?

include/monad/execution/transaction_processor.hpp
```diff
@@ -0,0 +1,179 @@
...
+struct TransactionProcessor
+{
+    enum class Status
+    {
+        SUCCESS,
+        INSUFFICIENT_BALANCE,
+        INVALID_GAS_LIMIT,
+        BAD_NONCE,
+        DEPLOYED_CODE,
+    };
```
zander-xyz: Are we satisfied with this being an enum class instead of a `tl::expected<void, Status>`? Not a huge deal either way - more just asking for the sake of consistency.

include/monad/execution/transaction_processor.hpp
```diff
@@ -0,0 +1,179 @@
...
+            .status = r.status_code == EVMC_SUCCESS ? 1u : 0u,
+            .gas_used = t.gas_limit - gas_remaining,
+            .type = t.type,
+            .logs = std::move(s.logs())};
+        return result;
+    }
+
+    template <class TState>
+    Status
+    validate(TState &state, Transaction const &t, uint64_t base_fee_per_gas)
```
zander-xyz: can the `state` here be marked const?

include/monad/execution/transaction_processor.hpp
```diff
@@ -0,0 +1,179 @@
...
+            intx::be::load<uint256_t>(s.get_balance(b.beneficiary));
+        s.set_balance(
+            b.beneficiary,
+            bene_balance + (t.per_gas_cost(b.base_fee_per_gas.value_or(0)) *
+                            (t.gas_limit - gas_remaining)));
+        const auto sender_balance =
+            intx::be::load<uint256_t>(s.get_balance(*t.from));
+        s.set_balance(
+            *t.from,
+            sender_balance + (t.per_gas_cost(b.base_fee_per_gas.value_or(0)) *
```
zander-xyz: does the compiler somehow optimize this so that the `t.per_gas_cost` doesn't get calculated twice?

include/monad/execution/evmc_host.hpp
```diff
@@ -0,0 +1,183 @@
...
+        address_t const &a, bytes32_t const &key,
+        bytes32_t const &value) noexcept override
+    {
+        return _state.set_storage(a, key, value);
+    }
+
+    virtual inline uint256be
+    get_balance(address_t const &a) const noexcept override
+    {
+        return _state.get_balance(a);
```
zander-xyz: does it make sense to do the intx::be::load here instead (and maybe store the result so you don't have to do it every time)? it seems like everywhere you call this function, you immediately pass it into intx::be::load anyways

include/monad/execution/evmc_host.hpp
sheapwilliams: Yes - I'll just implement it.  It just passes the info to the state.

include/monad/execution/evmc_host.hpp
sheapwilliams: I can't change the signature of that function - it is part of the `evmc::HostInterface`, so IIWII.

But yes, one thing we should change if we get away from evmc.

include/monad/execution/transaction_processor.hpp
sheapwilliams: Yup - good call.

include/monad/execution/transaction_processor.hpp
sheapwilliams: It should get an static single assignment in llvm ir, but I can make it an l-value just to ensure it.

include/monad/execution/transaction_processor.hpp
sheapwilliams: I wish.  `void` isn't a regular type, unfortunately.  So I can't take the `.value()` of that expected.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: Yup - exactly.  Frontier silently deployed an empty contract if there was an out-of-gas error.  They fixed that in homestead, but that is correct behavior for frontier.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: the bool is inverted success.  I can invert it to just indicate success.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: Screenshot 2023-03-15 at 5 29 33 PM

It is the first summation of Eqn. 60 in the yellow paper.  I can mark it with Eqn. 60 as a reference?

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: gas. The YP uses _g_ to refer to most gas calculations and gas constants.

include/monad/execution/evmc_front_half.hpp
sheapwilliams: The fork traits will be part of the back half too - specifically precompiles change between hard forks.  Also with the DAO hack, I think it will be used in the block processor...

I think I just haven't finished the concept yet is all - they should have the same *public* definitions.  It was mostly just playing with the idea to make sure the fake I programmed is correct.  Let's chat on zoom to figure out the best way to move this along when you get a chance.

include/monad/execution/evmc_front_half.hpp
sheapwilliams: None whatsoever.  I'll change to this.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: But to answer your question - for frontier, storing contract code never fails - it just runs out of gas, and consequently deploys an empty contract. 🤷

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: yes please!

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
 list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
 
 include(${PROJECT_SOURCE_DIR}/conan/cmake-conan/conan.cmake)
+include(${CMAKE_BINARY_DIR}/conan_paths.cmake)
```
zander-xyz: leaving comment here to get rid of this

include/monad/execution/evmc_front_half.hpp
sheapwilliams: I left the fake with under before so it screams.  All others changed to match the existing style.

include/monad/execution/evmc_front_half.hpp
sheapwilliams: https://github.com/sheapwilliams/monad/tree/complete-traits
I completed the traits, which requires the state to be injected everywhere.  But I think that is just how things are here - everything needs to interact with the state.  If you are ok with this, I'll just rebase into this PR.

sheapwilliams: Just re-pushed and completed the concept.

PR #53, opened 2023-04-03
Map empty address in txn decoding

kiblitz:
Bug fix: txn recipient is contract creation


include/monad/rlp/decode_helpers.hpp
```diff
@@ -36,6 +36,21 @@ decode_address(address_t &address, byte_string_view const enc)
     return decode_byte_array<20>(address.bytes, enc);
```
sheapwilliams, line 36: Do we need regular `decode_address` anymore?  Shouldn't we just always decode an optional address?

include/monad/rlp/decode_helpers.hpp
```diff
@@ -36,6 +36,21 @@ decode_address(address_t &address, byte_string_view const enc)
...
 }
 
+inline byte_string_view
+decode_address_optional(std::optional<address_t> &address, byte_string_view const enc)
+{
+    constexpr auto address_size = 20u;
```
sheapwilliams: This can just be `sizeof(address_t)` if you want to eliminate a bit of magic.

include/monad/rlp/decode_helpers.hpp
kiblitz: not all address decodings are into optional fields

sheapwilliams: I see...receipt logs and access lists.  Ok. :+1:

PR #56, opened 2023-04-05
Transaction type envelope RLP

kiblitz:
https://eips.ethereum.org/EIPS/eip-2718

Previously, we had `txn_type || RLP_list([..._txn_data_...])`. Per eip-2718, it should be `RLP_str(txn_type || RLP([..._txn_data_...]))`.

For testing, added real block `14000000` which was created post eip-2718.

eip-2718 also applies to receipts. Similarly, it takes the form of `txn_type || receipt_encoding`.


include/monad/core/transaction.hpp
kiblitz, line 1: https://eips.ethereum.org/EIPS/eip-2930 has txn type 1; https://eips.ethereum.org/EIPS/eip-1559 has txn type 2

src/monad/rlp/decode_helpers.cpp
```diff
@@ -165,97 +165,117 @@ decode_account(Account &acc, bytes32_t &code_root, byte_string_view const enc)
...
+    return rest_of_enc;
+}
+
+byte_string_view
+decode_transaction(Transaction &txn, byte_string_view const enc)
+{
+    MONAD_ASSERT(enc.size() > 0);
+
+    const uint8_t &first = enc[0];
+    if (first < 0xc0) // eip 2718 - typed transaction envelope
```
sheapwilliams, line 247: Technically, if it is less than 0x7f it is a typed transaction per EIP-2718.  We should probably assert if it is between 0x80 and 0xbf...that is a serious mistake.

src/monad/rlp/decode_helpers.cpp
```diff
@@ -265,6 +285,38 @@ byte_string_view decode_receipt(Receipt &receipt, byte_string_view const enc)
     return rest_of_enc;
 }
 
+byte_string_view decode_receipt(Receipt &receipt, byte_string_view const enc)
+{
+    MONAD_ASSERT(enc.size() > 0);
+
+    const uint8_t &first = enc[0];
+    receipt.type = Transaction::Type::eip155;
+    if (first < 0xc0) // eip 2718 - typed transaction envelope
```
sheapwilliams, line 295: ditto here.

src/monad/rlp/decode_helpers.cpp
kiblitz: `>= 0xc0` implies that this is a list RLP encoding and thus a legacy transaction. Otherwise, it is a string RLP encoding. The transaction type is indicated by `type` in #L253. The assertion that the type is either invalid or not handled is in the `default` case of the switch statement (`MONAD_ASSERT(false)`).

src/monad/rlp/decode_helpers.cpp
kiblitz: > `>= 0xc0` implies that this is a list RLP encoding and thus a legacy transaction. Otherwise, it is a string RLP encoding. The transaction type is indicated by `type` in #L253. The assertion that the type is either invalid or not handled is in the `default` case of the switch statement (`MONAD_ASSERT(false)`).

src/monad/rlp/decode_helpers.cpp
sheapwilliams: I see now.  :+1:

PR #59, opened 2023-04-05
Monad Run-loop Interface 

tongzhi1998:
Run-loop interface 
Rebased on current main
@kiblitz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ add_library(monad STATIC
     ${PROJECT_SOURCE_DIR}/include/monad/execution/evm.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/evmc_host.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/execution_model.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/execution/replay_block.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/transaction_processor.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/transaction_processor_data.hpp
     ${PROJECT_SOURCE_DIR}/src/monad/execution/create_contract_address.cpp
--- a/include/monad/execution/replay_block.hpp
+++ b/include/monad/execution/replay_block.hpp
@@ -1,0 +1,157 @@
+#pragma once
+
+#include <monad/db/block_db.hpp>
+#include <monad/db/state_db.hpp>
+#include <monad/execution/block_processor_interface.hpp>
+#include <monad/execution/config.hpp>
+#include <monad/execution/ethereum/fork_traits.hpp>
+#include <monad/execution/evm.hpp>
+#include <monad/execution/transaction_processor_data.hpp>
+#include <monad/trie/state_trie_interface.hpp>
+
+#include <filesystem>
+#include <iostream>
+
+MONAD_EXECUTION_NAMESPACE_BEGIN
+
+template <
+    class TState, class TBlockDb, class TBlockProcessor, class TStateTrie,
+    class TExecution>
+class ReplayBlock final
+{
+public:
+    enum class Status
+    {
+        END_OF_BLOCKDB, // success
+        COMPLETE, // success
+        START_BLOCK_NUMBER_OUTSIDE_DB,
+        INVALID_END_BLOCK_NUMBER,
+        ERROR // other error
+    };
+
+    struct Result
+    {
+        Status status;
+        block_num_t finished_block_number;
+    };
+
+    template <concepts::fork_traits<TState> TTraits>
+    [[nodiscard]] Result run_fork(
+        TState &state, TBlockDb const &block_db, TStateTrie &state_trie,
+        Block &block, block_num_t curr_block_number,
+        std::optional<block_num_t> until_block_number = std::nullopt)
+    {
+        using fiber_data_t = execution::TransactionProcessorFiberData<
+            TState,
+            TTraits,
+            execution::TransactionProcessor<TState, TTraits>,
+            execution::Evm<TState, TTraits>,
+            TExecution>;
+        auto loop_until_block_number = loop_until<TTraits>(until_block_number);
+        for (; curr_block_number < loop_until_block_number;
+             ++curr_block_number) {
+
+            if (!block_db.get(curr_block_number, block)) {
+                return Result{Status::END_OF_BLOCKDB, curr_block_number - 1u};
+            }
+
+            TBlockProcessor block_processor{};
+            [[maybe_unused]] auto receipts =
+                block_processor.template execute<TTraits, fiber_data_t>(
+                    block, state);
+
+            state_trie.incremental_update(state);
+            state.commit();
+        }
+
+        if (!until_block_number.has_value() ||
+            until_block_number.value() > curr_block_number) {
+            return run_fork<typename TTraits::next_fork_t>(
+                state,
+                block_db,
+                state_trie,
+                block,
+                curr_block_number,
+                until_block_number);
+        }
+        else {
+            return Result{Status::COMPLETE, curr_block_number - 1u};
+        }
+    }
+
+    [[nodiscard]] Result
+    run(TState &state, TBlockDb const &block_db, block_num_t start_block_number,
+        std::optional<block_num_t> until_block_number = std::nullopt)
+    {
+        TStateTrie state_trie{};
+        Block block{};
+
+        // invalid input
+        if (until_block_number.has_value() &&
+            (until_block_number <= start_block_number)) {
+            return Result{Status::INVALID_END_BLOCK_NUMBER, 0u};
+        }
+
+        if (!block_db.get(start_block_number, block)) {
+            return Result{Status::START_BLOCK_NUMBER_OUTSIDE_DB, 0u};
+        }
+
+        using fork_function_t = Result (ReplayBlock::*)(
+            TState &,
+            TBlockDb const &,
+            TStateTrie &,
+            Block &,
+            block_num_t,
+            std::optional<block_num_t>);
+
+        fork_function_t fork_functions[] = {
+            &ReplayBlock::run_fork<fork_traits::frontier>,
+            &ReplayBlock::run_fork<fork_traits::homestead>,
+            &ReplayBlock::run_fork<fork_traits::spurious_dragon>,
+            &ReplayBlock::run_fork<fork_traits::byzantium>,
+            &ReplayBlock::run_fork<fork_traits::istanbul>,
+            &ReplayBlock::run_fork<fork_traits::berlin>,
+            &ReplayBlock::run_fork<fork_traits::london>};
+
+        block_num_t fork_blocks[] = {
+            fork_traits::frontier::last_block_number,
+            fork_traits::homestead::last_block_number,
+            fork_traits::spurious_dragon::last_block_number,
+            fork_traits::byzantium::last_block_number,
+            fork_traits::istanbul::last_block_number,
+            fork_traits::berlin::last_block_number,
+            fork_traits::london::last_block_number};
+
+        uint64_t num_forks = sizeof(fork_functions) / sizeof(fork_function_t);
+
+        for (auto i = 0u; i < num_forks; i++) {
+            if (start_block_number <= fork_blocks[i]) {
+                return (this->*fork_functions[i])(
+                    state,
+                    block_db,
+                    state_trie,
+                    block,
+                    start_block_number,
+                    until_block_number);
+            }
+        }
+
+        return Result{Status::ERROR, 0u};
+    }
+
+    template <concepts::fork_traits<TState> TTraits>
+    [[nodiscard]] constexpr block_num_t
+    loop_until(std::optional<block_num_t> block_until)
+    {
+        if (!block_until.has_value()) {
+            return TTraits::last_block_number + 1u;
+        }
+        else {
+            return std::min(
+                block_until.value(),
+                static_cast<uint64_t>(TTraits::last_block_number + 1u));
+        }
+    }
+};
+
+MONAD_EXECUTION_NAMESPACE_END
--- a/src/monad/execution/test/CMakeLists.txt
+++ b/src/monad/execution/test/CMakeLists.txt
@@ -7,6 +7,7 @@ add_unit_test(
   ethereum_traits.cpp
   evm.cpp
   evmc_host.cpp
+  replay_block.cpp
   transaction_processor.cpp
   transaction_processor_data_succeed.cpp
   transaction_processor_data_fail_validation.cpp
--- a/src/monad/execution/test/replay_block.cpp
+++ b/src/monad/execution/test/replay_block.cpp
@@ -1,0 +1,175 @@
+#include <gtest/gtest.h>
+
+#include <monad/config.hpp>
+
+#include <monad/execution/replay_block.hpp>
+
+#include <monad/execution/block_processor_interface.hpp>
+#include <monad/execution/config.hpp>
+#include <monad/execution/execution_model.hpp>
+#include <monad/execution/test/fakes.hpp>
+#include <monad/execution/transaction_processor.hpp>
+#include <monad/execution/transaction_processor_data.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace db::fake
+{
+    class BlockDb : IBlockDb
+    {
+    public:
+        block_num_t block_number_threshold;
+
+        bool
+        get([[maybe_unused]] block_num_t block_number,
+            [[maybe_unused]] Block &block) const override
+        {
+            if (block_number < block_number_threshold) {
+                return true;
+            }
+            else {
+                return false;
+            }
+        }
+    };
+}
+
+namespace execution::fake
+{
+    class BlockProcessor final
+        : public execution::IBlockProcessor<execution::BoostFiberExecution>
+    {
+    public:
+        template <
+            concepts::fork_traits<execution::fake::State> TTraits,
+            class TFiberData>
+        std::vector<Receipt> execute(
+            [[maybe_unused]] const Block &block,
+            [[maybe_unused]] execution::fake::State &state)
+        {
+            return {};
+        }
+    };
+}
+
+namespace trie::fake
+{
+    class StateTrie final : public trie::IStateTrie<execution::fake::State>
+    {
+    public:
+        void incremental_update(
+            [[maybe_unused]] execution::fake::State &state) override
+        {
+            return;
+        }
+    };
+}
+
+using state_t = execution::fake::State;
+using traits_t = execution::fake::traits<state_t>;
+using block_db_t = db::fake::BlockDb;
+using block_processor_t = execution::fake::BlockProcessor;
+using state_trie_t = trie::fake::StateTrie;
+
+using replay_block_t = execution::ReplayBlock<
+    state_t, block_db_t, block_processor_t, state_trie_t,
+    execution::BoostFiberExecution>;
+
+TEST(ReplayBlock, start_block_number_outside_db)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = 0u;
+    state_t state;
+    replay_block_t replay_block;
+    auto [status, finished_block_number] = replay_block.run(state, block_db, 0);
+    EXPECT_EQ(status, replay_block_t::Status::START_BLOCK_NUMBER_OUTSIDE_DB);
+    EXPECT_EQ(finished_block_number, 0u);
+}
+
+TEST(ReplayBlock, invalid_end_block_number)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = 1'000u;
+    state_t state;
+    replay_block_t replay_block;
+    auto [status, finished_block_number] =
+        replay_block.run(state, block_db, 100, 100);
+    EXPECT_EQ(status, replay_block_t::Status::INVALID_END_BLOCK_NUMBER);
+    EXPECT_EQ(finished_block_number, 0u);
+}
+
+TEST(ReplayBlock, invalid_end_block_number_zero)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = 1'000u;
+    state_t state;
+    replay_block_t replay_block;
+    auto [status, finished_block_number] =
+        replay_block.run(state, block_db, 0u, 0u);
+    EXPECT_EQ(status, replay_block_t::Status::INVALID_END_BLOCK_NUMBER);
+    EXPECT_EQ(finished_block_number, 0u);
+}
+
+TEST(ReplayBlock, one_block)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = 1'000u;
+    state_t state;
+    replay_block_t replay_block;
+    auto [status, finished_block_number] =
+        replay_block.run(state, block_db, 100, 101);
+    EXPECT_EQ(status, replay_block_t::Status::COMPLETE);
+    EXPECT_EQ(finished_block_number, 100u);
+}
+
+TEST(ReplayBlock, frontier_run_from_zero)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = 1'234u;
+    state_t state;
+    replay_block_t replay_block;
+    auto [status, finished_block_number] = replay_block.run(state, block_db, 0);
+    EXPECT_EQ(status, replay_block_t::Status::END_OF_BLOCKDB);
+    EXPECT_EQ(finished_block_number, 1'233u);
+}
+
+TEST(ReplayBlock, frontier_to_homestead)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold =
+        fork_traits::frontier::last_block_number + 10;
+    state_t state;
+    replay_block_t replay_block;
+    auto [status, finished_block_number] = replay_block.run(
+        state, block_db, fork_traits::frontier::last_block_number - 10);
+    EXPECT_EQ(status, replay_block_t::Status::END_OF_BLOCKDB);
+    EXPECT_EQ(finished_block_number, 1'150'008u);
+}
+
+TEST(ReplayBlock, berlin_to_london)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold =
+        fork_traits::berlin::last_block_number + 10;
+    state_t state;
+    replay_block_t replay_block;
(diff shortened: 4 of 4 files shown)


include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+    END_OF_BLOCKDB, // success
+    REACH_END_BLOCK_NUMBER, // success
+    ERROR // need to specify the error later
+};
+
+template <class TState, class TBlockDb, class TBlockProcessor, class TStateTrie, class TExecution>
+class Blockchain final
+{
+public:
+    // this will be used for initialization (if we don't run from block 0)
```
kiblitz: Feels a bit clunky, especially since `transaction_processor_*_t` and `evm_*_t` are both only used for `fiber_data_*_t`. How do you feel about this?

```
template <concepts::fork_traits<TState> TTraits>
using fiber_data_t = execution::TransactionProcessorFiberData<TState, TTraits, execution::TransactionProcessor<TState, TTraits>, execution::Evm<TState, TTraits>, TExecution>;
```

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        // invalid input
+        if (until_block_num.has_value() &&
+            (until_block_num < start_block_num)) {
+            return RunLoopStatus::ERROR;
+        }
+
+        TStateTrie state_trie{};
+        Block block{};
+        if (start_block_num < fork_traits::frontier::block_number) {
```
kiblitz: To reduce redundant code, maybe we can assign a lambda in the `if` statements and just call the lambda with all the same parameters at the end?

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+            using next_transaction_processor_t = execution::
+                TransactionProcessor<TState, typename TTraits::next_fork_t>;
+            using next_evm_t =
+                execution::Evm<TState, typename TTraits::next_fork_t>;
+            using next_fiber_data_t = execution::TransactionProcessorFiberData<
+                TState,
+                typename TTraits::next_fork_t,
+                next_transaction_processor_t,
+                next_evm_t, TExecution>;
+            return run_fork<typename TTraits::next_fork_t, next_fiber_data_t>(
```
kiblitz: Using https://github.com/monad-crypto/monad/pull/59/files#r1159189816 I think you can just do `fiber_data_t<TTraits::next_fork_t>`

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -246,7 +281,12 @@ namespace fork_traits
 
     struct london : public berlin
     {
-        static constexpr auto block_number = 12'965'000;
+        // TODO: create a fake end state that satisfies TTraits
+        using next_fork_t = london;
+        static constexpr auto block_number =
```
kiblitz: Wherever `block_num_t` is defined, probably just define max as `-1` (underflow). Alternatively this could be an `optional`.

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,70 @@
...
+bool BlockDb::get(block_num_t const num, Block &block) const
+{
+    auto const key = std::to_string(num);
+    auto const result = db_.get(key.c_str());
+    if (!result.has_value()) {
+        return false;
+    }
+    byte_string_view view{
+        reinterpret_cast<uint8_t const *>(result->data()), result->length()};
+    size_t brotli_size = result->size() * 100; // TODO
```
kiblitz: Need to dynamically resize. i.e.
```
auto compression_factor = 100;
while (brotli_result != BROTLI_DECODER_RESULT_SUCCESS)
{
    ...
    brotli_result = BrotliDecoderDecompress(...);
    ...
    compression_factor *= 10;
}
```

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,70 @@
...
+        rlp::decode_block(block, view2);
+
+    return true;
+}
+
+void BlockDb::upsert(const block_num_t num, const Block &block)
+{
+    auto const key = std::to_string(num);
+    // TODO: encode_block is not implemented currently!
+    byte_string bytes = rlp::encode_block(block);
```
kiblitz: I feel like either we implement `encode_block` here in this PR or we just don't include `upsert` or `remove`. Leaning towards latter since I don't think we actually use those functions in this PR.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+            return run_fork<fork_traits::homestead, fiber_data_homestead_t>(
+                state,
+                block_db,
+                state_trie,
+                block,
+                start_block_num,
+                until_block_num);
+        }
+        else if (start_block_num < fork_traits::spurious_dragon::block_number) {
+            run_fork<
```
kiblitz: return

include/monad/blockchain/eth_replay_blockchain.hpp
kiblitz: sorry, meant function pointer

include/monad/blockchain/eth_replay_blockchain.hpp
sheapwilliams: `[[nodiscard]]` attribute can help find these sorts of errors.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+    std::string get_finished_fork_name() const
+    {
+        return this->finished_fork_name;
+    }
+
+private:
+    template <concepts::fork_traits<TState> TTraits, class TFiberData>
+    constexpr std::optional<block_num_t>
+    loop_until(std::optional<block_num_t> block_until)
```
sheapwilliams: similar...`[[nodiscard]]` throughout.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+        return RunLoopStatus::ERROR;
+    }
+
+    block_num_t get_finished_block_num() const
+    {
+        return this->finished_block_num;
+    }
+    std::string get_finished_fork_name() const
+    {
+        return this->finished_fork_name;
```
sheapwilliams: Don't need `this->`.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+        }
+        else {
+            return std::min(
+                block_until.value(),
+                static_cast<uint64_t>(TTraits::block_number));
+        }
+    }
+
+    std::string finished_fork_name = "";
+    block_num_t finished_block_num = 0u;
```
sheapwilliams: Suffix member identifiers with underscore to help readability throughout.

sheapwilliams: Also, it is slightly more idiomatic to use braced initialization unless there is a reason not to.
```suggestion
    block_num_t finished_block_num_ {};
```

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        return this->finished_block_num;
+    }
+    std::string get_finished_fork_name() const
+    {
+        return this->finished_fork_name;
+    }
+
+private:
+    template <concepts::fork_traits<TState> TTraits, class TFiberData>
```
sheapwilliams: These templates parameters are not used on the function...let's clean up all unused parameters.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+        std::optional<block_num_t> until_block_num = std::nullopt)
+    {
+        // invalid input
+        if (until_block_num.has_value() &&
+            (until_block_num < start_block_num)) {
+            return RunLoopStatus::ERROR;
+        }
+
+        TStateTrie state_trie{};
+        Block block{};
```
sheapwilliams: This can't be right...empty block?

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,243 @@
...
+    using transaction_processor_london_t =
+        execution::TransactionProcessor<TState, fork_traits::london>;
+    using evm_london_t = execution::Evm<TState, fork_traits::london>;
+    using fiber_data_london_t = execution::TransactionProcessorFiberData<
+        TState, fork_traits::london, transaction_processor_london_t,
+        evm_london_t, TExecution>;
+
+    template <concepts::fork_traits<TState> TTraits, class TFiberData>
+    RunLoopStatus run_fork(
+        TState &state, const TBlockDb &block_db, TStateTrie &state_trie,
```
sheapwilliams: make sure these files are formatted...AFAICT, the style in this codebase is east const.

include/monad/blockchain/eth_replay_blockchain.hpp
sheapwilliams: These probably shouldn't be defined in the Blockchain structure - Blockchain should ideally be agnostic of Ethereum vs. Monad.  Let's try to pull these out into a separate header, and see if we can't generalize the next fork logic here.

CMakeLists.txt
```diff
@@ -89,18 +98,25 @@ add_library(monad STATIC
...
     # execution
+    ${PROJECT_SOURCE_DIR}/include/monad/execution/block_processor_interface.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/config.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/create_contract_address.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/ethereum/fork_traits.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/evm.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/evmc_host.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/execution/execution_model.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/execution/transaction_processor.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/execution/transaction_processor_data.hpp
```
sheapwilliams, line 115: Many of these additions were bugs from my previous PRs.  Make the ones that aren't specifically part of this PR a separate commit.  They can stay in this PR tho.

include/monad/blockchain/eth_replay_blockchain.hpp
sheapwilliams: Also - why have getters?  If the answer is "for testing", let's chat about better ways to do that.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -86,8 +109,12 @@ namespace fork_traits
 
     struct homestead : public frontier
     {
+
+        using next_fork_t = spurious_dragon;
+
         // https://eips.ethereum.org/EIPS/eip-2
-        static constexpr auto block_number = 1'150'000u;
+        static constexpr auto block_number = 2'675'000u;
```
sheapwilliams: If we are going to change the semantic to be the last valid block number, then let's change the identifier to match that semantic too.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -3,17 +3,40 @@
...
 MONAD_NAMESPACE_BEGIN
 
 // https://ethereum.org/en/history/
 namespace fork_traits
 {
+    template <class T>
+    constexpr std::optional<block_num_t> fork_block_num()
+    {
+        return T::block_num;
+    }
```
sheapwilliams: Unnecessary?  And wrong?  The identifier is named `block_number`, not `block_num`, right?

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: Also, let's change the logic to be "last valid block", not 1 past the last valid block.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: I think `std::numeric_limits<int>::max` is the right thing. It should run forever if there are no new fork definitions.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -277,6 +317,8 @@ namespace fork_traits
     };
 
     // paris - 15'537'394
+
+    // upcoming shanghai?
```
sheapwilliams: They are calling it Shapella (Shanghai/Capella), and it is programmed for 12 Apr!  Exciting.

But, until we have a block number, let's remove the comment.

include/monad/execution/test/fakes.hpp
```diff
@@ -127,6 +129,7 @@ namespace fake
         inline int current_txn() { return _current_txn; }
 
         inline State get_copy() { return State(*this); }
+        void commit() { return; } // for compliance with the state interface (commit changes)
```
sheapwilliams: Let's try to reduce the number of extra comments.

src/monad/blockchain/test/blockchain.cpp
```diff
@@ -0,0 +1,110 @@
...
+        0u); // 0 represents not runned, will change to some status code later
+}
+
+TEST(Blockchain, one_block)
+{
+    block_db_t block_db;
+    block_db.block_num_threshold = 1'000u;
+    state_t state;
+    blockchain_t blockchain;
+    auto status = blockchain.run(state, block_db, 100, 100);
```
sheapwilliams: [100..100] should be an invalid range, I'd think? We should consider using half-open ranges to align with ... well, everything in C++.  We can discuss in zoom.

src/monad/blockchain/test/blockchain.cpp
```diff
@@ -0,0 +1,110 @@
...
+    EXPECT_EQ(status, RunLoopStatus::REACH_END_BLOCK_NUMBER);
```
sheapwilliams: [nit] Let's pick a more fluent status for this one, something along the lines of: `Blockchain::Status::Complete`, or something.

src/monad/blockchain/test/blockchain.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+    block_db_t block_db;
+    block_db.block_num_threshold = fork_traits::frontier::block_number + 10;
+    state_t state;
+    blockchain_t blockchain;
+    auto status = blockchain.run(
+        state, block_db, fork_traits::frontier::block_number - 10);
+    EXPECT_EQ(status, RunLoopStatus::END_OF_BLOCKDB);
+    EXPECT_EQ(blockchain.get_finished_block_num(), 1'150'009u);
+    EXPECT_EQ(blockchain.get_finished_fork_name(), "homestead");
```
sheapwilliams: I don't get the purpose of the name.  Also, if the member `block_num` is only for statusing the result, then let's make it part of the result instead of a member.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+        END_OF_BLOCKDB, // success
+        COMPLETE, // success
+        START_BLOCK_NUMBER_OUTSIDE_DB,
+        INVALID_END_BLOCK_NUMBER,
+        ERROR // other error
+    };
+
+    template <concepts::fork_traits<TState> TTraits>
+    [[nodiscard]] std::pair<Status, block_num_t> run_fork(
```
kiblitz: I personally think `struct` is preferable to `std::pair` since field names provide context.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,155 @@
...
+private:
+    template <concepts::fork_traits<TState> TTraits>
+    [[nodiscard]] constexpr block_num_t
+    loop_until(std::optional<block_num_t> block_until)
+    {
+        if (!block_until.has_value()) {
+            return TTraits::end_block_number;
+        }
+        else if (block_until.value() == 0) {
+            return 0u;
```
kiblitz: Semantics are inconsistent. In any other case, providing `x` means loop until `x` exclusive; but providing `0` means loop until `0` inclusive.

tongzhi1998: I agree, but I actually did this intentionally since we have checks for start <= end, return false. So if end if 0, then it will fail anyways. The reason to do this hack is that currently we set end of london to be uint_max. 
I actually decided to change the end of london to be uint_max - 1u, so now we don't need to worry about this anymore

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -246,7 +267,10 @@ namespace fork_traits
 
     struct london : public berlin
     {
-        static constexpr auto block_number = 12'965'000;
+        using next_fork_t = london;
```
kiblitz: Should `next_fork_t` be `london`? Also, any reason why `std::numeric_limits<uint64_t>::max() - 1u` instead of `std::numeric_limits<uint64_t>::max()`?

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,39 @@
...
+    size_t brotli_size = result->size() * 100; // TODO
+    byte_string brotli_buffer;
+    brotli_buffer.resize(brotli_size);
+    auto const brotli_result = BrotliDecoderDecompress(
+        view.size(), view.data(), &brotli_size, brotli_buffer.data());
+    brotli_buffer.resize(brotli_size);
+    MONAD_ASSERT(brotli_result == BROTLI_DECODER_RESULT_SUCCESS);
+    byte_string_view view2{brotli_buffer};
+
+    // TODO: tzhi: we don't have means of checking decoding_result
```
kiblitz: In our implementation, `decode_block(...)` returns the rest of the encoding that hasn't been decoded. You can assert that the value is the empty encoding.

include/monad/execution/ethereum/fork_traits.hpp
tongzhi1998: Hm... next_fork_t probably should be something else (a fake end_fork). We can discuss it tmr. For the limit, `loop_until(std::optional<block_num_t> block_until)
    {
        if (!block_until.has_value()) {
            return TTraits::end_block_number + 1u;
        }
        else {
            return std::min(
                block_until.value(),
                static_cast<uint64_t>(TTraits::end_block_number + 1u));
        }
    }`
There is a +1 here, otherwise it would overflow.

include/monad/db/block_db.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <monad/db/block_db_interface.hpp>
+#include <monad/db/file_db.hpp>
+
+#include <filesystem>
+#include <memory>
+
+MONAD_NAMESPACE_BEGIN
```
sheapwilliams: I think this namespace should match the directory structure:

```
MONAD_DB_NAMESPACE_BEGIN
```

include/monad/db/block_db_interface.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <monad/config.hpp>
+#include <monad/core/block.hpp>
+
+MONAD_NAMESPACE_BEGIN
```
sheapwilliams: Same throughout...

src/monad/db/block_db.cpp
sheapwilliams, line 1: While lifted from *place who shall not be named*, it needs some tests I think.

include/monad/db/state_db.hpp
```diff
@@ -24,7 +25,7 @@ namespace rocksdb
 
 MONAD_NAMESPACE_BEGIN
 
-class StateDb final
+class StateDb final : IState
```
sheapwilliams: You added an IState interface, but there is still no implementation of `commit`, making this class unable to be instantiated.

Let's get together over zoom - I don't think the StateDB is an IState.  I think an IState will use a StateDB to persist changes.

include/monad/core/concepts.hpp
```diff
@@ -14,7 +14,7 @@ namespace concepts
     {
         { T::intrinsic_gas(t) } -> std::convertible_to<uint64_t>;
         { T::starting_nonce() } -> std::convertible_to<uint64_t>;
-        { T::block_number } -> std::convertible_to<uint64_t>;
+        { T::end_block_number } -> std::convertible_to<uint64_t>;
```
sheapwilliams: [nit] "end" has scary semantics.  Let's use the language from [EIP-6122](https://eips.ethereum.org/EIPS/eip-6122), which is `first_block_number` and `last_block_number`.

include/monad/execution/config.hpp
sheapwilliams, line 1: I don't think we need this formatting change here in this commit...Why touch this file?

include/monad/core/concepts.hpp
sheapwilliams, line 1: The concept should also require the `next_fork_t` type for any fork trait.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: max is probably just fine for unsigned.  `next_fork_t` is trickier tho...need some way to end the recursion, and the "end this recursion" needs to meet the fork traits concept.  If not `london`, then it needs to be some sentinel type that recurses to itself...and isn't that just the same thing as `london`?

include/monad/execution/test/fakes.hpp
sheapwilliams, line 1: I think the fake changes belong with the test code that animates them, not the fork traits.  Also, the fork traits have test code that should have been updated with this commit.

include/monad/blockchain/eth_replay_blockchain.hpp
sheapwilliams: It would be best to name the fields if possible.  Better than `.first` and `.second`.  You often destructure them into named values in testing anyway - just make that part of the return type, I think.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,152 @@
...
+        ERROR // other error
+    };
+
+    template <concepts::fork_traits<TState> TTraits>
+    [[nodiscard]] std::pair<Status, block_num_t> run_fork(
+        TState &state, TBlockDb const &block_db, TStateTrie &state_trie,
+        Block &block, block_num_t curr_block_number,
+        std::optional<block_num_t> until_block_number = std::nullopt)
+    {
+        using fiber_data_t = execution::TransactionProcessorFiberData<
```
sheapwilliams: Any reason you didn't invert this?

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,152 @@
...
+                    block,
+                    start_block_number,
+                    until_block_number);
+            }
+        }
+
+        return std::make_pair(Status::ERROR, 0u);
+    }
+
+private:
```
sheapwilliams: I'd remove private here - no real reason to make it hidden with this state-free class.  👍

include/monad/blockchain/test/fake_block_processor.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <monad/execution/test/fakes.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace fake
+{
+
+    class BlockProcessor final
+        : public execution::IBlockProcessor<execution::BoostFiberExecution>
```
sheapwilliams: Declaring this fake with the real execution model limits its utility.  Let's jump on zoom and look at all these interfaces and why/if we need them and for what.

src/monad/blockchain/test/blockchain.cpp
sheapwilliams, line 1: I think the blockchain object belongs in the execution directory and namespace, not separate.

sheapwilliams, line 1: I really expect the fake blockDb to be here so I can see how these tests are working...

src/monad/blockchain/test/blockchain.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+TEST(Blockchain, frontier_to_homestead)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = fork_traits::frontier::end_block_number + 10;
+    state_t state;
+    blockchain_t blockchain;
+    auto [status, finished_block_number] = blockchain.run(
+        state, block_db, fork_traits::frontier::end_block_number - 10);
```
sheapwilliams: Is the name of the test wrong here?  This is only running frontier blocks, AFAICT.

src/monad/blockchain/test/blockchain.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+TEST(Blockchain, berlin_to_london)
+{
+    block_db_t block_db;
+    block_db.block_number_threshold = fork_traits::berlin::end_block_number + 10;
+    state_t state;
+    blockchain_t blockchain;
+    auto [status, finished_block_number] = blockchain.run(
+        state, block_db, fork_traits::berlin::end_block_number - 10);
```
sheapwilliams: Same.  And the one below - Let's review these on zoom.  I don't understand them to be doing what it says they are.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: That said - I'm ok with `london`, or a `no_next_fork_t` that is defined to be the last known fork.

include/monad/blockchain/eth_replay_blockchain.hpp
```diff
@@ -0,0 +1,152 @@
...
+        for (; curr_block_number < loop_until_block_number;
+             ++curr_block_number) {
+
+            if (!block_db.get(curr_block_number, block)) {
+                return std::make_pair(
+                    Status::END_OF_BLOCKDB, curr_block_number - 1u);
+            }
+
+            TBlockProcessor block_processor{};
+            block_processor.template execute<TTraits, fiber_data_t>(
```
sheapwilliams: Should be marked `[[nodiscard]]`, and gather the receipts into the receipt trie.

PR #60, opened 2023-04-10
Remove unnecessary state from Evm

kiblitz:
Make EVM static

### DIFF
--- a/include/monad/execution/evm.hpp
+++ b/include/monad/execution/evm.hpp
@@ -35,10 +35,11 @@ struct Evm
         TState &s, evmc_message const &m, address_t const &to) noexcept
     {
         auto const value = intx::be::load<uint256_t>(m.value);
-        auto balance = intx::be::load<uint256_t>(s.get_balance(m.sender));
-        s.set_balance(m.sender, balance - value);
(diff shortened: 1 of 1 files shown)


include/monad/execution/evm.hpp
```diff
@@ -15,107 +15,105 @@ MONAD_EXECUTION_NAMESPACE_BEGIN
...
     {
         auto const value = intx::be::load<uint256_t>(m.value);
-        auto balance = intx::be::load<uint256_t>(state_.get_balance(m.sender));
-        state_.set_balance(m.sender, balance - value);
-        balance = intx::be::load<uint256_t>(state_.get_balance(to));
-        state_.set_balance(to, balance + value);
+        auto balance = intx::be::load<uint256_t>(s.get_balance(m.sender));
+        s.set_balance(m.sender, balance - value);
+        balance = intx::be::load<uint256_t>(s.get_balance(to));
+        s.set_balance(to, balance + value);
```
sheapwilliams: Unrelated, but relevant nit: James suggested we don't reuse the `balance` l-value, but instead use 2 `const` l-values for `to_balance` and `from_balance`.  

Would you mind fixing that here?

PR #63, opened 2023-04-12
Static precompiles

kiblitz:
Framework for adding ETH precompiles. Groundwork for dynamic precompiles (amortize call cost through contract JIT compilation).

### DIFF
--- a/include/monad/execution/test/fakes.hpp
+++ b/include/monad/execution/test/fakes.hpp
@@ -175,58 +175,61 @@ namespace fake
         }
     };
 
-    template <class TState>
-    struct traits
+    namespace traits
     {
-        static inline uint64_t _sd_refund{};
-        static inline uint64_t block_number{};
-        static inline uint64_t static_precompiles{1};
-        static inline uint64_t _intrinsic_gas{21'000u};
-        static inline uint64_t _max_refund_quotient{2u};
-        static inline bool _fail_store_contract{};
-        static inline uint64_t _gas_creation_cost{};
-        static inline uint64_t _create_address{};
-        static inline uint64_t _echo_gas_cost{10};
-        static inline auto intrinsic_gas(Transaction const &)
+        template <class TState>
+        struct alpha
         {
-            return _intrinsic_gas;
-        }
-        static inline auto starting_nonce() { return 1u; }
-        static inline auto max_refund_quotient()
-        {
-            return _max_refund_quotient;
-        }
-        static inline auto echo_gas_cost() { return _echo_gas_cost; }
-        static inline auto get_selfdestruct_refund(TState const &)
-        {
-            return _sd_refund;
-        }
-        static inline void destruct_touched_dead(TState &s)
-        {
-            s.destruct_touched_dead();
-        }
-        static constexpr inline bool
-        store_contract_code(TState &, address_t const &a, evmc_result &r)
-        {
-            r.gas_left -= _gas_creation_cost;
-            if (!_fail_store_contract) {
-                r.create_address = a;
+            static inline uint64_t _sd_refund{};
+            static inline uint64_t block_number{};
+            static inline uint64_t static_precompiles{1};
+            static inline uint64_t _intrinsic_gas{21'000u};
+            static inline uint64_t _max_refund_quotient{2u};
+            static inline bool _fail_store_contract{};
+            static inline uint64_t _gas_creation_cost{};
+            static inline uint64_t _create_address{};
+            static inline uint64_t _echo_gas_cost{10};
+            static inline auto intrinsic_gas(Transaction const &)
+            {
+                return _intrinsic_gas;
             }
-            return _fail_store_contract;
-        }
-    };
+            static inline auto starting_nonce() { return 1u; }
+            static inline auto max_refund_quotient()
+            {
+                return _max_refund_quotient;
+            }
+            static inline auto echo_gas_cost() { return _echo_gas_cost; }
+            static inline auto get_selfdestruct_refund(TState const &)
+            {
+                return _sd_refund;
+            }
+            static inline void destruct_touched_dead(TState &s)
+            {
+                s.destruct_touched_dead();
+            }
+            static constexpr inline bool
+            store_contract_code(TState &, address_t const &a, evmc_result &r)
+            {
+                r.gas_left -= _gas_creation_cost;
+                if (!_fail_store_contract) {
+                    r.create_address = a;
+                }
+                return _fail_store_contract;
+            }
+        };
 
-    template <class TState>
-    struct next_traits : public traits<TState>
-    {
-        static inline uint64_t block_number{10};
-        static inline uint64_t static_precompiles{2};
-        static inline uint64_t _echo_gas_cost{15};
-        static inline auto echo_gas_cost() { return _echo_gas_cost; }
-    };
+        template <class TState>
+        struct beta : public alpha<TState>
+        {
+            static inline uint64_t block_number{10};
+            static inline uint64_t static_precompiles{2};
+            static inline uint64_t _echo_gas_cost{15};
+            static inline auto echo_gas_cost() { return _echo_gas_cost; }
+        };
 
-    static_assert(concepts::fork_traits<traits<State>, State>);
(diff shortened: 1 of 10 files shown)


include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams, line 1: These are fine.  Let's add their requirement to the concept.

include/monad/execution/static_precompiles.hpp
```diff
@@ -0,0 +1,47 @@
...
+    static constexpr exec_func_t precompile_execs[] = {
+        TPrecompiles::execute...};
+
+    [[nodiscard]] static inline tl::optional<exec_func_t>
+    static_precompile_exec_func(address_t const &addr) noexcept
+    {
+        const auto static_precompile_idx =
+            [&]() -> tl::optional<const unsigned> {
+            const auto last_address_i = sizeof(address_t) - 1u;
+            for (unsigned i = 0; i < last_address_i; ++i) {
```
sheapwilliams: Avoid the implicit conversion...
```suggestion
            for (auto i = 0u; i < last_address_i; ++i) {
```

or

```suggestion
            for (unsigned i = 0u; i < last_address_i; ++i) {
```

include/monad/execution/static_precompiles.hpp
```diff
@@ -0,0 +1,47 @@
...
+                const auto &b = addr.bytes[i];
+                if (b) {
```
sheapwilliams, line 31: We should look at the assembly - if it is actually testing byte by byte, we can probably do better.

include/monad/execution/test/fakes.hpp
```diff
@@ -174,45 +175,80 @@ namespace fake
...
+        template <class TState, concepts::fork_traits<TState> TTraits>
+        struct Echo
+        {
+            static evmc_result execute(const evmc_message &m) noexcept
+            {
+                const int64_t gas =
+                    (const int64_t)m.input_size *
+                    (TTraits::block_number > traits::alpha<TState>::block_number
+                         ? 15
+                         : 10);
```
sheapwilliams: These are from the alpha and beta, or something else?  Can we use them directly if they are?

include/monad/execution/test/fakes.hpp
```diff
@@ -174,45 +175,80 @@ namespace fake
...
 
-    static_assert(concepts::fork_traits<traits<State>, State>);
+        static_assert(concepts::fork_traits<alpha<State>, State>);
+        static_assert(concepts::fork_traits<beta<State>, State>);
+    }
+
+    namespace static_precompiles
+    {
+        template <class TState, concepts::fork_traits<TState> TTraits>
+        struct Echo
```
sheapwilliams: It would be nice to have greater than 1 precompile to make sure you are testing the expansion logic.

src/monad/execution/test/evmc_host.cpp
```diff
@@ -203,3 +208,83 @@ TEST(EvmcHost, revert_create_account)
...
+    fake::Evm e{};
+
+    evmc_host_t host{b, t, s, e};
+    next_evmc_host_t next_host{b, t, s, e};
+
+    evmc_message m{
+        .kind = EVMC_CALL,
+        .gas = 400,
+        .sender = from,
+        .input_data = reinterpret_cast<const unsigned char *>("hello world"),
```
sheapwilliams: I'm a bit fuzzy on these rules, but I don't think this is the right use of `reinterpret_cast`...let's make an `input_data` compatible pointer near line 225 and set it to the const string "hello world".  Then we can assign `.input_data` to that.

src/monad/execution/test/evmc_host.cpp
```diff
@@ -203,3 +208,83 @@ TEST(EvmcHost, revert_create_account)
...
+    auto const result = host.call(m);
+    auto const next_result = next_host.call(m);
+
+    EXPECT_EQ(result.status_code, EVMC_SUCCESS);
+    EXPECT_EQ(result.gas_left, 290);
+    EXPECT_EQ(result.output_size, 11u);
+    EXPECT_EQ(
+        strcmp(
+            reinterpret_cast<const char *>(result.output_data), "hello world"),
+        0);
```
sheapwilliams: I think there are lifetime requirements for this....IIRC, the `result.output_data` needs to be heap allocated, so that the `evmc::Result` "RAII" wrapper can free them.  We should double check this.

include/monad/execution/test/fakes.hpp
kiblitz: Are you talking about the `15` and `10`? They are arbitrary numbers. This ternary operation is to mimic when a fork changes the gas calculation equation. I.e. https://eips.ethereum.org/EIPS/eip-1108

src/monad/execution/test/evmc_host.cpp
kiblitz: Not able to have `const unsigned char *data = "hello world"`

include/monad/execution/test/fakes.hpp
kiblitz: Now that I think about it these can just be static fields part of the fork traits. Maybe that's what you meant.

include/monad/execution/evmc_host.hpp
```diff
@@ -154,10 +154,16 @@ struct EvmcHost : public evmc::HostInterface
...
-        // execute on backend, just this for now
-        evmc_result const result = {.status_code = EVMC_SUCCESS, .gas_left = m.gas};
-
-        if(result.status_code == EVMC_REVERT) {
+        evmc_result const result =
+            TStaticPrecompiles::static_precompile_exec_func(m.code_address)
+                .transform([&](auto static_precompile_execute) {
+                    return static_precompile_execute(m);
+                })
+                // glee PR-soon: dynamic precompile and regular smart contract
```
sheapwilliams: We can axe this comment, I think.  Next step should be evmone integration, as we discussed.

include/monad/execution/static_precompiles.hpp
sheapwilliams: Gabe and I looked at this together in godbolt.  Compiler sees the the contiguous byte array, and correctly widens the 0 test for the top 16-bits.

include/monad/execution/test/fakes.hpp
```diff
@@ -174,45 +175,104 @@ namespace fake
...
+        };
+
+        static_assert(concepts::fork_traits<alpha<State>, State>);
+        static_assert(concepts::fork_traits<beta<State>, State>);
+    }
+
+    namespace static_precompiles
+    {
+        template <class TState, concepts::fork_traits<TState> TTraits>
+        struct Echo
```
sheapwilliams, line 237: I think there is an identity precompile, which this sounds similar to...  Would it make sense to lift this out of test code and make it part of the precompile framework?  Can easily be done in the next PR.

include/monad/execution/test/fakes.hpp
```diff
@@ -174,45 +175,104 @@ namespace fake
...
+                unsigned char *output_data =
+                    (unsigned char *)std::malloc(m.input_size);
+                std::memcpy(output_data, m.input_data, m.input_size);
+                return {
+                    .status_code = EVMC_SUCCESS,
+                    .gas_left = m.gas - gas,
+                    .output_data = output_data,
+                    .output_size = m.input_size,
+                    .release = [](const evmc_result *result) {
+                        std::free((unsigned char *)result->output_data);
```
sheapwilliams, line 255: :+1:

include/monad/execution/test/fakes.hpp
```diff
@@ -174,45 +175,104 @@ namespace fake
...
+                    }};
             }
-            return _fail_store_contract;
-        }
-    };
+        };
 
-    static_assert(concepts::fork_traits<traits<State>, State>);
+        template <class TState, concepts::fork_traits<TState> TTraits>
+        struct DoNothing
```
sheapwilliams: :+1:  Maybe the name is "OneHundredGas", or something?

include/monad/execution/test/fakes.hpp
sheapwilliams: It is. 😄

PR #65, opened 2023-04-14
BlockDb Implementation & Tests

tongzhi1998:
Add BlockDb Implementation (only the 'get' method for now)
Combine tests for BlockDb with tests for rlp::decode


src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,40 @@
...
+        reinterpret_cast<uint8_t const *>(result->data()), result->length()};
+
+    byte_string brotli_buffer;
+    size_t brotli_size = 1'048'576u; // 1M
+    brotli_buffer.resize(brotli_size);
+
+    auto brotli_result = BrotliDecoderDecompress(
+        view.length(), view.data(), &brotli_size, brotli_buffer.data());
+
+    brotli_buffer.resize(brotli_size);
```
sheapwilliams: Any reason you moved away from a stack allocation instead of a 1MB heap allocation?

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,40 @@
...
+    MONAD_ASSERT(brotli_result == BROTLI_DECODER_RESULT_SUCCESS);
```
sheapwilliams: Off hand, it seems asserting success is too strong?  You already return a success bool - any reason not to use that?  Are you trying to assert against specific errors?

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,40 @@
...
+    byte_string_view view2{brotli_buffer};
+
+    auto const decoding_result = rlp::decode_block(block, view2);
+    MONAD_ASSERT(decoding_result.size() == 0);
```
sheapwilliams: Similar with block decode...is asserting too strong here?

src/monad/db/test/block_db.cpp
```diff
@@ -1,52 +1,123 @@
...
+using namespace monad;
+using namespace db;
 
-#include <gtest/gtest.h>
+auto monad_path = std::filesystem::current_path()
+                      .parent_path()
+                      .parent_path()
+                      .parent_path()
+                      .parent_path()
+                      .parent_path();
```
sheapwilliams: I know this was part of the previous commit, but this needs to be more flexible.  This binary can't only be run from a single directory for other tools to work well with it.

src/monad/db/test/block_db.cpp
```diff
@@ -1,52 +1,123 @@
...
 
-using namespace monad;
-using namespace monad::rlp;
+auto dir_path =
+    monad_path / "src" / "monad" / "rlp" / "test" / "assets" / "raw_blocks";
```
sheapwilliams: Are "raw_block" really just compressed blocks?

src/monad/db/test/block_db.cpp
```diff
@@ -1,52 +1,123 @@
-#include <monad/rlp/decode_helpers.hpp>
+#include <monad/db/block_db.hpp>
```
sheapwilliams, line 1: Moving this file doesn't seem right...This is an rlp test, and should probably stay in rlp/test.

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,44 @@
+#include <monad/core/assert.h>
+#include <monad/db/block_db.hpp>
+#include <monad/rlp/decode_helpers.hpp>
+#include <monad/rlp/encode_helpers.hpp>
+
+#include <brotli/decode.h>
+#include <brotli/encode.h>
+
+#include <cstdint>
+#include <iostream>
```
sheapwilliams: Unused?

src/monad/db/block_db.cpp
```diff
@@ -0,0 +1,44 @@
+#include <monad/core/assert.h>
+#include <monad/db/block_db.hpp>
+#include <monad/rlp/decode_helpers.hpp>
+#include <monad/rlp/encode_helpers.hpp>
```
sheapwilliams: unused?  Let's go through each cpp file and make sure we only IWYU.

src/monad/rlp/test/decode_block.cpp
```diff
@@ -1,52 +1,135 @@
...
-#include <monad/core/transaction.hpp>
+#include <gtest/gtest.h>
 
 #include <filesystem>
 #include <fstream>
-
-#include <intx/intx.hpp>
-
-#include <gtest/gtest.h>
+#include <iostream>
```
sheapwilliams: unused?

src/monad/rlp/test/decode_block.cpp
```diff
@@ -1,52 +1,132 @@
...
+        auto dir_path = monad_path / "src" / "monad" / "rlp" / "test" /
+                        "assets" / "compressed_blocks";
+
+        return dir_path;
     }
-    return output;
 }
 
-TEST(Rlp_Block, DecodeBlock2730000)
+TEST(BlockDb, ReadBlock46402)
```
sheapwilliams: You have the name changed here.  If so, then put this in db/test folder, so there isn't a BlockDB test showing up in the monad-rlptest executable.

Same for the assets - have them under db if that is what this test is called.

PR #69, opened 2023-05-04
WIP: In memory state

sheapwilliams:
Just for review.  2 different ways of creating a change_set that manages deletions and additions - I'm partial to the way in account_state.hpp instead of account_storage.hpp, but the wasted 16 bytes of optional is a shame...

Anyway, they both come together in the State object in in_memory_state.hpp.  Lemme know your thoughts.

### DIFF
--- a/include/monad/db/in_memory_state.hpp
+++ b/include/monad/db/in_memory_state.hpp
@@ -1,0 +1,189 @@
+#pragma once
+
+#include <monad/core/account.hpp>
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/receipt.hpp>
+
+#include <monad/db/config.hpp>
+
+#include <set>
+#include <unordered_map>
+#include <vector>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TAccounts, class TAccountStorage>
+struct State
+{
+    struct WorkingCopy
+    {
+        TAccounts::WorkingCopy accounts_;
+        TAccountStorage storage_;
+        std::vector<Receipt::Log> logs_{};
+        unsigned int id_{};
+        uint64_t _refund{};
+
+        WorkingCopy(
+            unsigned int i, TAccounts::WorkingCopy &&a,
+            TAccountStorage &&s)
+            : accounts_{a}
+            , storage_{s}
+            , id_{i}
+        {
+        }
+
+        inline unsigned int id() const noexcept { return id_; }
+        inline void create_contract(address_t const &a) noexcept
+        {
+            accounts_.create_contract(a);
+        }
+
+        [[nodiscard]] inline bool account_exists(address_t const &a) const
+        {
+            return accounts_.account_exists(a);
+        }
+        inline evmc_access_status access_account(address_t const &a) noexcept
+        {
+            return accounts_.access_account(a);
+        }
+
+        [[nodiscard]] inline bytes32_t
+        get_balance(address_t const &a) const noexcept
+        {
+            return accounts_.get_balance(a);
+        }
+
+        inline void set_balance(address_t const &a, uint256_t const &b)
+        {
+            accounts_.set_balance(a, b);
+        }
+
+        [[nodiscard]] inline auto get_nonce(address_t const &a) const noexcept
+        {
+            return accounts_.get_nonce(a);
+        }
+
+        inline void set_nonce(address_t const &a, uint64_t nonce) noexcept
+        {
+            accounts_.set_nonce(a, nonce);
+        }
+
+        [[nodiscard]] inline bytes32_t
+        get_code_hash(address_t const &a) const noexcept
+        {
+            return accounts_.get_code_hash(a);
+        }
+
+        inline void selfdestruct(address_t const &a, address_t const &b)
+        {
+            accounts_.selfdestruct(a, b);
+        }
+
+        inline void destruct_suicides() { accounts_.destruct_suicides(); }
+
+        inline void destruct_touched_dead()
+        {
+            accounts_.destruct_touched_dead();
+        }
+
+        inline uint64_t total_selfdestructs() const noexcept
+        {
+            return accounts_.total_selfdestructs();
+        }
+
+        inline evmc_access_status
+        access_storage(address_t const &a, bytes32_t const &key)
+        {
+            return storage_.access_storage(a, key);
+        }
+
+        [[nodiscard]] inline bytes32_t
+        get_storage(address_t const &a, bytes32_t const &key) const noexcept
+        {
+            return storage_.get_storage(a, key);
+        }
+
+        [[nodiscard]] inline evmc_storage_status set_storage(
+            address_t const &a, bytes32_t const &key, bytes32_t const &value)
+        {
+            return storage_.set_storage(a, key, value);
+        }
+
+        // Account contract accesses
+        void set_code(address_t const &, byte_string const &) {}
+
+        [[nodiscard]] size_t get_code_size(address_t const &) const noexcept
+        {
+            return 0u;
+        }
+
+        [[nodiscard]] size_t
+        copy_code(address_t const &, size_t, uint8_t *, size_t) const noexcept
+        {
+            return 0u;
+        }
+
+        inline void revert()
+        {
+            accounts_.revert();
+            storage_.revert();
+        }
+
+        // Logs
+        void store_log(Receipt::Log &&l) { logs_.emplace_back(l); }
+
+        std::vector<Receipt::Log> &logs() { return logs_; }
+
+        uint64_t get_refund() const noexcept { return _refund; }
+    };
+
+    enum class MergeStatus
+    {
+        WILL_SUCCEED,
+        TRY_LATER,
+        COLLISION_DETECTED,
+    };
+
+    TAccounts &accounts_;
+    TAccountStorage &storage_;
+    unsigned int current_txn_{};
+
+    State(TAccounts &a, TAccountStorage &s)
+        : accounts_{a}
+        , storage_{s}
+    {
+    }
+
+    [[nodiscard]] bytes32_t get_block_hash(int64_t) { return {}; }
+
+    inline unsigned int current_txn() { return current_txn_; }
+
+    inline WorkingCopy get_working_copy(unsigned int id)
+    {
+        return WorkingCopy(
+            id,
+            accounts_.get_working_copy(),
+            storage_.get_copy());
+    }
+
+    inline MergeStatus can_merge_changes(WorkingCopy const &c)
+    {
+        if (current_txn() != c.id()) { return MergeStatus::TRY_LATER; }
+
+        if (accounts_.can_merge(c.accounts_) && storage_.can_merge(c.storage_))
+        {
+            return MergeStatus::WILL_SUCCEED;
+        }
+        return MergeStatus::COLLISION_DETECTED;
+    }
+
+    inline void merge_changes(WorkingCopy &c)
+    {
+        accounts_.merge_changes(c.accounts_);
+        storage_.merge_touched(c.storage_);
+        ++current_txn_;
+    }
+};
+
+MONAD_DB_NAMESPACE_END
--- a/src/monad/db/test/state.cpp
+++ b/src/monad/db/test/state.cpp
@@ -1,0 +1,328 @@
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+
+#include <monad/db/account_state.hpp>
+#include <monad/db/account_storage.hpp>
+#include <monad/db/in_memory_state.hpp>
+
+#include <gtest/gtest.h>
+
+#include <unordered_map>
+
+using namespace monad;
+using namespace monad::db;
+
+static constexpr auto a = 0x5353535353535353535353535353535353535353_address;
+static constexpr auto b = 0xbebebebebebebebebebebebebebebebebebebebe_address;
+static constexpr auto c = 0xa5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5_address;
+// static constexpr auto d = 0xb5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5_address;
+// static constexpr auto e = 0xc5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5_address;
+// static constexpr auto f = 0xd5d5d5d5d5d5d5d5d5d5d5d5d5d5d5d5d5d5d5d5_address;
+// static constexpr auto hash1 =
+// 0x1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c_bytes32;
+// static constexpr auto hash2 =
+// 0x5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b_bytes32;
+
+static constexpr auto key1 =
+    0x00000000000000000000000000000000000000000000000000000000cafebabe_bytes32;
+static constexpr auto key2 =
+    0x1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c_bytes32;
+// static constexpr auto key3 =
+// 0x5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b5b_bytes32;
+static constexpr auto value1 =
+    0x0000000000000000000000000000000000000000000000000000000000000003_bytes32;
+static constexpr auto value2 =
+    0x0000000000000000000000000000000000000000000000000000000000000007_bytes32;
+// static constexpr auto value3 =
+// 0x000000000000000000000000000000000000000000000000000000000000000a_bytes32;
+static constexpr auto null =
+    0x0000000000000000000000000000000000000000000000000000000000000000_bytes32;
+
+using account_store_t = std::unordered_map<address_t, Account>;
+using account_storage_t = std::unordered_map<bytes32_t, bytes32_t>;
+using account_storage_store_t =
+    std::unordered_map<address_t, account_storage_t>;
+
+TEST(State, get_working_copy)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State as{accounts, storage};
+    account_db[a] = {.balance = 10'000};
+
+    [[maybe_unused]] auto bs = as.get_working_copy(0);
+    [[maybe_unused]] auto cs = as.get_working_copy(1);
+
+    bs.access_account(a);
+    bs.set_balance(a, 20'000);
+
+    cs.access_account(a);
+    cs.set_balance(a, 30'000);
+
+    EXPECT_TRUE(bs.account_exists(a));
+    EXPECT_FALSE(bs.account_exists(b));
+    EXPECT_TRUE(cs.account_exists(a));
+    EXPECT_FALSE(cs.account_exists(b));
+    EXPECT_EQ(bs.get_balance(a), bytes32_t{20'000});
+    EXPECT_EQ(cs.get_balance(a), bytes32_t{30'000});
+}
+
+TEST(State, can_merge_fresh)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[b] = {.balance = 40'000u};
+    account_db[c] = {.balance = 50'000u};
+    storage_db[b].insert({key1, value1});
+    storage_db[b].insert({key2, value2});
+    storage_db[c].insert({key1, value1});
+    storage_db[c].insert({key2, value2});
+
+    auto s = t.get_working_copy(0);
+
+    s.create_contract(a);
+    s.set_nonce(a, 1);
+    s.set_balance(a, 38'000);
+    EXPECT_EQ(s.set_storage(a, key2, value1), EVMC_STORAGE_ADDED);
+    EXPECT_EQ(s.set_storage(a, key1, value1), EVMC_STORAGE_ADDED);
+
+    s.access_account(b);
+    s.set_balance(b, 42'000);
+    s.set_nonce(b, 3);
+    EXPECT_EQ(s.set_storage(b, key1, value2), EVMC_STORAGE_MODIFIED);
+    EXPECT_EQ(s.set_storage(b, key2, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(s.set_storage(b, key2, value2), EVMC_STORAGE_DELETED_RESTORED);
+
+    s.access_account(c);
+    EXPECT_EQ(s.set_storage(c, key1, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(s.set_storage(c, key2, null), EVMC_STORAGE_DELETED);
+    s.selfdestruct(c, b);
+    s.destruct_suicides();
+
+    EXPECT_EQ(t.can_merge_changes(s), decltype(t)::MergeStatus::WILL_SUCCEED);
+}
+
+TEST(State, can_merge_same_account_different_storage)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[b] = {.balance = 40'000u};
+    account_db[c] = {.balance = 50'000u};
+    storage_db[b].insert({key1, value1});
+    storage_db[b].insert({key2, value2});
+    storage_db[c].insert({key1, value1});
+    storage_db[c].insert({key2, value2});
+
+    auto bs = t.get_working_copy(0);
+    auto cs = t.get_working_copy(1);
+
+    bs.access_account(b);
+    EXPECT_EQ(bs.set_storage(b, key1, value2), EVMC_STORAGE_MODIFIED);
+
+    EXPECT_EQ(t.can_merge_changes(bs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(bs);
+
+    cs.access_account(b);
+    EXPECT_EQ(cs.set_storage(b, key2, null), EVMC_STORAGE_DELETED);
+
+    EXPECT_EQ(t.can_merge_changes(cs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(cs);
+}
+
+TEST(State, cant_merge_colliding_storage)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[b] = {.balance = 40'000u};
+    storage_db[b].insert({key1, value1});
+
+    auto bs = t.get_working_copy(0);
+    auto cs = t.get_working_copy(1);
+
+    {
+        bs.access_account(b);
+        EXPECT_EQ(bs.set_storage(b, key1, value2), EVMC_STORAGE_MODIFIED);
+
+        EXPECT_EQ(
+            t.can_merge_changes(bs), decltype(t)::MergeStatus::WILL_SUCCEED);
+        t.merge_changes(bs);
+    }
+    {
+        cs.access_account(b);
+        EXPECT_EQ(cs.set_storage(b, key1, null), EVMC_STORAGE_DELETED);
+
+        EXPECT_EQ(
+            t.can_merge_changes(cs),
+            decltype(t)::MergeStatus::COLLISION_DETECTED);
+    }
+
+    // Need to rerun txn 1 - get new working copy
+    auto ds = t.get_working_copy(1);
+
+    ds.access_account(b);
+    EXPECT_EQ(ds.set_storage(b, key1, null), EVMC_STORAGE_DELETED);
+
+    EXPECT_EQ(t.can_merge_changes(ds), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(ds);
+}
+
+TEST(State, merge_txn0_and_txn1)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[a] = {.balance = 30'000u};
+    account_db[b] = {.balance = 40'000u};
+    account_db[c] = {.balance = 50'000u};
+    storage_db[b].insert({key1, value1});
+    storage_db[b].insert({key2, value2});
+    storage_db[c].insert({key1, value1});
+    storage_db[c].insert({key2, value2});
+
+    auto bs = t.get_working_copy(0);
+    auto cs = t.get_working_copy(1);
+
+    bs.access_account(b);
+    bs.set_balance(b, 42'000);
+    bs.set_nonce(b, 3);
+    EXPECT_EQ(bs.set_storage(b, key1, value2), EVMC_STORAGE_MODIFIED);
+    EXPECT_EQ(bs.set_storage(b, key2, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(bs.set_storage(b, key2, value2), EVMC_STORAGE_DELETED_RESTORED);
+
+    EXPECT_EQ(t.can_merge_changes(bs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(bs);
+
+    cs.access_account(a);
+    cs.access_account(c);
+    EXPECT_EQ(cs.set_storage(c, key1, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(cs.set_storage(c, key2, null), EVMC_STORAGE_DELETED);
+    cs.selfdestruct(c, a);
+    cs.destruct_suicides();
+
+    EXPECT_EQ(t.can_merge_changes(cs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(cs);
+}
+
+TEST(State, cant_merge_txn1_collision_need_to_rerun)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[b] = {.balance = 40'000u};
+    account_db[c] = {.balance = 50'000u};
+    storage_db[b].insert({key1, value1});
+    storage_db[b].insert({key2, value2});
+    storage_db[c].insert({key1, value1});
+    storage_db[c].insert({key2, value2});
+
+    auto bs = t.get_working_copy(0);
+    auto cs = t.get_working_copy(1);
+
+    bs.access_account(b);
+    bs.set_balance(b, 42'000);
+    bs.set_nonce(b, 3);
+    EXPECT_EQ(bs.set_storage(b, key1, value2), EVMC_STORAGE_MODIFIED);
+    EXPECT_EQ(bs.set_storage(b, key2, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(bs.set_storage(b, key2, value2), EVMC_STORAGE_DELETED_RESTORED);
+
+    EXPECT_EQ(t.can_merge_changes(bs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(bs);
+
+    cs.access_account(b);
+    cs.access_account(c);
+    EXPECT_EQ(cs.set_storage(c, key1, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(cs.set_storage(c, key2, null), EVMC_STORAGE_DELETED);
+    cs.selfdestruct(c, b);
+    cs.destruct_suicides();
+
+    EXPECT_EQ(
+        t.can_merge_changes(cs), decltype(t)::MergeStatus::COLLISION_DETECTED);
+
+    // Need to rerun txn 1 - get new working copy
+    auto ds = t.get_working_copy(1);
+
+    ds.access_account(b);
+    ds.access_account(c);
+    EXPECT_EQ(ds.set_storage(c, key1, null), EVMC_STORAGE_DELETED);
+    EXPECT_EQ(ds.set_storage(c, key2, null), EVMC_STORAGE_DELETED);
+    ds.selfdestruct(c, b);
+    ds.destruct_suicides();
+
+    EXPECT_EQ(t.can_merge_changes(ds), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(cs);
+}
+
+TEST(State, merge_txn1_try_again_merge_txn0_then_txn1)
+{
+    account_store_t account_db{};
+    Accounts accounts{account_db};
+    account_storage_store_t storage_db{};
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[a] = {.balance = 30'000u};
+    account_db[b] = {.balance = 40'000u};
+    account_db[c] = {.balance = 50'000u};
+    storage_db[b].insert({key1, value1});
+    storage_db[b].insert({key2, value2});
+    storage_db[c].insert({key1, value1});
+    storage_db[c].insert({key2, value2});
+
+    auto bs = t.get_working_copy(0);
+    auto cs = t.get_working_copy(1);
+
+    {
+        // Txn 0
+        bs.access_account(b);
+        bs.set_balance(b, 42'000);
+        bs.set_nonce(b, 3);
+        EXPECT_EQ(bs.set_storage(b, key1, value2), EVMC_STORAGE_MODIFIED);
+        EXPECT_EQ(bs.set_storage(b, key2, null), EVMC_STORAGE_DELETED);
+        EXPECT_EQ(
+            bs.set_storage(b, key2, value2), EVMC_STORAGE_DELETED_RESTORED);
+    }
+    {
+        // Txn 1
+        cs.access_account(a);
+        cs.access_account(c);
+        EXPECT_EQ(cs.set_storage(c, key1, null), EVMC_STORAGE_DELETED);
+        EXPECT_EQ(cs.set_storage(c, key2, null), EVMC_STORAGE_DELETED);
+        cs.selfdestruct(c, a);
+        cs.destruct_suicides();
+    }
+    EXPECT_EQ(t.can_merge_changes(cs), decltype(t)::MergeStatus::TRY_LATER);
+    EXPECT_EQ(t.can_merge_changes(bs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(bs);
+    EXPECT_EQ(t.can_merge_changes(cs), decltype(t)::MergeStatus::WILL_SUCCEED);
+    t.merge_changes(cs);
+}
+
+TEST(State, can_commit) {}
+
+TEST(State, cant_commit_colliding_accounts) {}
+
+TEST(State, cant_commit_colliding_storage) {}
+
+TEST(State, commit) {}
+
+TEST(State, commit_twice) {}
--- a/src/monad/db/test/CMakeLists.txt
+++ b/src/monad/db/test/CMakeLists.txt
@@ -5,5 +5,6 @@ add_unit_test(
   account_storage.cpp
   account_state.cpp
   datum.cpp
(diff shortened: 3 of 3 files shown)


include/monad/db/datum.hpp
```diff
@@ -0,0 +1,151 @@
+#pragma once
+
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/db/config.hpp>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+namespace fnv1a
```
tongzhi1998, line 11: I think evmc has a similar implementation of this in https://github.com/ethereum/evmc/blob/05819e343de23fae05fda0b5a23eeb0da3a320b7/include/evmc/evmc.hpp#L160. Do we need the ability to to hash an arbitrary byte string or just uint64_t? If the later one, then using evmc's implementation seems fine to me.

sheapwilliams: Nice - I missed that.  I'll take a look.

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TAccounts>
+struct Accounts
+{
+    struct ChangeSet
+    {
+        std::unordered_map<address_t, diff<Account>> accounts_{};
+
+        void inline clear() { accounts_.clear(); }
```
zander-xyz, line 22: is the inlining on a class member function really necessary? my understanding is that it's a tossup on what the compiler chooses to do

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+    };
+
+    struct WorkingCopy : public Accounts
+    {
+        ChangeSet changed_;
+        uint64_t total_selfdestructs_{};
```
zander-xyz, line 28: According to our slack discussions, these should be
```suggestion
        ChangeSet changed;
        uint64_t total_selfdestructs{};
```

ditto below

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+        [[nodiscard]] bool account_exists(address_t const &a) const
+        {
+            if (changed_.accounts_.contains(a)) {
+                if (changed_.accounts_.at(a).updated.has_value()) {
+                    return true;
+                }
+                return false;
+            }
+            return Accounts::account_exists(a);
+        }
```
zander-xyz, line 39: Based on slack discussions, shouldn't all of these functions exist as free-standing?

ditto below

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+        uint64_t total_selfdestructs() const noexcept
+        {
+            return total_selfdestructs_;
+        }
+
+        void revert() { changed_.clear(); }
+    };
+
+    // TODO Irrevocable change separated out to avoid reversion
```
zander-xyz, line 120: Can you explain this TODO?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+    struct WorkingCopy : public Accounts
+    {
+        ChangeSet changed_;
+        uint64_t total_selfdestructs_{};
+
+        [[nodiscard]] bool account_exists(address_t const &a) const
+        {
+            if (changed_.accounts_.contains(a)) {
+                if (changed_.accounts_.at(a).updated.has_value()) {
```
zander-xyz, line 33: Got nerd sniped on this one...

When I saw this code, my instinct was to suggest the below (to avoid a double lookup)
```suggestion
            if (auto it = changed_.accounts_.find(a); it != changes_.accounts_.end()) {
                if (it->second.updated.has_value()) {
```
However, under debug, the performances here are actually comparable.

Interestingly, with clang15 and O3, it seems like the compiler is able to evaluate my small `Find` example at compile time whereas it wasn't able to do so with the contains-at pattern. 

One other reason you might stray away from contains is that it's not constexpr. Though in my experience you have to be quite intentional when programming in constexpr so perhaps this alone is not compelling enough. 

Happy to hear other opinions too

Quick-bench: https://quick-bench.com/q/mWMz10iftoul_LHXzQgSbnBlg6k

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                }
+                return false;
+            }
+            return Accounts::account_exists(a);
+        }
+
+        void create_contract(address_t const &a) noexcept
+        {
+            auto const [_, inserted] =
+                changed_.accounts_.emplace(std::make_pair(
```
zander-xyz, line 44: if you emplace, you don't need the make_pair

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                    diff<Account>{get_committed_storage(a), Account{}}));
+            MONAD_ASSERT(inserted);
+        }
+
+        evmc_access_status access_account(address_t const &a) noexcept
+        {
+            if (changed_.accounts_.contains(a)) {
+                return EVMC_ACCESS_WARM;
+            }
+            changed_.accounts_.insert(std::make_pair(
```
zander-xyz, line 55: should we emplace here to match the above?

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+        {
+            return fnv1a::hash(a);
+        }
+    };
+};
+
+template <class T>
+struct diff
+{
+    std::optional<T> const orig{};
```
zander-xyz, line 105: does this not make the move constructor do a copy instead of a move here? are we okay with this?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                a,
+                diff<Account>{
+                    get_committed_storage(a), *get_committed_storage(a)}));
```
zander-xyz, line 58: doing this call twice?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+        void create_contract(address_t const &a) noexcept
+        {
+            auto const [_, inserted] =
+                changed_.accounts_.emplace(std::make_pair(
+                    a,
+                    diff<Account>{get_committed_storage(a), Account{}}));
+            MONAD_ASSERT(inserted);
+        }
+
+        evmc_access_status access_account(address_t const &a) noexcept
```
zander-xyz, line 50: The name of the function and what it actually does does not follow to me. Here, we are indicating that we are accessing the account, but neither the account gets returned nor does it get changed. In fact, the only thing you actually care about seems to be this evmc_access_status thing.

It seems like this function has too many responsibilities.

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                a,
+                diff<Account>{
+                    get_committed_storage(a), *get_committed_storage(a)}));
+            return EVMC_ACCESS_COLD;
+        }
+
+        [[nodiscard]] bytes32_t get_balance(address_t const &a) const noexcept
+        {
+            return intx::be::store<bytes32_t>(
+                changed_.accounts_.at(a).updated.value_or(Account{}).balance);
```
zander-xyz, line 65: so changed_ is actually the combination of committed storage AND the actual changes? i think we're conflating terms here

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        void set_balance(address_t const &address, uint256_t new_balance)
+        {
+            changed_.accounts_.at(address).updated.value().balance =
```
zander-xyz, line 70: are you doing the `value()` here (instead of an optional assignment) to be more clear of intent? 

same question to your other `has_value()` incantations

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                new_balance;
+        }
+
+        [[nodiscard]] auto get_nonce(address_t const &address) const noexcept
```
zander-xyz, line 74: you specify return types in your other functions so can we also do the same here?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+        void destruct_suicides() {}
+
+        void destruct_touched_dead()
+        {
+            for (auto &i : changed_.accounts_) {
+                if (i.second.updated.has_value() && 
+                    i.second.updated.value().balance == 0 &&
+                    i.second.updated.value().nonce == 0 &&
+                    i.second.updated.value().code_hash == NULL_HASH) {
```
zander-xyz, line 106:
```suggestion
                if (i.second.updated == Account{}) {
```

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+        void selfdestruct(address_t const &a, address_t const &beneficiary)
+        {
+            changed_.accounts_.at(beneficiary).updated.value().balance +=
+                changed_.accounts_.at(a).updated.value().balance;
+            changed_.accounts_.at(a).updated.reset();
+            ++total_selfdestructs_;
+        }
+
+        void destruct_suicides() {}
```
zander-xyz, line 98: Is this supposed to be empty?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                    i.second.updated.value().code_hash == NULL_HASH) {
+                    i.second.updated.reset();
+                }
+            }
+        }
+
+        uint64_t total_selfdestructs() const noexcept
+        {
+            return total_selfdestructs_;
+        }
```
zander-xyz, line 115: do we really need a getter for a public member variable?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+        {
+            changed_.accounts_.at(beneficiary).updated.value().balance +=
+                changed_.accounts_.at(a).updated.value().balance;
+            changed_.accounts_.at(a).updated.reset();
+            ++total_selfdestructs_;
+        }
+
+        void destruct_suicides() {}
+
+        void destruct_touched_dead()
```
zander-xyz, line 100: can this be noexcept?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                }
+            }
+        }
+
+        uint64_t total_selfdestructs() const noexcept
+        {
+            return total_selfdestructs_;
+        }
+
+        void revert() { changed_.clear(); }
```
zander-xyz, line 117: ditto on noexcept

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+#include <unordered_map>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TAccounts>
+struct Accounts
+{
+    struct ChangeSet
+    {
+        std::unordered_map<address_t, diff<Account>> accounts_{};
```
zander-xyz, line 20: you might consider the usage absl::flat_hash_map here (std::flat_map not yet available in compilers)

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+    inline std::optional<Account>
+    get_committed_storage(address_t const &a) const
+    {
+        if (merged_.accounts_.contains(a)) {
+            return merged_.accounts_.at(a).updated;
+        }
+        if (accounts_.contains(a)) {
+            return {accounts_.at(a)};
+        }
+        return {};
```
zander-xyz, line 152: can we please spell this as `return std::nullopt`

so that it doesn't get misinterpreted as _return a default constructed Account object_

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+    // TODO Irrevocable change separated out to avoid reversion
+    TAccounts &accounts_;
+    ChangeSet merged_{};
+
+    Accounts(TAccounts &a)
+        : accounts_{a}
+    {
+    }
+
+    inline bool committed_storage_contains(address_t const &a)
```
zander-xyz, line 129: nit: can be const

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+    {
+        if (merged_.accounts_.contains(a)) {
+            if (merged_.accounts_.at(a).updated.has_value()) {
+                return true;
+            }
+            return false;
+        }
```
zander-xyz, line 136: so if it's in the change set but not yet in `accounts`, it still counts as committed? wouldn't that be more accurately described as temporal state since we've modified the account but have not persisted it?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+        if (merged_.accounts_.contains(a)) {
+            return merged_.accounts_.at(a).updated;
+        }
+        if (accounts_.contains(a)) {
+            return {accounts_.at(a)};
+        }
+        return {};
+    }
+
+    // EVMC Host Interface
```
zander-xyz, line 155: thanks for the comment here, this was helpful

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+            }
+            return false;
+        }
+        return accounts_.contains(a);
+    }
+
+    // EVMC Host Interface
+    evmc_access_status access_account(address_t const &) noexcept
+    {
+        return EVMC_ACCESS_COLD;
```
zander-xyz, line 170: Okay, I don't understand this. So the account is always cold if called from EVMC land?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+    [[nodiscard]] bytes32_t get_code_hash(address_t const &a) const noexcept
+    {
+        return get_committed_storage(a).value_or(Account{}).code_hash;
+    }
+
+    inline WorkingCopy get_working_copy()
+    {
+        WorkingCopy a(*this);
+        return a;
+    }
```
zander-xyz, line 190: So you're forcing the WorkingCopy copy constructor to be invoked here by downcasting `*this`?

Why not just let the caller downcast themselves and copy if they need to?

Also, how are you enforcing that the cast will work?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                }
+                if (get_committed_storage(a) != acct.orig) {
+                    return false;
+                }
+            }
+            else {
+                if (committed_storage_contains(a)) {
+                    return false;
+                }
+            }
```
zander-xyz, line 207: So this is only permitting the two working copies to be merged if the original state of each address is equal?

Can you explain why this shouldn't this fail if **any** of the addresses changed overlap?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+    inline bool can_commit()
+    {
+        for (auto const &[a, A] : merged_.accounts_) {
+            if (A.orig.has_value()) {
+                if (!accounts_.contains(a)) {
+                    return false;
+                } else if (A.orig.value() != accounts_.at(a)) {
+                    return false;
+                }
```
zander-xyz, line 233: any thoughts against combining these two statements?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+            }
+            else if (accounts_.contains(a)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    inline void commit_all_merged()
```
zander-xyz, line 242: Does this need `can_commit` to be true? if so, can we assert that?

include/monad/db/account_state.hpp
```diff
@@ -0,0 +1,261 @@
...
+                merged_.accounts_.insert({a, A});
+            }
+        }
+    }
+
+    inline bool can_commit()
+    {
+        for (auto const &[a, A] : merged_.accounts_) {
+            if (A.orig.has_value()) {
+                if (!accounts_.contains(a)) {
```
zander-xyz, line 229: How does this conditional happen? ie. why would the change set have an orig that's different than what is in accounts?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+    struct InnerStorage
+    {
+        std::unordered_map<address_t, account_storage_t> storage_{};
+        std::unordered_map<
+            address_t,
+            std::unordered_set<
+                deleted_key, deleted_key::hash, deleted_key::equality>>
+            deleted_storage_{};
+
+        inline bool
```
zander-xyz, line 35: I personally would also make these member functions but then this points back to our slack discussion where I thought we agreed these would be free standing. Similar to my other comment in account_state.hpp

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+    InnerStorage merged_{};
+    std::unordered_map<address_t, std::unordered_set<bytes32_t>>
+        accessed_storage_{};
+
+    inline void remove_key(address_t const &a, bytes32_t const &key)
+    {
+        storage_.at(a).erase(key);
+        if (storage_.at(a).empty()) {
+            storage_.erase(a);
+        }
```
zander-xyz, line 66: This works now since our underlying storage is just a std container, but once it becomes a real database, we'll need to think harder about whether our usage pattern needs to change here or if we need do some implicit caching to mitigate the overhead of going through the database. For now, this is fine.

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    inline bool
```
zander-xyz, line 69: nodiscard?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+    remove_merged_key_if_present(address_t const &a, bytes32_t const &key)
+    {
+        if (merged_.contains_key(a, key)) {
+            merged_.storage_.at(a).erase(key);
+            if (merged_.storage_.at(a).empty()) {
```
zander-xyz, line 74: you're relying on the compiler here to merge this into one incantation of `at`?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+                merged_.storage_.erase(a);
+            }
+            return true;
+        }
+        return false;
+    }
+
+    inline void remove_touched_key(address_t const &a, bytes32_t const &key)
```
zander-xyz, line 82: okay, so what's the difference between the wording touched and accessed?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        }
+    }
+
+    inline bool storage_contains_key(
+        address_t const &a, bytes32_t const &key) const noexcept
+    {
+        return storage_.contains(a) && storage_.at(a).contains(key);
+    }
+
+    inline bool storage_or_merged_contains_key(
```
zander-xyz, line 96: Can we change this to be `key_exists_in_committed_storage`? We already use `contains` to indicate the presence of a key in an InnerStorage, but not to denote whether the key actually exists (it may have been deleted).

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        if (merged_.deleted_storage_contains_key(a, key)) {
+            return {};
+        }
+        if (merged_.contains_key(a, key)) {
+            return merged_.storage_.at(a).at(key).value;
+        }
+        if (storage_contains_key(a, key)) {
+            return storage_.at(a).at(key);
+        }
+        return {};
```
zander-xyz, line 136:
```suggestion
        return get_committed_storage(a, key);
```

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    [[nodiscard]] inline evmc_storage_status
+    zero_out_key(address_t const &a, bytes32_t const &key) noexcept
+    {
+        // Assume empty (zero) storage is not stored in storage_
```
zander-xyz, line 142: Can we please spell this as an assertion? Might need to define a new noexcept...that is only evaluated in release

```c++
#ifdef NDEBUG
#  define MONAD_NO_EXCEPT noexcept(false)
#else
#  define MONAD_NO_EXCEPT noexcept(true)
#endif
```

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+            return merged_.storage_.at(a).at(key).value;
+        }
+        if (storage_contains_key(a, key)) {
+            return storage_.at(a).at(key);
+        }
+        return {};
+    }
+
+    [[nodiscard]] inline evmc_storage_status
+    zero_out_key(address_t const &a, bytes32_t const &key) noexcept
```
zander-xyz, line 140: wait is this also part of the EVMC interface?

if so, can we document this?

if not, can we change the name of this function to `delete_key`?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+            if (touched_.contains_key(a, key)) {
+                if (get_committed_storage(a, key) ==
+                    touched_.storage_.at(a).at(key)) {
+                    remove_touched_key(a, key);
+                    return EVMC_STORAGE_DELETED;
+                }
+                else {
+                    remove_touched_key(a, key);
+                    touched_.deleted_storage_[a].insert(
+                        deleted_key{storage_.at(a).at(key), key});
```
zander-xyz, line 153: nit: could emplace to avoid the extra deleted_key qualification here

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        return EVMC_STORAGE_ASSIGNED;
+    }
+
+    [[nodiscard]] inline evmc_storage_status set_current_value(
+        address_t const &a, bytes32_t const &key,
+        bytes32_t const &value) noexcept
+    {
+        if (storage_or_merged_contains_key(a, key)) {
+            if (touched_.contains_key(a, key)) {
+                if (touched_.storage_[a][key].value == value) {
```
zander-xyz, line 177: switching to the operator [] syntax instead of at?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+                if (get_committed_storage(a, key) == value) {
+                    remove_touched_key(a, key);
+                    return EVMC_STORAGE_MODIFIED_RESTORED;
+                }
+
+                touched_.storage_[a][key].value = value;
+                return EVMC_STORAGE_MODIFIED;
+            }
+
+            touched_.storage_[a].insert(
```
zander-xyz, line 190: ditto comment about emplace

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        if (value == bytes32_t{}) {
+            return zero_out_key(a, key);
+        }
+        return set_current_value(a, key, value);
+    }
+
+    evmc_access_status
+    access_storage(address_t const &a, bytes32_t const &key) noexcept
+    {
+        auto const &[_, inserted] = accessed_storage_[a].insert(key);
```
zander-xyz, line 232: Is this literally the only usage of accessed? And how does this differ from it's meaning in account_state?

If this is a true indication of some storage needing to be fetched soon, should we be prefetching?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        }
+    }
+
+    inline void revert_touched()
+    {
+        touched_.clear();
+        accessed_storage_.clear();
+    }
+
+    inline AccountStorage get_copy() const
```
zander-xyz, line 290: What's the purpose of having `get_copy` instead of just relying on a copy constructor? This will elide anyway and have the same functionality as just copying, but it's more verbose and not necessary more expressive.

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
+#pragma once
```
zander-xyz, line 1: I haven't thought very deeply about this, but in general I see some parallels between account_storage and account_state. Is it possible to extract some of the common parts? Also happy to hear why this suggestion is totally unreasonable :)

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+    {
+        for (auto const &[a, key_set] : diffs.touched_.deleted_storage_) {
+            for (auto const &k : key_set) {
+                if (k.orig_value != get_committed_storage(a, k.key)) {
+                    return false;
+                }
+
+                if (merged_.deleted_storage_contains_key(a, k.key)) {
+                    return false;
+                }
```
zander-xyz, line 306: why can't we merge here?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        return true;
+    }
+
+    inline void merge_touched(AccountStorage &diffs)
+    {
+        for (auto &[a, key_set] : diffs.touched_.deleted_storage_) {
+            for (auto const &key : key_set) {
+                if (auto const removed =
+                        remove_merged_key_if_present(a, key.key);
+                    removed) {
```
zander-xyz, line 326:
```suggestion
                if (remove_merged_key_if_present(a, key.key)) {
```

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    inline void commit_all_merged()
+    {
+        MONAD_ASSERT(can_commit());
```
zander-xyz, line 269: MONAD_ASSERT?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+
+        for (auto &[addr, acct_storage] : diffs.touched_.storage_) {
+            if (!merged_.storage_.contains(addr)) {
+                merged_.storage_.emplace(
+                    std::make_pair(addr, std::move(acct_storage)));
+                continue;
+            }
+
+            for (auto const &[key, value] : acct_storage) {
+                MONAD_ASSERT(value != bytes32_t{});
```
zander-xyz, line 346: Typically I would say `assert` is good enough here but I'm not sure what our philosophy on assertions is.

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+        }
+    };
+
+    TStorage &storage_;
+    InnerStorage touched_{};
+    InnerStorage merged_{};
+    std::unordered_map<address_t, std::unordered_set<bytes32_t>>
+        accessed_storage_{};
+
+    inline void remove_key(address_t const &a, bytes32_t const &key)
```
zander-xyz, line 61: does this not need to look at touched or merged? and how does this differ from zero_out_key

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+template <class T>
+struct diff_value
+{
+    T const orig{};
+    T value{};
+
+    diff_value() = default;
+    diff_value(diff_value const &v) = default;
+    diff_value(diff_value &&v) noexcept = default;
+    diff_value(T const &o, T const &v)
```
zander-xyz, line 70: constexpr all the things? :)

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+    void operator=(T &&b) { value = std::move(b); }
+
+    struct equality
+    {
+        inline bool operator()(
+            diff_value const &first, diff_value const &second) const noexcept
+        {
+            return first.value == second.value;
+        }
+    };
```
zander-xyz, line 91: Why a separate struct for this?

include/monad/db/account_storage.hpp
```diff
@@ -0,0 +1,353 @@
...
+
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TStorage>
+struct AccountStorage
+{
+    using account_storage_t =
+        std::unordered_map<bytes32_t, diff_value<bytes32_t>>;
+
+    AccountStorage(TStorage &store)
```
zander-xyz, line 21: explicit?

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+    std::optional<T> updated{};
+
+    diff() = default;
+    diff(diff const &v) = default;
+    diff(diff &&v) noexcept = default;
+    diff(T const &o, T const &v)
+        : orig{o}
+        , updated{v}
+    {
+    }
```
zander-xyz, line 115: Is this constructor needed given we have the below `std::optional, T` overload?

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+    deleted_key() = default;
+    deleted_key(deleted_key const &v) = default;
+    deleted_key(deleted_key &&v) noexcept = default;
+    explicit deleted_key(bytes32_t const &k)
+        : orig_value{}
+        , key{k}
+    {
+    }
+    deleted_key(bytes32_t const &b, bytes32_t const &k)
```
zander-xyz, line 175: what's with the variable names here

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+        : orig_value{b}
+        , key{k}
+    {
+    }
+
+    // void operator=(deleted_key const &d) { key = d.key; }
```
zander-xyz, line 181: remove if not needed?

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,207 @@
...
+    void operator=(bytes32_t const &b) { key = b; }
+
+    struct equality
+    {
+        inline bool operator()(
+            deleted_key const &first, deleted_key const &second) const noexcept
+        {
+            return first.key == second.key;
```
zander-xyz, line 189: Side note: does it make sense if two `deleted_key`s have the same key but different `orig_value`s?

include/monad/core/account.hpp
```diff
@@ -17,4 +17,9 @@ struct Account
 static_assert(sizeof(Account) == 80);
 static_assert(alignof(Account) == 8);
 
+inline bool operator==(Account const &lhs, Account const &rhs)
+{
```
zander-xyz, line 21: nit: can you add a comment, something like 

"notably without incarnation" so that someone doesn't get confused and try to default this operator in the future?

include/monad/db/in_memory_state.hpp
```diff
@@ -0,0 +1,189 @@
...
+        TAccounts::WorkingCopy accounts_;
+        TAccountStorage storage_;
+        std::vector<Receipt::Log> logs_{};
+        unsigned int id_{};
+        uint64_t _refund{};
+
+        WorkingCopy(
+            unsigned int i, TAccounts::WorkingCopy &&a,
+            TAccountStorage &&s)
+            : accounts_{a}
```
zander-xyz, line 30: std::move? a is an lvalue here

include/monad/db/in_memory_state.hpp
```diff
@@ -0,0 +1,189 @@
...
+            , storage_{s}
+            , id_{i}
+        {
+        }
+
+        inline unsigned int id() const noexcept { return id_; }
```
zander-xyz, line 36: similar comment as elsewhere, getter for for public member?

include/monad/db/in_memory_state.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+#include <monad/db/config.hpp>
+
+#include <set>
+#include <unordered_map>
+#include <vector>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TAccounts, class TAccountStorage>
```
zander-xyz, line 16: `in_memory_state` implies that this class is meant to be a helper for unit tests. Can we move this to `src/monad/db/test`? 

if indeed this class is only needed for testing, is it still necessary for it to be templated on TAccounts and TAccountStorage?

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -78,10 +78,14 @@ struct alignas(64) TransactionProcessorFiberData
...
             auto txn_state = s_.get_copy();
-            TEvm e{};
+            TEvm e{}; // e needs to be constructed with the working copy of state
             result_ = p.execute(txn_state, e, bh_, txn_);
 
             if (auto const applied = s_.apply_state(txn_state); applied) {
+				// apply_state -> can_merge_changes
+				// Can merge needs to be in yield while loop while receiving TRY_AGAIN
+				// When WILL_SUCCEED is returned, merge, and return; if ERROR is 
+				// received, then error out
```
zander-xyz, line 88: weird indentation here

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -78,10 +78,14 @@ struct alignas(64) TransactionProcessorFiberData
             }
 
             auto txn_state = s_.get_copy();
-            TEvm e{};
+            TEvm e{}; // e needs to be constructed with the working copy of state
             result_ = p.execute(txn_state, e, bh_, txn_);
 
             if (auto const applied = s_.apply_state(txn_state); applied) {
```
zander-xyz, line 84: Do we really need the `applied` here?

src/monad/db/test/state.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+TEST(State, can_commit) {}
+
+TEST(State, cant_commit_colliding_accounts) {}
+
+TEST(State, cant_commit_colliding_storage) {}
+
+TEST(State, commit) {}
+
+TEST(State, commit_twice) {}
```
zander-xyz, line 328: can we mark these all as TODO?

src/monad/db/test/state.cpp
```diff
@@ -0,0 +1,328 @@
...
+    AccountStorage storage{storage_db};
+    State t{accounts, storage};
+
+    account_db[a] = {.balance = 30'000u};
+    account_db[b] = {.balance = 40'000u};
+    account_db[c] = {.balance = 50'000u};
+    storage_db[b].insert({key1, value1});
+    storage_db[b].insert({key2, value2});
+    storage_db[c].insert({key1, value1});
+    storage_db[c].insert({key2, value2});
```
zander-xyz, line 197: FYI you can use TEST_F and setup your own fixtures so that you don't have this setup code happening in every test

include/monad/db/account_state.hpp
sheapwilliams: Ur right - member function implementations in headers are already marked inline.  I'll remove where appropriate.

include/monad/db/account_state.hpp
sheapwilliams: Discussed offline.  We'll keep no underscore for dumb structs, and underscore suffix for things that act more as classes.

include/monad/db/account_state.hpp
sheapwilliams: This is an EVMC Host Interface function - we don't have flexibility to change it.  I'll comment it as such.

include/monad/db/account_state.hpp
sheapwilliams: Modified the benchmark so it isn't optimized away - they show the same results:
https://quick-bench.com/q/AfnQpEHaH1PkrwHk4s8rSqlgz44

We can leave this for now, but note to reviewers to watch for non-const calls between 2x lookups will lose the optimization.

include/monad/db/account_state.hpp
sheapwilliams: I can go through and scrub some of these.

include/monad/db/account_state.hpp
sheapwilliams: Same - I'll take a look.

include/monad/db/datum.hpp
sheapwilliams: I'll make sure that all uses of this are not moveable...I believe that is the case currently.

include/monad/db/account_state.hpp
sheapwilliams: I'm pretty sure this will be optimized away.  I can look tho.

include/monad/db/account_state.hpp
sheapwilliams: Discussed on zoom.  This function is part of the EVMC Host interface.  Perhaps changing `changed_` to `touched_` might help here?

include/monad/db/account_state.hpp
sheapwilliams: Discussed on zoom.  Better understanding of pre-block state vs. committed vs. working changes.

include/monad/db/account_state.hpp
sheapwilliams: Discussed on zoom.  I used value() functions for consistency with all optional accesses.

include/monad/db/account_state.hpp
sheapwilliams: Yea, sure.  I can change this.

include/monad/db/account_state.hpp
sheapwilliams: Yes, as discussed, this is what is written in the YP, so it is carried as an API to the state.  But this particular implementation of the state doesn't track destructs separately from other account deletions, so there is nothing to particularly do when this is called.  It is just an artifact of the implementation of this particular account state.

include/monad/db/account_state.hpp
sheapwilliams: I believe it can.  I'll change it.

include/monad/db/account_state.hpp
sheapwilliams: Probably - I was intentionally avoiding thinking about incarnation, but I'll take a look at that before deciding this here.

include/monad/db/account_state.hpp
sheapwilliams: Not really, but this is used by the fork traits to calculate the destruct refund.  So I have it as an API to keep the fork traits and state decoupled.

include/monad/db/account_state.hpp
sheapwilliams: Yea, probably.

include/monad/db/account_state.hpp
sheapwilliams: Discussed in person - this is just a reminder to avoid reverting the irrevocable change for every transaction.

include/monad/db/account_state.hpp
sheapwilliams: Discussed in person - merged and touched use both the same ChangeSet type.

include/monad/db/account_state.hpp
sheapwilliams: Why not - it makes it clearer.

include/monad/db/account_state.hpp
sheapwilliams: Discussed on zoom - I can remove this once I make the transaction validation work off a working copy instead of a non-working copy.  I'll mark this as a TODO.

include/monad/db/account_state.hpp
sheapwilliams: Discussed in zoom - perhaps a better name would be `construct_working_copy()`?  Or perhaps make a account state and storage working copies construct from state and storage respectively?

include/monad/db/account_state.hpp
sheapwilliams: Discussed in zoom - merged_ is the current working state of the block through the transaction list, and we are testing for merging a working copy into the merged_ change set.

Perhaps a better name for `commit_all_merged()` would be `write_block_changeset()` or something?

include/monad/db/account_state.hpp
sheapwilliams: You are right - it shouldn't.  This is debug/validation functionality that shouldn't really run in prod, but should during validation.  I'd be happy if this entire thing was some sort of assert in the below function.  Perhaps I'll move this functionality out...

include/monad/db/account_state.hpp
sheapwilliams: Yup - I'll do that.

include/monad/db/account_state.hpp
sheapwilliams: Yea, like above, I'll move the above into this function as a programmable assert.

include/monad/db/account_storage.hpp
sheapwilliams: Discussed on zoom - just like previous comments, this thing acts more like a class.

include/monad/db/in_memory_state.hpp
sheapwilliams: I'll fix the file name to match the `State` class name.

include/monad/db/account_storage.hpp
sheapwilliams: I dunno about this one!  We should check it out.

include/monad/db/datum.hpp
sheapwilliams: LOL - that's cause it used to be called `before`.  I can change it.

include/monad/db/datum.hpp
sheapwilliams: Not just not needed - doesn't compile with `orig_value` as `const`.  I'll remove.

include/monad/db/in_memory_state.hpp
sheapwilliams: Yea, same thing as previous.  This is to separate the API of its use over in `transaction_processor_data.hpp` from its implementation here.  I'd rather not have that know about these internals.

include/monad/execution/transaction_processor_data.hpp
sheapwilliams: clang-format  🤷

include/monad/execution/transaction_processor_data.hpp
sheapwilliams: The comment below mentions the forthcoming refactor - I'll fix up the syntax when I do that.

src/monad/db/test/state.cpp
sheapwilliams: The PR is marked as a WIP.  I think that counts. 😉

src/monad/db/test/state.cpp
sheapwilliams: Each test really is testing different states...I think only these 2 tests are copied.  I usually use a fixture when I get to 3.

src/monad/db/test/state.cpp
zander-xyz: When/if we merge this in, the PR will no longer exist

include/monad/db/account_storage.hpp
sheapwilliams: As we discussed on zoom, I'll take a look at collapsing some of this into common code.

include/monad/db/account_storage.hpp
sheapwilliams: It is used to remove keys on commit...but the logic might be duped.  After refactor, I'll take a look at these and reduce any duplication I find.

include/monad/db/account_storage.hpp
sheapwilliams: Discussed in zoom - I think there is some confusion here, but hopefully the refactor will address it.

include/monad/db/account_storage.hpp
sheapwilliams: Not part of the EVMC interface.  Yea, we can probably change the name.

include/monad/db/account_storage.hpp
sheapwilliams: Agree - we can write the assertion out.

I think we should be `noexcept()` correct, and when we inject a new/different database, we'll probably have to put in some type traits to make these noexcept correct.  But we can approach that at that time.

include/monad/db/account_storage.hpp
sheapwilliams: I think I should make this consistent...

include/monad/db/account_storage.hpp
sheapwilliams: Discussed 2 ways of doing this on zoom - one would have to scrub the zeros out of the ChangeSet on merge.  Or we can track them separately.  This is only cause we have to appropriately mark as warmed any value, even if it is zero.  I'll think on the other way when refactoring.

include/monad/db/account_storage.hpp
sheapwilliams: Ur right - this should just be assert.

include/monad/db/account_storage.hpp
sheapwilliams: We talked about this - Will make a working copy when I refactor.

include/monad/db/account_storage.hpp
sheapwilliams: Discussed on zoom - we can assert that deleted_storage_ will never have a orig_value that is zero, so this will *always* be a value we are expecting to delete.

include/monad/db/datum.hpp
sheapwilliams: That is the interface for `unordered_map`.

include/monad/db/datum.hpp
sheapwilliams: I can check - I think we do still need this.

include/monad/db/in_memory_state.hpp
sheapwilliams: Ur right - I need to forward it.

include/monad/execution/transaction_processor.hpp
```diff
@@ -101,8 +101,8 @@ struct TransactionProcessor
             s, b, t, static_cast<uint64_t>(result.gas_left));
```
zander-xyz, line 101: Is `TransactionProcessor` and `TransactionProcessorData` unit tested currently? If not, what's our plan to verify it's correctness?

PR #70, opened 2023-05-16
system rocksdb, compiler upgrade

jhunsaker:


include/monad/db/config.hpp
```diff
@@ -7,5 +7,5 @@
     {
 
 #define MONAD_DB_NAMESPACE_END                                                 \
-    MONAD_NAMESPACE_END                                                        \
-    }
\ No newline at end of file
+    }                                                                          \
+    MONAD_NAMESPACE_END
```
zander-xyz, line 11: was this change intended?

PR #78, opened 2023-05-23
:new: EVMC Host Interface State

sheapwilliams:
This should incorporate most of the feedback left on #69.

### DIFF
--- a/include/monad/db/state.hpp
+++ b/include/monad/db/state.hpp
@@ -1,0 +1,217 @@
+#pragma once
+
+#include <monad/core/account.hpp>
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/receipt.hpp>
+
+#include <monad/db/config.hpp>
+
+#include <vector>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TAccountStore, class TValueStore, class TCodeStore>
+struct State
+{
+    struct WorkingCopy
+    {
+        typename TAccountStore::WorkingCopy accounts_;
+        typename TValueStore::WorkingCopy storage_;
+        typename TCodeStore::WorkingCopy code_;
+        std::vector<Receipt::Log> logs_{};
+        unsigned int txn_id_{};
+
+        WorkingCopy(
+            unsigned int i, typename TAccountStore::WorkingCopy &&a,
+            typename TValueStore::WorkingCopy &&s,
+            typename TCodeStore::WorkingCopy &&c)
+            : accounts_{std::move(a)}
+            , storage_{std::move(s)}
+            , code_{std::move(c)}
+            , txn_id_{i}
+        {
+        }
+
+        unsigned int txn_id() const noexcept { return txn_id_; }
+        void create_contract(address_t const &a) noexcept
+        {
+            accounts_.create_contract(a);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] bool account_exists(address_t const &a) const
+        {
+            return accounts_.account_exists(a);
+        }
+
+        // EVMC Host Interface
+        evmc_access_status access_account(address_t const &a) noexcept
+        {
+            return accounts_.access_account(a);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] bytes32_t get_balance(address_t const &a) const noexcept
+        {
+            return accounts_.get_balance(a);
+        }
+
+        void set_balance(address_t const &a, uint256_t const &b)
+        {
+            accounts_.set_balance(a, b);
+        }
+
+        [[nodiscard]] auto get_nonce(address_t const &a) const noexcept
+        {
+            return accounts_.get_nonce(a);
+        }
+
+        void set_nonce(address_t const &a, uint64_t nonce) noexcept
+        {
+            accounts_.set_nonce(a, nonce);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] bytes32_t get_code_hash(address_t const &a) const noexcept
+        {
+            return accounts_.get_code_hash(a);
+        }
+
+        void selfdestruct(address_t const &a, address_t const &b)
+        {
+            accounts_.selfdestruct(a, b);
+        }
+
+        void destruct_suicides() { accounts_.destruct_suicides(); }
+
+        void destruct_touched_dead() { accounts_.destruct_touched_dead(); }
+
+        uint64_t total_selfdestructs() const noexcept
+        {
+            return accounts_.total_selfdestructs();
+        }
+
+        // EVMC Host Interface
+        evmc_access_status
+        access_storage(address_t const &a, bytes32_t const &key)
+        {
+            return storage_.access_storage(a, key);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] bytes32_t
+        get_storage(address_t const &a, bytes32_t const &key) const noexcept
+        {
+            return storage_.get_storage(a, key);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] evmc_storage_status set_storage(
+            address_t const &a, bytes32_t const &key, bytes32_t const &value)
+        {
+            return storage_.set_storage(a, key, value);
+        }
+
+        // Account contract accesses
+        void set_code(address_t const &a, byte_string const &c)
+        {
+            code_.set_code(a, c);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] size_t get_code_size(address_t const &a) const noexcept
+        {
+            return code_.get_code_size(a);
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] size_t copy_code(
+            address_t const &a, size_t offset, uint8_t *buffer,
+            size_t size) const noexcept
+        {
+            return code_.copy_code(a, offset, buffer, size);
+        }
+
+        void revert() noexcept
+        {
+            accounts_.revert();
+            storage_.revert();
+            code_.revert();
+        }
+
+        // Logs
+        void store_log(Receipt::Log &&l) { logs_.emplace_back(l); }
+
+        std::vector<Receipt::Log> &logs() { return logs_; }
+    };
+
+    enum class MergeStatus
+    {
+        WILL_SUCCEED,
+        TRY_LATER,
+        COLLISION_DETECTED,
+    };
+
+    TAccountStore &accounts_;
+    TValueStore &storage_;
+    TCodeStore &code_;
+    unsigned int current_txn_{};
+
+    State(TAccountStore &a, TValueStore &s, TCodeStore &c)
+        : accounts_{a}
+        , storage_{s}
+        , code_{c}
+    {
+    }
+
+    [[nodiscard]] bytes32_t get_block_hash(int64_t) const { return {}; }
+
+    unsigned int current_txn() const { return current_txn_; }
+
+    WorkingCopy get_working_copy(unsigned int id) const
+    {
+        return WorkingCopy(
+            id,
+            typename TAccountStore::WorkingCopy{accounts_},
+            typename TValueStore::WorkingCopy{storage_},
+            typename TCodeStore::WorkingCopy{code_});
+    }
+
+    MergeStatus can_merge_changes(WorkingCopy const &c) const
+    {
+        if (current_txn() != c.txn_id()) {
+            return MergeStatus::TRY_LATER;
+        }
+
+        if (accounts_.can_merge(c.accounts_) &&
+            storage_.can_merge(c.storage_) && code_.can_merge(c.code_)) {
+            return MergeStatus::WILL_SUCCEED;
+        }
+        return MergeStatus::COLLISION_DETECTED;
+    }
+
+    void merge_changes(WorkingCopy &c)
+    {
+        accounts_.merge_changes(c.accounts_);
+        storage_.merge_touched(c.storage_);
+        code_.merge_changes(c.code_);
+        ++current_txn_;
+    }
+
+    bool can_commit() const
+    {
+        return accounts_.can_commit() && storage_.can_commit() &&
+               code_.can_commit();
+    }
+
+    void commit()
+    {
+        accounts_.commit_all_merged();
+        storage_.commit_all_merged();
+        code_.commit_all_merged();
+        current_txn_ = 0;
+    }
+};
+
+MONAD_DB_NAMESPACE_END
--- a/src/monad/db/test/CMakeLists.txt
+++ b/src/monad/db/test/CMakeLists.txt
@@ -6,6 +6,7 @@ add_unit_test(
   block_db.cpp
   code_store.cpp
   datum.cpp
+  state.cpp
   value_store.cpp
   LIBRARIES
   monad)
--- a/src/monad/db/test/state.cpp
+++ b/src/monad/db/test/state.cpp
@@ -1,0 +1,422 @@
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+
+#include <monad/db/account_store.hpp>
+#include <monad/db/code_store.hpp>
+#include <monad/db/value_store.hpp>
+#include <monad/db/state.hpp>
+
+#include <gtest/gtest.h>
+
+#include <unordered_map>
+
+using namespace monad;
+using namespace monad::db;
+
+static constexpr auto a = 0x5353535353535353535353535353535353535353_address;
+static constexpr auto b = 0xbebebebebebebebebebebebebebebebebebebebe_address;
+static constexpr auto c = 0xa5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5_address;
+static constexpr auto key1 =
+    0x00000000000000000000000000000000000000000000000000000000cafebabe_bytes32;
+static constexpr auto key2 =
+    0x1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c_bytes32;
+static constexpr auto value1 =
+    0x0000000000000000000000000000000000000000000000000000000000000003_bytes32;
+static constexpr auto value2 =
+    0x0000000000000000000000000000000000000000000000000000000000000007_bytes32;
+static constexpr auto null =
+    0x0000000000000000000000000000000000000000000000000000000000000000_bytes32;
+static constexpr auto c1 =
+    byte_string{0x65, 0x74, 0x68, 0x65, 0x72, 0x6d, 0x69};
+
+using account_db_t = std::unordered_map<address_t, Account>;
+using key_value_db_t = std::unordered_map<bytes32_t, bytes32_t>;
+using value_db_t =
+    std::unordered_map<address_t, key_value_db_t>;
+using code_db_t = std::unordered_map<address_t, byte_string>;
+
+TEST(State, get_working_copy)
+{
+    account_db_t account_db{};
+    AccountStore accounts{account_db};
+    value_db_t value_db{};
+    ValueStore values{value_db};
+    code_db_t code_db{};
+    CodeStore code{code_db};
+    State as{accounts, values, code};
+    account_db[a] = {.balance = 10'000};
+
+    [[maybe_unused]] auto bs = as.get_working_copy(0);
+    [[maybe_unused]] auto cs = as.get_working_copy(1);
+
+    bs.access_account(a);
+    bs.set_balance(a, 20'000);
+
+    cs.access_account(a);
+    cs.set_balance(a, 30'000);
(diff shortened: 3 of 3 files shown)


include/monad/core/account.hpp
```diff
@@ -17,4 +17,9 @@ struct Account
 static_assert(sizeof(Account) == 80);
 static_assert(alignof(Account) == 8);
 
+inline bool operator==(Account const &lhs, Account const &rhs)
+{
+    return lhs.balance == rhs.balance && lhs.code_hash == rhs.code_hash &&
+           lhs.nonce == rhs.nonce;
+}
```
zander-xyz: You can default this once you get rid of incarnation

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+            assert(inserted);
+        }
+
+        // EVMC Host Interface
+        evmc_access_status access_account(address_t const &a) noexcept
+        {
+            if (changed_.contains(a)) {
+                return EVMC_ACCESS_WARM;
+            }
+            changed_.emplace(
```
zander-xyz, line 53: emplace can still throw, so marking this as noexcept is misleading.

might want to audit other cases of noexcept

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+                }
+                return false;
+            }
+            return AccountStore::account_exists(a);
+        }
+
+        void create_contract(address_t const &a) noexcept
+        {
+            auto const [_, inserted] = changed_.emplace(
+                a, diff_t{get_committed_storage(a), Account{}});
```
zander-xyz, line 43: Is it possible for `get_committed_storage` here to return anything other than a `std::nullopt`?

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+        {
+            changed_.at(beneficiary).updated.value().balance +=
+                changed_.at(a).updated.value().balance;
+            changed_.at(a).updated.reset();
+            ++total_selfdestructs_;
+        }
+
+        void destruct_suicides() const noexcept {}
+
+        void destruct_touched_dead() noexcept
```
zander-xyz, line 100: Is touched the vocabulary that the YP uses?

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+        {
+            for (auto &i : changed_) {
+                if (i.second.updated.has_value() &&
+                    i.second.updated.value().balance == 0 &&
+                    i.second.updated.value().nonce == 0 &&
+                    i.second.updated.value().code_hash == NULL_HASH) {
```
zander-xyz: In the other PR I suggested `i.second.updated == Account{}`.

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    // EVMC Host Interface
+    [[nodiscard]] bool account_exists(address_t const &a) const noexcept
+    {
+        if (merged_.contains(a)) {
+            if (merged_.at(a).updated.has_value()) {
+                return true;
+            }
+            return false;
```
zander-xyz: super nit
```suggestion
            return merged_.at(a).updated.has_value();
```

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+            else {
+                if (committed_storage_contains(a)) {
+                    return false;
+                }
+            }
+        }
+
+        return true;
```
zander-xyz: I think this whole thing is something like

```c++
return std::ranges::all_of(diffs.changed_, [](auto const& p) {
    return p.second == get_committed_storage(p.first);
});
```

include/monad/db/state.hpp
```diff
@@ -0,0 +1,217 @@
...
+        , code_{c}
+    {
+    }
+
+    [[nodiscard]] bytes32_t get_block_hash(int64_t) const { return {}; }
+
+    unsigned int current_txn() const { return current_txn_; }
+
+        WorkingCopy
+        get_working_copy(unsigned int id) const
```
zander-xyz: Did clang format really do this?

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    void merge_changes(WorkingCopy &diffs)
+    {
```
zander-xyz, line 189: Should we `assert(can_merge(diffs))`?

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+            else if (db_.contains(a)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    void commit_all_merged()
+    {
+        for (auto const &[a, A] : merged_) {
```
zander-xyz, line 217: I thought we were going to `assert(can_commit())` here?

include/monad/db/account_store.hpp
```diff
@@ -0,0 +1,257 @@
...
+            if (A.orig.has_value()) {
+                if (A.updated.has_value()) {
+                    db_[a] = A.updated.value();
+                }
+                else {
+                    db_.erase(a);
+                }
+            }
+            else {
```
zander-xyz, line 226: `assert(A.update.has_value())`?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
+#pragma once
```
zander-xyz, line 1: clang format this file?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TCodeDB>
+struct CodeStore
+{
+    using map_t = std::unordered_map<address_t, byte_string const>;
+
+    struct WorkingCopy : public CodeStore
+    {
+        WorkingCopy(CodeStore const &c) : CodeStore(c) {}
```
zander-xyz: explicit

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+        map_t code_{};
+        uint64_t total_selfdestructs_{};
```
zander-xyz: can we place these at the top of the struct?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+        byte_string const& code_at(address_t const &a) const
+        {
+            if(code_.contains(a))
+                return code_.at(a);
+            return CodeStore::code_at(a);
+        }
+
+        void set_code(address_t const &a, byte_string const &code)
+        {
+            if(code.size() == 0) { return; }
```
zander-xyz: nit
```suggestion
            if(code.empty()) { return; }
```

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+            code_.insert({a, code});
```
zander-xyz: assert successful insertion?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+        [[nodiscard]] size_t get_code_size(address_t const &a) const noexcept
+        {
+            return code_at(a).size();
+        }
+
+        // EVMC Host Interface
+        [[nodiscard]] size_t
+        copy_code(address_t const &a, size_t offset, uint8_t *buffer, size_t buffer_size)
+        {
+            auto const code = code_at(a);
```
zander-xyz:
```suggestion
            auto const& code = code_at(a);
```

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+    inline bool can_merge(WorkingCopy const &w)
+    {
+        for (auto &[a, _] : w.code_)
+        {
+            if (merged_.contains(a) || db_.contains(a))
+            {
+                return false;
+            }
+        }
+        return true;
```
zander-xyz:
```c++
return std::ranges::none_of(w.code_, [](auto const& a) {
    return merged_.contains(a) || db_.contains(a);
}, &map_t::value_type::first);
```

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    byte_string const& code_at(address_t const &a) const
+    {
+        if (merged_.contains(a)) { return merged_.at(a); }
+        if (db_.contains(a)) { return db_.at(a); }
+        return empty;
+    }
+
+    inline WorkingCopy get_working_copy()
```
zander-xyz: drop inlines here and below

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+    {
+        WorkingCopy a(*this);
+        return a;
+    }
+
+    inline bool can_merge(WorkingCopy const &w)
```
zander-xyz:
```suggestion
    [[nodiscard]] bool can_merge(WorkingCopy const &w) const
```

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+            if (merged_.contains(a) || db_.contains(a))
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    inline void merge_changes(WorkingCopy &w)
+    {
```
zander-xyz: assert(can_merge());

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+        WorkingCopy a(*this);
+        return a;
+    }
+
+    inline bool can_merge(WorkingCopy const &w)
+    {
+        for (auto &[a, _] : w.code_)
+        {
+            if (merged_.contains(a) || db_.contains(a))
```
zander-xyz: Given that code is immutable, is this even possible?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+    inline void merge_changes(WorkingCopy &w)
+    {
+        for (auto &a : w.code_)
+        {
+            merged_.insert(a);
+        }
+    }
+
+    inline bool can_commit()
```
zander-xyz: confirming this is just a debug function right?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+        return true;
+    }
+
+    inline void commit_all_merged()
+    {
+        for (auto &a : merged_)
+        {
+            db_.insert(a);
```
zander-xyz: Assert successful insertion?

include/monad/db/code_store.hpp
```diff
@@ -0,0 +1,123 @@
...
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    inline void commit_all_merged()
+    {
+        for (auto &a : merged_)
```
zander-xyz:
```suggestion
        for (auto const&a : merged_)
```

include/monad/db/datum.hpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+    }
+    deleted_key(bytes32_t const &b, bytes32_t const &k)
+        : orig_value{b}
+        , key{k}
+    {
+    }
+
+    // void operator=(deleted_key const &d) { key = d.key; }
+    void operator=(bytes32_t const &b) { key = b; }
```
zander-xyz: So this is someone changing the key that gets deleted, but the original value remains the same? How is that possible?

include/monad/db/state.hpp
```diff
@@ -0,0 +1,217 @@
...
+MONAD_DB_NAMESPACE_BEGIN
+
+template <class TAccounts, class TValueStore, class TCodeStore>
+struct State
+{
+    struct WorkingCopy
+    {
+        typename TAccounts::WorkingCopy accounts_;
+        TValueStore storage_;
+        typename TCodeStore::WorkingCopy code_;
```
zander-xyz: Does gcc complain about a missing initializer here?

include/monad/db/state.hpp
```diff
@@ -0,0 +1,217 @@
...
+            unsigned int i, typename TAccounts::WorkingCopy &&a,
+            TValueStore &&s, typename TCodeStore::WorkingCopy &&c)
+            : accounts_{std::move(a)}
+            , storage_{std::move(s)}
+            , code_{std::move(c)}
+            , id_{i}
+        {
+        }
+
+        unsigned int id() const noexcept { return id_; }
```
zander-xyz: is id the current txn? If so, can we rename it appropriately?

include/monad/db/value_store.hpp
```diff
@@ -0,0 +1,345 @@
+#pragma once
```
zander-xyz, line 1: Personally not the biggest fan of the rename to value_store, but I won't make a big fuss about it. 

General request here to make this look like the WorkingCopy design you have for account and code

I'll review this file in more depth once we move over to that design.

include/monad/db/value_store.hpp
```diff
@@ -0,0 +1,345 @@
...
+        touched_.storage_.at(a).erase(key);
+        if (touched_.storage_.at(a).empty()) {
+            touched_.storage_.erase(a);
+        }
+    }
+
+    bool db_contains_key(
+        address_t const &a, bytes32_t const &key) const noexcept
+    {
+        return db_.contains(a) && db_.at(a).contains(key);
```
zander-xyz, line 215: In the future, we may create a db wrapper that takes in an `address_t` and a `bytes32_t` and returns true or false. This gives the wrapper the flexibility of crafting the raw database key rather than adhering strictly to a map interface.

include/monad/db/value_store.hpp
```diff
@@ -0,0 +1,345 @@
...
+                deleted_key, deleted_key::hash, deleted_key::equality>>
+            deleted_storage_{};
+
+        bool
+        contains_key(address_t const &a, bytes32_t const &key) const noexcept
+        {
+            return storage_.contains(a) && storage_.at(a).contains(key);
+        }
+
+        bool deleted_db_contains_key(
```
zander-xyz: nit
```suggestion
        bool deleted_contains_key(
```

include/monad/db/value_store.hpp
```diff
@@ -0,0 +1,345 @@
...
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    void commit_all_merged()
+    {
+        MONAD_ASSERT(can_commit());
```
zander-xyz: this need not be a MONAD_ASSERT

include/monad/db/account_store.hpp
sheapwilliams: All true, but I left that cause that is the signature of the `HostInterface` in EVMC...  Technically, this isn't inheriting so it could just be removed.  But it will be somewhere else up the line...not quite sure what the right answer is there.

include/monad/db/account_store.hpp
sheapwilliams: creation race/collision between previously ordered transactions...That will fail in line 80 of evmc.hpp where EIP-684 checks detect this.

include/monad/db/account_store.hpp
sheapwilliams: Y - https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md#specification, part d.

include/monad/db/account_store.hpp
zander-xyz: https://github.com/ethereum/evmc/blob/master/include/evmc/evmc.hpp#L80

That points to a comment here?

include/monad/db/account_store.hpp
zander-xyz: You have no choice but to keep it at the top level function, so it seems reasonable to have it there but not here.

include/monad/db/account_store.hpp
sheapwilliams: I think it is:
```c++
return std::ranges::all_of(diffs.changed_, [&](auto const &p) {
    if (p.second.orig.has_value()) {
        return get_committed_storage(p.first) == p.second;
    }
    return not committed_storage_contains(p.first);
});
```
But I can't get this to work for some reason.  I'll play with it.

include/monad/db/code_store.hpp
sheapwilliams: Correct.

include/monad/db/account_store.hpp
sheapwilliams: This found a testing bug. 🎉

include/monad/db/account_store.hpp
zander-xyz: Ah, yeah apologies. But shouldn't it be
```c++
return std::ranges::all_of(diffs.changed_, [&](auto const &p) {
    return p.second.orig == get_committed_storage(p.first);
});
```
If `!p.second.orig.has_value()` then the predicate will only be true iff get_committed_storage returns a nullopt

include/monad/db/account_store.hpp
sheapwilliams: Sure, but I don't think that fixes anything.  All the top level function does is call this...same core issue tho, yea?  I'm fine to remove it here, but I don't think it is a fix.

include/monad/db/account_store.hpp
sheapwilliams: Here: https://github.com/monad-crypto/monad/blob/main/include/monad/execution/evm.hpp#L80

I misspelled...sorry about that.

include/monad/db/state.hpp
sheapwilliams: It is the transaction id - I can hame it that.

include/monad/db/account_store.hpp
zander-xyz: It's not a fix, you're right, but it is _more correct_.

include/monad/db/account_store.hpp
zander-xyz: Okay, so it should never return anything other than a `std::nullopt`. So can we assert that `get_commited_storage(a)` is that and just replace the first argument with `std::nullopt` directly?

sheapwilliams: Again, not quite.  We expect creation races when running optimistically - they are part of our design.  We don't want to short-circuit any detection there.

include/monad/db/code_store.hpp
sheapwilliams: test/debug code - only for testing invariants, and added as `assert(...)` to `merge_changes`.

include/monad/db/code_store.hpp
sheapwilliams: I clear `merged_` right after this, so I'm moving it now.

include/monad/db/datum.hpp
sheapwilliams: You can change storage values (and account values to a lesser degree, mostly just nonce after the initial balance changes) all you want.  You can update keys several times per contract if you wanted.  Orig will always be the exact same tho - it captures what is in storage when you were optimistically executed, which won't change no matter how many times storage is updated.  Yea?

include/monad/db/account_store.hpp
zander-xyz: What does it mean to have two separate transactions that create the same contract? Is that possible?

include/monad/db/datum.hpp
zander-xyz: Updating the storage in this context is just deleting the key right? If you change the key that gets deleted, won't the original value be different? For example "I want to delete key 1" and "I want to delete key 2" sounds like we should have two separate `deleted_key` objects with the original value of the key and their respective keys. It seems unintuitive to have a deleted_key for key 1 and then reassign the deleted_key to key 2 but keep the original value of key 1. 

I'm not understanding when you would want to change the key of a deleted key but not the original value

include/monad/db/state.hpp
sheapwilliams: Pleasantly surprised...no.  It does not. :+1:

include/monad/db/account_store.hpp
sheapwilliams: Discussed in person.

include/monad/db/value_store.hpp
```diff
@@ -0,0 +1,342 @@
...
+    {
+        for (auto const &[a, key_set] : diffs.touched_.deleted_storage_) {
+            for (auto const &k : key_set) {
+                if (k.orig_value != get_merged_value(a, k.key)) {
+                    return false;
+                }
+
+                if (merged_.deleted_contains_key(a, k.key)) {
+                    return false;
+                }
```
zander-xyz:
```suggestion
                if (k.orig_value != get_merged_value(a, k.key)) {
                    return false;
                }
```

If the first predicate happens, why even check the second?

PR #85, opened 2023-05-31
:recycle: Refactor EvmcHost object to have single responsibility

sheapwilliams:
Problem:
- EvmcHost is the required interface for EVMOne, and it requires many different responsibilities.
- EvmcHost currently is responsible for some EVM functions, and the TEvm object for the others.
- Current split of EVM logic isn't defensible.

Solution:
- Make EvmcHost more SRP by moving all EVM functionality into the Evm object.
- Move precomplies from being injected into EvmcHost to injected into Evm
- Save `get_tx_context`, EvmcHost is now just a composition object that delegates functionality.
- Remove unnecessary inline keyword.


include/monad/execution/evm.hpp
```diff
@@ -12,13 +12,58 @@
...
+    class TStaticPrecompiles>
 struct Evm
 {
     using result_t = tl::expected<evmc_message, evmc_result>;
     using new_address_t = tl::expected<address_t, evmc_result>;
     using unexpected_t = tl::unexpected<evmc_result>;
 
+    template <class TEvmHost>
+    [[nodiscard]] static evmc::Result
+    create_contract_account([[maybe_unused]] TEvmHost *host, TState &state, evmc_message const &m) noexcept
```
zander-xyz: You could do the following to avoid the `[[maybe_unused]]`, which is really a misnomer here, and make it clear that host is not being used
```suggestion
    create_contract_account(TEvmHost *, TState &state, evmc_message const &m) noexcept
```
Ditto below

include/monad/execution/evm.hpp
```diff
@@ -12,13 +12,58 @@
...
+    {
+        auto const contract_address = make_account_address(state, m);
```
zander-xyz, line 28: Side note: is there any reason to use `tl::expected` at this point? We recently moved to clang 16 which should support `std::expected`. Thoughts on just using that moving forward?

include/monad/execution/evm.hpp
```diff
@@ -12,13 +12,58 @@
...
+            return evmc::Result{result};
+        }
+        evmc_result const result =
+            TStaticPrecompiles::static_precompile_exec_func(m.code_address)
+                .transform([&](auto static_precompile_execute) {
+                    return static_precompile_execute(m);
+                })
+                // execute on backend, just this for now
+                .value_or(evmc_result{
+                    .status_code = EVMC_SUCCESS, .gas_left = m.gas});
```
zander-xyz, line 58: Wait, so we return success if `TStaticPrecompiles::static_precompile_exec_func` returns a `nullopt`? Is that intended?

include/monad/execution/replay_block_db.hpp
```diff
@@ -145,9 +145,9 @@ class ReplayFromBlockDb
     template <
         concepts::fork_traits<TState> TTraits,
         template <typename, typename> class TTxnProcessor,
-        template <typename, typename> class TEvm,
+        template <typename, typename, typename> class TEvm,
```
zander-xyz, line 148: TEvm is never used? Either we remove it or the `TEvmHost` needs to take in TEvm instead of Evm directly below.

include/monad/execution/evm.hpp
rgarc: I can't get this `std::expected` snippet to compile D:
https://godbolt.org/z/jd6s48Gva

zander-xyz: It works with `-stdlib=libc++`, looks like libc++ is not the system default on whatever box godbolt is running on.

include/monad/execution/evm.hpp
sheapwilliams: Sure.

include/monad/execution/evm.hpp
sheapwilliams: That is a stub for the interpreter, and that is what the comment above is trying to point out.  It is removed in my WIP interpreter work.

include/monad/execution/replay_block_db.hpp
sheapwilliams: That is an oversight.  It should be used there and will be fixed when Tong begins integrating in the next steps.

include/monad/execution/evm.hpp
sheapwilliams: I think we want to remove it when we turn on C++23 and get full library support for the monadic extensions we use in both compilers.  Not until then tho.

zander-xyz: Clang 17 it is...

PR #86, opened 2023-06-01
Logger and Behave

tongzhi1998:
- Add `Quill logger` for logging; Have the flexibility of turning it completely off or log in different levels
- Add Integration tests in using `Gherkin` and Behave for `ReplayEthOnlyBlockDb`
- To run behave, currently have to do: `pip install behave`
- Use `black` for python format and `flake8` for linter

Prev #79

### DIFF
--- a/features/ethereum/read_block_history.feature
+++ b/features/ethereum/read_block_history.feature
@@ -1,0 +1,42 @@
+Feature: Read Ethereum Block History
+
+  Scenario: Run replay_ethereum_block_db with invalid end block number
+    Given I run replay_ethereum_block_db with BlockDb path = "erigon/block_db"
+    And I run replay_ethereum_block_db with start block number = "1"
+    And I run replay_ethereum_block_db with finish block number = "1"
+    When I start "replay_ethereum_block_db"
+    Then the output should not contain any block level information
+
+  Scenario: Run replay_ethereum_block_db with correct configuration, 1 Block, no hash validation
+    Given I run replay_ethereum_block_db with BlockDb path = "erigon/block_db"
+    And I run replay_ethereum_block_db with start block number = "1"
+    And I run replay_ethereum_block_db with finish block number = "2"
+    When I start "replay_ethereum_block_db"
+    Then the output should contain "1" "Computed Transaction Root"
+    And the output should contain "1" "Expected Transaction Root"
+    And the output should contain "1" "Computed Receipt Root"
+    And the output should contain "1" "Expected Receipt Root"
+    And the output should contain "1" "Computed State Root"
+    And the output should contain "1" "Expected State Root"
+
+  Scenario: Run replay_ethereum_block_db with correct configuration, 1000 Blocks, no hash validation
+    Given I run replay_ethereum_block_db with BlockDb path = "erigon/block_db"
+    And I run replay_ethereum_block_db with start block number = "10"
+    And I run replay_ethereum_block_db with finish block number = "1010"
+    And I run replay_ethereum_block_db with log level = "4"
+    When I start "replay_ethereum_block_db"
+    Then the output should contain "1000" "Computed Transaction Root"
+    And the output should contain "1000" "Expected Transaction Root"
+    And the output should contain "1000" "Computed Receipt Root"
+    And the output should contain "1000" "Expected Receipt Root"
+    And the output should contain "1000" "Computed State Root"
+    And the output should contain "1000" "Expected State Root"
+
+  Scenario: Run replay_ethereum_block_db with correct configuration, 1 Block, with hash validation
+    Given I run replay_ethereum_block_db with BlockDb path = "erigon/block_db"
+    And I run replay_ethereum_block_db with start block number = "1"
+    And I run replay_ethereum_block_db with finish block number = "2"
+    When I start "replay_ethereum_block_db"
+    Then the "Transaction Root" should match
+    And the "Receipt Root" should match
+    And the "State Root" should match
--- a/features/ethereum/steps/read_block_history.py
+++ b/features/ethereum/steps/read_block_history.py
@@ -1,0 +1,70 @@
+import os
+import re
+import subprocess
+from behave import given, when, then
+
+
+@given('I run replay_ethereum_block_db with BlockDb path = "{block_db_path}"')
+def given_block_db_path(context, block_db_path):
+    context.block_db_path = block_db_path
+
+
+@given(
+    'I run replay_ethereum_block_db with start block number = "{start_block_number}"'
+)
+def given_start_block_number(context, start_block_number):
+    context.start_block_number = start_block_number
+
+
+@given(
+    'I run replay_ethereum_block_db with finish block number = "{finish_block_number}"'
+)
+def given_finish_block_number(context, finish_block_number):
+    context.finish_block_number = finish_block_number
+
+
+@given('I run replay_ethereum_block_db with log level = "{log_level}"')
+def given_log_level(context, log_level):
+    context.log_level = log_level
+
+
+@when('I start "{program}"')
+def when_start(context, program):
+    executable_path = os.getcwd() + "/build/src/monad/execution/ethereum/" + program
+    if not hasattr(context, "block_db_path"):
+        raise ValueError("Block Db path is required")
+    if not hasattr(context, "start_block_number"):
+        raise ValueError("Start block number is required")
+    args = [
+        executable_path,
+        "--block-db",
+        os.getcwd() + "/../../" + context.block_db_path,
+        "--start",
+        context.start_block_number,
+    ]
+    if hasattr(context, "finish_block_number"):
+        args += ["--finish", context.finish_block_number]
+    if hasattr(context, "log_level"):
+        args += ["--log-level", context.log_level]
+    result = subprocess.run(args, capture_output=True, text=True)
+    context.output = result.stdout
+
+
+@then('the output should contain "{number}" "{value}"')
+def then_output_contain(context, number, value):
+    assert context.output.count(value) == int(number)
+
+
+@then("the output should not contain any block level information")
+def then_output_empty(context):
+    assert "block_logger" not in context.output
+
+
+@then('the "{root_type}" should match')
+def then_root_match(context, root_type):
+    lines = context.output.splitlines()
+
+    for line in lines:
+        if root_type in line:
+            hashes = re.findall(r"\[([^\]]*)\]", line)
+            assert hashes[1] == hashes[2]


conan/conanfile.txt
```diff
@@ -1,6 +1,7 @@
 [requires]
 tl-expected/1.0.0
 tl-optional/1.0.0
+quill/[>=1.2.3]
```
tongzhi1998: For the version, I am using >= [Earliest version on https://conan.io/center/]
Same for the CLI11 package later

CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ function(monad_compile_options target)
     set_property(TARGET ${target} PROPERTY C_STANDARD_REQUIRED ON)
     set_property(TARGET ${target} PROPERTY CXX_STANDARD 20)
     set_property(TARGET ${target} PROPERTY CXX_STANDARD_REQUIRED ON)
-    target_compile_options(${target} PRIVATE -Wall -Wextra -Werror -Wconversion -Wpedantic)
+    target_compile_options(${target} PRIVATE -Wall -Wextra -Werror -Wconversion -Wpedantic -Wno-gnu-zero-variadic-macro-arguments)
```
tongzhi1998: https://github.com/monad-crypto/monad/pull/86/commits/5fdbc7111e342fe5978a7495c0a6f03f91d22440#diff-1d8eb1e00ed7bce28c0c8e3bd8000fdf63907942a6dd26c47492ba57196189e2R3
This VA_ARGS seems to be an GNU extension that is not part of C++ standard. As a result, with -Wall, I think Clang is not happy with it. I think we can have 2 ways to resolve this:

1. The one I used, adding -Wno-gnu-zero-variadic-macro-arguments
2. Do something like: target_compile_options (${target} PUBLIC $<$<CXX_COMPILER_ID:CLANG>:-Wno-gnu-zero-variadic-macro-arguments>)

Let me know what you think of this.

include/monad/execution/replay_block_db.hpp
```diff
@@ -2,22 +2,30 @@
...
 #include <optional>
-#include <ostream>
+
+#ifdef FAKELOGGER
+    #include <monad/logging/fake_log_macros.hpp>
+    #include <monad/logging/fake_logger.hpp>
+#else
+    #include <monad/logging/quill_log_macros.hpp>
+    #include <monad/logging/quill_logger.hpp>
+#endif
```
sheapwilliams: Let's put this ifdef in a separate header file, and every user can just simply use the header.  Will make it easier to evolve if this isn't throughout the codebase.

conan/conanfile.txt
tongzhi1998: Used latest version. Subscribed to quill update mailing list

CMakeLists.txt
```diff
@@ -121,6 +122,14 @@ add_library(monad STATIC
     ${PROJECT_SOURCE_DIR}/include/monad/fiber/properties.hpp
     ${PROJECT_SOURCE_DIR}/src/monad/fiber/algorithm.cpp
 
+    # logging
+    ${PROJECT_SOURCE_DIR}/include/monad/logging/config.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/logging/fake_logger.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/logging/fake_log_macros.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/logging/include_switch.hpp
```
sheapwilliams: Let's just call this `include/monad/logging/monad_log.hpp`.

include/monad/execution/replay_block_db.hpp
```diff
@@ -102,14 +103,31 @@ class ReplayFromBlockDb
...
-                output << "State Root: " << state_root.bytes << "\n";
+                [[maybe_unused]] auto block_logger =
+                    TLogger::get_logger("block_logger");
+
+                [[maybe_unused]] auto const transaction_root =
+                    transaction_trie.root_hash();
+                [[maybe_unused]] auto const receipt_root =
+                    receipt_trie.root_hash();
+                [[maybe_unused]] auto const state_root =
+                    state_trie.incremental_update(state);
```
sheapwilliams, line 114: Let's have the `MONAD_LOG_...` macros suppress the usage of the variables as we discussed in the execution channel instead of marking the l-values here as `[[maybe_unused]]`.

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include <monad/logging/config.hpp>
+
+#include <quill/Quill.h>
```
sheapwilliams: The fake doesn't need quill...Just define fakes for the `quill::Logger*`, etc.

src/monad/execution/ethereum/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+find_package(CLI11 REQUIRED)
+
+add_executable(replay_ethereum_only_block_db replay_ethereum_only_block_db.cpp)
```
sheapwilliams: [nit] can we remove the `..._only_...` part from these file and class names?

src/monad/execution/ethereum/replay_ethereum_only_block_db.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+#include <monad/execution/block_processor.hpp>
+#include <monad/execution/evmc_host.hpp>
+#include <monad/execution/replay_block_db.hpp>
+#include <monad/execution/static_precompiles.hpp>
+#include <monad/execution/test/fakes.hpp>
+
+// Substitute these includes to get a different logger
+#include <monad/logging/quill_log_macros.hpp>
+#include <monad/logging/quill_logger.hpp>
```
sheapwilliams: Let's use the monad_log.hpp file here instead, yea?

src/monad/execution/ethereum/replay_ethereum_only_block_db.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+#include <CLI/CLI.hpp>
+#include <filesystem>
+#include <quill/Quill.h>
```
sheapwilliams: We should group the standard library separate from vendor'ed packages with a space.

src/monad/execution/ethereum/test/replay_eth_block_db.cpp
```diff
@@ -6,6 +6,8 @@
 
 #include <monad/execution/test/fakes.hpp>
 
+#include <monad/logging/fake_logger.hpp>
+
```
sheapwilliams, line 11: Same here - use the monad_log.hpp file.

src/monad/execution/test/replay_block_db.cpp
```diff
@@ -6,6 +6,8 @@
 
 #include <monad/execution/test/fakes.hpp>
 
+#include <monad/logging/fake_logger.hpp>
+
```
sheapwilliams, line 10: Same - throughout...

features/read_from_block_db.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Read from BlockDb
```
sheapwilliams:
```suggestion
Feature: Read Ethereum Block History
```

Also, let's put this feature in an ethereum subdirectory: `features/ethereum/read_block_history.feature` or something...

features/steps/read_from_block_db.py
```diff
@@ -0,0 +1,45 @@
...
+def given_finish_block_number(context, finish_block_number):
+    context.finish_block_number = finish_block_number
+
+@given(u'I run ReplayEthereumOnlyBlockDb with log level = "{log_level}"')
+def given_log_level(context, log_level):
+    context.log_level = log_level
+
+@when('I start ReplayEthereumOnlyBlockDb')
+def when_start(context): 
+    executable_path = os.getcwd() + '/build/src/monad/execution/ethereum/replay_ethereum_only_block_db'
```
sheapwilliams: This suffers from the same fragility of location as the other tests do.  Let's engineer a way to make this work from anywhere, either by cmake configuration or environmental.  If we choose environmental, then we can do this when we do cmake integration.

features/read_from_block_db.feature
```diff
@@ -0,0 +1,42 @@
...
```
zander-xyz: nit: can we put these feature files in a nested `behave/features` instead?

CMakeLists.txt
tongzhi1998: Used pragma push & pragma pop to solve the issue locally

include/monad/execution/replay_block_db.hpp
```diff
@@ -102,14 +103,33 @@ class ReplayFromBlockDb
...
 
-                auto const transaction_root = transaction_trie.root_hash();
-                auto const receipt_root = receipt_trie.root_hash();
-                auto const state_root = state_trie.incremental_update(state);
-
-                output << "Transaction Root: " << transaction_root.bytes
-                       << "\n";
-                output << "Receipt Root: " << receipt_root.bytes << "\n";
-                output << "State Root: " << state_root.bytes << "\n";
+                auto block_logger = TLogger::get_logger("block_logger");
```
zander-xyz: Prefer not to hide the pointiness of the type in `auto`
```suggestion
                auto* block_logger = TLogger::get_logger("block_logger");
```

include/monad/execution/replay_block_db.hpp
zander-xyz: Agree with the above.

Let's also write a custom formatter for `bytes32_t` so that you don't have to type `x.bytes` every single time.

include/monad/execution/replay_block_db.hpp
```diff
@@ -102,14 +103,33 @@ class ReplayFromBlockDb
...
+                    state_trie.incremental_update(state);
+
+                MONAD_LOG_INFO(block_logger, "Block {}", current_block_number);
+
+                MONAD_LOG_INFO(
+                    block_logger,
+                    "Computed Transaction Root: {}, Expected Transaction Root: "
+                    "{}",
+                    transaction_root.bytes,
+                    block.header.transactions_root.bytes);
```
zander-xyz: What does this log actually appear as in the output?

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return nullptr;
+    }
+
+    static void *create_logger(
+        [[maybe_unused]] std::string const &logger_name,
+        [[maybe_unused]] std::optional<FakeTimestampClockType *>
+            timestamp_clock_type = std::nullopt,
+        [[maybe_unused]] std::optional<FakeTimestampClock *> timestamp_clock =
+            std::nullopt)
```
zander-xyz: Here and everywhere else please, to more accurately reflect intention: these are not [[maybe_unused]], these are definitely not used.
```suggestion
        std::string const &,
        std::optional<FakeTimestampClockType *> = std::nullopt,
        std::optional<FakeTimestampClock *> = std::nullopt)
```

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,65 @@
...
+};
+
+enum class FakeLogLevel
+{
+    FAKELEVEL
+};
+
+struct FakeEmptyLogger
+{
+    static void start() {}
```
zander-xyz: Can we `constexpr` all of the things?

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return nullptr;
+    }
+
+    static void set_log_level(
+        [[maybe_unused]] char const *logger_name = nullptr,
+        [[maybe_unused]] FakeLogLevel log_level = FakeLogLevel::FAKELEVEL)
+    {
+        return;
+    }
```
zander-xyz: Where do you actually need this stub to exist?

include/monad/logging/monad_log.hpp
```diff
@@ -0,0 +1,7 @@
+#ifdef FAKELOGGER
```
zander-xyz: Can we name this something more expressive? Maybe like
```c++
#ifdef ENABLE_LOGGING
/*real quill logger*/
#else
/*fake*/
#endif
```

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+        ->required();
+    auto finish_block_number_opt = cli.add_option(
+        "-f, --finish", finish_block_number, "1 pass the last executed block");
+
+    // Should support different log-level for different part of program later
+    // on; for simplicity, now only supports 1 log level throughout the program
+    cli.add_option("-l, --log-level", log_level, "level of logging");
+    cli.parse(argc, argv);
+
+    using logger_t = monad::log::QuillLogger;
```
zander-xyz: This should be grabbing the logger from `monad_log.hpp` instead of directly referencing `QuillLogger`. My suspicion is that once we do this, many other things will be uncovered.

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
zander-xyz, line 1: Side note: have we confirmed that this all compiles away?

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,65 @@
...
+struct FakeHandler
+{
+};
+
+enum class FakeLogLevel
+{
+    FAKELEVEL
+};
+
+struct FakeEmptyLogger
```
zander-xyz, line 25: `noexcept` where appropriate please (which should be everything in this file)

include/monad/logging/quill_logger.hpp
```diff
@@ -0,0 +1,52 @@
...
+struct QuillLogger
+{
+    static void start() { quill::start(); }
+
+    static quill::Logger *get_logger(char const *logger_name = nullptr)
+    {
+        return quill::get_logger(logger_name);
+    }
+
+    static quill::Logger *create_logger(
```
zander-xyz: ``[[nodiscard]]`` here and everywhere in this file please

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+using eth_start_fork = fork_traits::frontier;
+
+template <class TTraits, class TState, class TEvm, class TStaticPrecompiles>
+struct fakeEvmHost
+{
+    evmc_result _result{};
+    Receipt _receipt{};
+
+    [[nodiscard]] static constexpr inline evmc_message
+    make_msg_from_txn(Transaction const &)
```
zander-xyz: this can be marked noexcept

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+    {
+        return {.kind = EVMC_CALL};
+    }
+
+    [[nodiscard]] constexpr inline Receipt make_receipt_from_result(
```
zander-xyz, line 37: this too

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+    template <class TEvmHost>
+    Receipt execute(
+        TState &, TEvmHost &, BlockHeader const &, Transaction const &) const
+    {
+        return {};
+    }
+
+    Status validate(TState const &, Transaction const &, uint64_t)
+    {
+        return {};
```
zander-xyz: Can we be explicit in returning `SUCCESS` here? Also, noexcept in this class too. In general, it's good to try to give the compiler as many hints as possible. Realistically, the compiler will be able to optimize this function away, but still a good habit to get into.

Will stop commenting on these fakes now.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+    monad::block_num_t start_block_number;
+    monad::block_num_t finish_block_number;
+    quill::LogLevel log_level = quill::LogLevel::Info;
+
+    cli.add_option("-b, --block-db", block_db_path, "block_db directory")
+        ->required();
+    cli.add_option("-s, --start", start_block_number, "start block numer")
+        ->required();
+    auto finish_block_number_opt = cli.add_option(
+        "-f, --finish", finish_block_number, "1 pass the last executed block");
```
zander-xyz: If we make `finish_block_number` a `std::optional`, can we do away with the `finish_block_number_opt ` and just look at `finish_block_number` directly?

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    logger_t::start();
+
+    // create all the loggers needed for the program
+    [[maybe_unused]] auto *main_logger = logger_t::create_logger("main_logger");
+    [[maybe_unused]] auto *block_logger =
+        logger_t::create_logger("block_logger");
+    [[maybe_unused]] auto *txn_logger = logger_t::create_logger("txn_logger");
+    [[maybe_unused]] auto *state_logger =
+        logger_t::create_logger("state_logger");
```
zander-xyz: These are definitely unused, so the lvalue is not needed? This probably actually means that you don't want the nodiscard on these functions.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+    else {
+        MONAD_LOG_INFO(
+            main_logger,
+            "Running with block_db = {}, start block number = {}, finish block "
+            "number = {}",
+            block_db_path,
+            start_block_number,
+            finish_block_number);
+    }
```
zander-xyz: Are `std::optional`s not loggy? If not, we should make it so as long as the value type is also loggy. Then you can collapse these two logs into one.

```cpp
        MONAD_LOG_INFO(
            main_logger,
            "Running with block_db = {}, start block number = {}, finish block "
            "number = {}",
            block_db_path,
            start_block_number,
            finish_block_number);
```

For the optional formatting, maybe something like
```cpp
Ok{...} // for engaged optionals
Nullptr{} // for unengaged
```

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,240 @@
...
+    else {
+        [[maybe_unused]] auto result = replay_eth.run<
+            monad::eth_start_fork,
+            monad::fakeEmptyTP,
+            monad::fakeEmptyEvm,
+            monad::execution::StaticPrecompiles,
+            monad::fakeEmptyEvmHost,
+            monad::fakeEmptyFiberData>(
+            state, state_trie, block_db, receipt_collector, start_block_number);
+    }
```
zander-xyz: These two can be collapsed once we make finish_block_number a std::optional.

Also, should we be asserting the result of the run? It is [[nodiscard]] afterall...

cmake/test.cmake
```diff
@@ -14,5 +14,7 @@ function(add_unit_test)
         GTest::Main
         ${ADD_UNIT_TEST_LIBRARIES}
     )
+
+    target_compile_definitions(${ADD_UNIT_TEST_TARGET} PRIVATE FAKELOGGER)
```
zander-xyz: What if I do want logging turned on for my unit tests? Thoughts on making this user specified instead of hardcoded?

src/monad/execution/ethereum/test/replay_eth_block_db.cpp
```diff
@@ -162,29 +164,29 @@ using eth_start_fork = fork_traits::frontier;
 
 using state_t = execution::fake::State;
 using receipt_collector_t = std::vector<std::vector<Receipt>>;
+using logger_t = log::FakeEmptyLogger;
```
zander-xyz: We shouldn't have any of these `logger_t`'s anywhere in our codebase. The type of the logger should be directly inferred from `monad_log.hpp` (that's why that file exists).

src/monad/execution/test/replay_block_db.cpp
```diff
@@ -155,28 +157,29 @@ struct fakeEmptyFiberData
 using state_t = execution::fake::State;
 using traits_t = execution::fake::traits::alpha<state_t>;
 using receipt_collector_t = std::vector<std::vector<Receipt>>;
+using logger_t = log::FakeEmptyLogger;
```
zander-xyz: ditto

features/ethereum/read_block_history.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Read Ethereum Block History
```
zander-xyz, line 1: nit: can we please avoid pronouns in our feature tests?

For example, instead of 

> I run ReplayEthereumBlockDb with BlockDb path = "erigon/block_db"

prefer

> ReplayEthereumBlockDb is executed with BlockDb path = "erigon/block_db"

include/monad/logging/fake_log_macros.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#define MONAD_LOG_TRACE_L3(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_TRACE_L2(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_TRACE_L1(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_DEBUG(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_INFO(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_WARNING(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_ERROR(logger, fmt, ...) (void *)(logger)
+#define MONAD_LOG_CRITICAL(logger, fmt, ...) (void *)(logger)
```
zander-xyz: Any reason to do `(void*)(logger)` instead of `((void)0)`. I'm nervous about `logger` still being evaluated in this context.

features/steps/read_from_block_db.py
zander-xyz: Please not environmental!

include/monad/logging/fake_log_macros.hpp
tongzhi1998: The reason for doing so is to avoid marking logger as [[maybe_unused]].

include/monad/logging/fake_logger.hpp
tongzhi1998: You mean the `set_log_level` function? It is needed since Quill's way of using `set_log_level ` is `logger->set_log_level(LOGLEVEL)`. However, since our fake logger is sometimes a nullptr, we can't call `nullptr->set_log_level()`. So I just wrap it in a function like this.

zander-xyz: Right, but where do you currently call it?

include/monad/logging/fake_log_macros.hpp
zander-xyz: Gotcha, that makes sense. Can you look at MONAD_DEBUG_ASSERT to see how we do it there? That guarantees that everything gets compiled away while also giving us the luxury of not needing the [[maybe_unused]] everywhere.

include/monad/execution/replay_block_db.hpp
sheapwilliams: @zander-xyz Are you volunteering to write a custom formatter? 😄  Agree it would be super useful, I've never done one tho.  My impulse is that should be in a separate PR.

features/steps/read_from_block_db.py
sheapwilliams: We can use the same cmake `configure_file` trick, but it is a little trickier to get python paths setup correctly.  Not a big deal tho.

features/read_from_block_db.feature
sheapwilliams: N - these feature files are not specific to behave.  You can animate feature files with lots of different things.

include/monad/execution/replay_block_db.hpp
zander-xyz: It's just overloading operator<<, per the quill documentation.

features/read_from_block_db.feature
zander-xyz: Ah, that's right. I forgot about GUnit!

include/monad/logging/fake_logger.hpp
sheapwilliams: _Way_ too overstated I think...The C++ naming is terrible (no defending `maybe_unused`), but there are 2 schools of thought: 1) Name your parameters to give them meaning, and mark `[[maybe_unused]]` if unused for any reason - naming is as important as any kind of semantic meaning we can give it, so don't shirk that duty. 2) Obey the precise syntax of `maybe_unused` and don't name your parameters, and don't worry about conveying semantic information.

I've worked on different project that used both systems, and they both have value.

That said, I'm ok with the 2nd system on this project.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
sheapwilliams: I kinda agree.  Something like:
```suggestion
    using logger_t = monad::log::logger_t;
```
and `monad::log::logger_t` is defined in the header file based on the `ENABLE_LOGGING` switch.  :+1:

That way there is a single way to configure logging in ur translation unit.

include/monad/logging/fake_logger.hpp
zander-xyz: Yeah, I agree with that. There are cases where the type is enough to determine what the variable is

Example: `std::optional<FakeTimestampClockType>` is clearly an optional fake timestamp clock type
Counter-example: `std::string const&` could be anything!

So yeah, I was definitely overstated with my original comment. Use your best judgement, as always.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
zander-xyz: +1

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
sheapwilliams: I think we should log the result of the run. I don't expect clean runs all the time, and I don't want the program terminating if it is unclean.

features/ethereum/read_block_history.feature
sheapwilliams: https://cucumber.io/docs/gherkin/reference/

Pronouns are widely accepted in gherkin-style testing...I don't see any particular reason to avoid them.

That said, if anyone wants to help out and comment with good replacements for any of the lines that's cool.  I wouldn't sweat it too much otherwise.

sheapwilliams: I think the reason `I run` or `I start` is so popular is that it is easier for test drivers to match the phrase.  If you start with a wildcard, it makes it impossible to match correctly.

cmake/test.cmake
sheapwilliams: It will be easy enough to extend the `add_unit_test` function to add a logging parameter when we want one.

features/ethereum/read_block_history.feature
zander-xyz: Yeah, whatever we land on should be fine. I'm not particularly tied to a particular prose, I just have personal opinions on what reads better. But I can deal with that.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
zander-xyz: Why not?

include/monad/logging/fake_logger.hpp
tongzhi1998: Oh, sorry it is used at a later commit. [https://github.com/monad-crypto/monad/pull/86/files#diff-a7ed8486d3449b40e0787a8fad5d0f053acc765ad3c57f59af2f186b4672ad05R174](url)

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
tongzhi1998: Yes, std::optional is not loggable, so that's the reason why I didn't mark `finish_block_number` as `std::optional.` However, I think your suggestion makes sense and I will see what I can do for it.

include/monad/execution/replay_block_db.hpp
tongzhi1998: 12:19:37.429681677 [918967] replay_block_db.hpp:122      LOG_INFO      block_logger Computed Transaction Root: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], Expected Transaction Root: [86, 232, 31, 23, 27, 204, 85, 166, 255, 131, 69, 230, 146, 192, 248, 110, 91, 72, 224, 27, 153, 108, 173, 192, 1, 98, 47, 181, 227, 99, 180, 33]

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
tongzhi1998: These are actually used in the `set_log_level` function below. What `set_log_level` does is that it firsts gets the logger using its main and then set the appropriate log_level for it.

include/monad/logging/monad_log.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#ifdef ENABLE_LOGGING
+    #include <monad/logging/quill_log_macros.hpp>
+    #include <monad/logging/quill_logger.hpp>
+using logger_t = monad::log::QuillLogger;
```
tongzhi1998: The formatting here (the using line w/o indentation) looks weird, but clang-format suggests this

CMakeLists.txt
```diff
@@ -168,9 +177,12 @@ target_link_libraries(monad
     PUBLIC intx::intx
     PUBLIC tl::expected
     PUBLIC tl::optional
+    PUBLIC quill::quill
 )
 monad_compile_options(monad)
 
+target_compile_definitions(monad PRIVATE ENABLE_LOGGING)
```
sheapwilliams: Is `PRIVATE` the right visibility here?  Also, I don't think we want to force everyone to use logging that is linking against the monad library, yea?

CMakeLists.txt
```diff
@@ -182,3 +194,5 @@ elseif(UBSAN)
 else()
   target_link_options(monad PRIVATE LINKER:-z,defs)
 endif()
+
+add_subdirectory(${PROJECT_SOURCE_DIR}/src/monad/execution/ethereum)
```
sheapwilliams: Any reason not to put these up with the other `add_subdirectory` directives?

features/ethereum/steps/read_block_history.py
```diff
@@ -0,0 +1,68 @@
...
+
+@given('I run ReplayEthereumBlockDb with log level = "{log_level}"')
+def given_log_level(context, log_level):
+    context.log_level = log_level
+
+
+@when("I start ReplayEthereumBlockDb")
+def when_start(context):
+    executable_path = (
+        os.getcwd() + "/build/src/monad/execution/ethereum/replay_ethereum_block_db"
```
sheapwilliams: As a general note, the name of this program can be derived from the BDD statement:
```suggestion
@when("I start {program}")
def when_start(context):
    executable_path = (
        os.getcwd() + "/build/src/monad/execution/" + program
```
That will generalize the steps to other programs too.

include/monad/execution/replay_block_db.hpp
sheapwilliams: Ah - drop a link next time.  I assumed it was libfmt style custom formatters...

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,58 @@
...
+    static constexpr void *create_logger(
+        std::string const &,
+        std::optional<FakeTimestampClockType *> = std::nullopt,
+        std::optional<FakeTimestampClock *> = std::nullopt) noexcept
+    {
+        return nullptr;
+    }
+
+    static constexpr void *create_logger(
+        std::string const &, std::shared_ptr<FakeHandler> &&,
```
sheapwilliams: [IWYU] Need to include `<memory>` for `shared_ptr`

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,58 @@
...
+        std::optional<FakeTimestampClockType> = std::nullopt,
```
sheapwilliams: [IWYU] And `<optional>`...

include/monad/logging/monad_log.hpp
sheapwilliams: Yea, technically preprocessing macros need to have the hash in column 0 per the spec, but I think every preprocessor conforms to extensions that are more flexible.  I'm fine if this is what `clang-format` is outputting.

include/monad/logging/monad_log.hpp
```diff
@@ -0,0 +1,11 @@
...
+#else
+    #include <monad/logging/fake_log_macros.hpp>
+    #include <monad/logging/fake_logger.hpp>
+using logger_t = monad::log::FakeEmptyLogger;
```
sheapwilliams, line 9: This logger_t just wants to be in the `monad::log` namespace.  Feel free to define a `logger_t` type locally in ur main.cpp as needed.  But we shouldn't pollute the global namespace.  

Looks great otherwise!

include/monad/logging/fake_logger.hpp
sheapwilliams: This is all transparent as written.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
sheapwilliams: Because I don't expect cleanliness for a long while, and I don't want perfection being the enemy of good.

include/monad/execution/replay_block_db.hpp
zander-xyz: Haha, yeah those are not the most ergonomic for sure.

https://quillcpp.readthedocs.io/en/latest/tutorial.html#user-defined-types for future reference, for whenever we decide to make those changes.

include/monad/logging/formatter.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+        template <typename FormatContext>
+        auto format(const std::optional<uint64_t> &opt, FormatContext &ctx)
+        {
+            if (opt.has_value()) {
+                return fmt::format_to(ctx.out(), "{}", opt.value());
+            }
+            else {
+                return fmt::format_to(ctx.out(), "No Value");
+            }
```
zander-xyz: Do you have any thoughts on doing something like the below?
```cpp
Ok{__value__} // for engaged optionals
None{} // for unengaged
```
Couple reasons I can think of to do this:
- Logs are easier to read if the formatting is consistently `<engaged_or_not_engaged>{<value if exists>}` rather than `<value>`or `No Value`.
- As is, engaged `optional<uint64_t>` values are logged exactly the same as non-optional `uint64_t` values. My guess is we'd probably like to delineate between the two since they are different types.

include/monad/logging/formatter.hpp
```diff
@@ -0,0 +1,61 @@
...
+        {
+            for (const auto &byte : value.bytes) {
+                fmt::format_to(ctx.out(), "{:02x}", byte);
+            }
+            return ctx.out();
+        }
+    };
+
+    template <>
+    struct formatter<std::optional<uint64_t>>
```
zander-xyz: If you do something like
```cpp
template <typename T>
struct formatter<std::optional<T>>;
```
you should be able to enable logging for any value T that is also fmt'able.

Godbolt link as an example: https://godbolt.org/z/3sdb4fcKe

If you're feeling fancy you might even define a concept to require this and make the compiler error prettier, but I'll leave that as an exercise to the reader.

src/monad/execution/ethereum/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+find_package(CLI11 REQUIRED)
+
+add_executable(replay_ethereum_block_db replay_ethereum_block_db.cpp)
+monad_compile_options(replay_ethereum_block_db)
+
+target_compile_definitions(replay_ethereum_block_db PRIVATE ENABLE_LOGGING)
```
zander-xyz: You seem to be ifdef'ing on `DISABLE_LOGGING`, so this is extraneous

include/monad/core/bytes.hpp
```diff
@@ -4,6 +4,8 @@
 
 #include <evmc/evmc.hpp>
 
+#include <ostream>
```
zander-xyz: This does not seem to be needed

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,223 @@
...
+        monad::log::logger_t::create_logger("main_logger");
+    [[maybe_unused]] auto *block_logger =
+        monad::log::logger_t::create_logger("block_logger");
+    [[maybe_unused]] auto *txn_logger =
+        monad::log::logger_t::create_logger("txn_logger");
+    [[maybe_unused]] auto *state_logger =
+        monad::log::logger_t::create_logger("state_logger");
+
+    // set loggers logging level, there is only 1 log level for now
+    // TODO: flexible log levels for different logger
```
zander-xyz, line 177: I like the comment. Additionally, does anyone have any thoughts on using Github Issues to track enhancement `TODO`s like these?

@tongzhi1998 @sheapwilliams

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,223 @@
...
+        execution_t,
+        monad::fakeEmptyBP,
+        monad::fakeEmptyStateTrie,
+        transaction_trie_t,
+        receipt_trie_t,
+        receipt_collector_t,
+        monad::log::logger_t>
+        replay_eth;
+
+    [[maybe_unused]] auto result = replay_eth.run<
```
zander-xyz, line 208: Can we please log the result? As discussed earlier, no need to do any real error handling for now.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+using fakeState = execution::fake::State;
+using receiptCollector = std::vector<std::vector<Receipt>>;
+using eth_start_fork = fork_traits::frontier;
+
+template <class TTraits, class TState, class TEvm, class TStaticPrecompiles>
+struct fakeEvmHost
+{
+    evmc_result _result{};
+    Receipt _receipt{};
```
zander-xyz, line 29: nit: Just looking at some other files, seems like we prefer the underscore suffix as opposed to the prefix.

For style related comments, the only strong opinion I have is that we should be consistent, so, in general, feel free to look around and just choose the one that "fits in the best". 

Ditto down below.

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,60 @@
...
+    static constexpr void start() noexcept {}
+
+    static constexpr void *get_logger(char const * = nullptr) noexcept
+    {
+        return nullptr;
+    }
+
+    static constexpr void *create_logger(
+        std::string const &,
+        std::optional<FakeTimestampClockType *> = std::nullopt,
```
zander-xyz: As a case study, what actually happens if we start wanting to specify a `TimestampClockType`?

First, we should probably make the second argument a `std::optional<FakeTimestampClockType>`, to match the interface of `QuillLogger`. There, the `TimestampClockType` is not "pointy", so it should probably not be here as well.

Secondly, both types (`FakeTimestampClockType` and `quill::TimestampClockType`)  need to provide *some* common way for both to be constructed. There are a couple ways to address this.

1. Change the fake to match Quill.

That is,
```cpp
/**
 * Enum for the used clock type
 */
enum FakeTimestampClockType : uint8_t
{
  Tsc = 0,
  System,
  Custom
};
```
and then in `monad_log.hpp`
```cpp
#if DISABLE_LOGGING
using timestamp_clock_type_t = FakeTimestampClockType;
#else         
using timestamp_clock_type_t = quill::TimestampClockType;
#endif
```
Then our call site (if and when it exists), becomes
```cpp
monad::logger_t::create_logger("blah", monad::timestamp_clock_type_t::<...>);
```
2. Don't do anything about it since it's not being used anyways.

Either is a fine approach, I'll leave it up to you to decide which direction to go down.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
sheapwilliams: Either way - if we log a TODO, then putting the issue number there is way cool with me.

include/monad/logging/fake_logger.hpp
sheapwilliams: I agree with Alex here - the fake seems incorrectly programmed.  It should match the exact interface of the real object, especially if it is part of a new type like `optional` makes it.  I think we should do option 1.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
tongzhi1998: I think for fakes, we actually use prefix underscore instead of suffix underscore.

include/monad/logging/formatter.hpp
tongzhi1998: I just realized that in fmt/std.h, it has a formatter for std::optional<T>, so redefining it will cause error. As a result, I just use the formatter there, and the output looks like: `finish block number = optional(1)`

zander-xyz: Good find! That works too

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
tongzhi1998: Instead of writing a custom formatter for `Result` type, which would include forward declaration of the massive template type (e.g. `ReplayFromBlockDb<...>::Result`), I just choose to cast `Result.Status` (a enum) as an integer.

src/monad/execution/ethereum/replay_ethereum_block_db.cpp
sheapwilliams: Yea, that is true.  I was originally trying to make fake usage standout...I think it does based on this convo.

zander-xyz: Cool, I'm happy with that. Thanks guys.

include/monad/logging/formatter.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace quill
+{
+    template <>
+    struct copy_loggable<monad::bytes32_t> : std::true_type
+    {
+    };
+
+    template <typename T>
+    struct copy_loggable<std::optional<T>> : std::true_type
```
zander-xyz: It feels like `std::optional<T>` should be copy loggable iff `T` is also `coppy_loggable`. quill provides a trait to help facilitate this, `is_tagged_copyable_v`.

include/monad/logging/fake_logger.hpp
```diff
@@ -0,0 +1,63 @@
...
+};
+
+struct FakeHandler
+{
+};
+
+enum class FakeLogLevel
+{
+    FAKELEVEL
+};
```
zander-xyz: Per offline discussion, I think we should delete these until it's clear that we actually need them.

include/monad/logging/formatter.hpp
tongzhi1998: Makes sense! Actually here I use is_registered_copyable_v instead.

include/monad/logging/fake_logger.hpp
tongzhi1998: I only keep the FakeLogLevel because it is needed in main.

features/steps/read_from_block_db.py
tongzhi1998: Will work on this later

PR #90, opened 2023-06-02
skeleton for precompiled contracts

rgarc:
Problem: Gas cost for precompiled contracts changes across forks. With the current set of types in `/monad/execution`, we specify the list of precompiled contracts as a [variadic template](https://github.com/monad-crypto/monad/blob/328b2806d86eacdf60a3eb050fd7fbc83ac9abaf/include/monad/execution/static_precompiles.hpp#L15) `class... TPrecompiles`. This led me in my first attempt to define the precompiles per fork in [this commit](https://github.com/monad-crypto/monad/commit/520a2b5d43d4cc115736a234eda7c37e5ce94d9a) which is repetitive and error-prone.

Solution: Move from defining the list of precompiles as a `class... TPrecompiles` to a type list using [`boost::mp11::mp_list`](https://www.boost.org/doc/libs/master/libs/mp11/doc/html/mp11.html). We can then turn that into a list of function pointers with this snippet
```C++
template <typename... Types>
static constexpr auto
construct_precompile_array(boost::mp11::mp_list<Types...>)
{
    return std::array<exec_func_t, sizeof...(Types)>{Types::execute...};
}
```

### DIFF
--- a/include/monad/execution/test/fakes.hpp
+++ b/include/monad/execution/test/fakes.hpp
@@ -212,7 +212,7 @@ namespace fake
             static inline bool _success_store_contract{};
             static inline uint64_t _gas_creation_cost{};
             static inline uint64_t _create_address{};
-            static inline uint64_t _echo_gas_cost{10};
+            static inline constexpr uint64_t _echo_gas_cost{10};
             static inline auto intrinsic_gas(Transaction const &)
             {
                 return _intrinsic_gas;
@@ -222,7 +222,7 @@ namespace fake
             {
                 return _max_refund_quotient;
             }
-            static constexpr uint64_t echo_gas_cost() { return _echo_gas_cost; }
+            static consteval uint64_t echo_gas_cost() { return _echo_gas_cost; }
             static inline auto get_selfdestruct_refund(TState const &)
             {
                 return _sd_refund;
@@ -252,8 +252,8 @@ namespace fake
                 static_precompiles::Echo<beta>,
                 static_precompiles::OneHundredGas>;
 
-            static inline uint64_t _echo_gas_cost{15};
-            static constexpr uint64_t echo_gas_cost() { return _echo_gas_cost; }
+            static constexpr uint64_t _echo_gas_cost{15};
+            static consteval uint64_t echo_gas_cost() { return _echo_gas_cost; }
         };
     }
 
--- a/src/monad/execution/test/evm.cpp
+++ b/src/monad/execution/test/evm.cpp
@@ -392,7 +392,6 @@ TEST(Evm, out_of_gas_static_precompile_execution)
     fake::EvmHost h{};
     s._map.emplace(from, Account{.balance = 15'000});
     s._map.emplace(code_address, Account{.nonce = 6});
-    traits_t::_echo_gas_cost = 10;
 
     constexpr static char data[] = "hello world";
     constexpr static auto data_size = sizeof(data);
--- a/src/monad/execution/test/static_precompiles.cpp
+++ b/src/monad/execution/test/static_precompiles.cpp
@@ -29,7 +29,6 @@ TEST(StaticPrecompiles, execution_echo)
         0x0000000000000000000000000000000000000001_address};
     auto exec_func =
         alpha_static_precompiles_t::static_precompile_exec_func(code_address);
-    alpha_traits_t::_echo_gas_cost = 10;
 
     const auto data = "hello world";
     const auto data_size = 11u;
@@ -55,7 +54,6 @@ TEST(StaticPrecompiles, beta_traits_execution_echo)
         0x0000000000000000000000000000000000000001_address};
     auto exec_func =
         beta_static_precompiles_t::static_precompile_exec_func(code_address);
-    beta_traits_t::_echo_gas_cost = 15;
 
     const auto data = "hello world";
     const auto data_size = 11u;


include/monad/core/concepts.hpp
```diff
@@ -16,7 +16,6 @@ namespace concepts
         typename T::next_fork_t;
         { T::intrinsic_gas(t) } -> std::convertible_to<uint64_t>;
         { T::starting_nonce() } -> std::convertible_to<uint64_t>;
-        { T::static_precompiles } -> std::convertible_to<uint64_t>;
```
sheapwilliams, line 19: It is good to remove this, but we should test for the new type like on line 16.  You can add it into the next commit where you extend the fork_traits.

include/monad/execution/test/fakes.hpp
```diff
@@ -214,7 +222,7 @@ namespace fake
             {
                 return _max_refund_quotient;
             }
-            static inline auto echo_gas_cost() { return _echo_gas_cost; }
+            static constexpr uint64_t echo_gas_cost() { return _echo_gas_cost; }
```
sheapwilliams: We should make `_echo_gas_cost` `constexpr` as well...Perhaps this is set in some tests, but I think this should be a compile-time constant, and we should test that way.

src/monad/execution/ethereum/test/replay_eth_block_db.cpp
```diff
@@ -465,7 +467,7 @@ TEST(ReplayFromBlockDb_Eth, frontier_to_spurious_dragon)
         StaticPrecompiles,
         fakeEmptyEvmHost,
         fakeReceiptFiberData,
-        fake::static_precompiles::Echo>(
+        boost::mp11::mp_list<>>(
```
sheapwilliams: :+1: Nice easy fake....

[small nit] consider adding a declarative type at the beginning of the file, something like
```cpp
  using empty_precompile_list_t = boost::mp11::mp_list<>;
```
So it reads nicely at the usage site.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -106,6 +127,9 @@ namespace fork_traits
         // https://eips.ethereum.org/EIPS/eip-2
         static constexpr auto last_block_number = 2'674'999u;
 
+        using static_precompiles = frontier::static_precompiles;
```
sheapwilliams: You don't need to add this - if it is unchanged from frontier, it will be inherited here without change as well.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -149,6 +173,9 @@ namespace fork_traits
 
         static constexpr auto last_block_number = 4'369'999u;
 
+        using static_precompiles = homestead::static_precompiles;
```
sheapwilliams: ....same here, and throughout...

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -220,9 +256,19 @@ namespace fork_traits
...
 
+        using static_precompiles = type_list_t<
+            contracts::EllipticCurveRecover<istanbul>,
+            contracts::Sha256Hash<istanbul>, contracts::Ripemd160Hash<istanbul>,
+            contracts::Identity<istanbul>,
+            contracts::ModularExponentiation<istanbul>,
+            contracts::BigNumberAdd<istanbul>,
+            contracts::BigNumberMultiply<istanbul>,
+            contracts::BigNumberPairing<istanbul>,
+            contracts::Blake2F<istanbul>>;
```
sheapwilliams: If none of the constants changed from Constantinople/byzantium to Istanbul, then you can use `mp_push_back` to convey that.  We only need to re-declare the types if some of the static constants have changed between forks, and we need the declaration to pick up the new static constants.  Hopefully that makes sense...

include/monad/core/concepts.hpp
rgarc: Yeah that's a good point. I think I initially didn't do this because I wasn't sure what to put in the constraint if we are using a type list
```C++
{ T::static_precompiles } -> mp_list< ... ???? >
```
but I realize now, by "test for the new type", you literally mean something like this?
```diff
diff --git a/include/monad/core/concepts.hpp b/include/monad/core/concepts.hpp
index e0291b1..3125601 100644
--- a/include/monad/core/concepts.hpp
+++ b/include/monad/core/concepts.hpp
@@ -14,9 +14,9 @@ namespace concepts
                                    address_t const &a, evmc_result &r)
     {
         typename T::next_fork_t;
+        typename T::static_precompiles_t;
         { T::intrinsic_gas(t) } -> std::convertible_to<uint64_t>;
         { T::starting_nonce() } -> std::convertible_to<uint64_t>;
-        { T::static_precompiles } -> std::convertible_to<uint64_t>;
         { T::last_block_number } -> std::convertible_to<uint64_t>;
         { T::get_selfdestruct_refund(s) } -> std::convertible_to<uint64_t>;
         { T::max_refund_quotient() } -> std::convertible_to<int>;
```

sheapwilliams: Yup - that is it.  You only add the type to the fork_traits in the second commit, so fine to add them there.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -188,9 +215,18 @@ namespace fork_traits
     struct byzantium : spurious_dragon
     {
         using next_fork_t = istanbul;
-        static constexpr auto static_precompiles = 8u;
         static constexpr auto last_block_number = 9'068'999u;
 
+        using static_precompiles = type_list_t<
```
zander-xyz: Minor, but I'm noticing a pattern where our `type_list_t` is a list of precompiles, where all are templated on the same fork. As a result, the developer needs to type `contract<fork>`, where fork is unchanged for each precompile of the same fork. It would be nice if you only had to specify the fork once, and have all of the precompiles pick it up. That way, things are (ever-so-slightly) less verbose, but more importantly, harder to get wrong. 

There are limitations to this approach (namely, exacting the template list and arg for every static precompile), but here's one way to define `type_list_t`
```cpp
template <typename TFork, template <typename> typename... Types>
using type_list_t = boost::mp11::mp_list<Types<TFork>...>;
```

which can be used like
```cpp
struct Anago {};

template <typename T>
struct Dummy {};

template <typename T>
struct AnotherDummy{};

using list = type_list_t<Anago, Dummy, AnotherDummy>;
```

For type lists with heterogenous forks, you can use `boost::mp11::mp_append` to craft static precompiles from different forks like
```cpp
#include <boost/mp11/mpl_list.hpp>

struct Anago {};
struct Stormy {};

template <typename T>
struct Dummy {};

template <typename T>
struct AnotherDummy{};

template <typename T, template <typename> typename... Types>
using type_list_t = boost::mp11::mp_list<Types<T>...>;

int main()
{
    using list = type_list_t<Anago, Dummy, AnotherDummy>;
    using another_list = type_list_t<Stormy, Dummy, AnotherDummy>;
    using combined_list = boost::mp11::mp_append<list, another_list>;
    static_assert(std::is_same_v<combined_list,
                                 boost::mp11::mp_list<Dummy<Anago>,
                                 AnotherDummy<Anago>, 
                                 Dummy<Stormy>,
                                 AnotherDummy<Stormy>>>);
}
```

And here's a godbolt link in case you're interested in playing around with it: https://godbolt.org/z/q44G1hKEa

include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <monad/execution/ethereum/config.hpp>
+#include <monad/execution/ethereum/fork_traits.hpp>
+
+MONAD_EXECUTION_ETHEREUM_NAMESPACE_BEGIN
+
+namespace static_precompiles
+{
+    template <class TState>
```
zander-xyz: Is this also-known-as the below?
```cpp
template <monad::concepts::fork_traits TFork>
```
Same question for each of the other static precompiles.

include/monad/execution/test/fakes.hpp
zander-xyz: I guess we could `consteval` it if we really want to be sure about this 🤷‍♂️

include/monad/execution/ethereum/fork_traits.hpp
rgarc: Thanks, I really like this.

include/monad/core/concepts.hpp
rgarc: fixed in 391b7eb1fa81f2fd9c80a47d177a679fc5203718

include/monad/execution/ethereum/fork_traits.hpp
rgarc: fixed in 391b7eb1fa81f2fd9c80a47d177a679fc5203718

include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
rgarc: You are correct in that we want to pass in the fork when instantiating the precompile structs, so `TFork` is a much better name to convey this instead of `TState` so I will fix that. However, I was not able to constrain each precompile's template parameter by the `fork_traits` concept.

zander-xyz: Hmm, that seems strange. What's the error?

rgarc: It was easier to repro on the fake forks used for testing. I couldn't get something like this to work 3ff17cf7882402241966a6e44d213977b83e3e82

src/monad/execution/ethereum/test/replay_eth_block_db.cpp
rgarc: fixed in 1be5e5a54f75f54e00d806b6e7aed8744c6d159b

include/monad/execution/test/fakes.hpp
rgarc: fixed in db374a48c3fdb7a62bad955ffbef229007761554

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: Since the precompiles are templated on the fork, and assuming we still care about the fork being accurate for the succeeding precompiles, we'd need to transform the preceding list to be templated on the current fork. With boost mp11, this involves a mp_transform with a subsequent mp_replace_front.

And like Shea said, this should cut down on some of the verbosity involved in carrying precompiles over between forks.

See below for a modified example of how to do this.

```cpp
#include <boost/mp11/mpl_list.hpp>

struct Anago {};
struct Stormy {};

template <typename T>
struct Dummy {};

template <typename T>
struct AnotherDummy{};

template <typename T, template <typename> typename... Types>
using type_list_t = boost::mp11::mp_list<Types<T>...>;

template <typename T>
using switch_fork = boost::mp11::mp_replace_front<T, Stormy>;

int main()
{
    using list = type_list_t<Anago, Dummy, AnotherDummy>;
    using same_but_different_list = boost::mp11::mp_transform<switch_fork, list>;
    static_assert(std::is_same_v<
        same_but_different_list,
        boost::mp11::mp_list<Dummy<Stormy>, AnotherDummy<Stormy>>>);
}
```

include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
zander-xyz: Okay, let's put a pin on this one. Thanks for investigating Rene.

include/monad/execution/ethereum/fork_traits.hpp
rgarc: fixed in a4aa7564c15b66a03a08a76a207b589632dc0399

include/monad/execution/ethereum/fork_traits.hpp
rgarc: Thanks for showing what all is possible with `boost::mp11`. As this PR currently stands, it is not at all apparent which constants change across forks, but it will be when I go to fill in the implementations in a separate PR.

PR #91, opened 2023-06-05
Add In Memory Db

zander-xyz:
The idea is that we should be able to use dependency injection to swap out the backend of the db interface. It's currently designed to be compatible with an in memory and rocksdb backend. Will tackle interface with our custom trie db when it comes


include/monad/trie/rocks_writer.hpp
```diff
@@ -7,51 +7,50 @@
...
-    }
-
-    void write()
-    {
-        auto const res = db->Write(rocksdb::WriteOptions{}, &batch);
-        MONAD_ROCKS_ASSERT(res);
-        batch.Clear();
-    }
-};
+// struct RocksWriter
```
rgarc, line 16: Is there a reason to keep this commented code lying around as opposed to just deleting it?

include/monad/db/trie_db.hpp
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        lc.prev();
+
+        using namespace std::placeholders;
+
+        auto const left =
+            lc.key()
+                .transform(&TCursor::Key::path)
+                .transform(std::bind(trie::longest_common_prefix_size, _1, lk));
```
rgarc: small nit: is there a reason to use `std::bind` and `std::placeholders` anymore instead of lambdas? so something like this
```diff
diff --git a/include/monad/db/trie_db.hpp b/include/monad/db/trie_db.hpp
index ca8a627..3496ee2 100644
--- a/include/monad/db/trie_db.hpp
+++ b/include/monad/db/trie_db.hpp
@@ -93,12 +93,10 @@ struct TrieDB
 
         lc.prev();
 
-        using namespace std::placeholders;
-
         auto const left =
             lc.key()
                 .transform(&TCursor::Key::path)
-                .transform(std::bind(trie::longest_common_prefix_size, _1, lk));
+                .transform([&lk](const trie::Nibbles& nibbles) { return trie::longest_common_prefix_size(nibbles, lk); });
 
         lc.next();
         MONAD_DEBUG_ASSERT(lc.key().transform(&TCursor::Key::path) == lk);
@@ -107,7 +105,7 @@ struct TrieDB
         auto const right =
             lc.key()
                 .transform(&TCursor::Key::path)
-                .transform(std::bind(trie::longest_common_prefix_size, lk, _1));
+                .transform([&lk](const trie::Nibbles& nibbles) { return trie::longest_common_prefix_size(lk, nibbles); });
 
         auto const tk =
             !left.has_value() && !right.has_value()
```

src/monad/trie/test/helpers.hpp
```diff
@@ -15,213 +15,212 @@
...
-    {
-        trie_.process_updates(updates);
-
-        writer_.write();
-        leaves_cursor_.take_snapshot();
-        trie_cursor_.take_snapshot();
-    }
-};
-
+// TODO: these no longer compile due to interface changes. Fix when we
```
rgarc, line 18: ditto as the other commented chunk of code

zander-xyz: We likely need this code again (albeit, fixed up), but I think the actual fix-up belongs in a separate PR, when we add in rocksdb support.

include/monad/db/state.hpp
```diff
@@ -207,8 +207,8 @@ struct State
 
     void commit()
     {
-        accounts_.commit_all_merged();
         storage_.commit_all_merged();
+        accounts_.commit_all_merged();
```
sheapwilliams, line 214: Why this change?

include/monad/db/trie_db.hpp
```diff
@@ -0,0 +1,369 @@
...
+#include <map>
+#include <monad/core/account.hpp>
+#include <monad/core/assert.h>
+#include <monad/db/config.hpp>
+#include <monad/rlp/decode_helpers.hpp>
+#include <monad/trie/in_memory_comparator.hpp>
+#include <monad/trie/in_memory_cursor.hpp>
+#include <monad/trie/in_memory_writer.hpp>
+#include <monad/trie/trie.hpp>
+#include <variant>
```
sheapwilliams: include ordering? `variant` and `map` unused?  include `<unordered_map>`?

include/monad/db/trie_db.hpp
```diff
@@ -0,0 +1,369 @@
...
+        trie::Trie<cursor_t, writer_t> trie;
+
+        std::unordered_map<address_t, std::optional<Account>> updates;
+
+        Accounts()
+            : leaves_cursor{leaves_storage}
+            , leaves_writer{leaves_storage}
+            , trie_cursor{trie_storage}
+            , trie_writer{trie_storage}
+            , trie{leaves_cursor, trie_cursor, leaves_writer, trie_writer}
```
sheapwilliams: Are some of these members left uninitialized? `leaves_storage` and `trie_storage`, in particular? And `updates`?  Is that all ok?

include/monad/db/trie_db.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+        Storage()
+            : leaves_cursor{leaves_storage}
+            , leaves_writer{leaves_storage}
+            , trie_cursor{trie_storage}
+            , trie_writer{trie_storage}
+            , trie{leaves_cursor, trie_cursor, leaves_writer, trie_writer}
+        {
+        }
+    } storage;
```
sheapwilliams: How do the `accounts` member and the `storage` member differ?  Is is just the internal `updates` member?  They look so similar...Is there a generalization of these things?  Something that helps me understand the difference?

include/monad/db/state.hpp
zander-xyz: Discussed over zoom, will place a comment here for clarity.

include/monad/db/trie_db.hpp
zander-xyz: Yep, I'll reorder these according to the below.

monad
external
stdlib

include/monad/db/trie_db.hpp
zander-xyz: I can be more explicit here for clarity.

include/monad/db/trie_db.hpp
zander-xyz: Discussed over zoom. Yes, a large part of these classes is just plumbing, with the main difference being the updates.

PR #94, opened 2023-06-09
Add fiber aware async DB querying

zander-xyz:
This adds the capability to query the DB in an async fashion. Currently, behavior can be specialized using an executor. The inline executor invokes f with no arguments on the current thread whilst the single thread executor assigns f to be invoked on a different thread that it maintains. The executor model can be easily extended to support thread pools, if we ever need that.

The current unit tests I have are quite basic, so I'm open to any suggestions on how we might unit test this better.

### DIFF
--- a/include/monad/db/trie_db.hpp
+++ b/include/monad/db/trie_db.hpp
@@ -3,6 +3,7 @@
 #include <monad/core/account.hpp>
 #include <monad/core/assert.h>
 #include <monad/db/config.hpp>
+#include <monad/execution/execution_model.hpp>
 #include <monad/rlp/decode_helpers.hpp>
 #include <monad/trie/in_memory_comparator.hpp>
 #include <monad/trie/in_memory_cursor.hpp>
@@ -22,8 +23,7 @@ MONAD_DB_NAMESPACE_BEGIN
 
 namespace impl
 {
-
-    template <typename TDBImpl>
+    template <typename TDBImpl, typename TExecution>
     struct DBInterface
     {
         struct Updates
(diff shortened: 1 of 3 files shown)


include/monad/db/trie_db.hpp
```diff
@@ -34,12 +34,25 @@ namespace impl
                 storage;
         } updates{};
 
+        decltype(TExecution::get_executor()) executor{
+            TExecution::get_executor()};
+
```
sheapwilliams, line 39: Can't this just be `auto executor = TExecution::get_executor()`?

include/monad/execution/execution_model.hpp
```diff
@@ -9,7 +9,18 @@ MONAD_EXECUTION_NAMESPACE_BEGIN
...
-    static inline void yield() { boost::this_fiber::yield(); }
+    static void yield() noexcept { boost::this_fiber::yield(); }
+
+    [[nodiscard]] constexpr static auto get_executor() noexcept
+    {
+        return [](auto &&f) {
+            using result_t = decltype(std::declval<decltype(f)>()());
+            boost::fibers::promise<result_t> promise;
+            auto future = promise.get_future();
+            std::jthread thread{[&promise, f] { promise.set_value(f()); }};
```
sheapwilliams, line 22: `#include <thread>`?

include/monad/db/trie_db.hpp
zander-xyz: Not if I keep it as is (a member variable).

sheapwilliams: Ah - I missed that. :+1:

PR #95, opened 2023-06-09
implement precompiles through homestead

rgarc:
This PR implements the first four precompiled contracts by thinly wrapping [silkpre](https://github.com/torquem-ch/silkpre) into our repo.

### DIFF
--- a/include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
@@ -10,7 +10,7 @@ namespace static_precompiles
     template <class TFork>
     struct BigNumberAdd
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        static evmc::Result execute(const evmc_message &m) noexcept;
     };
 }
 
--- a/include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
@@ -12,23 +12,24 @@ namespace static_precompiles
     struct EllipticCurveRecover
     {
         using gas_cost = typename TFork::elliptic_curve_recover_gas_t;
-        static evmc_result execute(const evmc_message &message) noexcept
+        static evmc::Result execute(const evmc_message &message) noexcept
         {
             constexpr auto cost = gas_cost::base::value;
 
             if (message.gas < cost) {
-                return {.status_code = evmc_status_code::EVMC_OUT_OF_GAS};
+                return evmc::Result{evmc_result{
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS}};
             }
 
             auto const result =
                 silkpre_ecrec_run(message.input_data, message.input_size);
 
-            return {
+            return evmc::Result{evmc_result{
                 .status_code = evmc_status_code::EVMC_SUCCESS,
                 .gas_left = message.gas - cost,
                 .output_data = result.data,
                 .output_size = result.size,
-                .release = evmc_free_result_memory};
+                .release = evmc_free_result_memory}};
         }
     };
 }
--- a/include/monad/execution/ethereum/static_precompiles/identity.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/identity.hpp
@@ -11,20 +11,21 @@ namespace static_precompiles
     struct Identity
     {
         using gas_cost = typename TFork::identity_gas_t;
-        static evmc_result execute(const evmc_message &message) noexcept
+        static evmc::Result execute(const evmc_message &message) noexcept
         {
             auto const cost = gas_cost::compute(message.input_size);
 
             if (message.gas < cost) {
-                return {.status_code = evmc_status_code::EVMC_OUT_OF_GAS};
+                return evmc::Result{evmc_result{
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS}};
             }
 
-            return evmc_make_result(
+            return evmc::Result{
                 evmc_status_code::EVMC_SUCCESS,
                 /* gas_left= */ message.gas - cost,
                 /* gas_refund= */ 0,
                 message.input_data,
-                message.input_size);
+                message.input_size};
         }
     };
 }
--- a/include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
@@ -12,23 +12,24 @@ namespace static_precompiles
     struct Ripemd160Hash
     {
         using gas_cost = typename TFork::ripemd160_gas_t;
-        static evmc_result execute(const evmc_message &message) noexcept
+        static evmc::Result execute(const evmc_message &message) noexcept
         {
             auto const cost = gas_cost::compute(message.input_size);
             if (message.gas < cost) {
-                return {.status_code = evmc_status_code::EVMC_OUT_OF_GAS};
+                return evmc::Result{evmc_result{
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS}};
             }
 
             bytes32_t output{};
             silkpre_rmd160(
                 output.bytes, message.input_data, message.input_size);
 
-            return evmc_make_result(
+            return evmc::Result{
                 evmc_status_code::EVMC_SUCCESS,
                 /* gas_left= */ message.gas - cost,
                 /* gas_refund= */ 0,
                 output.bytes,
-                sizeof(bytes32_t));
+                sizeof(bytes32_t)};
         }
     };
 }
--- a/include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
@@ -12,12 +12,13 @@ namespace static_precompiles
     struct Sha256Hash
     {
         using gas_cost = typename TFork::sha256_gas_t;
-        static evmc_result execute(const evmc_message &message) noexcept
+        static evmc::Result execute(const evmc_message &message) noexcept
         {
             auto const cost = gas_cost::compute(message.input_size);
 
             if (message.gas < cost) {
-                return {.status_code = evmc_status_code::EVMC_OUT_OF_GAS};
+                return evmc::Result{evmc_result{
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS}};
             }
 
             bytes32_t output{};
@@ -27,12 +28,12 @@ namespace static_precompiles
                 message.input_size,
                 /* use_cpu_extensions */ true);
 
-            return evmc_make_result(
+            return evmc::Result{
                 evmc_status_code::EVMC_SUCCESS,
                 /* gas_left= */ message.gas - cost,
                 /* gas_refund= */ 0,
                 output.bytes,
-                sizeof(bytes32_t));
+                sizeof(bytes32_t)};
         }
     };
 }
--- a/include/monad/execution/ethereum/static_precompiles/big_number_multiply.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/big_number_multiply.hpp
@@ -10,7 +10,7 @@ namespace static_precompiles
     template <class TFork>
     struct BigNumberMultiply
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        static evmc::Result execute(const evmc_message &m) noexcept;
     };
 }
 
--- a/include/monad/execution/ethereum/static_precompiles/big_number_pairing.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/big_number_pairing.hpp
@@ -10,7 +10,7 @@ namespace static_precompiles
     template <class TFork>
     struct BigNumberPairing
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        static evmc::Result execute(const evmc_message &m) noexcept;
     };
 }
 
--- a/include/monad/execution/ethereum/static_precompiles/blake2f.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/blake2f.hpp
@@ -10,7 +10,7 @@ namespace static_precompiles
     template <class TFork>
     struct Blake2F
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        static evmc::Result execute(const evmc_message &m) noexcept;
     };
 }
 
--- a/include/monad/execution/ethereum/static_precompiles/modular_exponentiation.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/modular_exponentiation.hpp
@@ -10,7 +10,7 @@ namespace static_precompiles
     template <class TFork>
     struct ModularExponentiation
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        static evmc::Result execute(const evmc_message &m) noexcept;
     };
 }
 
--- a/include/monad/execution/evm.hpp
+++ b/include/monad/execution/evm.hpp
@@ -48,20 +48,21 @@ struct Evm
             result.status_code != EVMC_SUCCESS) {
             return evmc::Result{result};
         }
-        evmc_result const result =
+        evmc::Result result =
             TStaticPrecompiles::static_precompile_exec_func(m.code_address)
                 .transform([&](auto static_precompile_execute) {
                     return static_precompile_execute(m);
                 })
                 // execute on backend, just this for now
-                .value_or(evmc_result{
-                    .status_code = EVMC_SUCCESS, .gas_left = m.gas});
+                .disjunction(evmc::Result{evmc_result{
+                    .status_code = EVMC_SUCCESS, .gas_left = m.gas}})
+                .value();
 
         if (result.status_code == EVMC_REVERT) {
             state.revert();
         }
 
-        return evmc::Result{result};
+        return result;
     }
 
     [[nodiscard]] static result_t
--- a/include/monad/execution/static_precompiles.hpp
+++ b/include/monad/execution/static_precompiles.hpp
@@ -17,7 +17,7 @@ template <
     class TState, concepts::fork_traits<TState> TTraits, class TPrecompiles>
 struct StaticPrecompiles
 {
-    using exec_func_t = evmc_result (*)(const evmc_message &) noexcept;
+    using exec_func_t = evmc::Result (*)(const evmc_message &) noexcept;
 
     template <typename... Types>
     static constexpr auto
--- a/include/monad/execution/test/fakes.hpp
+++ b/include/monad/execution/test/fakes.hpp
@@ -262,39 +262,41 @@ namespace fake
         template <typename T>
         struct Echo
         {
-            static evmc_result execute(const evmc_message &m) noexcept
+            static evmc::Result execute(const evmc_message &m) noexcept
             {
                 const int64_t gas =
                     (const int64_t)(m.input_size * T::echo_gas_cost());
                 if (m.gas < gas) {
-                    return {.status_code = EVMC_OUT_OF_GAS};
+                    return evmc::Result{
+                        evmc_result{.status_code = EVMC_OUT_OF_GAS}};
                 }
                 unsigned char *output_data =
                     (unsigned char *)std::malloc(m.input_size);
                 std::memcpy(output_data, m.input_data, m.input_size);
-                return {
+                return evmc::Result{evmc_result{
                     .status_code = EVMC_SUCCESS,
                     .gas_left = m.gas - gas,
                     .output_data = output_data,
                     .output_size = m.input_size,
                     .release = [](const evmc_result *result) {
                         std::free((unsigned char *)result->output_data);
-                    }};
+                    }}};
             }
         };
 
         struct OneHundredGas
         {
-            static evmc_result execute(const evmc_message &m) noexcept
+            static evmc::Result execute(const evmc_message &m) noexcept
             {
                 const int64_t gas = 100;
                 if (m.gas < gas) {
-                    return {.status_code = EVMC_OUT_OF_GAS};
+                    return evmc::Result{
+                        evmc_result{.status_code = EVMC_OUT_OF_GAS}};
                 }
-                return {
+                return evmc::Result{evmc_result{
                     .status_code = EVMC_SUCCESS,
                     .gas_left = m.gas - gas,
-                    .output_size = 0u};
+                    .output_size = 0u}};
             }
         };
     }
--- a/src/monad/execution/test/evm.cpp
+++ b/src/monad/execution/test/evm.cpp
@@ -415,16 +415,17 @@ namespace revert_test
 {
     struct AlwaysRevertForAThousand
     {
-        static evmc_result execute(const evmc_message &m) noexcept
+        static evmc::Result execute(const evmc_message &m) noexcept
         {
             const int64_t gas = 1000;
             if (m.gas < gas) {
-                return {.status_code = EVMC_OUT_OF_GAS};
+                return evmc::Result{
+                    evmc_result{.status_code = EVMC_OUT_OF_GAS}};
             }
-            return {
+            return evmc::Result{evmc_result{
                 .status_code = EVMC_REVERT,
                 .gas_left = m.gas - gas,
-                .output_size = 0u};
+                .output_size = 0u}};
         }
     };
 


include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
```diff
@@ -10,7 +11,25 @@ namespace static_precompiles
...
+        static evmc_result execute(const evmc_message &message) noexcept
+        {
+            constexpr auto cost = typename gas_cost::base::value_type();
+            if (std::cmp_less(message.gas, cost)) {
+                return {
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS,
+                    .gas_left = message.gas};
+            }
+
+            const auto result =
```
sheapwilliams: Remember - This project uses east const 😄.  Seriously tho, we can setup [a clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifieralignment) option for this so we don't have to keep this meaningless detail in our brains.

include/monad/execution/ethereum/static_precompiles/identity.hpp
```diff
@@ -10,7 +10,25 @@ namespace static_precompiles
...
+                return {
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS,
+                    .gas_left = message.gas};
+            }
+
+            return evmc_make_result(
+                evmc_status_code::EVMC_SUCCESS,
+                message.gas - cost,
+                0,
+                message.input_data,
```
sheapwilliams, line 27: This guy assumes the `evmc_message`'s `input_data` lifetime lasts as long as the evmc_result will need to be around...That is true for our code, but I'm not sure if that is true for EVM1 code...Something to consider.

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
```diff
@@ -10,7 +11,32 @@ namespace static_precompiles
...
+            if (std::cmp_less(message.gas, cost)) {
+                return {
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS,
+                    .gas_left = message.gas};
+            }
+
+            static constexpr auto NUM_HASH_BYTES = 20;
+            using hash = monad::byte_string_fixed<NUM_HASH_BYTES>;
+
+            hash output;
```
sheapwilliams: Initialize?

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
```diff
@@ -10,7 +11,32 @@ namespace static_precompiles
...
+        {
+            const auto cost = gas_cost::compute(
+                static_cast<gas_cost::integer_type>(message.input_size));
+            if (std::cmp_less(message.gas, cost)) {
+                return {
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS,
+                    .gas_left = message.gas};
+            }
+
+            static constexpr auto NUM_HASH_BYTES = 20;
```
sheapwilliams: Is this related to the `size(address_t)` at all?  Where does 20 come from?

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
```diff
@@ -10,7 +11,32 @@ namespace static_precompiles
...
+
+            silkpre_rmd160(
+                output.data(), message.input_data, message.input_size);
+
+            return evmc_make_result(
+                evmc_status_code::EVMC_SUCCESS,
+                message.gas - cost,
+                0,
+                output.data(),
+                NUM_HASH_BYTES);
```
sheapwilliams: In general, these factory result maker functions are ok, but I have no idea what the 3rd 0 is without looking it up...would designated initializers like you have on line 20 be easier to follow?  Thoughts?

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
```diff
@@ -10,7 +11,36 @@ namespace static_precompiles
...
+            const auto cost = gas_cost::compute(
+                static_cast<gas_cost::integer_type>(message.input_size));
+
+            if (std::cmp_less(message.gas, cost)) {
+                return {
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS,
+                    .gas_left = message.gas};
+            }
+
+            static constexpr auto NUM_HASH_BYTES = 32;
```
sheapwilliams: Is this just `size(bytes32_t)`?  How is what `silkpre_sha256` returns different than `bytes32_t`?  Which byte order is it in?

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
```diff
@@ -10,7 +11,36 @@ namespace static_precompiles
...
+            using hash = monad::byte_string_fixed<NUM_HASH_BYTES>;
+
+            hash output;
```
sheapwilliams: Initialize?  And throughout...

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
```diff
@@ -10,7 +11,36 @@ namespace static_precompiles
...
+
+            silkpre_sha256(
+                output.data(),
+                message.input_data,
+                message.input_size,
+                /* use_cpu_extensions */ false);
```
sheapwilliams: Why false here?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -41,10 +41,40 @@ namespace fork_traits
...
     struct frontier
     {
         using next_fork_t = homestead;
         static constexpr auto last_block_number = 1'149'999u;
+
+        using elliptic_curve_recover_gas_t = gas_thunk<int64_t, 3000, 0>;
+        using sha256_gas_t = gas_thunk<int64_t, 60, 12>;
+        using ripemd160_gas_t = gas_thunk<int64_t, 600, 120>;
+        using identity_gas_t = gas_thunk<int64_t, 15, 3>;
+
```
sheapwilliams, line 86: So is this only for frontier then?  Other hard forks are forthcoming?

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
```diff
@@ -10,7 +11,32 @@ namespace static_precompiles
...
-        static evmc_result execute(const evmc_message &m) noexcept;
+        using gas_cost = typename TFork::ripemd160_gas_t;
+        static evmc_result execute(const evmc_message &message) noexcept
+        {
+            const auto cost = gas_cost::compute(
+                static_cast<gas_cost::integer_type>(message.input_size));
+            if (std::cmp_less(message.gas, cost)) {
+                return {
+                    .status_code = evmc_status_code::EVMC_OUT_OF_GAS,
+                    .gas_left = message.gas};
```
sheapwilliams: When there is insufficient gas, does it take the remaining gas, or just return it like you have here?  I don't recall...

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
sheapwilliams: Of course! 160 bits...my bad.  :+1:

include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
rgarc: Fixed up

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
rgarc: Fixed up with
```c++
hash output{};
```

include/monad/execution/ethereum/static_precompiles/identity.hpp
rgarc: I think I'm still confused about the ownership model here. Let's chat over zoom to clear it up.

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
rgarc: I used the factory result maker function because I'm confused about the ownership model.

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: Fixed up

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: Mostly to err on the side of safety for now. I don't really have the best confidence in their CPU-specific intrinsic code, and I figured this would only be worth revisiting once we are targeting performance.

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
rgarc: It's correct that it should take the remaining gas. Fixed in 27c82c01fbd0fb96bc12d450c35b7d9a82aaca22 by letting the designated initializer clear it out.

include/monad/execution/ethereum/fork_traits.hpp
rgarc: Correct, these will be good through (including) homestead.

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
```diff
@@ -10,7 +11,34 @@ namespace static_precompiles
...
     struct Sha256Hash
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        using gas_cost = typename TFork::sha256_gas_t;
+        static evmc_result execute(const evmc_message &message) noexcept
+        {
+            auto const cost = gas_cost::compute(
+                static_cast<gas_cost::integer_type>(message.input_size));
+
+            if (std::cmp_less(message.gas, cost)) {
```
zander-xyz: Not a big deal, but asking because this seems like a very pointed decision.

I'm curious why you chose to use `cmp_less` here instead of the operator. As far as I can tell, these are both signed integers, and we have `-Wconversion` to prevent implicit integer promotions anyways.

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
zander-xyz: Why don't you have the best confidence?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -41,10 +41,40 @@ namespace fork_traits
...
     namespace contracts = execution::ethereum::static_precompiles;
 
+    /**
+     * Gas cost for many precompiles is computed as Base + PerWord * (number of words in input)
+     * @tparam IntegerType
+     * @tparam Base
+     * @tparam PerWord
+     */
+    template <typename IntegerType, IntegerType Base, IntegerType PerWord>
+    struct gas_thunk
```
zander-xyz: Is `gas_thunk` an ethereum term? Is there a more descriptive name for me to better understand what this does?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -41,10 +41,40 @@ namespace fork_traits
...
+        using base = std::integral_constant<IntegerType, Base>;
+        using per_word = std::integral_constant<IntegerType, PerWord>;
+
+        /**
+         * @param size of input message in bytes
+         * @return the total gas cost
+         */
+        static constexpr IntegerType compute(IntegerType size)
+        {
+            return static_cast<IntegerType>((size + 31) / 32 * per_word()) +
```
zander-xyz: I'm guessing this equation is in the ethereum yellow paper somewhere, can you please leave that as a comment here?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -41,10 +41,40 @@ namespace fork_traits
 
     namespace contracts = execution::ethereum::static_precompiles;
 
+    /**
+     * Gas cost for many precompiles is computed as Base + PerWord * (number of words in input)
+     * @tparam IntegerType
+     * @tparam Base
+     * @tparam PerWord
+     */
+    template <typename IntegerType, IntegerType Base, IntegerType PerWord>
```
zander-xyz: Based on how you designed this, I'm guessing we'll need different IntegerTypes at some point? Will it always be signed? Is there a concept that can better describe this?

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: They are both
```C++
    uint8_t[32]
```
the final hash value outputted by `silkpre_sha256` is in big-endian order

src/monad/core/test/sha256.cpp
```diff
@@ -0,0 +1,49 @@
+#include <gtest/gtest.h>
```
zander-xyz: what about unit tests for all the other precompiles?

include/monad/execution/ethereum/static_precompiles/identity.hpp
```diff
@@ -10,7 +10,23 @@ namespace static_precompiles
     template <class TFork>
     struct Identity
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        using gas_cost = typename TFork::identity_gas_t;
+        static evmc_result execute(const evmc_message &message) noexcept
+        {
+            auto const cost = gas_cost::compute(
+                static_cast<gas_cost::integer_type>(message.input_size));
```
zander-xyz: input_size is a `size_t` whose width is platform specific and is an unsigned variable. are you absolutely sure this cast is OK?

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
zander-xyz: I think a comment here would be nice though.

include/monad/execution/ethereum/static_precompiles/identity.hpp
zander-xyz: `evmc_make_result` does a copy so should be okay after this point, I'd think

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: Purely a gut feeling I suppose. 

I think this is slightly harder to read (thus more probable to contain bugs)
```c++
// The following function was adapted from https://github.com/noloader/SHA-Intrinsics/blob/master/sha256-x86.c
/*   Intel SHA extensions using C intrinsics               */
/*   Written and place in public domain by Jeffrey Walton  */
/*   Based on code from Intel, and by Sean Gulley for      */
/*   the miTLS project.                                    */
__attribute__((target("sha,sse4.1"))) static void sha_256_x86_sha(uint32_t h[8], const void* input, size_t len) {
    __m128i STATE0, STATE1;
    __m128i MSG, TMP;
    __m128i MSG0, MSG1, MSG2, MSG3;
    __m128i ABEF_SAVE, CDGH_SAVE;
    const __m128i MASK = _mm_set_epi64x(0x0c0d0e0f08090a0bULL, 0x0405060700010203ULL);

    /* Load initial values */
    TMP = _mm_loadu_si128((const __m128i*)&h[0]);
    STATE1 = _mm_loadu_si128((const __m128i*)&h[4]);

    TMP = _mm_shuffle_epi32(TMP, 0xB1);          /* CDAB */
    STATE1 = _mm_shuffle_epi32(STATE1, 0x1B);    /* EFGH */
    STATE0 = _mm_alignr_epi8(TMP, STATE1, 8);    /* ABEF */
    STATE1 = _mm_blend_epi16(STATE1, TMP, 0xF0); /* CDGH */

    struct buffer_state state;
    init_buf_state(&state, input, len);

    /* 512-bit chunks is what we will operate on. */
    uint8_t chunk[CHUNK_SIZE];

    while (calc_chunk(chunk, &state)) {
        /* Save current state */
        ABEF_SAVE = STATE0;
        CDGH_SAVE = STATE1;

        /* Rounds 0-3 */
        MSG = _mm_loadu_si128((const __m128i*)(chunk + 0));
        MSG0 = _mm_shuffle_epi8(MSG, MASK);
        MSG = _mm_add_epi32(MSG0, _mm_set_epi64x(0xE9B5DBA5B5C0FBCFULL, 0x71374491428A2F98ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);

        /* Rounds 4-7 */
        MSG1 = _mm_loadu_si128((const __m128i*)(chunk + 16));
        MSG1 = _mm_shuffle_epi8(MSG1, MASK);
        MSG = _mm_add_epi32(MSG1, _mm_set_epi64x(0xAB1C5ED5923F82A4ULL, 0x59F111F13956C25BULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG0 = _mm_sha256msg1_epu32(MSG0, MSG1);

        /* Rounds 8-11 */
        MSG2 = _mm_loadu_si128((const __m128i*)(chunk + 32));
        MSG2 = _mm_shuffle_epi8(MSG2, MASK);
        MSG = _mm_add_epi32(MSG2, _mm_set_epi64x(0x550C7DC3243185BEULL, 0x12835B01D807AA98ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG1 = _mm_sha256msg1_epu32(MSG1, MSG2);

        /* Rounds 12-15 */
        MSG3 = _mm_loadu_si128((const __m128i*)(chunk + 48));
        MSG3 = _mm_shuffle_epi8(MSG3, MASK);
        MSG = _mm_add_epi32(MSG3, _mm_set_epi64x(0xC19BF1749BDC06A7ULL, 0x80DEB1FE72BE5D74ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG3, MSG2, 4);
        MSG0 = _mm_add_epi32(MSG0, TMP);
        MSG0 = _mm_sha256msg2_epu32(MSG0, MSG3);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG2 = _mm_sha256msg1_epu32(MSG2, MSG3);

        /* Rounds 16-19 */
        MSG = _mm_add_epi32(MSG0, _mm_set_epi64x(0x240CA1CC0FC19DC6ULL, 0xEFBE4786E49B69C1ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG0, MSG3, 4);
        MSG1 = _mm_add_epi32(MSG1, TMP);
        MSG1 = _mm_sha256msg2_epu32(MSG1, MSG0);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG3 = _mm_sha256msg1_epu32(MSG3, MSG0);

        /* Rounds 20-23 */
        MSG = _mm_add_epi32(MSG1, _mm_set_epi64x(0x76F988DA5CB0A9DCULL, 0x4A7484AA2DE92C6FULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG1, MSG0, 4);
        MSG2 = _mm_add_epi32(MSG2, TMP);
        MSG2 = _mm_sha256msg2_epu32(MSG2, MSG1);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG0 = _mm_sha256msg1_epu32(MSG0, MSG1);

        /* Rounds 24-27 */
        MSG = _mm_add_epi32(MSG2, _mm_set_epi64x(0xBF597FC7B00327C8ULL, 0xA831C66D983E5152ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG2, MSG1, 4);
        MSG3 = _mm_add_epi32(MSG3, TMP);
        MSG3 = _mm_sha256msg2_epu32(MSG3, MSG2);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG1 = _mm_sha256msg1_epu32(MSG1, MSG2);

        /* Rounds 28-31 */
        MSG = _mm_add_epi32(MSG3, _mm_set_epi64x(0x1429296706CA6351ULL, 0xD5A79147C6E00BF3ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG3, MSG2, 4);
        MSG0 = _mm_add_epi32(MSG0, TMP);
        MSG0 = _mm_sha256msg2_epu32(MSG0, MSG3);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG2 = _mm_sha256msg1_epu32(MSG2, MSG3);

        /* Rounds 32-35 */
        MSG = _mm_add_epi32(MSG0, _mm_set_epi64x(0x53380D134D2C6DFCULL, 0x2E1B213827B70A85ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG0, MSG3, 4);
        MSG1 = _mm_add_epi32(MSG1, TMP);
        MSG1 = _mm_sha256msg2_epu32(MSG1, MSG0);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG3 = _mm_sha256msg1_epu32(MSG3, MSG0);

        /* Rounds 36-39 */
        MSG = _mm_add_epi32(MSG1, _mm_set_epi64x(0x92722C8581C2C92EULL, 0x766A0ABB650A7354ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG1, MSG0, 4);
        MSG2 = _mm_add_epi32(MSG2, TMP);
        MSG2 = _mm_sha256msg2_epu32(MSG2, MSG1);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG0 = _mm_sha256msg1_epu32(MSG0, MSG1);

        /* Rounds 40-43 */
        MSG = _mm_add_epi32(MSG2, _mm_set_epi64x(0xC76C51A3C24B8B70ULL, 0xA81A664BA2BFE8A1ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG2, MSG1, 4);
        MSG3 = _mm_add_epi32(MSG3, TMP);
        MSG3 = _mm_sha256msg2_epu32(MSG3, MSG2);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG1 = _mm_sha256msg1_epu32(MSG1, MSG2);

        /* Rounds 44-47 */
        MSG = _mm_add_epi32(MSG3, _mm_set_epi64x(0x106AA070F40E3585ULL, 0xD6990624D192E819ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG3, MSG2, 4);
        MSG0 = _mm_add_epi32(MSG0, TMP);
        MSG0 = _mm_sha256msg2_epu32(MSG0, MSG3);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG2 = _mm_sha256msg1_epu32(MSG2, MSG3);

        /* Rounds 48-51 */
        MSG = _mm_add_epi32(MSG0, _mm_set_epi64x(0x34B0BCB52748774CULL, 0x1E376C0819A4C116ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG0, MSG3, 4);
        MSG1 = _mm_add_epi32(MSG1, TMP);
        MSG1 = _mm_sha256msg2_epu32(MSG1, MSG0);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);
        MSG3 = _mm_sha256msg1_epu32(MSG3, MSG0);

        /* Rounds 52-55 */
        MSG = _mm_add_epi32(MSG1, _mm_set_epi64x(0x682E6FF35B9CCA4FULL, 0x4ED8AA4A391C0CB3ULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG1, MSG0, 4);
        MSG2 = _mm_add_epi32(MSG2, TMP);
        MSG2 = _mm_sha256msg2_epu32(MSG2, MSG1);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);

        /* Rounds 56-59 */
        MSG = _mm_add_epi32(MSG2, _mm_set_epi64x(0x8CC7020884C87814ULL, 0x78A5636F748F82EEULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        TMP = _mm_alignr_epi8(MSG2, MSG1, 4);
        MSG3 = _mm_add_epi32(MSG3, TMP);
        MSG3 = _mm_sha256msg2_epu32(MSG3, MSG2);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);

        /* Rounds 60-63 */
        MSG = _mm_add_epi32(MSG3, _mm_set_epi64x(0xC67178F2BEF9A3F7ULL, 0xA4506CEB90BEFFFAULL));
        STATE1 = _mm_sha256rnds2_epu32(STATE1, STATE0, MSG);
        MSG = _mm_shuffle_epi32(MSG, 0x0E);
        STATE0 = _mm_sha256rnds2_epu32(STATE0, STATE1, MSG);

        /* Combine state  */
        STATE0 = _mm_add_epi32(STATE0, ABEF_SAVE);
        STATE1 = _mm_add_epi32(STATE1, CDGH_SAVE);
    }

    TMP = _mm_shuffle_epi32(STATE0, 0x1B);       /* FEBA */
    STATE1 = _mm_shuffle_epi32(STATE1, 0xB1);    /* DCHG */
    STATE0 = _mm_blend_epi16(TMP, STATE1, 0xF0); /* DCBA */
    STATE1 = _mm_alignr_epi8(STATE1, TMP, 8);    /* ABEF */

    /* Save state */
    _mm_storeu_si128((__m128i*)&h[0], STATE0);
    _mm_storeu_si128((__m128i*)&h[4], STATE1);
}
```
as opposed to
```C++
static inline ALWAYS_INLINE void sha_256_implementation(uint32_t h[8], const void* input, size_t len) {
    /*
     * Note 1: All integers (expect indexes) are 32-bit unsigned integers and addition is calculated modulo 2^32.
     *
     * Note 2: For each round, there is one round constant k[i] and one entry in the message schedule array w[i],
     * 0 = i = 63
     *
     * Note 3: The compression function uses 8 working variables, a through h
     *
     * Note 4: Big-endian convention is used when expressing the constants in this pseudocode,
     *     and when parsing message block data from bytes to words, for example,
     *     the first word of the input message "abc" after padding is 0x61626380
     */

    struct buffer_state state;
    init_buf_state(&state, input, len);

    /* 512-bit chunks is what we will operate on. */
    uint8_t chunk[CHUNK_SIZE];

    while (calc_chunk(chunk, &state)) {
        unsigned i, j;

        uint32_t ah[8];
        /* Initialize working variables to current hash value: */
        for (i = 0; i < 8; i++) {
            ah[i] = h[i];
        }

        const uint8_t* p = chunk;

        /*
         * The w-array is really w[64], but since we only need
         * 16 of them at a time, we save stack by calculating
         * 16 at a time.
         *
         * This optimization was not there initially and the
         * rest of the comments about w[64] are kept in their
         * initial state.
         */

        /*
         * create a 64-entry message schedule array w[0..63] of 32-bit words
         * (The initial values in w[0..63] don't matter, so many implementations zero them here)
         * copy chunk into first 16 words w[0..15] of the message schedule array
         */
        uint32_t w[16];

        /* Compression function main loop: */
        for (i = 0; i < 4; i++) {
            for (j = 0; j < 16; j++) {
                if (i == 0) {
                    w[j] = (uint32_t)p[0] << 24 | (uint32_t)p[1] << 16 | (uint32_t)p[2] << 8 | (uint32_t)p[3];
                    p += 4;
                } else {
                    /* Extend the first 16 words into the remaining 48 words w[16..63] of the message schedule array: */
                    const uint32_t s0 =
                        right_rot(w[(j + 1) & 0xf], 7) ^ right_rot(w[(j + 1) & 0xf], 18) ^ (w[(j + 1) & 0xf] >> 3);
                    const uint32_t s1 =
                        right_rot(w[(j + 14) & 0xf], 17) ^ right_rot(w[(j + 14) & 0xf], 19) ^ (w[(j + 14) & 0xf] >> 10);
                    w[j] = w[j] + s0 + w[(j + 9) & 0xf] + s1;
                }
                const uint32_t s1 = right_rot(ah[4], 6) ^ right_rot(ah[4], 11) ^ right_rot(ah[4], 25);
                const uint32_t ch = (ah[4] & ah[5]) ^ (~ah[4] & ah[6]);
                const uint32_t temp1 = ah[7] + s1 + ch + k[i << 4 | j] + w[j];
                const uint32_t s0 = right_rot(ah[0], 2) ^ right_rot(ah[0], 13) ^ right_rot(ah[0], 22);
                const uint32_t maj = (ah[0] & ah[1]) ^ (ah[0] & ah[2]) ^ (ah[1] & ah[2]);
                const uint32_t temp2 = s0 + maj;

                ah[7] = ah[6];
                ah[6] = ah[5];
                ah[5] = ah[4];
                ah[4] = ah[3] + temp1;
                ah[3] = ah[2];
                ah[2] = ah[1];
                ah[1] = ah[0];
                ah[0] = temp1 + temp2;
            }
        }

        /* Add the compressed chunk to the current hash value: */
        for (i = 0; i < 8; i++) {
            h[i] += ah[i];
        }
    }
}
```

zander-xyz: I am inclined to say that if we're ever not confident with any part of our code, we haven't unit tested it enough. Of course that is easier said than done. Devil is always in the details.

include/monad/execution/ethereum/fork_traits.hpp
rgarc: Are you thinking something like this?
```diff
diff --git a/include/monad/execution/ethereum/fork_traits.hpp b/include/monad/execution/ethereum/fork_traits.hpp
index e92ec87..08c23d0 100644
--- a/include/monad/execution/ethereum/fork_traits.hpp
+++ b/include/monad/execution/ethereum/fork_traits.hpp
@@ -47,7 +47,8 @@ namespace fork_traits
      * @tparam Base
      * @tparam PerWord
      */
-    template <typename IntegerType, IntegerType Base, IntegerType PerWord>
+    template <
+        std::signed_integral IntegerType, IntegerType Base, IntegerType PerWord>
     struct gas_thunk
     {
         using integer_type = IntegerType;

```

include/monad/execution/ethereum/static_precompiles/identity.hpp
rgarc: This cast is indeed suspicious. But I'm also not sure how to best handle it. Only idea I have is to check if `message.input_size` is in range of `std::numeric_limits<IntegerType>` and perhaps return `EVMC_PRECOMPILE_FAILURE` if it isn't?
```C++
    /**
     * A call to a precompiled or system contract has ended with a failure.
     *
     * An example: elliptic curve functions handed invalid EC points.
     */
    EVMC_PRECOMPILE_FAILURE = 12,
```

include/monad/execution/ethereum/fork_traits.hpp
rgarc: Gave a more descriptive name in https://github.com/monad-crypto/monad/pull/95/commits/1ba2ad021999dd5a4ac3b208860f8dbf7c2759cd

include/monad/execution/ethereum/fork_traits.hpp
rgarc: LInked to YP in https://github.com/monad-crypto/monad/pull/95/commits/1ba2ad021999dd5a4ac3b208860f8dbf7c2759cd

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: Discussed over zoom. Perhaps this does not need to be templated on IntegerType

include/monad/execution/ethereum/static_precompiles/identity.hpp
zander-xyz: Discussed over zoom

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
sheapwilliams: We shouldn't really unit test vendor code - just integration test it.  But we also shouldn't be eye-balling code (or anything for that matter).  Eye-balls regularly deceive us; in tests we trust.

If we have any real suspicions about vendor code, let's elevate and discuss.

Wrt this code, I don't worry about it by itself, only if it doesn't have decent tests.

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
sheapwilliams: `bytes32_t` is similarly BE.  Are these actually different underlying types?  IF they are the same underlying, then it would be nice to just use the bytes32_t alias.  No worries otherwise.

include/monad/execution/ethereum/static_precompiles/identity.hpp
sheapwilliams: Ah - I didn't really look inside the factory fcn... 👌

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
rgarc: fixed in c437a5dfda79032ad9746561c9812fd56664cb07

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: I agree and fixed in 9232bf002a14ea31aaa45931667ef21bf1530c58

include/monad/execution/ethereum/fork_traits.hpp
rgarc: fixed in bda7f93a78c9f60834a60a0336af332e8bc35904

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: It was an artifact from when I was still messing around with the signedness of the gas type. I've gone back to operator< in 51b9c2e92d0d97786f6c180d1176271b4700e871

include/monad/execution/ethereum/static_precompiles/sha256_hash.hpp
rgarc: Flipped to true in 0a2faf6273bab0056aaeba5a4cac97610672cfa1

src/monad/core/test/sha256.cpp
rgarc: replaced and supplemented with more proper tests in 030eb045bdb43378223e040c33f4017be95d1ac8

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -41,10 +41,49 @@ namespace fork_traits
...
+        using per_word = std::integral_constant<int64_t, PerWordCost>;
+
+        /**
+         * Implements the generic form of YP Appendix E Eq 221
+         * @param size of input message in bytes
+         * @return the total gas cost
+         */
+        static constexpr int64_t compute(size_t size)
+        {
+            auto constexpr word_size = sizeof(monad::bytes32_t);
```
rgarc, line 68: I'm still not sure of a reasonable runtime bound I can put on `size` so that the arithmetic expression this function returns is sound.

zander-xyz: https://twitter.com/SeanParent/status/1647018829892169730

There's still a chance that we're okay!

In all seriousness, if `size` causes our signed int arithmetic to be in UB territory, something wrong has happened. If the YP doesn't put a reasonable bound on size, I'm inclined to say that we should just leave this as is. 

Side note: integer division in c++ involves discarding the fractional portion of the result. Just looking at the YP, it seems like they define the division in this particular equation as taking the ceiling, which is not what's happening here. Is that okay? If not, can we fix and add a unit test which tests this?

rgarc: I accomplish taking the ceiling by adding `word_size - 1` to the numerator, but perhaps it's not the worst thing to make that more explicit with a proper call to some ceiling function.

zander-xyz: Ah, nice! TIL :)

include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
```diff
@@ -10,7 +11,24 @@ namespace static_precompiles
...
+                return {.status_code = evmc_status_code::EVMC_OUT_OF_GAS};
+            }
+
+            auto const result =
+                silkpre_ecrec_run(message.input_data, message.input_size);
+
+            return {
+                .status_code = evmc_status_code::EVMC_SUCCESS,
+                .gas_left = message.gas - cost,
+                .output_data = result.data,
```
zander-xyz, line 30: Does this leak memory at all? `silkpre_erec_run` allocates and then returns the pointer (what's stored in result). Then we do a shallow copy of that pointer into `evmc_result`.

I guess ASAN would be able to tell us the answer here.

rgarc: I had the same question when writing this. I'm pretty sure you're right that in isolation, this code leaks. I justified this to myself by checking the [call-site](https://github.com/monad-crypto/monad/blob/1210d20c9f990ba10167c4f5a073da5d5421b6f0/include/monad/execution/evm.hpp#L51-L58). We wrap the `evmc_result` in a [`evmc::Result`](https://evmc.ethereum.org/classevmc_1_1Result.html#aae63ab0def71eae3c0d6e5e60627d399) type [here](https://github.com/monad-crypto/monad/blob/1210d20c9f990ba10167c4f5a073da5d5421b6f0/include/monad/execution/evm.hpp#L64) which allegedly has proper RAII semantics and assumes ownership.

zander-xyz: Hmm, are you sure? I don't see `evmc::Result::release` being specialized here. Also, thoughts on just having this return an `evmc::Result` directly?

rgarc: I thought `evmc::Result::release` gets called in the [destructor](https://evmc.ethereum.org/classevmc_1_1Result.html#a4d78e0a5f1e1a32be7fa630fd2936072). I'm on board with returning an `evmc::Result` directly so someone does not have to think about this in the future.

zander-xyz: it does if you set it!

rgarc: Oh, you're absolutely right. It is never being set. Thanks for closely reading, because I would have missed this otherwise.




AddressSanitizer? More like AlexSanitizer.

rgarc: I'm pretty sure this is the only way to fix 
```diff
diff --git a/include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp b/include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
index f55be1c..0412baf 100644
--- a/include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
+++ b/include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
@@ -27,7 +27,8 @@ namespace static_precompiles
                 .status_code = evmc_status_code::EVMC_SUCCESS,
                 .gas_left = message.gas - cost,
                 .output_data = result.data,
-                .output_size = result.size};
+                .output_size = result.size,
+                .release = evmc_free_result_memory};
         }
     };
 }
```

zander-xyz: Thanks for making that fix! Without the wrapping with `evmc::Result`, this still leaks in the unit tests.

Just curious, are you able to verify this with ASAN or Valgrind?

include/monad/execution/ethereum/static_precompiles/ripemd160_hash.hpp
```diff
@@ -10,7 +11,25 @@ namespace static_precompiles
...
+            auto const cost = gas_cost::compute(message.input_size);
+            if (message.gas < cost) {
+                return {.status_code = evmc_status_code::EVMC_OUT_OF_GAS};
+            }
+
+            bytes32_t output{};
+            silkpre_rmd160(
+                output.bytes, message.input_data, message.input_size);
+
+            return evmc_make_result(
```
zander-xyz: Kind of unfortunate that `evmc_make_result` always allocates :(

include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
rgarc: Yeah the unit tests definitely leak which isn't surprising. Verified with valgrind.
```
==2003398== LEAK SUMMARY:
==2003398==    definitely lost: 224 bytes in 7 blocks
==2003398==    indirectly lost: 0 bytes in 0 blocks
==2003398==      possibly lost: 0 bytes in 0 blocks
==2003398==    still reachable: 1,128,488 bytes in 376 blocks
==2003398==         suppressed: 0 bytes in 0 blocks
```
After commenting out the tests, I get
```
==2003994== LEAK SUMMARY:
==2003994==    definitely lost: 0 bytes in 0 blocks
==2003994==    indirectly lost: 0 bytes in 0 blocks
==2003994==      possibly lost: 0 bytes in 0 blocks
==2003994==    still reachable: 14,152 bytes in 375 blocks
==2003994==         suppressed: 0 bytes in 0 blocks
```

rgarc: @sheapwilliams do you have any objections to me refactoring such that
```diff
diff --git a/include/monad/execution/static_precompiles.hpp b/include/monad/execution/static_precompiles.hpp
index 7d37f57..fbea844 100644
--- a/include/monad/execution/static_precompiles.hpp
+++ b/include/monad/execution/static_precompiles.hpp
@@ -17,7 +17,7 @@ template <
     class TState, concepts::fork_traits<TState> TTraits, class TPrecompiles>
 struct StaticPrecompiles
 {
-    using exec_func_t = evmc_result (*)(const evmc_message &) noexcept;
+    using exec_func_t = evmc::Result (*)(const evmc_message &) noexcept;
 
     template <typename... Types>
     static constexpr auto
```
I think that line was last touched by an intern Gabriel Lee.

rgarc: I rolled the aforementioned refactor at b9ebd8b93671a45990a2d0c29a7b5764fb1cbd37 and confirmed unit tests do not leak

docker/Dockerfile
```diff
@@ -24,14 +24,16 @@ RUN yum -y install \
...
+  gmp
 
 RUN yum -y install \
   abseil-cpp-devel \
   boost-devel \
   brotli-devel \
   gtest-devel \
-  rocksdb-devel
+  rocksdb-devel \
+  gmp-devel
```
sheapwilliams: [nit] can you ABC order these and above?

include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
```diff
@@ -10,7 +10,7 @@ namespace static_precompiles
     template <class TFork>
     struct BigNumberAdd
     {
-        static evmc_result execute(const evmc_message &m) noexcept;
+        static evmc::Result execute(const evmc_message &m) noexcept;
```
sheapwilliams, line 13: These are fine for now...the RAII wrapper has disabled copies, so this might lead to other issues later in integration.  But we can go with this for now and try to make it all work with move semantics.

include/monad/execution/ethereum/static_precompiles/elliptic_curve_recover.hpp
sheapwilliams: As I mentioned above, no objections.  This disables copy semantics, but we should be able to make that all work.

include/monad/execution/ethereum/static_precompiles/big_number_add.hpp
rgarc: Fair point. It took me way too long to figure out how to accomplish `value_or` once the type is no longer copyable. 😮‍💨 
```diff
diff --git a/include/monad/execution/evm.hpp b/include/monad/execution/evm.hpp
index c4c59f5..166dac5 100644
--- a/include/monad/execution/evm.hpp
+++ b/include/monad/execution/evm.hpp
@@ -48,20 +48,21 @@ struct Evm
             result.status_code != EVMC_SUCCESS) {
             return evmc::Result{result};
         }
-        evmc_result const result =
+        evmc::Result result =
             TStaticPrecompiles::static_precompile_exec_func(m.code_address)
                 .transform([&](auto static_precompile_execute) {
                     return static_precompile_execute(m);
                 })
                 // execute on backend, just this for now
-                .value_or(evmc_result{
-                    .status_code = EVMC_SUCCESS, .gas_left = m.gas});
+                .disjunction(evmc::Result{evmc_result{
+                    .status_code = EVMC_SUCCESS, .gas_left = m.gas}})
+                .value();

         if (result.status_code == EVMC_REVERT) {
             state.revert();
         }

-        return evmc::Result{result};
+        return result;
     }

     [[nodiscard]] static result_t
```

sheapwilliams: That works!

PR #98, opened 2023-06-09
Fix gcc build issue

tongzhi1998:
Problem:
- Gcc build fails due to the clang pragma & macro definition

Solution:
- Add CMake condition to determine if Clang is the compiler, if so, add flag
- Add newline at the end of macro defs


include/monad/logging/fake_log_macros.hpp
```diff
@@ -46,4 +46,4 @@
     do {                                                                       \
         (void)sizeof(logger);                                                  \
     }                                                                          \
-    while(0)
\ No newline at end of file
+    while(0)
```
tongzhi1998, line 49: This newline here is important. For some reason gcc compiler needs a newline after having \ in the macro definition.

include/monad/logging/quill_log_macros.hpp
```diff
@@ -1,7 +1,9 @@
 #pragma once
 
+#ifdef __clang__
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+#endif
```
sheapwilliams: I would have thought the `#pragma clang ...` part would accomplish the ifdef part.  There is something here I need to learn... 🤔

include/monad/logging/fake_log_macros.hpp
jhunsaker: what was the error?

include/monad/logging/quill_log_macros.hpp
jhunsaker: I think it warns about unknown macro? can you post the error?

include/monad/logging/fake_log_macros.hpp
tongzhi1998: error: backslash-newline at end of file [-Werror]
[build]    45 | #define MONAD_LOG_CRITICAL(logger, fmt, ...)                                   \

tongzhi1998: I don't understand why this error occurs (and why specific to gcc), but adding newline at EOF seems to fix the issue.

include/monad/logging/quill_log_macros.hpp
tongzhi1998: quill_log_macros.hpp:4: error: ignoring ‘#pragma clang diagnostic’ [-Werror=unknown-pragmas]
[build]     4 | #pragma clang diagnostic push

But it does have this line:
cc1plus: all warnings being treated as errors

Because we compile with -Werror

jhunsaker: just add a conditional in the cmake for clang as this is fixed in c++20

https://stackoverflow.com/a/68853995/19035862
https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

include/monad/logging/fake_log_macros.hpp
zander-xyz: At a certain point the standard defined non-empty files without a newline as undefined behavior. Since C++11, the standard spells out that these files shall be processed as if a newline was appended to the end of the file. 

https://eel.is/c++draft/lex.phases#1.2

My guess is that no one has bothered to remove this warning from gcc.

PR #99, opened 2023-06-12
Add logging for trie updates and more root hash unit tests

zander-xyz:
- Added logging of trie updates
- Made logging optionally enabled (with default being disabled)
- Add additional assertions for root hash

Will continue to beef up the unit tests as we integrate

Resolves

https://github.com/monad-crypto/monad/issues/97

### DIFF
--- a/include/monad/db/trie_db.hpp
+++ b/include/monad/db/trie_db.hpp
@@ -234,19 +234,19 @@ namespace impl
 
     template <
         typename TAccountsCursor, typename TStorageCursor,
-        typename TAccountsWriter, typename TStorageWriter, typename TExecutor>
+        typename TAccountsWriter, typename TStorageWriter, typename TExecution>
     struct TrieDBImpl
         : public DBInterface<
               TrieDBImpl<
                   TAccountsCursor, TStorageCursor, TAccountsWriter,
-                  TStorageWriter, TExecutor>,
-              TExecutor>
+                  TStorageWriter, TExecution>,
+              TExecution>
     {
         using DBInterface<
             TrieDBImpl<
                 TAccountsCursor, TStorageCursor, TAccountsWriter,
-                TStorageWriter, TExecutor>,
-            TExecutor>::updates;
+                TStorageWriter, TExecution>,
+            TExecution>::updates;
 
         template <typename TCursor, typename TWriter>
         struct Trie
--- a/include/monad/logging/config.hpp
+++ b/include/monad/logging/config.hpp
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <monad/config.hpp>
 
 #define MONAD_LOG_NAMESPACE_BEGIN MONAD_NAMESPACE_BEGIN namespace log {


test/include/monad/test/environment.hpp
```diff
@@ -0,0 +1,25 @@
...
+public:
+    void SetUp() override
+    {
+        monad::log::logger_t::start();
+
+        auto *logger = monad::log::logger_t::create_logger("trie_db_logger");
+        MONAD_ASSERT(logger);
+
+        monad::log::logger_t::set_log_level(
+            "trie_db_logger", monad::log::level_t::Info);
```
sheapwilliams: Off-hand, this seems too specific to be in test/include/...  What is the thinking of putting this here?

test/include/monad/test/hard_updates.hpp
```diff
@@ -1,13 +1,13 @@
 #pragma once
 
+#include <monad/test/config.hpp>
+
```
sheapwilliams: Similarly, this seemed at home in monad/trie/test to me.  When you get a chance, let's chat so I understand what you are up thinking.

src/monad/db/test/CMakeLists.txt
```diff
@@ -10,4 +11,5 @@ add_unit_test(
   value_store.cpp
   trie_db.cpp
   LIBRARIES
-  monad)
+  monad
+)
```
sheapwilliams: I just use `cmake-format` for these, so we don't have to think about hand-formatting them different ways in different commits.  We all should, I think.

include/monad/logging/fake_logger.hpp
```diff
@@ -10,15 +10,15 @@ MONAD_LOG_NAMESPACE_BEGIN
...
-    FakeBacktrace,
+    TraceL3,
+    TraceL2,
+    TraceL1,
+    Debug,
+    Info,
+    Warning,
+    Error,
+    Critical,
+    Backtrace,
```
sheapwilliams, line 21: :+1:

include/monad/logging/formatter.hpp
```diff
@@ -27,23 +61,104 @@ namespace quill
...
+    {
+        template <typename FormatContext>
+        auto format(monad::bytes32_t const &value, FormatContext &ctx) const
+        {
+            fmt::format_to(ctx.out(), "0x");
+            for (auto const &v : value.bytes) {
+                fmt::format_to(ctx.out(), "{:02x}", v);
+            }
+            return ctx.out();
+            return ctx.out();
```
tongzhi1998: Typo here? Remove one `return ctx.out()`

include/monad/logging/monad_log.hpp
```diff
@@ -1,17 +1,20 @@
 #pragma once
 
+#include "monad/logging/fake_logger.hpp"
```
tongzhi1998: Do we need this include?

include/monad/db/trie_db.hpp
```diff
@@ -276,6 +277,10 @@ namespace impl
         std::vector<trie::Update> account_trie_updates{};
         std::vector<trie::Update> storage_trie_updates{};
 
+        decltype(monad::log::logger_t::get_logger()) logger =
```
tongzhi1998, line 280: Seems like a good way to avoid injecting the `TLogger` dependency into every place where we need it. If we decide to do this, then I could reformat my previous code to remove `TLogger`. 
Any thought? @zander-xyz @sheapwilliams

include/monad/logging/fake_logger.hpp
tongzhi1998: Thanks for fixing this. Realized I was doing it wrong...

include/monad/logging/formatter.hpp
```diff
@@ -1,18 +1,52 @@
 #pragma once
 
 #include <monad/execution/replay_block_db.hpp>
```
tongzhi1998, line 3: I fix it in my upcoming PR, but we really shouldn't include `replay_block_db.hpp` here since it will create circular dependency. Just delete this line or replace it with `<monad/core/bytes.hpp>` if you encounter any error (if not, then I will fix it in my PR).

include/monad/logging/monad_log.hpp
zander-xyz: Nope, my LSP strikes again...adding random includes. Thanks

include/monad/db/trie_db.hpp
zander-xyz: That sounds like a good route to go down. Feel free to add that as an issue and someone can get around to doing it when they get free cycles.

include/monad/logging/formatter.hpp
zander-xyz: I'll let ya fix it in your PR :)

test/include/monad/test/environment.hpp
zander-xyz: Discussed over zoom. 

Put this all in `test/unit/common` to make intention of usage clear.

test/include/monad/test/hard_updates.hpp
zander-xyz: Discussed over zoom. Moving this back to trie

include/monad/db/trie_db.hpp
sheapwilliams: Discussed in zoom - I think this sort of accessor for service plane elements is just fine, as longs as it is built to be faked at some level like the logger currently is.

PR #101, opened 2023-06-14
Add Logger to BP and FiberData; Update Behave test cases

tongzhi1998:
- Adding Logger to both Block Processor and Transaction Fiber Data
- Adding Behave test cases accordingly


include/monad/execution/transaction_processor_data.hpp
```diff
@@ -59,6 +63,11 @@ struct alignas(64) TransactionProcessorFiberData
     inline void validate_execute_and_apply_state_diff()
     {
         TTxnProcessor p{};
+        auto *txn_logger = log::logger_t::get_logger("txn_logger");
+        std::chrono::_V2::steady_clock::time_point const start_time =
```
rgarc: I haven't seen the `_V2` namespace used before in code. What is this for?

tongzhi1998: I guess I just copied some code online for the timer. You are right, it seems we don't really need _V2 (and I just realized it's not part of c++ standard).

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -71,6 +80,9 @@ struct alignas(64) TransactionProcessorFiberData
                     (s_.current_txn() == id_ &&
                      status != TxnReadyStatus::WILL_SUCCEED) ||
                     status == TxnReadyStatus::ERROR) {
+
+                    MONAD_LOG_WARNING(
```
sheapwilliams: This should probably be info, not a warning.

include/monad/execution/block_processor.hpp
```diff
@@ -16,6 +19,44 @@ struct AllTxnBlockProcessor
...
+            b.header.transactions_root,
+            b.header.receipts_root,
+            log::fmt_byte_string_fixed_wrapper<256>(b.header.logs_bloom),
+            b.header.difficulty,
+            b.header.number,
+            b.header.gas_limit,
+            b.header.gas_used,
+            b.header.timestamp,
+            log::fmt_byte_string_wrapper{b.header.extra_data});
+
```
sheapwilliams, line 30: Let's move the long log into a helper function to keep the signature with the logic as best as possible.

features/ethereum/steps/read_block_history.py
```diff
@@ -66,5 +75,34 @@ def then_root_match(context, root_type):
...
+
+@then("the output should contain block header details")
+def then_no_blockheader_details(context):
+    assert "BlockHeader Fields" in context.output
+
+
+@then("the output should contain transaction scheduling information")
+def then_transaction_schedule(context):
+    assert "Start executing Transaction" in context.output
+    assert "Finish executing Transaction" in context.output
```
sheapwilliams: I think this can be generalized as we discussed on zoom.

include/monad/execution/block_processor.hpp
```diff
@@ -38,6 +79,19 @@ struct AllTxnBlockProcessor
         for (auto &d : data) {
             r.push_back(d.get_receipt());
         }
+
+        std::chrono::steady_clock::time_point const finished_time =
+            std::chrono::steady_clock::now();
+        [[maybe_unused]] uint64_t elapsed_ms = static_cast<uint64_t>(
```
sheapwilliams: Let's look at some variadic preprocessor magic to see if we can't get rid of these `[[maybe_unused]]` parts of logs.

include/monad/logging/formatter.hpp
```diff
@@ -160,4 +210,20 @@ namespace fmt
             return ctx.out();
         }
     };
+
+    template <unsigned N>
+    struct formatter<intx::uint<N>> : public monad::log::basic_formatter
```
sheapwilliams, line 349: Can these formatters be `constexpr`?  While `intx::to_string` isn't constexpr, perhaps many of these other formatters can be?

include/monad/execution/block_processor.hpp
zander-xyz: Possibly also write an `fmt::formatter` if we actually do want to log the entire thing. It also has the nice side effect that formatting happens in the background thread rather than here.

include/monad/execution/block_processor.hpp
```diff
@@ -16,6 +19,44 @@ struct AllTxnBlockProcessor
...
+            "Gas Used = {}, "
+            "Timestamp = {}, "
+            "Extra Data = {} ",
+            b.header.parent_hash,
+            b.header.ommers_hash,
+            b.header.beneficiary,
+            b.header.state_root,
+            b.header.transactions_root,
+            b.header.receipts_root,
+            log::fmt_byte_string_fixed_wrapper<256>(b.header.logs_bloom),
```
zander-xyz: It is a little bit unfortunate that the size can not be deduced. We can fix that by writing a constructor for the wrapper,
```cpp
fmt_byte_string_fixed_wrapper(byte_string_fixed<N> const& arg) : bsf(arg) {}
```
which then lets us spell the instantiation like
```cpp
log::fmt_byte_string_fixed_wrapper(b.header.logs_bloom)
```

include/monad/logging/formatter.hpp
```diff
@@ -1,11 +1,16 @@
 #pragma once
 
-#include <monad/execution/replay_block_db.hpp>
+#include <monad/core/address.hpp>
+#include <monad/core/byte_string.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/core/int.hpp>
+
```
zander-xyz, line 12: Can we also include `<monad/logging/config.h>`?

include/monad/execution/block_processor.hpp
zander-xyz: I'd think that all we really needed to do here was
```suggestion
        [[maybe_unused]] uint64_t elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(finished_time - start_time);
```
No need for the `count()` or the static_cast. `fmt/chrono.h` should contain the formatters we desire

zander-xyz: +1

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -87,8 +99,20 @@ struct alignas(64) TransactionProcessorFiberData
                 // Can merge needs to be in yield while loop while receiving
                 // TRY_AGAIN When WILL_SUCCEED is returned, merge, and return;
                 // if ERROR is received, then error out
+                std::chrono::steady_clock::time_point const finished_time =
+                    std::chrono::steady_clock::now();
+                [[maybe_unused]] uint64_t elapsed_ms = static_cast<uint64_t>(
```
zander-xyz: Similar comment here as above.

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -59,6 +63,11 @@ struct alignas(64) TransactionProcessorFiberData
     inline void validate_execute_and_apply_state_diff()
```
zander-xyz: sorry, while we're here can we get rid of this inline and above?

include/monad/logging/formatter.hpp
```diff
@@ -90,34 +120,54 @@ namespace fmt
...
             }
-            fmt::format_to(ctx.out(), "}}");
             return ctx.out();
         }
     };
 
-    template <>
-    struct formatter<monad::log::fmt_byte_string_wrapper>
+    template <size_t N>
+    struct formatter<monad::log::fmt_byte_string_fixed_wrapper<N>>
```
zander-xyz, line 210: Can we make the logging match that of `byte_string`? The only difference between the two is that one is a fixed size, so it's a bit jarring if their formatting is different.

include/monad/logging/formatter.hpp
zander-xyz: The parse can be constexpr for sure, but unfortunately the `format_to` is not :(

features/ethereum/configure_logger.feature
```diff
@@ -0,0 +1,48 @@
...
+    Scenario: Using default log level
+        Given I run with BlockDb path = "erigon/block_db"
+        And I run with start block number = "3000000"
+        And I run with finish block number = "3000001"
+        And I run with default log level
+        When I start "replay_ethereum"
+        Then the output should not contain block header details
+        And the output should contain transaction scheduling information
+
+    Scenario: Setting block_logger log_level = 3
```
zander-xyz: In order to figure out what `log_level = 3` means, I have to look at the code, which is one step of indirection I can avoid if we use the named enum here and then do the translation in the background to hide away the ugliness from the feature.

include/monad/execution/block_processor.hpp
tongzhi1998: I might be wrong, but I think `std::chrono::duration_cast<...>` returns a time duration, and I need to call `.count()` to get an integer. Furthermore, the return type of .count() is long, but I guess I could make `elapsed_time` as auto instead of enforcing it to be `uint64_t`.

zander-xyz: Try this

```cpp
#include <fmt/format.h>
#include <fmt/chrono.h>
#include <chrono>

int main() {
    auto const start = std::chrono::steady_clock::now();
    auto const ms = std::chrono::duration_cast<std::chrono::milliseconds>(
        std::chrono::steady_clock::now() - start);
    fmt::print("Elapsed: {}", ms);
}
```

include/monad/execution/block_processor.hpp
```diff
@@ -38,6 +50,17 @@ struct AllTxnBlockProcessor
         for (auto &d : data) {
             r.push_back(d.get_receipt());
         }
+
+        auto const finished_time = std::chrono::steady_clock::now();
+        auto elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
```
zander-xyz: in general, try to const any variables that you do not plan on changing. One gotcha with this advice is if you decide that you need move semantics.
```suggestion
        auto const elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
```

include/monad/logging/fake_log_macros.hpp
```diff
@@ -1,47 +1,64 @@
 #pragma once
 
+#define MONAD_LOG_TRACE_L3_HELPER(...) (void)(sizeof(__VA_ARGS__))
```
zander-xyz: All of these helpers macros do the same thing so why don't we just define the one?

include/monad/logging/fake_log_macros.hpp
```diff
@@ -1,47 +1,64 @@
...
 
+#define MONAD_LOG_TRACE_L2_HELPER(...) (void)(sizeof(__VA_ARGS__))
+
 #define MONAD_LOG_TRACE_L2(logger, fmt, ...)                                   \
     do {                                                                       \
         (void)sizeof(logger);                                                  \
     }                                                                          \
     while(0)
 
+#define MONAD_LOG_TRACE_L1_HELPER(...) (void)(sizeof(__VA_ARGS__))
```
zander-xyz: Seems like many of these helper macros are defined but never used, should we use them?

include/monad/logging/formatter.hpp
```diff
@@ -76,31 +133,54 @@ namespace fmt
...
+                bh.state_root,
+                bh.transactions_root,
+                bh.receipts_root,
+                monad::log::fmt_byte_string_fixed_wrapper(bh.logs_bloom),
+                bh.difficulty,
+                bh.number,
+                bh.gas_limit,
+                bh.gas_used,
+                bh.timestamp,
+                monad::log::fmt_byte_string_wrapper{bh.extra_data});
```
zander-xyz: For simple types like this wrapper, using designated initializers doesn't really help us _that_ much. But in general with structs, it's good practice to use them whenever possible for both readability and robustness. It's a lot harder to initialize structs incorrectly when the arguments are initialized by name rather than by position in the argument list. 
```suggestion
                monad::log::fmt_byte_string_wrapper{.bs=bh.extra_data});
```

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+        template <typename FormatContext>
+        auto format(monad::Receipt const &r, FormatContext &ctx) const
+        {
+            fmt::format_to(
+                ctx.out(),
+                "\"Bloom\": {}, "
+                "\"Status\": {}, "
+                "\"Gas Used\": {}, "
+                "\"Transaction Type\": {}, "
+                "\"Logs\": [",
```
zander-xyz: Do you really need the quotes around each member name?

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+    template <>
+    struct formatter<monad::Receipt::Log> : public monad::log::basic_formatter
+    {
+        template <typename FormatContext>
+        auto format(monad::Receipt::Log const &l, FormatContext &ctx) const
+        {
+            fmt::format_to(
+                ctx.out(),
+                "\"Data\": {}, ",
+                monad::log::fmt_byte_string_wrapper(l.data));
```
zander-xyz: designated initializers please
```suggestion
                monad::log::fmt_byte_string_wrapper{.bs=l.data});
```

include/monad/logging/formatter.hpp
```diff
@@ -23,17 +32,59 @@ struct fmt_byte_string_wrapper
 {
     byte_string bs;
 };
+
+template <size_t N>
+struct fmt_byte_string_fixed_wrapper
+{
+    byte_string_fixed<N> bsf;
+    fmt_byte_string_fixed_wrapper(byte_string_fixed<N> const &arg)
```
zander-xyz: Mark explicit please (this is to avoid accidental implement conversions)

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+            fmt::format_to(ctx.out(), "[");
+            for (auto i = 0u; i < v.size(); ++i) {
+                if (i != 0)
+                    fmt::format_to(ctx.out(), ", ");
+                fmt::format_to(ctx.out(), "{}", v[i]);
+            }
+            fmt::format_to(ctx.out(), "]");
+            return ctx.out();
+        }
+    };
```
zander-xyz: This formatter should not be needed. `std::vector<T>` should be fmt'able as long as `T` is as well.

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+
+    template <>
+    struct formatter<monad::Receipt> : public monad::log::basic_formatter
```
zander-xyz, line 227: For more complex types, what are your thoughts on always wrapping the string with `Type{...}`? So in this case, `Receipt{...}`. In my (subjective) opinion, this is more readable.

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+    {
+        template <typename FormatContext>
+        auto format(monad::Receipt const &r, FormatContext &ctx) const
+        {
+            fmt::format_to(
+                ctx.out(),
+                "\"Bloom\": {}, "
```
zander-xyz: I think I'd rather just the space separation as opposed to the space-and-comma separation. Both can act as a separator, so we should just choose one, and I think the space makes the log more readable rather than the comma.

In the same vein, I think I'd prefer `key=value` rather than `key: value` for similar reasons.

So putting this all together, something like:

`Type{key1=value1 key2=value2 key3=value3 ...}`

Thoughts?

@sheapwilliams @tongzhi1998

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+                "\"Status\": {}, "
+                "\"Gas Used\": {}, "
+                "\"Transaction Type\": {}, "
+                "\"Logs\": [",
+                monad::log::fmt_byte_string_fixed_wrapper(r.bloom),
+                r.status,
+                r.gas_used,
+                static_cast<int>(r.type));
```
zander-xyz: I think we should probably write a formatter for `Transaction::type` so that this is human readable.

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+            for (auto i = 0u; i < r.logs.size(); ++i) {
```
zander-xyz: You can format the vector of logs directly in the above `format_to` instead of manually doing it here.

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+            fmt::format_to(
+                ctx.out(),
+                "\"Data\": {}, ",
+                monad::log::fmt_byte_string_wrapper(l.data));
+            fmt::format_to(ctx.out(), "\"Topics\": [");
+            for (auto i = 0u; i < l.topics.size(); ++i) {
+                if (i != 0)
+                    fmt::format_to(ctx.out(), ", ");
+                fmt::format_to(ctx.out(), "{}", l.topics[i]);
+            }
```
zander-xyz: Same comment as elsewhere: as long as `T` is formattable, there is no need to manually format the `std::vector<T>`, just let the library do all the work :).

include/monad/logging/formatter.hpp
```diff
@@ -122,6 +202,107 @@ namespace fmt
...
+        {
+            fmt::format_to(ctx.out(), "Nibbles{{0x");
+            for (uint8_t i = 0; i < n.size(); ++i) {
+                MONAD_DEBUG_ASSERT(n[i] <= 0xf);
+                fmt::format_to(ctx.out(), "{:01x}", n[i]);
+            }
+            fmt::format_to(ctx.out(), "}}");
+            return ctx.out();
+        }
+    };
```
zander-xyz, line 307: Weird that this is popping up as an addition in your PR...do you perhaps need to rebase?

include/monad/logging/formatter.hpp
```diff
@@ -160,4 +341,20 @@ namespace fmt
...
         }
     };
+
+    template <unsigned N>
+    struct formatter<intx::uint<N>> : public monad::log::basic_formatter
+    {
+        template <typename FormatContext>
+        auto format(intx::uint<N> const &value, FormatContext &ctx) const
+        {
+            // Seems redundent, actually not
```
zander-xyz:
```suggestion
            // Seems redundant, actually not
```

Also, I'm not sure I understand this comment. Can you explain?

include/monad/logging/formatter.hpp
```diff
@@ -160,4 +341,20 @@ namespace fmt
...
+            // intx::to_string() is not constexpr, so we can't use it directly
+            // And if we don't use to_string(), intx::uint<N> type doesn't have
+            // a way of outputting
+            std::string const temp = intx::to_string(value);
```
zander-xyz: If you're going to name this variable `temp`, why not just inline it?

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -139,18 +138,37 @@ int main(int argc, char *argv[])
     std::filesystem::path block_db_path{};
     monad::block_num_t start_block_number;
     std::optional<monad::block_num_t> finish_block_number = std::nullopt;
-    quill::LogLevel log_level = quill::LogLevel::Info;
 
-    cli.add_option("-b, --block-db", block_db_path, "block_db directory")
+    monad::log::logger_t::start();
+
+    // create all the loggers needed for the program
+    [[maybe_unused]] auto *main_logger =
```
zander-xyz: nit: you don't need the maybe_unused for the main logger
```suggestion
    auto *main_logger =
```

include/monad/logging/fake_log_macros.hpp
tongzhi1998: Yeah, that totally makes sense. Don't know what I was thinking...

include/monad/logging/formatter.hpp
tongzhi1998: Hm... I think it is because I reorganize the sequence of those formatters, so that this shows up as new. 
To be more specific, Github thinks I change your original Nibble to bytes32_t and write a new Nibble, but what actually happens is that I swap the sequence...

include/monad/logging/formatter.hpp
tongzhi1998: I was actually wrong about this, will fix it

include/monad/logging/formatter.hpp
tongzhi1998: +1

include/monad/logging/fake_log_macros.hpp
tongzhi1998: Same as above...

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -87,13 +97,23 @@ struct alignas(64) TransactionProcessorFiberData
                 // Can merge needs to be in yield while loop while receiving
                 // TRY_AGAIN When WILL_SUCCEED is returned, merge, and return;
                 // if ERROR is received, then error out
+                auto const finished_time = std::chrono::steady_clock::now();
+                auto elapsed_ms =
```
zander-xyz: think you missed the const here

include/monad/logging/formatter.hpp
```diff
@@ -76,31 +136,55 @@ namespace fmt
...
+                bh.transactions_root,
+                bh.receipts_root,
+                monad::log::fmt_byte_string_fixed_wrapper(bh.logs_bloom),
+                bh.difficulty,
+                bh.number,
+                bh.gas_limit,
+                bh.gas_used,
+                bh.timestamp,
+                monad::log::fmt_byte_string_wrapper{.bs = bh.extra_data});
+            fmt::format_to(ctx.out(), "}}");
```
zander-xyz: nit: this is OK, but you don't need separate `fmt::format_to`s for the wrapper `BlockHeader{}`

test/features/ethereum/steps/read_block_history.py
```diff
@@ -0,0 +1,94 @@
...
+def when_start(context, program):
+    executable_path = os.getcwd() + "/build/src/monad/execution/ethereum/" + program
+    if not hasattr(context, "block_db_path"):
+        raise ValueError("Block Db path is required")
+    if not hasattr(context, "start_block_number"):
+        raise ValueError("Start block number is required")
+    args = [
+        executable_path,
+        "--block_db",
+        os.getcwd() + "/../../" + context.block_db_path,
```
zander-xyz, line 45: Is all of this stuff with `os.getcwd()` implying that we need to run behave from a certain folder?

tongzhi1998: Unfortunately yes. I am going to raise an issue for this. Shea and I discussed about this earlier and we thought it's better to postpone it to a later PR.

include/monad/logging/fake_log_macros.hpp
```diff
@@ -1,49 +1,59 @@
 #pragma once
 
+#define MONAD_LOG_HELPER(...) (void)(sizeof(__VA_ARGS__))
+
```
sheapwilliams, line 4: [praise] Nice!

include/monad/logging/formatter.hpp
```diff
@@ -1,13 +1,25 @@
...
+
 #include <monad/trie/nibbles.hpp>
 #include <monad/trie/update.hpp>
 
 #include <quill/Quill.h>
 
+#include <optional>
 #include <type_traits>
 
+#include <iostream>
```
sheapwilliams, line 21: Still need this guy?

PR #102, opened 2023-06-14
:new: Evmone

sheapwilliams:
Create the initial interpreter object.

Follow-on integration tests will be in a C++ BDD style for the state and host.

### DIFF
--- a/include/monad/execution/evmone_baseline_interpreter.hpp
+++ b/include/monad/execution/evmone_baseline_interpreter.hpp
@@ -1,0 +1,44 @@
+#pragma once
+
+#include <monad/core/byte_string.hpp>
+#include <monad/core/concepts.hpp>
+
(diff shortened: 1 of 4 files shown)


src/monad/db/test/code_store.cpp
```diff
@@ -132,7 +132,8 @@ TEST(CodeStore, merge_changes)
         EXPECT_TRUE(s.can_merge(t));
         s.merge_changes(t);
     }
-    EXPECT_EQ(0, std::memcmp(s.code_at(b).c_str(), c2.c_str(), c2.size()));
+    //EXPECT_EQ(0, std::memcmp(s.code_at(b).c_str(), c2.c_str(), c2.size()));
```
sheapwilliams: Removing....

PR #103, opened 2023-06-15
:new: Evmone

sheapwilliams:
Create the initial interpreter object.

Follow-on integration tests will be in a C++ BDD style for the state and host.


include/monad/db/code_store.hpp
```diff
@@ -32,15 +32,15 @@ struct CodeStore
     {
     }
 
-    [[nodiscard]] byte_string const &code_at(address_t const &a) const
+    [[nodiscard]] byte_string_view const code_at(address_t const &a) const
```
zander-xyz: If we move to byte_string_view then we probably don't need to qualify the return type with `const`

include/monad/execution/evm.hpp
```diff
@@ -23,26 +23,27 @@ struct Evm
...
     {
         auto const contract_address = make_account_address(state, m);
         if (!contract_address) {
             return evmc::Result{contract_address.error()};
         }
-        // evmone execute, just this for now
-        evmc_result res = {.status_code = EVMC_SUCCESS, .gas_left = 12'000};
+
+        auto interpreter = TInterpreter{};
+        auto result = interpreter.execute(host, m);
```
zander-xyz: Thoughts on making `execute` a static function instead so that we don't need to instantiate the interpreter first? This seems consistent with what Rene is doing with the static precompiles.

include/monad/execution/evmone_baseline_interpreter.hpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        auto const code =
+            s.code_at(m.sender); // call needs to be plumbed through state...
+        if (code.empty()) {
+            return {.status_code = EVMC_SUCCESS, .gas_left = m.gas};
+        }
+
+        evmone::VM v{};
+        evmone::ExecutionState es{
+            m, TTraits::rev, *reinterpret_cast<evmc_host_interface *>(h), nullptr, byte_string_view{code}, {}};
```
zander-xyz: What's your plan for `fake::EvmHost` here? It seems like it doesn't implement the `evmc_host_interface` currently. Is it just that the host is never being used in a meaningful way?

What do we expect the host to look like in the future? Shall we discuss over zoom?

Also, clang-format?

include/monad/execution/evmone_baseline_interpreter.hpp
```diff
@@ -0,0 +1,44 @@
...
+        evmone::baseline::CodeAnalysis ca{
+            evmone::baseline::analyze(TTraits::rev, byte_string_view{code})};
```
zander-xyz: You changed code_at to return the byte_string_view directly already (ditto above)

src/monad/core/receipt.cpp
```diff
@@ -9,16 +9,16 @@ MONAD_NAMESPACE_BEGIN
 void set_3_bits(Receipt::Bloom &bloom, byte_string_view const bytes)
 {
     auto const h{ethash::keccak256(bytes.cbegin(), bytes.size())};
```
zander-xyz, line 12: is this part of the YP? if so, small request to add that as a comment (similar to how you've done it elsewhere)

src/monad/execution/test/evm.cpp
```diff
@@ -304,14 +306,16 @@ TEST(Evm, revert_create_account)
     s._map.emplace(from, Account{.balance = 10'000});
     traits_t::_gas_creation_cost = 10'000;
     traits_t::_success_store_contract = false;
+    fake::Interpreter::_result = {
+        .status_code = EVMC_REVERT, .gas_left = 11'000};
```
zander-xyz: are you using this?

src/monad/execution/test/evm.cpp
```diff
@@ -324,6 +328,8 @@ TEST(Evm, call_evm)
     fake::EvmHost h{};
     s._map.emplace(from, Account{.balance = 50'000u});
     s._map.emplace(to, Account{.balance = 50'000u});
+    fake::Interpreter::_result = {
+        .status_code = EVMC_SUCCESS, .gas_left = 7'000};
```
zander-xyz: same question here

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -135,14 +140,14 @@ namespace fork_traits
             return 0u;
         }
 
-        [[nodiscard]] static constexpr inline auto
+        [[nodiscard]] static constexpr auto
```
zander-xyz, line 181: 👍

include/monad/execution/evm.hpp
sheapwilliams: Interpreter _may_ eventually have locals (maybe the host i/f? integration will tell), or use a static pool for construction efficiency...  I don't think I know the final form of this guy just yet, and fully expect it to change.  So I'm convinced it will change shortly.

But I can make it that way for now...why not?

include/monad/execution/evmone_baseline_interpreter.hpp
sheapwilliams: Yea, we'll know more in integration, and be better able to make better choices then.

include/monad/execution/evmone_baseline_interpreter.hpp
sheapwilliams: Right - of course. :+1:

src/monad/execution/test/evm.cpp
sheapwilliams: Y - this is testing that a contract reversion will revert the created address.  [This](https://github.com/monad-crypto/monad/pull/103/files/a84af19ed4ccba183f449d7436872d5ddbcfa8a3#diff-10ec1f4f7233893915d6dcddab9f8465b1038d3eb37d8863e6e386aa51c5a533R317) tests it.

src/monad/execution/test/evm.cpp
sheapwilliams: Same thing - this just tests that the interpreter result is plumbed thru.

include/monad/execution/evmone_baseline_interpreter.hpp
zander-xyz: Sorry, what I was trying to get at was that you don't need the explicit cast to a byte_string_view
```suggestion
            evmone::baseline::analyze(TTraits::rev, code)};
```

sheapwilliams: Fixed both. :+1:

PR #105, opened 2023-06-15
Add persistent storage

zander-xyz:
Currently our database interface only supports an in memory backend, so there isn't any concept of data persistence. This PR implements a rocksdb backend.

Currently, the persistent store creates a new database on every run. I've opened https://github.com/monad-crypto/monad/issues/104, which I'll resolve in a separate PR to open an existing database.

### DIFF
--- a/include/monad/core/assert.h
+++ b/include/monad/core/assert.h
@@ -14,18 +14,20 @@ void monad_assertion_failed(
 }
 #endif
 
-#define MONAD_ASSERT(expr)                                                     \
-    (MONAD_LIKELY(!!(expr))                                                    \
-         ? ((void)0)                                                           \
-         : monad_assertion_failed(                                             \
+// clang-format off
+#define MONAD_ASSERT(expr)                                                       \
+    (MONAD_LIKELY(!!(expr))                                                      \
+         ? ((void)0)                                                             \
+         : monad_assertion_failed(                                               \
                #expr, __PRETTY_FUNCTION__, __FILE__, __LINE__))
 
 #ifdef NDEBUG
-    #define MONAD_DEBUG_ASSERT(x)                                              \
-        do {                                                                   \
-            (void)sizeof(x);                                                   \
-        }                                                                      \
+    #define MONAD_DEBUG_ASSERT(x)                                                \
+        do {                                                                     \
+            (void)sizeof(x); /* NOLINT: suppressing bugprone-sizeof-expression*/ \
+        }                                                                        \
         while (0)
 #else
     #define MONAD_DEBUG_ASSERT(x) MONAD_ASSERT(x)
 #endif
+// clang-format on
--- a/include/monad/db/trie_db.hpp
+++ b/include/monad/db/trie_db.hpp
@@ -3,6 +3,7 @@
 #include <monad/core/account.hpp>
 #include <monad/core/assert.h>
 #include <monad/db/config.hpp>
+#include <monad/db/util.hpp>
 #include <monad/execution/execution_model.hpp>
 #include <monad/logging/monad_log.hpp>
 #include <monad/rlp/decode_helpers.hpp>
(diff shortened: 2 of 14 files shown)


include/monad/core/assert.h
```diff
@@ -14,18 +14,20 @@ void monad_assertion_failed(
 }
 #endif
 
-#define MONAD_ASSERT(expr)                                                     \
-    (MONAD_LIKELY(!!(expr))                                                    \
-         ? ((void)0)                                                           \
-         : monad_assertion_failed(                                             \
+// clang-format off
```
zander-xyz, line 17: otherwise clang-format might move the NOLINT around

include/monad/db/trie_db.hpp
```diff
@@ -272,10 +264,196 @@ namespace impl
...
+            : accounts{}
+            , storage{}
+        {
+        }
+
+        void take_snapshot() {}
+        void release_snapshot() {}
+    };
+
+    struct PersistentTrieSetup
```
sheapwilliams: Any allergy to just calling this a RocksDBTrieSetup?  We'll have at least to persistent tries.

include/monad/db/trie_db.hpp
```diff
@@ -236,6 +240,199 @@ namespace impl
...
+    make_basic_storage_key(address_t const &a, bytes32_t const &k)
+    {
+        byte_string_fixed<sizeof(address_t) + sizeof(bytes32_t)> key;
+        std::copy_n(a.bytes, sizeof(address_t), key.data());
+        std::copy_n(k.bytes, sizeof(bytes32_t), &key[sizeof(address_t)]);
+        return key;
+    }
+
+    template <typename TExecution>
+    struct PersistentDBImpl
```
sheapwilliams: Similar Q here...Just name it Rocks...?

include/monad/db/trie_db.hpp
```diff
@@ -236,6 +240,199 @@ namespace impl
...
+        std::shared_ptr<rocksdb::DB> db;
+        rocksdb::WriteBatch batch;
+
+        PersistentDBImpl(
+            std::filesystem::path name = std::filesystem::absolute("db"))
+            : name(name)
+            , options([]() {
+                rocksdb::Options ret;
+                ret.IncreaseParallelism(2);
+                ret.OptimizeLevelStyleCompaction();
```
sheapwilliams, line 270: Do we want these configs to be injected into these objects at some point?

include/monad/db/trie_db.hpp
```diff
@@ -348,17 +544,28 @@ namespace impl
...
+                    std::filesystem::create_directory(name);
+                }
+
                 rocksdb::DB *db = nullptr;
 
-                rocksdb::Status const s =
-                    rocksdb::DB::Open(options, name, cfds, &cfs, &db);
+                rocksdb::Status const s = rocksdb::DB::Open(
+                    options,
+                    name / fmt::format("{}", std::chrono::system_clock::now()),
```
sheapwilliams, line 564: I suppose this is similar - we want it injected, but sounds like you intend to follow-up with that.

include/monad/db/trie_db.hpp
zander-xyz: Yep, we definitely want to do that.

Modified https://github.com/monad-crypto/monad/issues/104 to widen it's scope to include injecting the `rocksdb::Options` (as opposed to just the database name)

include/monad/db/trie_db.hpp
```diff
@@ -723,4 +926,7 @@ using RocksTrieDB = impl::TrieDBImpl<
 // Database impl without trie root generating logic, backed by stl
 using InMemoryDB = impl::InMemoryDBImpl<monad::execution::BoostFiberExecution>;
 
+// Database impl without trie root generating logic, backed by rocksdb
+using RocksDB = impl::RocksDBDBImpl<monad::execution::BoostFiberExecution>;
```
sheapwilliams: [no action] Mildly unfortunate name...but I understand it. 😄

PR #120, opened 2023-06-19
:recycle: Replace contains-at access pattern with a db query

zander-xyz:
Problem:
- There are many instances where we follow a contains incantation with a subsequent at. While the double call may be optimized away by compilers for STL backends, they will likely not optimize away for a real database backend. Moreover, the contains and at call is (currently) not fiber aware whereas query is.

Solution:
- Replace pattern with a call to query, which will asynchronously fetch the value and yield the current fiber

Related to https://github.com/monad-crypto/monad/issues/96

### DIFF
--- a/include/monad/db/value_store.hpp
+++ b/include/monad/db/value_store.hpp
@@ -69,17 +69,11 @@ struct ValueStore
         return false;
     }
 
-    bool
-    db_contains_key(address_t const &a, bytes32_t const &key) const noexcept
-    {
-        return db_.contains(a, key);
-    }
-
     bool db_or_merged_contains_key(
         address_t const &a, bytes32_t const &key) const noexcept
     {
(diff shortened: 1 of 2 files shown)


include/monad/db/value_store.hpp
```diff
@@ -91,22 +85,15 @@ struct ValueStore
...
             return merged_.storage_.at(a).at(key).updated;
         }
-        if (db_contains_key(a, key)) {
-            return db_.at(a, key);
-        }
-        return {};
+        return db_.query(a, key).value_or(bytes32_t{});
     }
 
+    // Note: just for unit testing
```
sheapwilliams: Technically it is for debug asserting, but I don't think that distinction is super important.  Just noting here for posterity.

zander-xyz: Fixed up the comment

Issue #113, opened 2023-06-18
CMake config Python files to make BDD tests able to run from anywhere


tongzhi1998:
- Currently Behave test cases can be only run from the `${PROJECT_SOURCE_DIR} `due to the usage of `os.getcwd()` in python steps files
- There seems to be also restriction from behave itself such that if the folder to steps directory is not immediately findable, behave won't work (e.g. we can only do `behave test/features/ethereum`, not `behave test`)
- We need to refactor it to make it able to run from anywhere

PR #126, opened 2023-06-21
:sparkles: Add trie debugger (tdb)

zander-xyz:
Problem:
- If there is a state root mismatch between the ethereum block and what we calculated, it is hard to identify the root cause

Solution:
- Introduce a tdb to help separate bugs in the trie from everything else

### DIFF
--- a/src/monad/db/test/CMakeLists.txt
+++ b/src/monad/db/test/CMakeLists.txt
@@ -1,5 +1,8 @@
 add_unit_test(
-    # ENABLE_LOGGING
+    # TODO: disable logging again when behave features are switched over to use
+    # replay_ethereum. For now, this needs to be enabled for those behave tests
+    # to pass
+    ENABLE_LOGGING
     TARGET
     monad-dbtests
     SOURCES
--- a/test/features/tdb/root_hash.feature
+++ b/test/features/tdb/root_hash.feature
@@ -1,0 +1,12 @@
+Feature: TDB Root Hash
+
+    # TODO: make this generate the data from replay_ethereum when that becomes ready
+    Scenario: Verify state root hash from TrieDBTest/1.ModifyStorageOfAccount
+        Given I run db unit test TrieDBTest/1.ModifyStorageOfAccount
+        And I run tdb with the output
+        Then the output should contain "Root hash 0x0169f0b22c30d7d6f0bb7ea2a07be178e216b72f372a6a7bafe55602e5650e60"  
+
+    Scenario: Verify storage root hash from TrieDBTest/1.ModifyStorageOfAccount
+        Given I run db unit test TrieDBTest/1.ModifyStorageOfAccount
+        And I run tdb with the output and account 0x5353535353535353535353535353535353535353 
+        Then the output should contain "Root hash 0xf0481d515d9698359549c6d184c2fd05e17a35a0a2cb4726c5442e715158b9bd"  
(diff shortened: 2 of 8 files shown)


cmake/test.cmake
```diff
@@ -7,18 +7,22 @@ function(add_unit_test)
...
+            ${ADD_UNIT_TEST_TARGET} PUBLIC DISABLE_LOGGING
+        )
     endif()
 
-    target_link_libraries(${ADD_UNIT_TEST_TARGET}
-        GTest::GTest
-        monad_unit_test_common
+    target_link_libraries(
+        ${ADD_UNIT_TEST_TARGET} GTest::GTest monad_unit_test_common
         ${ADD_UNIT_TEST_LIBRARIES}
```
sheapwilliams: You didn't change anything here, correct?  Is this just formatting changes?

src/monad/db/test/CMakeLists.txt
```diff
@@ -13,3 +14,13 @@ add_unit_test(
...
     monad
 )
+
+# TODO: make this generate the data from replay_ethereum when that becomes
+# ready
+add_custom_command(
+    TARGET monad-dbtests
+    POST_BUILD
+    COMMAND $<TARGET_FILE:monad-dbtests> --gtest_filter=InMemoryTrieDB.erase >
+            ${PROJECT_SOURCE_DIR}/monad/tdb/test_tdb.data
```
sheapwilliams: I think this shouldn't be in the PROJECT_SOURCE_DIR, but in the build directory somewhere.  I expect removing the build directory removed everything.

test/features/tdb/steps/tdb.py
```diff
@@ -0,0 +1,37 @@
...
+import os
+
+
+@given("I run tdb with the output")
+def given_run_tdb(context):
+    with tempfile.NamedTemporaryFile() as tp:
+        tp.write(context.output)
+        tp.flush()
+        result = subprocess.run(
+            ["python3", os.getcwd() + "/monad/tdb/tdb.py", "--log", tp.name],
```
sheapwilliams, line 14: We should add the same issue here: #113.

test/features/tdb/steps/tdb.py
```diff
@@ -0,0 +1,37 @@
...
+
+@given("I run tdb with the output and account {account}")
+def given_run_tdb_with_account(context, account):
+    with tempfile.NamedTemporaryFile() as tp:
+        tp.write(context.output)
+        tp.flush()
+        result = subprocess.run(
+            [
+                "python3",
+                os.getcwd() + "/monad/tdb/tdb.py",
```
sheapwilliams, line 30: ...here too.

test/features/tdb/root_hash.feature
```diff
@@ -0,0 +1,12 @@
+Feature: TDB Root Hash
+
+    # TODO: make this generate the data from replay_ethereum when that becomes ready
+    Scenario: Verify state root hash from TrieDBTest/1.ModifyStorageOfAccount
+        Given I run db unit test TrieDBTest/1.ModifyStorageOfAccount
+        And I run tdb with the output
+        Then the output should contain "Root hash 0x0169f0b22c30d7d6f0bb7ea2a07be178e216b72f372a6a7bafe55602e5650e60"
```
sheapwilliams, line 7: This is a good start. I'd like to see the `Then` parts of this grow to detect specific differences in runs, but this is a great start.

PR #128, opened 2023-06-22
Integration at TP level; Add recover_sender functionality

tongzhi1998:
- Replace fake TP with real TP
- Add support for recover_sender
- Fix bugs in rlp_encode and rlp_decode

### DIFF
--- a/include/monad/core/transaction.hpp
+++ b/include/monad/core/transaction.hpp
@@ -1,4 +1,5 @@
 #pragma once
+
 #include <monad/config.hpp>
 #include <monad/core/address.hpp>
 #include <monad/core/assert.h>
@@ -53,16 +54,18 @@ static_assert(sizeof(Transaction) == 248);
 static_assert(alignof(Transaction) == 8);
 
 // EIP-1559
-inline uint64_t per_gas_priority_fee(
-    Transaction const &t, uint64_t const base_fee_per_gas)
+constexpr uint64_t
+per_gas_priority_fee(Transaction const &t, uint64_t const base_fee_per_gas)
 {
     return std::min(t.priority_fee, t.gas_price - base_fee_per_gas);
 }
 
-inline uint64_t
+constexpr uint64_t
 per_gas_cost(Transaction const &t, uint64_t const base_fee_per_gas)
 {
     return per_gas_priority_fee(t, base_fee_per_gas) + base_fee_per_gas;
 }
+
+[[nodiscard]] std::optional<address_t> recover_sender(Transaction const &t);
 
 MONAD_NAMESPACE_END
--- a/include/monad/execution/transaction_processor.hpp
+++ b/include/monad/execution/transaction_processor.hpp
@@ -9,6 +9,8 @@
 
 #include <monad/execution/config.hpp>
 
+#include <monad/rlp/encode_helpers.hpp>
+
 #include <intx/intx.hpp>
 
 #include <algorithm>
--- a/include/monad/execution/transaction_processor_data.hpp
+++ b/include/monad/execution/transaction_processor_data.hpp
@@ -29,13 +29,13 @@ struct alignas(64) TransactionProcessorFiberData
     using txn_processor_status_t = typename TTxnProcessor::Status;
 
     TState &s_;
-    Transaction const &txn_;
+    Transaction &txn_;
     BlockHeader const &bh_;
     unsigned int id_;
     Receipt result_{};
 
     TransactionProcessorFiberData(
-        TState &s, Transaction const &t, BlockHeader const &b, unsigned int id)
+        TState &s, Transaction &t, BlockHeader const &b, unsigned int id)
         : s_{s}
         , txn_{t}
         , bh_{b}
@@ -66,7 +66,14 @@ struct alignas(64) TransactionProcessorFiberData
         auto *txn_logger = log::logger_t::get_logger("txn_logger");
         auto const start_time = std::chrono::steady_clock::now();
 
-        MONAD_LOG_INFO(txn_logger, "Start executing Transaction {}", id_);
+        txn_.from = recover_sender(txn_);
+
+        MONAD_LOG_INFO(
+            txn_logger,
+            "Start executing Transaction {}, from = {}, to = {}",
+            id_,
+            txn_.from,
+            txn_.to);
 
         while (true) { // retry until apply state cleanly
             while (true) { // spin until *could be* successful
--- a/include/monad/rlp/encode_helpers.hpp
+++ b/include/monad/rlp/encode_helpers.hpp
@@ -54,6 +54,7 @@ inline byte_string encode_address(std::optional<address_t> const &a)
 
 byte_string encode_account(Account const &, bytes32_t const &storage_root);
 byte_string encode_transaction(Transaction const &);
+byte_string encode_transaction_for_signing(Transaction const &);
 byte_string encode_access_list(Transaction::AccessList const &list);
 byte_string encode_topics(std::vector<bytes32_t> const &topics);
 byte_string encode_log(Receipt::Log const &log);
--- a/src/monad/core/transaction.cpp
+++ b/src/monad/core/transaction.cpp
@@ -1,0 +1,44 @@
+#include <monad/core/transaction.hpp>
+
+#include <monad/rlp/encode_helpers.hpp>
+
+#include <ethash/keccak.hpp>
+
+#include <silkpre/ecdsa.h>
+
+#include <memory>
+
+MONAD_NAMESPACE_BEGIN
+
+std::optional<address_t> recover_sender(Transaction const &t)
+{
+    if (t.from.has_value()) {
+        return t.from;
+    }
+
+    byte_string const txn_encoding = rlp::encode_transaction_for_signing(t);
+    ethash::hash256 const txn_encoding_hash{
+        ethash::keccak256(txn_encoding.data(), txn_encoding.size())};
+
+    uint8_t signature[sizeof(t.sc.r) * 2];
+    intx::be::unsafe::store(signature, t.sc.r);
+    intx::be::unsafe::store(signature + sizeof(t.sc.r), t.sc.s);
+
+    std::optional<address_t> res = evmc::address{};
+    std::unique_ptr<secp256k1_context, decltype(&secp256k1_context_destroy)>
+        context(
+            secp256k1_context_create(SILKPRE_SECP256K1_CONTEXT_FLAGS),
+            &secp256k1_context_destroy);
+    if (!silkpre_recover_address(
+            res->bytes,
+            txn_encoding_hash.bytes,
+            signature,
+            t.sc.odd_y_parity,
+            context.get())) {
+        return std::nullopt;
+    }
+
+    return res;
+}
+
+MONAD_NAMESPACE_END
--- a/src/monad/rlp/encode_helpers.cpp
+++ b/src/monad/rlp/encode_helpers.cpp
@@ -94,6 +94,64 @@ byte_string encode_transaction(Transaction const &txn)
     return {};
 }
 
+byte_string encode_transaction_for_signing(Transaction const &txn)
+{
+    if (txn.type == Transaction::Type::eip155) {
+        if (txn.sc.chain_id.has_value()) {
+            return encode_list(
+                encode_unsigned(txn.nonce),
+                encode_unsigned(txn.gas_price),
+                encode_unsigned(txn.gas_limit),
+                encode_address(txn.to),
+                encode_unsigned(txn.amount),
+                encode_string(txn.data),
+                encode_unsigned(txn.sc.chain_id.value_or(0)),
+                encode_unsigned(0u),
+                encode_unsigned(0u));
+        }
+        else {
+            return encode_list(
+                encode_unsigned(txn.nonce),
+                encode_unsigned(txn.gas_price),
+                encode_unsigned(txn.gas_limit),
+                encode_address(txn.to),
+                encode_unsigned(txn.amount),
+                encode_string(txn.data));
+        }
+    }
+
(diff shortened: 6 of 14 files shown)


include/monad/execution/transaction_processor.hpp
```diff
@@ -115,6 +121,33 @@ struct TransactionProcessor
...
+            secp256k1_context_create(SILKPRE_SECP256K1_CONTEXT_FLAGS)};
+        if (!silkpre_recover_address(
+                t.from->bytes,
+                txn_encoding_hash.bytes,
+                signature,
+                t.sc.odd_y_parity,
+                context)) {
+            t.from = std::nullopt;
+        }
+    }
```
sheapwilliams: I don't think recovering the signature is intrinsic to the tp, so it shouldn't be a member of transaction processor.  Recovering the signature can be done anywhere, yea?  I think it should be a free function, and just used wherever we decide.

include/monad/rlp/decode_helpers.hpp
```diff
@@ -24,6 +24,16 @@ decode_unsigned(T &u_num, byte_string_view const enc)
...
+inline constexpr byte_string_view
+decode_bool(bool &target, byte_string_view const enc)
+{
+    uint64_t i{0};
+    auto ret = decode_unsigned<uint64_t>(i, enc);
+    MONAD_ASSERT(i <= 1);
+    target = i;
+    return ret;
+}
+
```
sheapwilliams: Can you put the detail/fix/test of this bug in a separate commit so it is obvious what the issue was?

include/monad/rlp/decode_helpers.hpp
```diff
@@ -36,14 +46,16 @@ decode_address(address_t &address, byte_string_view const enc)
     return decode_byte_array<20>(address.bytes, enc);
 }
 
-inline byte_string_view
-decode_address_optional(std::optional<address_t> &address, byte_string_view const enc)
+inline byte_string_view decode_address_optional(
+    std::optional<address_t> &address, byte_string_view const enc)
```
sheapwilliams: I know this was here before, but I don't think we should put the parameter listing in the name.  Can we use a simple overload of `decode_address(std::optional<address_t>, byte_string_view const enc)` for this case?

include/monad/rlp/decode_helpers.hpp
```diff
@@ -36,14 +46,16 @@ decode_address(address_t &address, byte_string_view const enc)
...
 {
     byte_string_view payload{};
     auto const rest_of_enc = parse_string_metadata(payload, enc);
     if (payload.size() == sizeof(address_t)) {
+        address = address_t{};
         std::memcpy(address->bytes, payload.data(), sizeof(address_t));
-    } else {
+    }
+    else {
```
sheapwilliams, line 56: clang-format?

include/monad/rlp/encode_helpers.hpp
```diff
@@ -48,15 +48,15 @@ inline byte_string encode_address(address_t const &a)
 }
 
 byte_string encode_account(Account const &, bytes32_t const &storage_root);
-byte_string encode_transaction(Transaction const &);
+byte_string encode_transaction(Transaction const &, bool signing = false);
```
sheapwilliams: I don't think we want a bool `signing` parameter for this function...it can just be a separate function.  We are duplicating much of the code anyway in this function, so no need to cram all this into the same function I think.

include/monad/execution/transaction_processor.hpp
tongzhi1998: I ended up doing something that might seem weird: making `recover_sender` a free function but adding a function called `recover_sender_wrapper` in `TransactionProcessor`. The reason behind this is that if we just call the free function `recover_sender` in `FiberData`, then test cases would fail (because in test cases, we don't really have a well-formed transaction). By wrapping it up, we can create a fake `recover_sender_wrapper` in our test cases so that they can be the same as before. 
I think this is still somehow strange. Please let me know if you have any better solution!

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -66,7 +66,14 @@ struct alignas(64) TransactionProcessorFiberData
         auto *txn_logger = log::logger_t::get_logger("txn_logger");
         auto const start_time = std::chrono::steady_clock::now();
 
-        MONAD_LOG_INFO(txn_logger, "Start executing Transaction {}", id_);
+        p.recover_sender_wrapper(txn_);
```
sheapwilliams: Looks much better.  I think we don't need a wrapper in the tp either tho.  Something like this:

```cpp
   t.from = recover_sender(txn_);
```

Where it returns an `optional<address_t>`. We can have it in a namespace if we want, but truly free so that we can do different things with Transaction const-ness.  Maybe we push this to the validate function as opposed to the scheduled in the fiber?...I see lots of different ideas for this, so would just prefer the most flexibility.

include/monad/core/transaction.hpp
```diff
@@ -53,8 +54,8 @@ static_assert(sizeof(Transaction) == 248);
 static_assert(alignof(Transaction) == 8);
 
 // EIP-1559
-inline uint64_t per_gas_priority_fee(
-    Transaction const &t, uint64_t const base_fee_per_gas)
+inline uint64_t
+per_gas_priority_fee(Transaction const &t, uint64_t const base_fee_per_gas)
```
zander-xyz: I know you didn't add these functions, but can we make these functions `constexpr` (the one below too)? And then you can drop the inline since it becomes implicit with the constexpr

include/monad/execution/transaction_processor_data.hpp
zander-xyz: +1

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -87,12 +94,13 @@ struct alignas(64) TransactionProcessorFiberData
...
-            auto txn_state = s_.get_copy();
+            auto working_copy = s_.get_working_copy(id_);
             TEvm
                 e{}; // e needs to be constructed with the working copy of state
-            result_ = p.execute(txn_state, e, bh_, txn_);
+            result_ = p.execute(working_copy, e, bh_, txn_);
 
-            if (auto const applied = s_.apply_state(txn_state); applied) {
+            auto const can_merge_changes = s_.can_merge_changes(working_copy);
+            if (can_merge_changes == TState::MergeStatus::WILL_SUCCEED) {
```
zander-xyz: Any reason you prefer this as opposed to
```suggestion
            if (s_.can_merge_changes(working_copy) == TState::MergeStatus::WILL_SUCCEED) {
```

include/monad/rlp/decode_helpers.hpp
```diff
@@ -24,6 +24,16 @@ decode_unsigned(T &u_num, byte_string_view const enc)
     return rest_of_enc;
 }
 
+inline constexpr byte_string_view
```
zander-xyz: the inline is redundant (bc of the constexpr)

include/monad/rlp/decode_helpers.hpp
```diff
@@ -24,6 +24,16 @@ decode_unsigned(T &u_num, byte_string_view const enc)
...
+decode_bool(bool &target, byte_string_view const enc)
+{
+    uint64_t i{0};
+    auto ret = decode_unsigned<uint64_t>(i, enc);
+    MONAD_ASSERT(i <= 1);
```
zander-xyz: this feels more appropriate as a debug_assert, I think

include/monad/rlp/decode_helpers.hpp
```diff
@@ -24,6 +24,16 @@ decode_unsigned(T &u_num, byte_string_view const enc)
     return rest_of_enc;
 }
 
+inline constexpr byte_string_view
+decode_bool(bool &target, byte_string_view const enc)
```
zander-xyz: thanks for the separate commit for the RLP bugs, but did you end up adding tests for decode_bool? maybe I missed it but I couldn't find it in the RLP commit

src/monad/core/transaction.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+MONAD_NAMESPACE_BEGIN
+
+void recover_sender(Transaction &t)
+{
+    if (t.from.has_value()) {
+        return;
+    }
+
+    byte_string txn_encoding = rlp::encode_transaction_for_signing(t);
```
zander-xyz: const?

src/monad/core/transaction.cpp
```diff
@@ -0,0 +1,38 @@
...
+    ethash::hash256 txn_encoding_hash{
+        ethash::keccak256(txn_encoding.data(), txn_encoding.size())};
+
+    uint8_t signature[32 * 2];
```
zander-xyz: The size of this array seems better expressed as `sizeof(t.sc.r) + sizeof(t.sc.s)`. Typically with magic number you'd want to either add commentary explaining what they represent (see comments elsewhere linking to the YP), or try to represent it using code, which is my personal preference.

In this case, it didn't take me long to realize what you were trying to do, but with a small amount of effort, this can be way clearer.

src/monad/core/transaction.cpp
```diff
@@ -0,0 +1,38 @@
...
+    intx::be::unsafe::store(signature, t.sc.r);
+    intx::be::unsafe::store(signature + 32, t.sc.s);
```
zander-xyz: Same commentary for the 32 here

src/monad/core/transaction.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+    t.from = evmc::address{};
+    static secp256k1_context *context{
+        secp256k1_context_create(SILKPRE_SECP256K1_CONTEXT_FLAGS)};
```
zander-xyz: is this leaking memory (use valgrind to check)?

src/monad/core/transaction.cpp
```diff
@@ -0,0 +1,38 @@
...
+    if (!silkpre_recover_address(
+            t.from->bytes,
+            txn_encoding_hash.bytes,
+            signature,
+            t.sc.odd_y_parity,
+            context)) {
+        t.from = std::nullopt;
```
zander-xyz: Is this a valid state to be in? Should we DEBUG_ASSERT here?

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -45,73 +48,17 @@ struct fakeEvmHost
...
-    {
-        return {};
-    }
-
-    Status validate(TState const &, Transaction const &, uint64_t) noexcept
-    {
-        return Status::SUCCESS;
-    }
-};
-
```
zander-xyz, line 74: QQ: now that we have a real TransactionProcessor, do all of the other fake TPs still need to exist? I see some floating around in the various unit tests

tongzhi1998: I think they still do. I think the purpose of unit test is to test only one thing (not depending on the correctness of another object), for example, in TransactionProcessorFiberData tests, we still want the fake TransactionProcess because we don't want something wrong in TransactionProcessor to affect the correctness of FiberData.

include/monad/execution/transaction_processor_data.hpp
tongzhi1998: There are 3 different MergeStatus (1 successful, 2 unsuccessful). Not sure if we want to do things differently for 2 unsuccessful Status (currently I don't see if we want, but just in case we do, leaving the can_merge_change in a separate line is better)

src/monad/core/transaction.cpp
tongzhi1998: That's a good question. I think I need to read more about the spec but my intuition is that it's possible that the transaction is malformed, so that we can't recover sender's address. However, those transactions could still be included in a block, so DEBUG_ASSERT might not be appropriate? Also silkworm does something like `t.from = std::nullopt` too.

src/monad/core/transaction.cpp
tongzhi1998: There are some still_reachable blocks. It seems to have something to do with rocksdb and google test. I believe this function doesn't leak but I am not 100% confident.

include/monad/execution/transaction_processor_data.hpp
zander-xyz: It should be easy enough to modify this code in the future if the need arises.

src/monad/core/transaction.cpp
```diff
@@ -0,0 +1,44 @@
...
+    if (!silkpre_recover_address(
+            res->bytes,
+            txn_encoding_hash.bytes,
+            signature,
+            t.sc.odd_y_parity,
+            context)) {
+        secp256k1_context_destroy(context);
+        return std::nullopt;
+    }
+    secp256k1_context_destroy(context);
```
zander-xyz: minor: you can use a std::unique_ptr and override the Deleter to call destroy to avoid having to manage the cleanup yourself

src/monad/rlp/encode_helpers.cpp
```diff
@@ -96,6 +94,64 @@ byte_string encode_transaction(Transaction const &txn)
     return {};
 }
 
+byte_string encode_transaction_for_signing(Transaction const &txn)
```
zander-xyz, line 97: unit tests for this guy?

tongzhi1998: I have unit test for `recover_sender` under `src/monad/core/test/transaction.cpp`. I think that's pretty much the unit test for this function.

src/monad/core/transaction.cpp
sheapwilliams: Currently, we have a monolithic architecture.  That means there is no trust boundary between consensus' ledger and execution.  So we fully expect that consensus will have removed all unrecoverable transactions from the ledger for us.  This won't be the case in the future, but it is a fine assumption for now to believe that this won't happen.

PR #135, opened 2023-06-27
:new: Block header hash cache in BlockDB

sheapwilliams:
This is the final piece to support the EVMC Host interface.  I was trying to simplify the parallelism requirements, which meant caching the hash on block read before the massive parallelism for transactions.

On genesis startup, there should be no issues, as everything is already cached before use.  But on a startup with an existing database, we still will likely need thread-safe build of the cache so that missing caches aren't collided (although is should still generate a correct result I believe).  This will be future work.


include/monad/db/block_db.hpp
```diff
@@ -24,10 +25,40 @@ class BlockDb
         DECODE_ERROR
     };
 
-    [[nodiscard]] Status get(block_num_t const block_num, Block &block) const;
+    static constexpr auto null{
+        0x0000000000000000000000000000000000000000000000000000000000000000_bytes32};
+
+    [[nodiscard]] Status get(block_num_t const block_num, Block &block);
```
rgarc, line 32: This is more of a question than an actual suggestion. I've always thought it to be strange whenever a getter member function is not `const` given that it should just be reading data. I see that we lose `const`-ness for this getter only because we need to mutate to update the cache. I've seen people work around this by using the `mutable` keyword like so
```diff
diff --git a/include/monad/db/block_db.hpp b/include/monad/db/block_db.hpp
index bb6310a..8059f68 100644
--- a/include/monad/db/block_db.hpp
+++ b/include/monad/db/block_db.hpp
@@ -28,11 +28,11 @@ public:
     static constexpr auto null{
         0x0000000000000000000000000000000000000000000000000000000000000000_bytes32};
 
-    [[nodiscard]] Status get(block_num_t const block_num, Block &block);
-    [[nodiscard]] Status get_past_into_block_cache(block_num_t const num);
-    [[nodiscard]] bytes32_t get_block_hash(block_num_t n);
-    void
-    store_current_block_hash(byte_string_view const v, block_num_t n) noexcept;
+    [[nodiscard]] Status get(block_num_t const block_num, Block &block) const;
+    [[nodiscard]] Status get_past_into_block_cache(block_num_t const num) const;
+    [[nodiscard]] bytes32_t get_block_hash(block_num_t n) const;
+    void store_current_block_hash(
+        byte_string_view const v, block_num_t n) const noexcept;
     [[nodiscard]] bool should_be_in_cache(block_num_t n) const noexcept;
 
     [[nodiscard]] bool is_next_block(block_num_t n) const noexcept
@@ -57,8 +57,8 @@ public:
 
 private:
     FileDb db_;
-    bytes32_t recent_hashes_[number_of_hashes]{};
-    std::optional<block_num_t> current_block_{};
+    mutable bytes32_t recent_hashes_[number_of_hashes]{};
+    mutable std::optional<block_num_t> current_block_{};
 };
 
 MONAD_DB_NAMESPACE_END
diff --git a/src/monad/db/block_db.cpp b/src/monad/db/block_db.cpp
index 44c0ba9..d383d32 100644
--- a/src/monad/db/block_db.cpp
+++ b/src/monad/db/block_db.cpp
@@ -10,7 +10,7 @@
 
 MONAD_DB_NAMESPACE_BEGIN
 
-BlockDb::Status BlockDb::get(block_num_t const num, Block &block)
+BlockDb::Status BlockDb::get(block_num_t const num, Block &block) const
 {
     auto const key = std::to_string(num);
     auto const result = db_.get(key.c_str());
@@ -44,7 +44,7 @@ BlockDb::Status BlockDb::get(block_num_t const num, Block &block)
     return Status::SUCCESS;
 }
 
-BlockDb::Status BlockDb::get_past_into_block_cache(block_num_t const num)
+BlockDb::Status BlockDb::get_past_into_block_cache(block_num_t const num) const
 {
     // Read the child and use the parent hash from that block
     MONAD_DEBUG_ASSERT(num < current_block_);
@@ -94,7 +94,7 @@ bool BlockDb::should_be_in_cache(block_num_t n) const noexcept
 }
 
 void BlockDb::store_current_block_hash(
-    byte_string_view const v, block_num_t n) noexcept
+    byte_string_view const v, block_num_t n) const noexcept
 {
     auto const h = rlp::get_rlp_header_from_block(v);
 
@@ -106,7 +106,7 @@ void BlockDb::store_current_block_hash(
             : n;
 }
 
-[[nodiscard]] bytes32_t BlockDb::get_block_hash(block_num_t number)
+[[nodiscard]] bytes32_t BlockDb::get_block_hash(block_num_t number) const
 {
     if (should_be_in_cache(number)) {
         if (recent_hashes_[write_index(number)] == null) {
```
What are the down sides to this?

include/monad/db/block_db.hpp
```diff
@@ -24,10 +25,40 @@ class BlockDb
...
+                   ? 0u
+                   : (current_block_.value() - number_of_hashes);
+    }
+
+    [[nodiscard]] size_t write_index(block_num_t n) const noexcept
+    {
+        return n % number_of_hashes;
+    }
+
+    static constexpr auto number_of_hashes{256u};
```
rgarc, line 60: Is this 256 number coming from somewhere?

include/monad/db/block_db.hpp
```diff
@@ -24,10 +25,40 @@ class BlockDb
...
 
 private:
     FileDb db_;
+    bytes32_t recent_hashes_[number_of_hashes]{};
```
rgarc: Thoughts on using `std::array` instead?

src/monad/db/block_db.cpp
```diff
@@ -35,7 +37,85 @@ BlockDb::Status BlockDb::get(block_num_t const num, Block &block) const
...
+[[nodiscard]] bytes32_t BlockDb::get_block_hash(block_num_t number)
+{
+    if (should_be_in_cache(number)) {
+        if (recent_hashes_[write_index(number)] == null) {
+            auto const status = get_past_into_block_cache(number);
+            MONAD_DEBUG_ASSERT(status == Status::SUCCESS);
+        }
+        return recent_hashes_[write_index(number)];
+    }
+    return bytes32_t{};
```
rgarc, line 119: Why is this function returning `bytes32_t{} = null` if the block hash is not cached?

src/monad/rlp/decode_helpers.cpp
```diff
@@ -391,6 +391,15 @@ byte_string_view decode_block_header_vector(
     return rest_of_enc;
 }
 
+byte_string_view get_rlp_header_from_block(byte_string_view const block_encoding)
+{
+    byte_string_view rlp_block{};
+    (void)parse_list_metadata(rlp_block, block_encoding);
+    byte_string_view rlp_block_header{};
+    (void)parse_list_metadata(rlp_block_header, rlp_block);
+    return {rlp_block.data(), rlp_block_header.end()};
```
rgarc, line 402: Is this a typo? Mixing `.data()` and `.end()` calls from different objects is only safe if they are contiguous in memory.

include/monad/db/block_db.hpp
sheapwilliams: `mutable` & `const` is really just a regular "empty" decoration, so declaring with no decorations is to simply avoid the "double negative" problem we have in speech - it makes things less clear to have to cancel out the effects in your brain, and you have less to follow to get any one idea accomplished.  Recall the [7±2](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two) things you can focus on at the same time?  Using cancelling specifiers consumes 1-2 of those slots, and pushes out 1-2 others you were trying to focus on to get your task done, so you increase the burden to understanding of your object.  That is the general reason to avoid this.

Specifically, I've only used `mutable` `const` when it is unavoidable, like when using library code that is fixed or some other dependency.  I find it really messes with code expressions, and can lead to UB in several instances.

On the subject of architecture/patterns: perhaps a better pattern for a caching block db is a decorator pattern.  Decorate the block db with a pass-through cache, and then inject that if you want.  I'd prefer to have an on-demand quality of the block hash calculation, but I was wary of the synchronization issues we'll develop being on-demand.  So to avoid that I basically shoe-horned the functionality into the place where the raw data was most readily available - when getting the block.  I'm happy to revisit after we get a better sense of its usage.

include/monad/db/block_db.hpp
sheapwilliams: Yea, the YP - I'll try to find a good reference for that.

src/monad/db/block_db.cpp
sheapwilliams: That is the specification - only the last 256 are available, otherwise nulls.

src/monad/rlp/decode_helpers.cpp
sheapwilliams: Not a typo, this is correct. These are two different views over the same data/byte array.  1 is unwrapping the outer block list to get the start of the rlp-encoded block header.  The second is a view over the block-header-as-a-list, with the rlp header stripped.  So to get the RLP encoded block header we take the start from the decoded block, and the end from the block header decode.

It is sorta gross - I know.  But I was lacking imagination when I did this.  Lemme know if you want to pair on it and shape it into something more intuitive.

src/monad/db/block_db.cpp
sheapwilliams: I added a comment pointing at the condition that defines this.

PR #137, opened 2023-06-28
:heavy_plus_sign: Use vcpkg for deps being found via conan

zander-xyz:
Problem:
- We rely on multiple different dependency management mechanisms

Solution:
- Conslidate to vcpkg

Will work on migrating the rest bit-by-bit.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,8 +15,10 @@ RUN yum -y update
 RUN yum -y install \
   clang \
   cmake \
+  git \
   ninja-build \
-  python3-pip
+  python3-pip \
+  zip
 
 RUN yum -y install \
   abseil-cpp \


CMakeLists.txt
```diff
@@ -9,36 +9,29 @@ option(ASAN "Turn on Address Sanitizer")
...
-# conan
+# deps
 # ##############################################################################
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
-list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
+include(FetchContent)
 
-include(${PROJECT_SOURCE_DIR}/conan/cmake-conan/conan.cmake)
+FetchContent_Declare(
```
ned14: I would recommend making vcpkg a git subrepo rather than fetching it in cmake. Then we can easily fork it or customise it.

CMakeLists.txt
```diff
@@ -9,36 +9,29 @@ option(ASAN "Turn on Address Sanitizer")
...
-    ${settings}
-)
+if(NOT vcpkg_POPULATED)
+    message(FATAL_ERROR "Failed to fetch vcpkg")
+endif()
 
-# ##############################################################################
-# deps
-# ##############################################################################
+include(${vcpkg_SOURCE_DIR}/scripts/buildsystems/vcpkg.cmake)
```
ned14: Before this line you need to do a bit of helping to cope with cmake toolchain files e.g.

```
if(NOT CMAKE_TOOLCHAIN_FILE STREQUAL "${VCPKG_ROOT_DIR}/scripts/buildsystems/vcpkg.cmake")
  # This must be forced so dependency builds pick it up
  if(CMAKE_TOOLCHAIN_FILE)
    set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE "${CMAKE_TOOLCHAIN_FILE}" CACHE INTERNAL "vcpkg chained toolchain file")
  endif()
  set(CMAKE_TOOLCHAIN_FILE "${VCPKG_ROOT_DIR}/scripts/buildsystems/vcpkg.cmake" CACHE INTERNAL "vcpkg toolchain file" FORCE)
endif()
```

PR #145, opened 2023-06-29
✅ Add initial integration test of EVM, Host, and State

sheapwilliams:
More to follow, but I want to publish this to get some eyes on it.

Future work will look into getting GUnit to work with ctest. If I can, I'll refactor this before working on the TODO tests.

### DIFF
--- a/cmake/test.cmake
+++ b/cmake/test.cmake
@@ -26,8 +26,21 @@ function(add_unit_test)
     endif()
 
     target_link_libraries(
-        ${ADD_UNIT_TEST_TARGET} GTest::GTest monad_unit_test_common
-        ${ADD_UNIT_TEST_LIBRARIES}
+        ${ADD_UNIT_TEST_TARGET} PUBLIC GTest::GTest monad_unit_test_common
+                                       ${ADD_UNIT_TEST_LIBRARIES}
     )
     gtest_discover_tests(${ADD_UNIT_TEST_TARGET})
+endfunction()
+
+function(add_integration_test)
+    set(ONE_VALUE_ARGS TARGET)
+    cmake_parse_arguments(ADD_INT_TEST "" "${ONE_VALUE_ARGS}" "" ${ARGN})
+
+    add_unit_test(${ADD_INT_TEST_TARGET} ${ARGN})
+    target_link_libraries(${ADD_INT_TEST_TARGET} PUBLIC monad)
+
+    if(NOT TARGET integration_tests)
+        add_custom_target(integration_tests)
+    endif()
+    add_dependencies(integration_tests ${ADD_INT_TEST_TARGET})
 endfunction()
--- a/test/integration/evm_state_host/evm_state_host.cpp
+++ b/test/integration/evm_state_host/evm_state_host.cpp
@@ -1,0 +1,114 @@
+#include <monad/core/concepts.hpp>
+
+#include <monad/db/account_store.hpp>
+#include <monad/db/block_db.hpp>
+#include <monad/db/code_store.hpp>
+#include <monad/db/state.hpp>
+#include <monad/db/trie_db.hpp>
+#include <monad/db/value_store.hpp>
+
+#include <monad/execution/config.hpp>
+#include <monad/execution/ethereum/fork_traits.hpp>
+#include <monad/execution/evm.hpp>
+#include <monad/execution/evmc_host.hpp>
+#include <monad/execution/evmone_baseline_interpreter.hpp>
+
+#include <monad/execution/test/fakes.hpp>
+
+#include <evmc/evmc.hpp>
+
+#include <gtest/gtest.h>
+
+#include <test_resource_data.h>
+
+#include <unordered_map>
+
+using namespace monad;
+
+static constexpr auto from = 0x5353535353535353535353535353535353535353_address;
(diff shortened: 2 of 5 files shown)


cmake/test.cmake
```diff
@@ -7,19 +7,35 @@ function(add_unit_test)
...
+        ${ADD_UNIT_TEST_TARGET} PUBLIC GTest::GTest monad_unit_test_common
+                                       ${ADD_UNIT_TEST_LIBRARIES}
     )
     gtest_discover_tests(${ADD_UNIT_TEST_TARGET})
 endfunction()
+
+function(add_integration_test)
+    set(ONE_VALUE_ARGS TARGET)
+    cmake_parse_arguments(ADD_INT_TEST "" "${ONE_VALUE_ARGS}" "" ${ARGN})
+    message(STATUS "int target is: ${ADD_INT_TEST_TARGET}")
```
zander-xyz: Seems unnecessary, but if you want to keep this log around, can we spell out integration?

cmake/test.cmake
```diff
@@ -7,19 +7,35 @@ function(add_unit_test)
...
+
+    add_unit_test(${ADD_INT_TEST_TARGET} ${ARGN})
+    target_link_libraries(${ADD_INT_TEST_TARGET} PUBLIC monad)
+
+    add_custom_target(integration-tests)
```
zander-xyz: this fails if we have more than one integration test, I think you ought to check if integration-tests exists first before adding it

include/monad/db/state.hpp
```diff
@@ -149,6 +149,13 @@ struct State
             code_.revert();
         }
 
+        [[nodiscard]] bytes32_t get_block_hash(int64_t number) const noexcept
```
zander-xyz, line 152: yeah it really sucks that the evm host API takes block number as signed...i feel your pain

what do you think about pushing the signed-ness up to the evmc_host layer and doing the translation to unsigned there?

cmake/test.cmake
sheapwilliams: Nope, I'll remove it.

include/monad/db/state.hpp
sheapwilliams: Not the worst impulse, but I'd really like to leave the evmc host as simple composition.  Implementors can deal with the details.

zander-xyz: yeah i guess there are two schools of thought
1. use the the evmc_host as a layer to hide-away the warts of the interface itself
2. what you said

either is fine, your call

test/integration/evm_state_host/evm_state_host.cpp
```diff
@@ -0,0 +1,114 @@
...
+    0x000000000000000000000000000000000000000000000000000000000000004c_bytes32;
+
+using code_db_t = std::unordered_map<address_t, byte_string>;
+using account_store_db_t = db::InMemoryDB;
+
+template <class TState, concepts::fork_traits<TState> TTraits>
+using traits_templated_static_precompiles_t = execution::StaticPrecompiles<
+    TState, TTraits, typename TTraits::static_precompiles_t>;
+
+TEST(EvmInterpStateHost, return_existing_storage)
```
zander-xyz, line 43: thoughts on making this a typed test and doing it on all types of DBs? similar to the *store unit tests

include/monad/db/state.hpp
sheapwilliams: Yea - I think this is the right idea for now.  If we find ourselves with a different composition and copying the signed->unsigned code, then we can fix it then.  If that different composition doesn't have that (cause reasons) then we got it right.  We can figure it then.

test/integration/evm_state_host/evm_state_host.cpp
sheapwilliams: I'm not totally allergic to it, but permutating all the genetic possibilities of our integration and test plan can get out of hand, right?  I think the backend DB is a detail that the DBInterface hides away for this test - we are really just using the in-memory for testing, and integrating the evm, host, and state.  The back-end integration between DBInterface is taken care of by other tests.  They will come together at the next higher level, probably with the block playback.

So, I'm comfortable with only these types tested at this level of testing, esp. cause the trie building is the slowest part of our testing.

This is a subjective thing for sure to get the test coverage right while keeping pipelines and job durations reasonable.  Hit me up on zoom if you want to talk more about it.

zander-xyz: One alternative viewpoint could be formed by examining what this unit test is doing. This unit test is storing a key value pair into the storage of an account, then reading from the storage of that account, storing that into "memory", then returning it.

That involves the implementation of the DB interface to be correct. Contrast that to a fake, where you really don't care what the implementation is (in which case, I wouldn't care what backend you used). Here, you *really do care*. You care that the implementation is right. You want and demand it to be right through your assertions. And in our case, gtest makes it super easy for us to exercise both the "easy" case (InMemoryDb), and harder cases (*TrieDbs). 

Does this make sense?

sheapwilliams: If I worked on the DB, I might frame things around that POV.  But I didn't, and I don't.  Let's zoom if you want to chat test coverage.  It is subjective and not always easy to get right, so it is worth a discussion.

PR #154, opened 2023-07-06
ethereum tests

rgarc:
Utility to execute tests from https://github.com/ethereum/tests

TODOs
- [ ] make cloning the ethereum tests repo optional for CI as it is quite large

### DIFF
--- a/include/monad/state/account_state.hpp
+++ b/include/monad/state/account_state.hpp
@@ -168,7 +168,7 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
     {
         auto const [_, inserted] =
             changed_.emplace(a, diff_t{get_committed_storage(a), Account{}});
-        MONAD_DEBUG_ASSERT(inserted);
+        //        MONAD_DEBUG_ASSERT(inserted);
     }
 
     // EVMC Host Interface
@@ -193,7 +193,10 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
 
     void set_balance(address_t const &address, uint256_t new_balance) noexcept
     {
-        changed_.at(address).updated.value().balance = new_balance;
+        auto &x = changed_.at(address).updated;
+        if (x) {
+            x.value().balance = new_balance;
+        }
     }
 
     [[nodiscard]] uint64_t get_nonce(address_t const &address) const noexcept
--- a/test/statetest/CMakeLists.txt
+++ b/test/statetest/CMakeLists.txt
@@ -1,0 +1,33 @@
+# Modifications to the Original Code (or portions thereof):
+# Monad: 2023
+# - Fit naming and dependency management conventions of Monad
+# Original Code is licensed under the Apache 2.0 License
+# monad: Fast Ethereum Virtual Machine implementation
+# Copyright 2022 The monad Authors.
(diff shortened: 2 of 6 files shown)


test/compatibility/compatibility.cpp
```diff
@@ -0,0 +1,267 @@
...
+    if (s == "berlin") {
+        return 5;
+    }
+    if (s == "london") {
+        return 6;
+    }
+    return std::nullopt;
+}
+
+using state_t = monad::execution::fake::State;
```
rgarc: Need to plug in an actual in memory database. I previously thought a fake state would have sufficed, but that is not the case.

test/compatibility/from_json.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+            // TODO: parse the storage field
+            if (json.contains("storage") && json["storage"].is_object() &&
+                !json["storage"].empty()) {
+                throw std::runtime_error(
+                    "you forgot to parse the storage field on an object that "
+                    "has non-zero storage");
+            }
+        }
+
```
rgarc, line 94: I'm plugging in an account for the beneficiary (miner) when running transactions, but won't that affect our ability to actually match the test cases because this beneficiary account will mess with the post state hash.

sheapwilliams: Yea, it sounds like it will...I think I need to revisit this logic...stay tuned.

test/compatibility/compatibility.cpp
sheapwilliams: Yea, I think you'll need all real things...I'd be surprised and curious if you are using any fakes for anything.

test/compatibility/from_json.hpp
sheapwilliams: As discussed, I don't have this right.  I'll work to fix this right away.

rgarc: I think I found out what I was doing wrong. I took a look at the geth analog of this tool, and it looks like it might be using the `currentCoinbase` field as the beneficiary address from the test.json files
https://github.com/ethereum/go-ethereum/blob/c2db667c8f824ec0d025cbde95e6aaad67fc10ba/cmd/evm/internal/t8ntool/execution.go#L260C14-L260C14
I will look further and actually parse this field to see what's going on, but I'm pretty sure this is what is going on and will allow me to remove this hard-coded beneficiary.

CMakeLists.txt
```diff
@@ -16,21 +16,21 @@ include(cmake/test.cmake)
 
 set(VCPKG_ROOT_DIR "${PROJECT_SOURCE_DIR}/third_party/vcpkg")
 
-if(NOT CMAKE_TOOLCHAIN_FILE STREQUAL
-   "${VCPKG_ROOT_DIR}/scripts/buildsystems/vcpkg.cmake"
-)
+if (NOT CMAKE_TOOLCHAIN_FILE STREQUAL
```
rgarc, line 21: TODO: undo all of the whitespace changes

CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ endfunction()
 # unit tests
 # ##############################################################################
 
+add_subdirectory(${PROJECT_SOURCE_DIR}/test/compatibility)
```
rgarc, line 90: I think this is the only necessary diff

PR #161, opened 2023-07-10
👷 Make CI also compile and run with fuzztests

zander-xyz:
on top of #160

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ project(monad)
 
 option(ASAN "Turn on Address Sanitizer")
 option(UBSAN "Turn on Undefined Behavior Sanitizer")
+option(FUZZING "Enable fuzzing for clang" ON)
 
 include(cmake/test.cmake)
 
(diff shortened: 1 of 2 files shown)


cmake/toolchains/with-fuzztest.cmake
```diff
@@ -0,0 +1 @@
+set(FUZZTEST_FUZZING_MODE ON CACHE INTERNAL "turn on fuzztesting" FORCE)
```
sheapwilliams: These aren't real toolchain files.  We shouldn't be using toolchain files to accomplish this functionality, I think.

sheapwilliams: Discussed over zoom - I think presets are the right idea for different cmake configurations.  Let's add a preset file if that helps manage those.

CMakeLists.txt
```diff
@@ -105,15 +106,18 @@ add_subdirectory(${PROJECT_SOURCE_DIR}/test/integration)
...
-
+    if(FUZZING)
+        set(ANTLR_BUILD_CPP_TESTS
+            OFF
+            CACHE INTERNAL "do not build ANTLR tests" FORCE
+        )
+        set(FUZZTEST_FUZZING_MODE
+            ON
+            CACHE INTERNAL "enable fuzzing" FORCE
+        )
```
sheapwilliams, line 117: I don't think we need this variable anymore...do we?

zander-xyz: That's just convenience for the user so that fuzzing is enabled by default. Otherwise the user has to do `-DFUZZTEST_FUZZING_MODE=On` themselves if they want to actually fuzz (done with `--fuzz=<fuzz test name>`).

zander-xyz: FUZZING is a monad specific thing that doesn't actually do anything in google/fuzztest

sheapwilliams: Gotcha - thx.

PR #163, opened 2023-07-10
:recycle: Refactor EvmHost instantiation

tongzhi1998:
Problem:
- In TPFiberData, we instantiate a TEvmHost, but it takes in a reference to static Evm, which shouldn't be the case
- The current Evm and EvmHost fakes are not templated, causing a lot of inconsistentcy with real obj; Furthermore, all methods for Evm fake is not static

Solution:
- In the constructor for EvmHost, we delete the reference to Evm
- Rewrite the fakes to make them templated; Mark all functions in fake Evm as static

### DIFF
--- a/include/monad/execution/test/fakes.hpp
+++ b/include/monad/execution/test/fakes.hpp
@@ -144,10 +144,15 @@ namespace fake
         State get_working_copy(unsigned int) { return State(*this); }
     };
 
+    template <class TState, concepts::fork_traits<TState> TTraits, class TEvm>
     struct EvmHost
     {
         evmc_result _result{};
         Receipt _receipt{};
+
+        EvmHost() = default;
+
+        EvmHost(BlockHeader const &, Transaction const &, TState &) noexcept {}
 
         [[nodiscard]] static constexpr inline evmc_message
         make_msg_from_txn(Transaction const &)
@@ -173,37 +178,38 @@ namespace fake
         }
     };
 
+    template <
+        class TState, concepts::fork_traits<TState> TTraits,
+        class TStaticPrecompiles, class TInterpreter>
     struct Evm
     {
         using new_address_t = tl::expected<address_t, evmc_result>;
         using unexpected_t = tl::unexpected<evmc_result>;
 
-        new_address_t _result{};
-        evmc_result _e_result{};
-
-        [[nodiscard]] tl::expected<address_t, evmc_result>
+        [[nodiscard]] static tl::expected<address_t, evmc_result>
         make_account_address(evmc_message const &) noexcept
         {
-            return _result;
+            return new_address_t{};
         }
 
-        [[nodiscard]] evmc_result transfer_call_balances(evmc_message const &)
+        [[nodiscard]] static evmc_result
+        transfer_call_balances(evmc_message const &)
         {
-            return _e_result;
+            return evmc_result{};
         }
 
-        template <class TState, class TEvmHost>
-        [[nodiscard]] evmc::Result
+        template <class TEvmHost>
+        [[nodiscard]] static evmc::Result
         call_evm(TEvmHost *, TState &, evmc_message const &) noexcept
         {
-            return evmc::Result{_e_result};
+            return evmc::Result{};
         }
 
-        template <class TState, class TEvmHost>
-        [[nodiscard]] evmc::Result create_contract_account(
+        template <class TEvmHost>
+        [[nodiscard]] static evmc::Result create_contract_account(
             TEvmHost *, TState &, evmc_message const &) noexcept
         {
-            return evmc::Result{_e_result};
+            return evmc::Result{};
         }
     };
 
--- a/src/monad/execution/ethereum/replay_ethereum.cpp
+++ b/src/monad/execution/ethereum/replay_ethereum.cpp
@@ -7,6 +7,7 @@
 #include <monad/execution/evmc_host.hpp>
 #include <monad/execution/replay_block_db.hpp>
 #include <monad/execution/static_precompiles.hpp>
+#include <monad/execution/test/fakes.hpp>
 #include <monad/execution/transaction_processor.hpp>
 #include <monad/execution/transaction_processor_data.hpp>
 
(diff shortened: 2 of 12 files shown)


src/monad/execution/test/transaction_processor.cpp
```diff
@@ -72,7 +80,15 @@ TEST(
...
-    fake::EvmHost e{};
+    fake::EvmHost<
+        fake::State,
+        traits_t,
+        fake::Evm<
+            fake::State,
+            traits_t,
+            fake::static_precompiles::OneHundredGas,
+            fake::Interpreter>>
+        e{};
```
sheapwilliams: You can define this type (and perhaps more of the composed ones too) once at the beginning and just use the alias here for an easier read.

include/monad/execution/test/fakes.hpp
```diff
@@ -173,37 +185,38 @@ namespace fake
...
         }
 
-        template <class TState, class TEvmHost>
-        [[nodiscard]] evmc::Result create_contract_account(
+        template <class TEvmHost>
+        [[nodiscard]] static evmc::Result create_contract_account(
             TEvmHost *, TState &, evmc_message const &) noexcept
         {
-            return evmc::Result{_e_result};
+            return evmc::Result{};
```
sheapwilliams, line 212: This is removing fake functionality...I'm surprised this isn't used?  Makes me wonder if we are missing test coverage.

include/monad/execution/test/fakes.hpp
```diff
@@ -144,11 +151,16 @@ namespace fake
         State get_working_copy(unsigned int) { return State(*this); }
     };
 
+    template <class TState, concepts::fork_traits<TState> TTraits, class TEvm>
     struct EvmHost
     {
         evmc_result _result{};
         Receipt _receipt{};
 
+        EvmHost() {}
```
sheapwilliams: Do we need the default case at all?

include/monad/execution/test/fakes.hpp
```diff
@@ -80,12 +80,19 @@ namespace fake
...
+
             return _map.at(address).code_hash;
         }
 
         [[nodiscard]] byte_string_view
         get_code(address_t const &a) const noexcept
         {
+            if (_code.find(a) == _code.end()) {
+                return {};
+            }
```
sheapwilliams: How did the need for these changes come up?

include/monad/execution/test/fakes.hpp
tongzhi1998: I didn't see places where we use these variables (`_result` and `_e_result`).

include/monad/execution/test/fakes.hpp
tongzhi1998: Like here?
https://github.com/monad-crypto/monad/pull/163/files#diff-10ec1f4f7233893915d6dcddab9f8465b1038d3eb37d8863e6e386aa51c5a533R284

include/monad/execution/test/fakes.hpp
tongzhi1998: This is actually a partial fix for the behave tests. I should include this in a separate commit. I will remove it.

src/monad/execution/test/transaction_processor.cpp
tongzhi1998: Makes sense! However, I am probably going to refactor them again in the next PR. 
Also about the next PR, I found `State` & `WorkingCopy` are all under template type `TState`, but they shouldn't be. Like in `ReplayFromBlockDb` object, everything above `TP` should use `State` but everything below it should use `WorkingCopy`. However, since the fake `State` doesn't have a `WorkingCopy` inner class, all unit tests will fail. So I am currently refactoring all of those types. I am not sure if this is the right approach, we can discuss it more tmr.

include/monad/execution/test/fakes.hpp
zander-xyz: One argument could be that the real host doesn't have a default constructor so this one shouldn't either. In which case the fix is to just construct dummy arguments into the below constructor. 

Another argument is that this exists purely as a convenience constructor for testing.

Either is fine, but if you do choose to go with having an empty constructor, could you please default it?

```cpp
EvmHost() = default;
```

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -164,9 +134,9 @@ int main(int argc, char *argv[])
     [[maybe_unused]] auto result = replay_eth.run<
         monad::eth_start_fork,
         monad::execution::TransactionProcessor,
-        monad::fakeEmptyEvm,
+        monad::execution::fake::Evm,
         monad::execution::StaticPrecompiles,
-        monad::fakeEvmHost,
+        monad::execution::fake::EvmHost,
         monad::execution::TransactionProcessorFiberData,
         monad::fakeInterpreter,
```
zander-xyz, line 141: [needs no action]

kind of unfortunate that have `monad::fakeInterpreter` instead of `monad::fake::interpreter` to match the rest...

src/monad/execution/test/evm.cpp
```diff
@@ -273,7 +273,15 @@ TEST(Evm, create_contract_account)
...
     fake::State s{};
-    fake::EvmHost h{};
+    fake::EvmHost<
+        fake::State,
+        traits_t,
+        fake::Evm<
+            fake::State,
+            traits_t,
+            fake::static_precompiles::OneHundredGas,
+            fake::Interpreter>>
```
zander-xyz: Is this just `evm_t`? ditto below

src/monad/execution/test/evmone_baseline_interpreter.cpp
```diff
@@ -13,11 +13,21 @@ using namespace monad::execution;
...
     fake::State s{};
-    fake::EvmHost h{};
+    fake::EvmHost<
+        fake::State,
+        traits_t,
+        fake::Evm<
+            fake::State,
+            traits_t,
+            fake::static_precompiles::OneHundredGas,
+            fake::Interpreter>>
```
zander-xyz, line 29: could we define a using clause for this (similar to `evm_t` elsewhere) and use it here and below so that we're not repeating this?

src/monad/execution/test/transaction_processor.cpp
```diff
@@ -11,6 +11,13 @@ using namespace monad::execution;
 using traits_t = fake::traits::alpha<fake::State>;
 using processor_t = TransactionProcessor<fake::State, traits_t>;
 
+fake::EvmHost<
+    fake::State, traits_t,
+    fake::Evm<
+        fake::State, traits_t, fake::static_precompiles::OneHundredGas,
+        fake::Interpreter>>
+    e{};
```
zander-xyz: Why is this a global variable?

src/monad/execution/test/transaction_processor.cpp
zander-xyz: If you plan on refactoring them again in the next PR, that's probably even more of an argument to consolidate the definitions. Save your fingers some typing :)

src/monad/execution/test/evm.cpp
tongzhi1998: This is different than `evm_t`, the original `evm_t` isn't fake. I just declare the entire type as `evm_host_t`

zander-xyz: awesome, thanks. that works

src/monad/execution/test/transaction_processor.cpp
```diff
@@ -72,12 +79,12 @@ TEST(
     constexpr static auto bene{
         0x5353535353535353535353535353535353535353_address};
     fake::State s{};
-    fake::EvmHost e{};
     s._map[from] = {.balance = 56'000'000'000'000'000, .nonce = 25};
     s._map[bene] = {.balance = 0, .nonce = 0};
-    e._result = {
+    evm_host_t h{};
```
zander-xyz, line 84: Note: this breaks if we replace evm_host_t with the real thing. Are we happy with that?

tongzhi1998: I think we should be okay with it. As far as I understand, these unit tests are just testing one thing specifically (in this case, `TransactionProcess`), so we just keep everything else as fakes to avoid unnecessary compound of code.

zander-xyz: I guess the concern is more if at some point we'd like to plug something real in just-to-see-if-it-works. But sounds like you're not concerned about it 👍

PR #167, opened 2023-07-11
Address: bring hash along with address_t

Chen-Yifan:


include/monad/core/address.hpp
```diff
@@ -1,14 +1,41 @@
...
+{
+    using hash_t = bytes32_t;
+    hash_t hash;
+    address_t address;
+};
+
+static_assert(sizeof(Address) == 52);
+static_assert(alignof(Address) == 1);
+
+inline constexpr bool operator==(const Address &a, const Address &b) noexcept
```
ned14: Why not use `operator<=>`?

include/monad/core/address.hpp
```diff
@@ -1,14 +1,41 @@
...
+{
+    return a.hash == b.hash && a.address == b.address;
+}
+
+inline Address construct_address(const address_t &a)
```
ned14: Why not use `Address::Address()`?

include/monad/core/hash_fn.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <monad/core/address.hpp>
+
+#include <cstdint>
+
+MONAD_NAMESPACE_BEGIN
+
+class HashFn
```
ned14, line 9: The STL convention is for a free function called `hash_value()` which can get discovered via ADL.

include/monad/db/trie_db.hpp
```diff
@@ -37,9 +38,10 @@ namespace impl
     {
         struct Updates
         {
-            std::unordered_map<address_t, std::optional<Account>> accounts;
+            std::unordered_map<Address, std::optional<Account>, HashFn>
```
ned14, line 41: I really don't care for unordered maps in public APIs. At best they're just plain slow, at worst passing around maps of things in public APIs tends to suggest overly strong coupling between parts.

Don't get me wrong, sometimes it's unavoidable, but I'd always ask if it's the right approach? For example, generally for a list of something to add to something where that list will get destroyed after, I'd tend to use spans for the API with a vector or array used by the caller for storage. This is because building that list involves a few malloc-free, whereas building a map involves a malloc-free per item in the map. Also, contiguous storage is cache and CPU friendly, whereas maps are just about the least friendly data organisation possible to cache or CPU.

zander-xyz: This piece of code is due to change soon.

include/monad/core/address.hpp
Chen-Yifan: That's all we need to get the custom HashFn comparator working with stl unordered_map, just implemented the minimum to get things working.

include/monad/core/address.hpp
Chen-Yifan: It might lead to misuse of constructor in writing this as a constructor, we want to differentiate `construct_address()` with hash computation from constructor `Address::Address(const address_t& a)` that leaves hash empty.

zander-xyz: `explicit` can help with that. or tag dispatching if you really want to provide both constructors

include/monad/core/address.hpp
zander-xyz: even with spaceship you have to default the `operator==` (if that was the intended behavior)

but generally it's polite to provide the other operators if you provide one of them, and spaceship makes it easy for us to do that.

PR #168, opened 2023-07-12
♻️ consolidate all modifiers in DBInterface into commit() 

zander-xyz:
Problem:
- The DBInterface API is bloated with modifiers that complicates the
  interface

Solution:
- Consolidate all modifiers into a single commit function

On top of https://github.com/monad-crypto/monad/pull/165

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,8 @@ add_library(
     ${PROJECT_SOURCE_DIR}/include/monad/state/code_state.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/state/state.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/state/value_state.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/state/state_changes.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/state/concepts.hpp
     # trie
     ${PROJECT_SOURCE_DIR}/include/monad/trie/compact_encode.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/trie/nibbles.hpp
--- a/include/monad/state/concepts.hpp
+++ b/include/monad/state/concepts.hpp
@@ -1,0 +1,36 @@
+#pragma once
+
+#include <monad/core/account.hpp>
+#include <monad/core/address.hpp>
+#include <monad/state/config.hpp>
+
+#include <concepts>
+#include <optional>
+#include <span>
+
+MONAD_STATE_NAMESPACE_BEGIN
+
+// clang-format off
+
+template <typename T>
+concept account_changes = requires(T obj) {
+    { obj.account_changes } -> std::ranges::range;
+    { *obj.account_changes.begin() } -> std::convertible_to<std::pair<address_t, std::optional<Account>>>;
+    { obj.account_changes.empty() } -> std::same_as<bool>;
+};
+
+template <typename T>
+concept storage_changes = requires(T obj) {
+    { obj.storage_changes } -> std::ranges::range;
+    { obj.storage_changes.begin()->first } -> std::convertible_to<address_t>;
+    { obj.storage_changes.begin()->second } -> std::ranges::range;
+    { *obj.storage_changes.begin()->second.begin() } -> std::convertible_to<std::pair<bytes32_t, bytes32_t>>;
+    { obj.storage_changes.empty() } -> std::same_as<bool>;
+};
+
+template <typename T>
+concept changeset = account_changes<T> && storage_changes<T>;
+
+// clang-format on
+
+MONAD_STATE_NAMESPACE_END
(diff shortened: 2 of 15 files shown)


include/monad/state/state_changes.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <monad/core/address.hpp>
+#include <monad/core/bytes.hpp>
+#include <monad/state/config.hpp>
+
+#include <optional>
+#include <vector>
+
+MONAD_STATE_NAMESPACE_BEGIN
+
+struct StateChanges
```
zander-xyz, line 13: This is expected to change as @sheapwilliams concludes his work to handle call reversions.

include/monad/state/concepts.hpp
```diff
@@ -0,0 +1,41 @@
...
+    {
+        *obj.storage_changes.begin()->second.begin()
+    } -> std::convertible_to<std::pair<bytes32_t, bytes32_t>>;
+    {
+        obj.account_changes.empty()
+    } -> std::same_as<bool>;
+    {
+        obj.storage_changes.empty()
+    } -> std::same_as<bool>;
+};
```
sheapwilliams: formatting of concepts makes this nearly illegible...such a shame. 😦

CMakeLists.txt
```diff
@@ -209,6 +209,8 @@ add_library(
     ${PROJECT_SOURCE_DIR}/include/monad/state/code_state.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/state/state.hpp
     ${PROJECT_SOURCE_DIR}/include/monad/state/value_state.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/state/state_changes.hpp
+    ${PROJECT_SOURCE_DIR}/include/monad/state/concepts.hpp
```
sheapwilliams, line 213: Other new header files?

include/monad/state/concepts.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <monad/state/config.hpp>
+
+#include <concepts>
+#include <optional>
+#include <span>
+
+MONAD_STATE_NAMESPACE_BEGIN
+
+template <typename T>
+concept HasStateChanges = requires(T obj) {
```
sheapwilliams: [nit] `HasStateChanges`?  I was reading the usages of this concept, and I was expecting something different - I became a bit confused.  Is this just a ChangeSet?  Or a BlockChangeSet if we are tying the semantic of block id to this eventually?

include/monad/state/value_state.hpp
```diff
@@ -116,24 +118,21 @@ struct ValueState
...
-                }
-                else {
-                    db_.update(addr, key, value.updated);
-                }
+                sc.storage_changes[addr].emplace_back(key, value.updated);
             }
         }
         merged_.clear();
-        db_.commit_storage();
+        db_.commit(sc);
```
sheapwilliams, line 135: [discussion point] This is probably fine for now, but I feel like there is a bigger refactor lurking here - why not just store the changes as a ChangeSet/StateChanges?  Why build up a new data structure, when we could just account for the changes in the way the structure the DB accepts.  Thoughts?

src/monad/db/test/db.cpp
```diff
@@ -132,3 +117,19 @@ TEST(InMemoryTrieDB, erase)
     EXPECT_EQ(db.root_hash(), NULL_ROOT);
     EXPECT_EQ(db.root_hash(a), NULL_ROOT);
 }
+
+TYPED_TEST(TrieDBTest, ModifyStorageOfAccount)
```
sheapwilliams, line 121: [nit] this test name doesn't look like the rest of them...it looks like the test is binding a root hash?  Where did that hash come from?

CMakeLists.txt
zander-xyz: there are the only two

include/monad/state/concepts.hpp
zander-xyz: How does `HasChangeSet` sound? If we decide to define a concept in the future that cares about both the change set and the block number, we can do that.

src/monad/db/test/db.cpp
zander-xyz: There was a bug in a prior implementation where a modification to the storage (but not the account) resulted in an updated storage root but not an updated global state root. That's all that this tests.

include/monad/state/value_state.hpp
zander-xyz: Discussed over slack. This comment is correct in that this code is inefficient. We're going to put on this since we expect the design to change.

CMakeLists.txt
sheapwilliams: I see!

src/monad/db/test/db.cpp
sheapwilliams: I understand now.  It says that, but I didn't get it right away.  Thx.

PR #170, opened 2023-07-13
🐛 State fixes

sheapwilliams:
Fix several bugs after integrating with Rene's compatibility testing work:

1. Assert if accessing a non-existing account 
2. Set the code hash when setting the code
3. Don't look through self-destructed accounts when processing EIP-161 empty accounts

### DIFF
--- a/include/monad/state/account_state.hpp
+++ b/include/monad/state/account_state.hpp
@@ -213,7 +213,8 @@ struct AccountState<TAccountDB>::WorkingCopy : public AccountState<TAccountDB>
     void destruct_touched_dead() noexcept
     {
         for (auto &i : changed_) {
-            if (i.second.updated.value() == Account{}) {
+            if (i.second.updated.has_value() &&
+                i.second.updated.value() == Account{}) {
                 i.second.updated.reset();
             }
         }


include/monad/state/account_state.hpp
```diff
@@ -151,6 +151,7 @@ struct AccountState<TAccountDB>::WorkingCopy : public AccountState<TAccountDB>
     // EVMC Host Interface
     evmc_access_status access_account(address_t const &a)
     {
+        MONAD_DEBUG_ASSERT(account_exists(a));
```
tongzhi1998, line 154: Should we do the same thing for `get_balance(address_t const& a)`? 
I am asking this because I see in your beneficiary code, you call `get_balance()` for an account without checking if it exists or not (I saw there is `value_or(Account{}).balance `in AccountState, but just not completely sure about it). 
Also, is there any specific reason to mark `get_balance` in `State` and `AccountState` as `noexcept`? There is some map accessing code there so marking them as `noexcept` can be confusing

sheapwilliams: We don't get to mess with the signatures of the EVMC Host interface calls....There isn't a good answer to that, so for now they faithfully mimic the required interface - even if it is wrong.

We can add a debug assert - can't hurt anything save find API usage issues earlier. :+1:

tongzhi1998: I see. I am fine with either adding the DEBUG_ASSERT or not.

PR #173, opened 2023-07-14
:construction_worker: Allow both ASan and UBSan to work together

sheapwilliams:
Problem:
- Both ASan and UBSan can work together, but the current cmake does not allow it
- There is no reason not to allow it

Solution:
- Let both of the sanitizers be configured if desired.

@rgarc - this addresses the issue that came up during the compatibility integration

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,12 +262,14 @@ if(ASAN)
         monad PUBLIC "-fsanitize=address" "-fno-omit-frame-pointer"
     )
     target_link_options(monad PUBLIC "-fsanitize=address")
-elseif(UBSAN)
+endif()
+if(UBSAN)
     message("Undefined Behavior Sanitizer: On")
     target_compile_options(
-        monad PUBLIC "-fsanitize=undefined" "-fno-omit-frame-pointer"
+        monad PUBLIC "-fsanitize=undefined" "-fno-omit-frame-pointer" "-fPIC"
     )
     target_link_options(monad PUBLIC "-fsanitize=undefined")
-else()
+endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -262,12 +262,14 @@ if(ASAN)
         monad PUBLIC "-fsanitize=address" "-fno-omit-frame-pointer"
     )
     target_link_options(monad PUBLIC "-fsanitize=address")
-elseif(UBSAN)
+endif()
+if(UBSAN)
     message("Undefined Behavior Sanitizer: On")
     target_compile_options(
-        monad PUBLIC "-fsanitize=undefined" "-fno-omit-frame-pointer"
+        monad PUBLIC "-fsanitize=undefined" "-fno-omit-frame-pointer" "-fPIC"
```
zander-xyz, line 269: why the fPIC?

ned14: The sanitisers need position independent code.

That above is the wrong way to set position independent code. On the target, set in cmake the property https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html, or in the cmake toolchain file, set `CMAKE_POSITION_INDEPENDENT_CODE` to On.

zander-xyz: how does this actually manifest? I've definitely run UBSAN and ASAN without fPIC

ned14: Generally your compiler silently turns it on for you, which is fine if you build everything with sanitisers on (e.g. thanks to toolchain file).

If you mix non-sanitised and sanitised object files, then you get linker complaint, or sometimes when you go to execute it just dies with a very obscure failure.

PR #177, opened 2023-07-14
:new: Process the Block Beneficiary

sheapwilliams:
This is at least *more* correct than what was there before.  There is still some question of further gas reward to the miner that will be a follow-on.  But for now this at least gets Tong's work processing early blocks correctly.

### DIFF
--- a/include/monad/execution/block_processor.hpp
+++ b/include/monad/execution/block_processor.hpp
@@ -16,7 +16,7 @@ MONAD_EXECUTION_NAMESPACE_BEGIN
 template <class TExecution>
 struct AllTxnBlockProcessor
 {
-    template <class TState, class TFiberData>
+    template <class TState, class TTraits, class TFiberData>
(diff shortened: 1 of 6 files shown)


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -255,6 +282,10 @@ namespace fork_traits
 
         static constexpr evmc_revision rev = EVMC_BYZANTIUM;
         static constexpr auto last_block_number = 9'068'999u;
+        static constexpr uint256_t block_reward =
```
tongzhi1998, line 285: `next_fork_t` for `byzantium` should be changed to `constantinople` instead? 
By the way, this makes me wonder if our unit tests should cover each fork transition (Right now it only covers a few).

sheapwilliams: Yup - total miss.  Thx Tong!

PR #179, opened 2023-07-17
Replay Initial Ethereum Blocks With No Txns

tongzhi1998:
- Rewrite `replay_block_db` interface
- Replace all fakes with real object in `replay_ethereum.cpp`
- Add Genesis block support for `replay_block_db`
- Add BDD tests to assure successful run from block 0 to 46146

### DIFF
--- a/test/features/ethereum/configure_logger.feature
+++ b/test/features/ethereum/configure_logger.feature
@@ -1,40 +1,33 @@
 Feature: Configure Logger
 
-    Scenario: Using default log level
+    Background: BlockDb path setup
         Given I run with BlockDb path = "erigon/block_db"
         And I run with start block number = "3000000"
         And I run with finish block number = "3000001"
-        And I run with default log level
+
+    Scenario: Using default log level
+        Given I run with default log level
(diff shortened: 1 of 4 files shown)


include/monad/execution/replay_block_db.hpp
```diff
@@ -102,36 +151,22 @@ class ReplayFromBlockDb
                                 TInterpreter>>,
                         TExecution>>(state, block);
 
+                state.commit();
+
+                // TODO: How exactly do we calculate transaction root and
```
rgarc, line 156: Is there something wrong with how you're calculating them below?
```C++
                auto const transactions_root = transaction_trie.root_hash();
                auto const receipts_root = receipt_trie.root_hash();
```

tongzhi1998: It will have problem for later blocks. However, since all the blocks right now don't have any txns (thus no receipt too), the hash for `txn_root` and `rct_root` are just `NULL_ROOT_HASH`, which is defined in `bytes.hpp` and used in the fakes.

PR #180, opened 2023-07-18
🆕 Award gas used to the block beneficiary

sheapwilliams:
Currently, the gas used is lost, but the YP, Section 6.2 states:

> The Ether for the gas is given to the miner, whose address is specified as the beneficiary of the present block B. 

This supports up to London, where the gas used is now partially burnt.  That change will be forthcoming.

### DIFF
--- a/include/monad/execution/transaction_processor.hpp
+++ b/include/monad/execution/transaction_processor.hpp
@@ -107,6 +107,12 @@ struct TransactionProcessor
         TTraits::destruct_touched_dead(s);
         s.destruct_suicides();
 
+        TTraits::apply_txn_award(
+            s,
+            t,
+            base_fee_per_gas,
+            t.gas_limit - static_cast<uint64_t>(result.gas_left));
+
         auto receipt =
             h.make_receipt_from_result(result.status_code, t, gas_remaining);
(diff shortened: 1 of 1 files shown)


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -186,6 +186,14 @@ namespace fork_traits
         {
             apply_mining_award(s, b, block_reward, additional_ommer_reward);
         }
+
+        template <class TState>
+        static constexpr void apply_txn_award(
+            TState &s, Transaction const &, uint64_t base_fee_per_gas,
+            uint64_t gas_used)
+        {
+            s.add_txn_award(uint256_t{gas_used * base_fee_per_gas});
```
tongzhi1998: Hm... I am not too sure if this might incur any overflow problem? Like the product of 2 `uint64_t` might overflow `uint64_t`, and even though it is assigned to `uint256_t` later, will the first overflow still create problem?

sheapwilliams: Y - I'll cast both to a uint256_t before the multiplication. :+1:

PR #181, opened 2023-07-19
:bug: Commit storage and account changes together.

rgarc:
Problem:
- On the current main branch, assertions in [`TrieDBInterface::commit`]
  always fail because [`ValueState::commit_all_merged`] always calls
  into [`TrieDBInterface::commit`] with an empty list of account
  changes.

Solution:
- Collect the account and storage changes into the same `StateChanges`
  struct and commit them together so that the aforementioned assertions
  no longer fail.

Note:
- This PR directly contradicts [this comment], so this PR needs to be
  scrapped or the comment should be updated.

[`TrieDBInterface::commit`]: https://github.com/monad-crypto/monad/blob/e6c159c5e5ef834b241e5fa36d39ce487a3a0748/include/monad/db/trie_db_interface.hpp#L149-L250
[`ValueState::commit_all_merged`]: https://github.com/monad-crypto/monad/blob/e6c159c5e5ef834b241e5fa36d39ce487a3a0748/include/monad/state/value_state.hpp#L117-L136
[this comment]: https://github.com/monad-crypto/monad/blob/e6c159c5e5ef834b241e5fa36d39ce487a3a0748/include/monad/state/state.hpp#L238-L240

### DIFF
--- a/include/monad/state/account_state.hpp
+++ b/include/monad/state/account_state.hpp
@@ -124,18 +124,18 @@ struct AccountState
         });
     }
 
-    void commit_all_merged()
+    [[nodiscard]] StateChanges::AccountChanges gather_changes() const
     {
         assert(can_commit());
 
-        StateChanges sc;
+        StateChanges::AccountChanges account_changes;
         for (auto const &[addr, diff] : merged_) {
-            sc.account_changes.emplace_back(addr, diff.updated);
+            account_changes.emplace_back(addr, diff.updated);
         }
-
-        db_.commit(sc);
-        merged_.clear();
+        return account_changes;
     }
+
+    void clear_changes() { merged_.clear(); }
 
     [[nodiscard]] bytes32_t get_state_hash() const { return db_.root_hash(); }
 };
@@ -174,8 +174,7 @@ struct AccountState<TAccountDB>::WorkingCopy : public AccountState<TAccountDB>
         auto const account = get_committed_storage(a).has_value()
                                  ? get_committed_storage(a)
                                  : Account{};
-        changed_.emplace(
-            a, diff_t{get_committed_storage(a), account});
+        changed_.emplace(a, diff_t{get_committed_storage(a), account});
         return EVMC_ACCESS_COLD;
     }
 
--- a/include/monad/state/state.hpp
+++ b/include/monad/state/state.hpp
@@ -7,6 +7,7 @@
 #include <monad/core/receipt.hpp>
 
 #include <monad/state/config.hpp>
+#include <monad/state/state_changes.hpp>
 
 #include <ethash/keccak.hpp>
 
@@ -16,7 +17,8 @@
 MONAD_STATE_NAMESPACE_BEGIN
 
 template <
-    class TAccountState, class TValueState, class TCodeState, class TBlockCache>
+    class TAccountState, class TValueState, class TCodeState, class TBlockCache,
+    class TDatabase>
 struct State
 {
     struct WorkingCopy
@@ -184,13 +186,17 @@ struct State
     TValueState &storage_;
     TCodeState &code_;
     TBlockCache &block_cache_{};
+    TDatabase &db_{};
     unsigned int current_txn_{};
 
-    State(TAccountState &a, TValueState &s, TCodeState &c, TBlockCache &bc)
+    State(
+        TAccountState &a, TValueState &s, TCodeState &c, TBlockCache &bc,
+        TDatabase &db)
         : accounts_{a}
         , storage_{s}
         , code_{c}
         , block_cache_{bc}
+        , db_{db}
     {
     }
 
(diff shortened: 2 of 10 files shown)


include/monad/state/account_state.hpp
```diff
@@ -123,11 +123,10 @@ struct AccountState
...
 
-    void commit_all_merged()
+    void commit_all_merged(StateChanges &sc)
     {
         assert(can_commit());
 
-        StateChanges sc;
         for (auto const &[addr, diff] : merged_) {
             sc.account_changes.emplace_back(addr, diff.updated);
         }
```
sheapwilliams, line 134: We probably shouldn't have this composition be structural by letting the AccountState `commit()` the ChangeSet to the back-end.

Perhaps we can plumb a db `commit(...)` that takes a ChangeSet, and then have the State use that instead?  Something like that?

include/monad/state/state.hpp
```diff
@@ -238,8 +239,9 @@ struct State
         // Note: storage updates must be committed prior to the account
         // updates, since each account needs the most up-to-date storage
         // root
-        storage_.commit_all_merged();
-        accounts_.commit_all_merged();
+        StateChanges state_changes;
+        storage_.commit_all_merged(state_changes); // this WONT call commit()
+        accounts_.commit_all_merged(state_changes); // this WILL call commit()
```
sheapwilliams: If we don't make this composition structural, then we can remove these comments too...

include/monad/state/state.hpp
```diff
@@ -238,8 +239,9 @@ struct State
         // Note: storage updates must be committed prior to the account
         // updates, since each account needs the most up-to-date storage
         // root
```
zander-xyz: Discussed on zoom: we should remove this

include/monad/state/account_state.hpp
```diff
@@ -124,17 +124,16 @@ struct AccountState
         });
     }
 
-    void commit_all_merged()
+    StateChanges::AccountChanges gather_account_changes()
```
zander-xyz: Can this have `[[nodiscard]]` and `const`?

include/monad/state/state.hpp
```diff
@@ -246,12 +252,12 @@ struct State
 
     void commit()
     {
-        // Note: storage updates must be committed prior to the account
-        // updates, since each account needs the most up-to-date storage
-        // root
-        storage_.commit_all_merged();
-        accounts_.commit_all_merged();
+        StateChanges state_changes = {
```
zander-xyz: this one can be const

src/monad/state/test/state.cpp
```diff
@@ -437,14 +440,47 @@ TYPED_TEST(StateTest, can_commit)
...
 
+template <typename TDB>
+struct StateAndStorageTest : public testing::Test
+{
+};
+using StateAndStorageTestDBTypes =
+    ::testing::Types<db::InMemoryTrieDB, db::RocksTrieDB>;
+TYPED_TEST_SUITE(StateAndStorageTest, StateAndStorageTestDBTypes);
+
+TYPED_TEST(StateAndStorageTest, commit_account_and_storage_together_regression)
```
zander-xyz: Is this just testing that you can't commit storage changes for an account that doesn't exist? If so, the test name should probably be modified to more accurately reflect that (ie. `create_storage_for_missing_account`).

src/monad/state/test/state.cpp
```diff
@@ -437,14 +440,47 @@ TYPED_TEST(StateTest, can_commit)
...
+                 {c, {{key1, value1}, {key2, value2}}}}}),
+        std::bad_optional_access);
+#else
+    EXPECT_DEBUG_DEATH(
+        db.commit(StateChanges{
+            .storage_changes =
+                {{b, {{key1, value1}, {key2, value2}}},
+                 {c, {{key1, value1}, {key2, value2}}}}}),
+        ".*");
+#endif
```
zander-xyz: In this specific test case, it seems like you *just* care that the program dies. Would EXPECT_DEATH (or EXPECT_EXIT) be enough? That'll get rid of the macro magic you have here.

src/monad/state/test/state.cpp
```diff
@@ -437,14 +440,47 @@ TYPED_TEST(StateTest, can_commit)
     EXPECT_TRUE(t.can_commit());
 }
 
+template <typename TDB>
+struct StateAndStorageTest : public testing::Test
+{
+};
+using StateAndStorageTestDBTypes =
+    ::testing::Types<db::InMemoryTrieDB, db::RocksTrieDB>;
+TYPED_TEST_SUITE(StateAndStorageTest, StateAndStorageTestDBTypes);
```
zander-xyz: These seem more aptly named `TrieDBTest` and `TrieDBTypes` (see `test/db.cpp`)

include/monad/state/account_state.hpp
zander-xyz: I now see that we can't const this function because we clear from merged. Perhaps we should move the clearing functionality into it's own function, and have `state.hpp` clear the state once it commits to the database.

That narrows the scope of this function, and a nice side effect is that you can const it.

src/monad/state/test/state.cpp
rgarc: I think I tried both of those. I will try again because I agree this ifdef is overkill.

src/monad/state/test/value_state.cpp
```diff
@@ -685,5 +685,5 @@ TYPED_TEST(ValueStateTest, commit_all_merged)
         EXPECT_TRUE(s.can_commit());
     }
 
-    s.commit_all_merged();
+    auto _ = s.gather_changes();
```
zander-xyz, line 688: is this line of code needed?

rgarc: nope!

Issue #188, opened 2023-07-20
Paralleled recover_sender


tongzhi1998:
- Previously `recover_sender` is in `FiberData`, which works fine by itself but fails in paralleled execution. 
- Suspect that the problem is due to some static `secp` library functions which corrupts our heap memory
- The temporary solution is to move the `recover_sender` function to `BlockProcessor` instead, but that reduces parallelism
- Need to find a more elegant solution to this problem to take advantage of parallel execution 
- #189

PR #189, opened 2023-07-20
:recycle: Move recover_sender from FiberData to BP

tongzhi1998:
Problem:
- We currently call recover_sender from FiberData, which fails in parallel execution (maybe due to static secp library function)

Solution:
- Temporarily move this function to BlockProcessor
- Mark mem allocation part as static to avoid duplication

Future Work:
- Issue #188

### DIFF
--- a/include/monad/execution/transaction_processor_data.hpp
+++ b/include/monad/execution/transaction_processor_data.hpp
@@ -27,7 +27,7 @@ struct alignas(64) TransactionProcessorFiberData
     using txn_processor_status_t = typename TTxnProcessor::Status;
 
     TState &s_;
-    Transaction &txn_;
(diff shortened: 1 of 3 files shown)


include/monad/execution/transaction_processor_data.hpp
```diff
@@ -64,8 +64,6 @@ struct alignas(64) TransactionProcessorFiberData
         auto *txn_logger = log::logger_t::get_logger("txn_logger");
         auto const start_time = std::chrono::steady_clock::now();
 
-        txn_.from = recover_sender(txn_);
```
sheapwilliams, line 67: The Transaction can be const at this point, right?

PR #192, opened 2023-07-21
:new: Add database restart support for replay_ethereum

tongzhi1998:
- Add DB restart code
- Currently the DB state is stored under build/. So don't ever remove build/ if you don't want to lose all DB state.

### DIFF
--- a/src/monad/execution/ethereum/replay_ethereum.cpp
+++ b/src/monad/execution/ethereum/replay_ethereum.cpp
@@ -30,9 +30,14 @@
 #include <monad/state/state.hpp>
 #include <monad/state/value_state.hpp>
 
+#include <fmt/format.h>
+
 #include <CLI/CLI.hpp>
 
+#include <cerrno>
+#include <cstdlib>
 #include <filesystem>
+#include <optional>
 
 MONAD_NAMESPACE_BEGIN
 
@@ -53,13 +58,46 @@ public:
     bytes32_t root_hash() const { return NULL_ROOT; }
 };
 
+inline std::optional<std::filesystem::path> get_db_path(
+    block_num_t const start_block_number, block_num_t const finish_block_number,
+    std::filesystem::path state_db_dir, bool cleanup)
+{
+    if (!std::filesystem::exists(state_db_dir)) {
+        return std::nullopt;
+    }
+    state_db_dir = state_db_dir / "replay_ethereum_state_db";
+    auto const from_dir =
+        state_db_dir / fmt::format("{}", (start_block_number - 1u));
+    if (start_block_number != 0u && !std::filesystem::exists(from_dir)) {
+        return std::nullopt;
+    }
+
+    auto const to_dir =
+        state_db_dir / fmt::format("{}", finish_block_number - 1u);
+    if (std::filesystem::exists(to_dir)) {
+        std::filesystem::remove_all(to_dir);
+    }
+    std::filesystem::create_directories(to_dir);
+    if (start_block_number != 0u) {
+        std::filesystem::copy(
+            from_dir, to_dir, std::filesystem::copy_options::recursive);
+        if(cleanup){
+            std::filesystem::remove_all(from_dir);
+        }
+    }
+
+    return to_dir;
+}
+
 MONAD_NAMESPACE_END
 
 int main(int argc, char *argv[])
 {
-    CLI::App cli{"replay_ethereum_block_db"};
+    CLI::App cli{"replay_ethereum"};
 
     std::filesystem::path block_db_path{};
+    std::filesystem::path state_db_path{};
+    bool cleanup = false;
     monad::block_num_t start_block_number;
     std::optional<monad::block_num_t> finish_block_number = std::nullopt;
 
@@ -88,6 +126,9 @@ int main(int argc, char *argv[])
         ->required();
     cli.add_option(
         "-f, --finish", finish_block_number, "1 pass the last executed block");
+    cli.add_option(
+        "--state_db", state_db_path, "restart state_db directory (absolute or relative)")->required();
+    cli.add_flag("--cleanup", cleanup, "Clean up the original state_db");
 
     auto *log_levels = cli.add_subcommand("log_levels", "level of logging");
     log_levels->add_option("--main", main_log_level, "Log level for main");
@@ -101,7 +142,7 @@ int main(int argc, char *argv[])
 
     // Real Objects
     using code_db_t = std::unordered_map<monad::address_t, monad::byte_string>;
-    using db_t = monad::db::InMemoryTrieDB;
+    using db_t = monad::db::RocksTrieDB;
     using block_db_t = monad::db::BlockDb;
     using receipt_collector_t = monad::receiptCollector;
     using state_t = monad::state::State<
@@ -130,7 +171,18 @@ int main(int argc, char *argv[])
         finish_block_number);
 
     block_db_t block_db(block_db_path);
-    db_t db{};
+
+    auto const to_dir = monad::get_db_path(
+        start_block_number, finish_block_number.value(), state_db_path, cleanup);
+    if (!to_dir.has_value()) {
+        MONAD_LOG_ERROR(
+            main_logger,
+            "Can't create or read from rocks_db with start_block_number = {}",
+            start_block_number);
+        return EINVAL;
+    }
+    db_t db{to_dir.value()};
+
     code_db_t code_db{};
     monad::state::AccountState accounts{db};
     monad::state::ValueState values{db};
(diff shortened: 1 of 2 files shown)


src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -53,6 +57,28 @@ class fakeEmptyReceiptTrie
...
 
+inline std::pair<std::filesystem::path, bool> read_write_rocks_db(
+    block_num_t const start_block_number, block_num_t const finish_block_number)
+{
+    auto const dir = test_resource::build_dir / "rocksdb" / "replay_ethereum";
+    auto const from_dir = dir / fmt::format("{}", (start_block_number - 1u));
+    if (start_block_number != 0u && !std::filesystem::exists(from_dir)) {
+        return std::make_pair(dir / "error", false);
+    }
+
```
sheapwilliams: Can we perhaps modify the `replay_ethereum` executable to take a db from the command-line instead?  I'd really like to control where I store these dbs for different purposes.

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -88,6 +119,10 @@ int main(int argc, char *argv[])
         ->required();
     cli.add_option(
         "-f, --finish", finish_block_number, "1 pass the last executed block");
+    auto relative_path_option = cli.add_option(
+        "--state_db",
+        state_db_path_relative,
+        "state restart db directory (relative to project src)");
```
sheapwilliams: Can we use absolute paths instead?  I don't see any reason to tie this to the source root.  It will help get rid of the test_resource.h header from this program as well...

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -30,6 +30,10 @@
 #include <monad/state/state.hpp>
 #include <monad/state/value_state.hpp>
 
+#include <test_resource_data.h>
```
sheapwilliams: [nit] I think we should try to avoid having this file in this program - test resources should really be binned to the tests, I think.

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -101,7 +136,8 @@ int main(int argc, char *argv[])
 
     // Real Objects
     using code_db_t = std::unordered_map<monad::address_t, monad::byte_string>;
-    using db_t = monad::db::InMemoryTrieDB;
+    // Always use RocksTrieDB for now, which supports restart & trie-building
```
sheapwilliams: I think this change is fine - we don't need to support command-line in-memory database setup.  Something for testing and debugging only.  We can lose the comment.

src/monad/execution/ethereum/replay_ethereum.cpp
tongzhi1998: Makes sense. However, should we still set a default location for `state_db` storage? If so, since we don't want to use test_resource, should we use `getenv()` instead? (I imagine the default location should be something like user's home)

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -53,13 +57,40 @@ class fakeEmptyReceiptTrie
     bytes32_t root_hash() const { return NULL_ROOT; }
 };
 
+inline std::pair<std::filesystem::path, bool> read_write_rocks_db(
```
zander-xyz: A couple things with this function signature:
- `std::optional` seems to convey meaning a bit better than a pair of path and success boolean flag
- the name of this function seems more aptly named as `try_get_db_path` or something along those lines, since that is what this function is doing

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -53,13 +57,40 @@ class fakeEmptyReceiptTrie
...
+        std::filesystem::remove_all(to_dir);
+    }
+    std::filesystem::create_directories(to_dir);
+    if (start_block_number != 0u) {
+        std::filesystem::copy(
+            from_dir, to_dir, std::filesystem::copy_options::recursive);
+    }
+
+    return std::make_pair(to_dir, true);
+}
```
zander-xyz: Do you absolutely need this automated way of keeping track of block history? Any objections to keeping it simple and just taking in an input db path directly on the command line?

This method of keeping track of block history is fine if you absolutely need it, but eventually we'll add block history support natively into the db interface, and I'm not sure all of this file manipulation is easy to understand.

Of course this (for now) relies on the user specifying the correct db for the right block number. Eventually we'll need to store block number somehow, which will let us verify this programmatically.

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -130,7 +166,18 @@ int main(int argc, char *argv[])
...
-    db_t db{};
+
+    auto const res = monad::read_write_rocks_db(
+        start_block_number, finish_block_number.value(), state_db_path);
+    if (!res.second) {
+        MONAD_LOG_ERROR(
+            main_logger,
+            "Can't create or read from rocks_db with start_block_number = {}",
+            start_block_number);
+        return 1;
```
zander-xyz: return EINVAL?

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -53,13 +58,42 @@ class fakeEmptyReceiptTrie
...
+
 MONAD_NAMESPACE_END
 
 int main(int argc, char *argv[])
 {
-    CLI::App cli{"replay_ethereum_block_db"};
+    CLI::App cli{"replay_ethereum"};
 
     std::filesystem::path block_db_path{};
+    std::filesystem::path state_db_path{getenv("HOME")};
```
zander-xyz: why is the getenv needed?

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -88,6 +122,8 @@ int main(int argc, char *argv[])
         ->required();
     cli.add_option(
         "-f, --finish", finish_block_number, "1 pass the last executed block");
+    cli.add_option(
+        "--state_db", state_db_path, "restart state_db directory (absolute)");
```
zander-xyz: where do you actually require this to be an absolute path?

zander-xyz: also can we rename this option to `--db`? `state_db` is internal terminology that we use but I'm not sure it needs to be exposed outside of the implementation of the db interface.

src/monad/execution/ethereum/replay_ethereum.cpp
tongzhi1998: After thinking about it a bit more, I am not sure if we need to really merge this in as a PR (since this is temporary by nature). However, I think the above DB manipulation does save me some time by not having to do `cp -r ` stuff whenever I want to save the current DB.
Let me know how you guys think of this @zander-xyz @sheapwilliams

src/monad/execution/ethereum/replay_ethereum.cpp
tongzhi1998: 1. I can verify this is absolute path.
2. I am a bit hesitant to make this as `--db` because it might get confused with `block_db`

sheapwilliams: Yea, we'll probably have 3 or so dbs: state, block/transactions, and receipts.  We can probably just drop the "db" nomenclature as totally unhelpful if we are going to drop anything.  It time, the thing we probably will want to differentiate it between a RocksDB and a TrieDB...That will get confusing.

src/monad/execution/ethereum/replay_ethereum.cpp
sheapwilliams: We should definitely merge this - it isn't helpful for Tong to have the only canonical source about how to get thru block history.  He just just be keeping track of the hacks/bugs, I think.

That said, I'd be ok with a different way of keeping track if progression - perhaps a smaller "progress" table in the db.  I think that would be time well spent.  U 2 should consider jumping on a zoom and hash it out for us.

zander-xyz: Temporary code is ok. Let's merge this in

src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -88,6 +122,8 @@ int main(int argc, char *argv[])
         ->required();
     cli.add_option(
         "-f, --finish", finish_block_number, "1 pass the last executed block");
+    cli.add_option(
+        "--state_db", state_db_path, "restart state_db directory (absolute)")->required();
```
zander-xyz: >(absolute)

nit: wording

Are you sure this does not work with a relative path?

PR #196, opened 2023-07-26
:recycle: Fix trie_db_logger issue in behave test

tongzhi1998:
- In configure logger test, we didn't take care of trie_db_logger, thus causing us to fail tests

### DIFF
--- a/test/features/ethereum/configure_logger.feature
+++ b/test/features/ethereum/configure_logger.feature
@@ -2,33 +2,34 @@ Feature: Configure Logger
 
     Background: BlockDb path setup
         Given I run with BlockDb path = "erigon/block_db"
-        And I run with start block number = "3000000"
-        And I run with finish block number = "3000001"
+        And I run with start block number = "46147"
+        And I run with finish block number = "46148"
 
     Scenario: Using default log level
         Given I run with default log level
         When I start "replay_ethereum"
-        Then the output should not contain "BlockHeader Fields:"
-        And the output should not contain "Log"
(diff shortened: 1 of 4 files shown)


test/features/ethereum/configure_logger.feature
```diff
@@ -25,10 +26,12 @@ Feature: Configure Logger
         Then the output should not contain "BlockHeader Fields:"
         And the output should not contain "Log"
         And the output should not contain "Start executing Transaction"
+        But the output should contain "ACCOUNT_UPDATES"
```
sheapwilliams: 'But' is usually reserved for "not" statements, e.g.
> But the output should **not** contain "ACCOUNT_UPDATES"
as an example.

test/features/ethereum/configure_logger.feature
```diff
@@ -2,15 +2,16 @@ Feature: Configure Logger
...
+        And I run with start block number = "46147"
+        And I run with finish block number = "46148"
 
     Scenario: Using default log level
         Given I run with default log level
         When I start "replay_ethereum"
         Then the output should not contain "BlockHeader Fields:"
         And the output should not contain "Log"
         And the output should contain "Start executing Transaction"
+        And the output should contain "ACCOUNT_UPDATES"
```
sheapwilliams: I think this statement is a little weak - even though you are expecting the statement, you are expecting a lot of statements that you haven't bound here in this test.  I think we lose nothing if you just remove these 'Then' clauses.  Feel free to zoom me if I'm missing something.

PR #198, opened 2023-07-27
✨ Introduce block history to the DBInterface 

zander-xyz:
Problem:
- We need to access state from a previous block for various RPC methods

Solution:
- Add ability to save block history and prune off those who fall outside
  of the block history size

## Filesystem
Directory structure for history looks like `<root>/<block number>/<database type>` where `<root>` is user specified, `<block_number>` is the block number associated with the checkpoint(s), and `<database_type>` will eventually be one of `rocksdb` (?), `rockstriedb`, and `triedb`.

`<root>/CURRENT/<database type>` always contains the working database that the validator is operating on

```shell
.
├── 0
│   └── rockstriedb
│       ├── 000017.sst
│       ├── 000018.sst
│       ├── 000019.sst
│       ├── 000020.sst
│       ├── CURRENT
│       ├── MANIFEST-000005
│       └── OPTIONS-000015
├── 1
│   └── rockstriedb
│       ├── 000017.sst
│       ├── 000018.sst
│       ├── 000019.sst
│       ├── 000020.sst
│       ├── 000028.sst
│       ├── 000029.sst
│       ├── 000030.sst
│       ├── 000031.sst
│       ├── CURRENT
│       ├── MANIFEST-000024
│       └── OPTIONS-000026
└── CURRENT
    └── rockstriedb
        ├── 000017.sst
        ├── 000018.sst
        ├── 000019.sst
        ├── 000020.sst
        ├── 000023.log
        ├── CURRENT
        ├── IDENTITY
        ├── LOCK
        ├── LOG
        ├── MANIFEST-000024
        ├── OPTIONS-000015
        └── OPTIONS-000026

7 directories, 30 files
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -27,6 +27,7 @@ RUN yum -y install \
   brotli \
   gmp \
   gtest \
+  gmock \
   rocksdb
 
 RUN yum -y install \
@@ -35,6 +36,7 @@ RUN yum -y install \
   brotli-devel \
   gmp-devel \
   gtest-devel \
+  gmock-devel \
   rocksdb-devel
 
 RUN pip install -U conan==1.*


include/monad/db/create_and_prune_block_history.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include <iostream>
+#include <memory>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+struct RocksTrieDB;
+
+template <typename TDB>
+    requires std::same_as<TDB, db::RocksTrieDB>
+[[nodiscard]] inline tl::expected<void, std::string>
```
rgarc, line 22: It's probably fine here, but using `std::string` as an error type makes us lose type information on the error that occurred.

include/monad/db/util.hpp
```diff
@@ -7,6 +7,18 @@
...
 MONAD_DB_NAMESPACE_BEGIN
 
+struct RocksTrieDB;
+
+template <typename TDB>
+    requires std::same_as<TDB, db::RocksTrieDB>
+[[nodiscard]] constexpr std::string_view as_string() noexcept
+{
+    using namespace std::literals::string_view_literals;
+    if constexpr (std::same_as<TDB, db::RocksTrieDB>) {
```
rgarc, line 17: Why is this `if constexpr` needed given the `requires` clause above?

zander-xyz: In preparation for adding support for RocksDB

include/monad/db/create_and_prune_block_history.hpp
zander-xyz: I think that's probably okay here too - the explanation should be clear. And if it's not, we should change it.

PR #200, opened 2023-07-28
:recycle: Remove deleted_storage from ValueState

tongzhi1998:
- By `EVMC` convention, if we set `value = NULL`, then it is marked as deleted. Therefore, we don't really need `deleted_storage`
- Fix 1 incorrect test case for x-y-z type storage value change
- Add 1 additional test case to cover a previous existing bug

### DIFF
--- a/include/monad/state/value_state.hpp
+++ b/include/monad/state/value_state.hpp
@@ -23,11 +23,6 @@ struct ValueState
     struct InnerStorage
     {
         std::unordered_map<address_t, key_value_map_t> storage_{};
-        std::unordered_map<
-            address_t,
-            std::unordered_set<
-                deleted_key, deleted_key::hash, deleted_key::equality>>
-            deleted_storage_{};
 
         bool
         contains_key(address_t const &a, bytes32_t const &key) const noexcept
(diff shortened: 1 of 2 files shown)


include/monad/state/value_state.hpp
```diff
@@ -159,29 +100,14 @@ struct ValueState
...
         for (auto &[addr, acct_storage] : diffs.touched_.storage_) {
             if (!merged_.storage_.contains(addr)) {
                 merged_.storage_.emplace(addr, std::move(acct_storage));
                 continue;
             }
 
             for (auto const &[key, value] : acct_storage) {
-                assert(value != bytes32_t{});
-                merged_.storage_.at(addr).at(key).updated = value.updated;
+                merged_.storage_[addr][key] = value.updated;
```
sheapwilliams, line 110: Using `operator[]` instead of `at(...)` here means that this can default create, and that is *really bad* for merging.  Before changing any `at()` to `[]`, we should talk about it to make certain.

include/monad/state/value_state.hpp
```diff
@@ -218,30 +141,34 @@ struct ValueState<TValueDB>::WorkingCopy : public ValueState<TValueDB>
...
                 if (merged_value == touched_.storage_.at(a).at(key)) {
-                    remove_touched_key(a, key);
+                    touched_.storage_.at(a).at(key).updated = bytes32_t{};
                     return EVMC_STORAGE_DELETED;
                 }
-                else {
-                    remove_touched_key(a, key);
-                    touched_.deleted_storage_[a].insert(
-                        deleted_key{db_.at(a, key), key});
+                else if (touched_.storage_[a][key] != bytes32_t{}) {
```
sheapwilliams: Same here especially...

include/monad/state/value_state.hpp
```diff
@@ -107,7 +107,7 @@ struct ValueState
             }
 
             for (auto const &[key, value] : acct_storage) {
-                merged_.storage_.at(addr).at(key).updated = value.updated;
+                merged_.storage_[addr][key] = value.updated;
```
sheapwilliams, line 110: Regarding the last commit - is there a test case that we are missing?

PR #202, opened 2023-07-28
:bug: Fix self-transfer

tongzhi1998:
Problem:
- When an account transfers balance to itself, our code incorrectly set the balance

Solution:
- Change the sequence of operation such that we set sender's balance before getting receiver's balance

### DIFF
--- a/include/monad/execution/evm.hpp
+++ b/include/monad/execution/evm.hpp
@@ -81,8 +81,8 @@ struct Evm
         auto const value = intx::be::load<uint256_t>(m.value);
         auto const from_balance =
             intx::be::load<uint256_t>(s.get_balance(m.sender));
-        auto const to_balance = intx::be::load<uint256_t>(s.get_balance(to));
         s.set_balance(m.sender, from_balance - value);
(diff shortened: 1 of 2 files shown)


include/monad/execution/evm.hpp
```diff
@@ -78,6 +78,8 @@ struct Evm
     static void transfer_balances(
         TState &s, evmc_message const &m, address_t const &to) noexcept
     {
+        if (m.sender == to)
+            return;
         auto const value = intx::be::load<uint256_t>(m.value);
         auto const from_balance =
             intx::be::load<uint256_t>(s.get_balance(m.sender));
```
sheapwilliams: I think we should try to avoid the branch and generalize it ~ like so:
```suggestion
        auto const value = intx::be::load<uint256_t>(m.value);
        auto const from_balance =
            intx::be::load<uint256_t>(s.get_balance(m.sender));
        s.set_balance(m.sender, from_balance - value);
        auto const to_balance = intx::be::load<uint256_t>(s.get_balance(to));
        s.set_balance(to, to_balance + value);
```

Thoughts?  I don't imagine self-transfers are common whatsoever, so it doesn't feel like we should special case them to me.

PR #204, opened 2023-07-31
:new: Integrating database restart to replay_ethereum

tongzhi1998:
- Add database restart support for `replay_ethereum`
- Refactor behave tests to adapt to the new command line arguments
- Add different logger instantiation in `replay_ethereum`

### DIFF
--- a/test/features/ethereum/configure_logger.feature
+++ b/test/features/ethereum/configure_logger.feature
@@ -1,8 +1,9 @@
 Feature: Configure Logger
 
-    Background: BlockDb path setup
+    Background: Db path setup
         Given I run with BlockDb path = "erigon/block_db"
-        And I run with start block number = "46147"
+        And I run with StateDb path = "RocksTrieDB"
+        And I run with inferred start block number = "46147"
         And I run with finish block number = "46148"
 
     Scenario: Using default log level
--- a/test/features/ethereum/read_block_history.feature
+++ b/test/features/ethereum/read_block_history.feature
@@ -2,15 +2,16 @@ Feature: Read Ethereum Block History
 
   Background: BlockDb path setup
     Given I run with BlockDb path = "erigon/block_db"
+    And I run with StateDb path = "RocksTrieDB"
 
   Scenario: Run replay_ethereum with invalid end block number
-    Given I run with start block number = "1"
-    And I run with finish block number = "1"
+    Given I run with inferred start block number = "2"
+    And I run with finish block number = "2"
     When I start "replay_ethereum"
     Then the output should not contain "block_logger"
 
   Scenario: Run replay_ethereum with correct configuration, 1 Block, no hash validation
-    Given I run with start block number = "1"
+    Given I run with inferred start block number = "1"
     And I run with finish block number = "2"
     When I start "replay_ethereum"
     Then the output should contain "1" "Computed Transaction Root"
@@ -21,8 +22,8 @@ Feature: Read Ethereum Block History
     And the output should contain "1" "Expected State Root"
 
   Scenario: Run replay_ethereum with correct configuration, 1000 Blocks, no hash validation
-    Given I run with start block number = "10"
-    And I run with finish block number = "1010"
+    Given I run with inferred start block number = "0"
+    And I run with finish block number = "1000"
     When I start "replay_ethereum"
     Then the output should contain "1000" "Computed Transaction Root"
     And the output should contain "1000" "Expected Transaction Root"
--- a/test/features/ethereum/run_ethereum_blocks.feature
+++ b/test/features/ethereum/run_ethereum_blocks.feature
@@ -1,11 +1,12 @@
 Feature: Run Ethereum Block History
 
-    Background: BlockDb path setup
+    Background: DB path setup
         Given I run with BlockDb path = "erigon/block_db"
+        And I run with StateDb path = "RocksTrieDB"
 
     Scenario: Run ethereum block history for all initial blocks with no transaction
-        Given I run with start block number = "0"
-        And I run with finish block number = "46147"
+        Given I run with inferred start block number = "0"
+        Given I run with finish block number = "46147"
         And I run with "trie_db_logger" log level = "Critical"
         When I start "replay_ethereum"
         Then the "State Root" should match
--- a/test/features/ethereum/steps/read_block_history.py
+++ b/test/features/ethereum/steps/read_block_history.py
@@ -11,9 +11,27 @@ def given_block_db_path(context, block_db_path):
     context.block_db_path = block_db_path
 
 
-@given('I run with start block number = "{start_block_number}"')
+@given('I run with StateDb path = "{state_db_path}"')
+def given_block_db_path(context, state_db_path):
+    context.state_db_path = state_db_path
+    dir_path = os.getcwd() + "/test/features/ethereum/" + state_db_path
+    if os.path.exists(dir_path):
+        subprocess.run(["rm", "-rf", dir_path])
+    subprocess.run(["mkdir", dir_path])
+
+
+@given('I run with inferred start block number = "{start_block_number}"')
 def given_start_block_number(context, start_block_number):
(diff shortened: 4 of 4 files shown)


src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -73,13 +73,16 @@ int main(int argc, char *argv[])
         monad::log::logger_t::create_logger("txn_logger");
     [[maybe_unused]] auto *state_logger =
         monad::log::logger_t::create_logger("state_logger");
+    [[maybe_unused]] auto *change_set_logger =
+        monad::log::logger_t::create_logger("change_set_logger");
```
tongzhi1998, line 78: Sorry @rgarc I told you yesterday that your changes shouldn't impact replay_ethereum. Turns out it isn't the case. So I fix it here. Thanks for bringing it up!

rgarc: Yeah sorry about that. Just a heads up that you might have to do something similar again if this gets merged.
https://github.com/monad-crypto/monad/blob/b9ea5a237560bad485feb5e00e271e181e51765c/include/monad/execution/evmone_baseline_interpreter.hpp#L36-L38

tongzhi1998: Cool!

include/monad/db/prepare_state.hpp
```diff
@@ -69,4 +69,53 @@ prepare_state(TDB const &db, uint64_t block_number)
...
+        if (child_path.stem().string() != "CURRENT") {
+            sub_dir_cnt++;
+            latest_block_number = std::max(
+                latest_block_number, std::stoul(child_path.stem().string()));
+        }
+    }
+
+    if (sub_dir_cnt == 0u) {
+        return 0u;
+    }
```
zander-xyz: The logic you added here to determine the block number seems better in it's own free standing function `auto_detect_starting_block_number`. Then in the `RocksTrieDB` constructor, you can directly call into the constructor specifying the block number.

Side effect here is that there's less duplicated code.

include/monad/db/prepare_state.hpp
```diff
@@ -69,4 +69,53 @@ prepare_state(TDB const &db, uint64_t block_number)
...
+    if (fs::exists(current_dir)) {
+        fs::remove_all(current_dir);
+    }
+
+    fs::create_directories(current_dir);
+
+    auto const path = current_dir / monad::db::as_string<TDB>();
+
+    uint64_t latest_block_number = 0u;
+    uint64_t sub_dir_cnt = 0u;
```
zander-xyz: After we break this logic out into a separate function, just return `latest_block_number` directly rather than maintaining `sub_dir_cnt`

include/monad/db/prepare_state.hpp
```diff
@@ -69,4 +69,53 @@ prepare_state(TDB const &db, uint64_t block_number)
...
+    for (auto const &entry : std::filesystem::directory_iterator(db.root)) {
+        std::filesystem::path child_path =
+            entry.path().lexically_relative(db.root);
```
zander-xyz: Why was the `lexically_relative` part needed? Also, this can be `auto const`

include/monad/db/prepare_state.hpp
```diff
@@ -69,4 +69,53 @@ prepare_state(TDB const &db, uint64_t block_number)
...
+        if (child_path.stem().string() != "CURRENT") {
```
zander-xyz: Now that we have "CURRENT" used in multiple places, I think it's probably appropriate to include it at the top of this file like

```cpp
namespace detail
{
    constexpr auto CURRENT_DATABASE = "CURRENT"sv;
}
```

include/monad/db/rocks_trie_db.hpp
```diff
@@ -126,6 +128,55 @@ namespace detail
...
+
+                MONAD_ROCKS_ASSERT(s);
+                MONAD_ASSERT(cfds.size() == cfs.size());
+
+                return db;
+            }())
+            , accounts_trie(db, cfs[1], cfs[2])
+            , storage_trie(db, cfs[3], cfs[4])
+        {
+        }
```
zander-xyz: Instead of doing this, I would encourage you to do something like
```cpp
        RocksTrieDB(std::filesystem::path root, uint64_t block_history_size)
            : RocksTrieDB(root, auto_detect_starting_block_number(*this), block_history_size) {}
```
This reduces the amount of duplicate code that we have and need to maintain.

include/monad/db/rocks_trie_db.hpp
```diff
@@ -165,6 +216,11 @@ namespace detail
             MONAD_DEBUG_ASSERT(s.has_value());
         }
 
+        constexpr uint64_t get_start_block_number() const
+        {
+            return start_block_number;
+        }
```
zander-xyz: We eventually want to do work to add in block number into the DBInterface. That'll be a whole separate PR that I think we should not pollute this PR with. For now, can we rename this to `block_number` which returns `start_block_number`? Also it would be nice to add a TODO noting that this needs to be fixed when we add in wider `block_number` support.

include/monad/db/rocks_trie_db.hpp
```diff
@@ -165,6 +216,11 @@ namespace detail
             MONAD_DEBUG_ASSERT(s.has_value());
         }
 
+        constexpr uint64_t get_start_block_number() const
```
zander-xyz: `[[nodiscard]]` and `noexcept` please. Also, can you please move this function above the `DBInterface implementations` comment, since this isn't that?

include/monad/db/prepare_state.hpp
tongzhi1998: We actually don't need lexically_relative once we have .stem(). Fixed it!

include/monad/db/prepare_state.hpp
```diff
@@ -10,6 +10,8 @@
 #include <concepts>
 #include <filesystem>
 
+#include <iostream>
+
```
zander-xyz: why is this needed?

src/monad/db/test/db.cpp
```diff
@@ -166,21 +168,70 @@ TEST(RocksTrieDB, block_history)
...
+
+        EXPECT_EQ(
+            db.root_hash(a),
+            0x3f9802e4f21fce3d2b07d21c8f2b60b22f7c745c455e752728030580177f8e11_bytes32);
+        EXPECT_EQ(
+            db.root_hash(),
+            0x3f7578fb3acc297f8847c7885717733b268cb52dc6b8e5a68aff31c254b6b5b3_bytes32);
+    }
+
+    auto const root_for_second_constructor = test::make_db_root(
```
zander-xyz: Sorry, can you make this a separate unit test instead of appending onto the existing one?

src/monad/db/test/db.cpp
```diff
@@ -166,21 +168,70 @@ TEST(RocksTrieDB, block_history)
...
+            .account_changes = {{a, acct}},
+            .storage_changes = {{a, {{key1, value1}, {key2, value2}}}}});
+        db.create_and_prune_block_history(block_number++);
+
+        EXPECT_EQ(
+            db.root_hash(a),
+            0x3f9802e4f21fce3d2b07d21c8f2b60b22f7c745c455e752728030580177f8e11_bytes32);
+        EXPECT_EQ(
+            db.root_hash(),
+            0x3f7578fb3acc297f8847c7885717733b268cb52dc6b8e5a68aff31c254b6b5b3_bytes32);
```
zander-xyz, line 212: we also have `block_number()` so should we test that as well? ie. `db.block_number() == <my_exected_block_number>`

include/monad/db/prepare_state.hpp
tongzhi1998: I was using it to debug but forget to delete. Sorry about that.

PR #206, opened 2023-07-31
:sparkles: Add state_history client

zander-xyz:
Problem:
- Rpc server needs to query the database when servicing requests to its various endpoints

Solution:
- Create an interface which the rpc server may use to interact with the database from a separate process

This is by no means a complete interface, but should start to shape how the RPC server interacts with execution's persistent store.

Implementations and corresponding unit tests to arrive shortly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,3 +151,6 @@ dmypy.json
 
 # vcpkg
 vcpkg_installed/
+
+# rust
+Cargo.lock
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,8 @@ add_subdirectory(third_party/evmone)
 add_subdirectory(third_party/intx)
 add_subdirectory(third_party/silkpre)
 
+add_subdirectory(state_history)
+
 function(monad_compile_options target)
     set_property(TARGET ${target} PROPERTY C_STANDARD 11)
     set_property(TARGET ${target} PROPERTY C_STANDARD_REQUIRED ON)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,6 +2,8 @@
 
 FROM fedora:38
 
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+
 RUN yum -y install \
   dnf-plugins-core
 
--- a/scripts/cargo/clean_build_and_run.sh
+++ b/scripts/cargo/clean_build_and_run.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -Eeuo pipefail
+
+source "$HOME/.cargo/env"
+
+cargo clean
+cargo build
+cargo run
--- a/state_history/CMakeLists.txt
+++ b/state_history/CMakeLists.txt
@@ -1,0 +1,20 @@
+configure_file("build.rs.in" "${CMAKE_CURRENT_BINARY_DIR}/build.rs")
+
+add_custom_target(
+    copy_state_history_crate ALL
+    COMMENT "Copy state_history"
+    COMMAND cp -r "${PROJECT_SOURCE_DIR}/state_history"
+            "${CMAKE_CURRENT_BINARY_DIR}"
+    COMMAND ln -sf "${CMAKE_CURRENT_BINARY_DIR}/build.rs"
+            "${CMAKE_CURRENT_BINARY_DIR}/state_history/build.rs"
+    USES_TERMINAL VERBATIM
+)
+
+add_custom_target(
+    state_history_crate ALL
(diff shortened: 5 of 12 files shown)


librpc/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+
+        fn make_client(root: &CxxString) -> Result<UniquePtr<Client>>;
+        fn make_client_at_block_number(
+            root: &CxxString,
+            block_number: u64,
+        ) -> Result<UniquePtr<Client>>;
+
+        fn get_account(self: &Client, address: [u8; 20]) -> Result<UniquePtr<Account>>;
+        fn get_code(self: &Client, address: [u8; 20]) -> Result<&CxxString>;
+        fn get_storage_hash(self: &Client, address: [u8; 20]) -> Result<[u8; 32]>;
```
zander-xyz: If anyone knows the syntax for typedefs here, would appreciate a comment. Rust did not seem happy with my
```rust
type Bytes32 = [u8; 32]
```

librpc/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "librpc"
+version = "0.0.0"
+authors = ["David Tolnay <dtolnay@gmail.com>"]
+description = "Toy project from https://github.com/dtolnay/cxx"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+publish = false
+repository = "https://github.com/dtolnay/cxx"
```
sheapwilliams: This all certainly can't be right...shouldn't this be for the new library?

librpc/include/account.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "bytes.hpp"
+
+#include <monad/core/account.hpp>
+
+#include <array>
+#include <cstdint>
+
+namespace org::librpc
```
sheapwilliams: `org`?  That doesn't seem right...

Also, librpc is the use case, not what this is.  This is just a rust library for state history, right?  `libstate`?  Something else that isn't what it is used for?

librpc/include/client.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <filesystem>
+#include <memory>
+
+namespace org::librpc
+{
+    struct Client
+    {
+        using Address = std::array<uint8_t, sizeof(monad::address_t)>;
+
+        std::filesystem::path const root;
```
sheapwilliams: [question] I see this doesn't do anything yet, but do you know how are you going to manage concurrent accesses with a filesystem API that isn't concurrency safe?  Is the idea that the program will have 1 client, and multiple calls to getters from it?  Something else?

Feel free to fill me in over zoom if that is easier.

librpc/Cargo.toml
zander-xyz: Yeah deleted!

librpc/include/client.hpp
zander-xyz: RocksDB can be simultaneously opened by any number of accessors, as long as all participants are read-only. Last time I heard, we were planning on adding support for TrieDB as well.

librpc/include/account.hpp
zander-xyz: this account struct exists purely for librpc? yeah, I think org is the wrong thing...artifact of some online examples

sheapwilliams: I guess what I meant, is that I expect something called `librpc`, to do...well, remote procedure calls.  Right?

I try to name things what they are, not what they are used for, as that often changes.  This seems like a state history library, so maybe we should name it something like that?

zander-xyz: This type (as well as Bytes32) only exist because of the limitations of types we are allowed to pass across the FFI boundary. Would you be happy with libffi?

sheapwilliams: 😄 That is only slightly less bad?  It is still concerning itself with the details of what is happening too much.  Consider the usage:

```
... <in an rpc service handler>

librpc::db{...};
db.get_balance(...);
...
```

Same thing with ffi...the namespace isn't helpful as to what it is even doing - it looks like it is making an rpc call, or something else. Compare:

```
... <in an rpc service handler>

monad::state_history::db{...};
db.get_balance(...);
...
```

That is how I'm used to reading things - it says what it is, not the details about why it exists.  'Why' is most obvious from the context (servicing an rpc call, etc)...does that make sense?

Just my 2c.  I'd consider using the above namespace and library name, or some derivation therein.

librpc/src/lib.rs
michael-yxchen, line 1: I'm happy with the rust interface.

PR #207, opened 2023-08-01
:bug: Fix db to execute with executor

zander-xyz:
Problem:
- Previously the executor was never invoked unless the base class implementation was explicitly invoked

Solution:
- Always execute queries using the executor, and enforce this via unit tests

### DIFF
--- a/test/unit/common/include/monad/test/hijacked_db.hpp
+++ b/test/unit/common/include/monad/test/hijacked_db.hpp
@@ -1,0 +1,41 @@
+#pragma once
+
+#include <monad/db/config.hpp>
+#include <monad/db/in_memory_db.hpp>
+#include <monad/db/in_memory_trie_db.hpp>
+#include <monad/db/rocks_db.hpp>
+#include <monad/db/rocks_trie_db.hpp>
+#include <monad/test/config.hpp>
+
+MONAD_TEST_NAMESPACE_BEGIN
+
+namespace hijacked
+{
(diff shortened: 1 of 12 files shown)


include/monad/execution/execution_model.hpp
```diff
@@ -14,7 +14,7 @@ struct BoostFiberExecution
     using fiber_t = boost::fibers::fiber;
     static void yield() noexcept { boost::this_fiber::yield(); }
 
-    [[nodiscard]] static auto execute(std::invocable auto &&f)
+    [[nodiscard]] auto static execute(std::invocable auto &&f)
```
sheapwilliams: ❓ Is this really the way clang-format does this?  Or are you using a specific tidy config?

test/unit/common/include/monad/test/hijacked_db.hpp
sheapwilliams, line 1: :+1:

include/monad/execution/execution_model.hpp
zander-xyz: Nope, I just messed up the ordering. Seemingly clang-format has no opinion on the ordering. Fixed.

sheapwilliams: It seems that is can have a preference, if we want it to...

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifierorder

PR #208, opened 2023-08-01
:loud_sound: Enable VM tracing on debug mode.

rgarc:
Problem:
- The evmone interpreter ships with instruction-level tracing that is disabled by default (presumably for performance reasons).
- Having this tracing facilitates debugging the execution of transactions.

Solution:
- Enable VM tracing by calling `add_tracing` with an instruction tracer conditioned on whether or not the program was compiled in debug mode.

Note:
- See [evmone instruction tracer implementation](https://github.com/ethereum/evmone/blob/eda0bebe6b054ea575d0716caa1e2906bcf38068/lib/evmone/tracing.cpp#L69-L152)

### DIFF
--- a/include/monad/execution/evmone_baseline_interpreter.hpp
+++ b/include/monad/execution/evmone_baseline_interpreter.hpp
@@ -4,6 +4,7 @@
 #include <monad/core/concepts.hpp>
 
 #include <monad/execution/config.hpp>
+#include <monad/logging/monad_log.hpp>
 
 #include <evmone/baseline.hpp>
 
@@ -18,6 +19,11 @@
 
 #include <evmone/vm.hpp>
 
+#ifdef EVMONE_TRACING
+    #include <evmone/tracing.hpp>
+    #include <sstream>
+#endif
+
 MONAD_EXECUTION_NAMESPACE_BEGIN
 
 template <class TState, concepts::fork_traits<TState> TTraits>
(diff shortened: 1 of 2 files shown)


include/monad/execution/evmone_baseline_interpreter.hpp
```diff
@@ -27,23 +30,32 @@ struct EVMOneBaselineInterpreter
...
+                "evmone_baseline_interpreter_logger");
+        evmc::Result result{
+            evmc_result{.status_code = EVMC_SUCCESS, .gas_left = m.gas}};
         auto const code = s.get_code(m.code_address);
         if (code.empty()) {
             return result;
         }
 
         evmone::VM v{};
+#ifndef NDEBUG
```
sheapwilliams: I really like the functionality.  Can we not directly couple it to `NDEBUG` tho?  Perhaps a new macro (`EVMONE_TRACING`, or something), and just default set that to true in debug, and false otherwise?

I think you have the most common use case here (tracing enabled while in debug mode), but it would be nice to turn it off if needed.

include/monad/execution/evmone_baseline_interpreter.hpp
```diff
@@ -16,7 +17,9 @@
     #pragma GCC diagnostic pop
 #endif
 
+#include <evmone/tracing.hpp>
 #include <evmone/vm.hpp>
+#include <sstream>
```
sheapwilliams: ...and if you want to be really kind to compilation times, you can conditionally include these headers based on the same macro.

include/monad/execution/evmone_baseline_interpreter.hpp
rgarc: Thanks. I also had this question when writing this and you read my mind.

PR #216, opened 2023-08-03
:white_check_mark: Fix setting and clearing storage in same commit.

rgarc:
:bug: Fix handling of storage keys in ValueState
    
Problem:
- If a storage slot has not been previously commited and is set and then cleared within the same change set, we should no longer consider that slot touched when going to commit those changes to the database.

Solution:
- Replace the setting of the update state of the storage key with the removal of the touched key with `remove_touched_key`.
- Added assertion to check the original state of the storage key.
- Add a test case that sets and clears storage in the same commit.

### DIFF
--- a/include/monad/state/value_state.hpp
+++ b/include/monad/state/value_state.hpp
@@ -21,15 +21,13 @@ struct InnerStorage
 
     std::unordered_map<address_t, key_value_map_t> storage_{};
 
-    bool
-    contains_key(address_t const &a, bytes32_t const &key) const noexcept
+    bool contains_key(address_t const &a, bytes32_t const &key) const noexcept
     {
         return storage_.contains(a) && storage_.at(a).contains(key);
     }
 
     void clear() { storage_.clear(); }
 };
-
 
 template <class TValueDB>
 struct ValueState
@@ -168,11 +166,12 @@ struct ValueState<TValueDB>::ChangeSet : public ValueState<TValueDB>
         }
 
         if (touched_.contains_key(a, key)) {
-            if (touched_.storage_.at(a).at(key).updated != bytes32_t{}) {
-                touched_.storage_.at(a).at(key).updated = bytes32_t{};
-                return EVMC_STORAGE_ADDED_DELETED;
-            }
-            return EVMC_STORAGE_ASSIGNED;
+            MONAD_DEBUG_ASSERT(
+                touched_.storage_.at(a).at(key).orig == bytes32_t{});
+            MONAD_DEBUG_ASSERT(
+                touched_.storage_.at(a).at(key).updated != bytes32_t{});
+            remove_touched_key(a, key);
+            return EVMC_STORAGE_ADDED_DELETED;
         }
         return EVMC_STORAGE_ASSIGNED;
     }
--- a/src/monad/state/test/state.cpp
+++ b/src/monad/state/test/state.cpp
@@ -470,6 +470,27 @@ TYPED_TEST(TrieDBTest, commit_storage_and_account_together_regression)
     t.commit();
 }
 
+TYPED_TEST(TrieDBTest, set_and_then_clear_storage_in_same_commit)
+{
+    using namespace intx;
+    auto db = test::make_db<TypeParam>();
+    AccountState accounts{db};
+    ValueState values{db};
+    code_db_t code_db{};
+    CodeState code{code_db};
+    State t{accounts, values, code, block_cache, db};
+
+    auto changeset = t.get_new_changeset(0u);
+    using namespace intx;
+    changeset.create_account(a);
+    EXPECT_EQ(changeset.set_storage(a, key1, value1), EVMC_STORAGE_ADDED);
+    EXPECT_EQ(changeset.set_storage(a, key1, null), EVMC_STORAGE_ADDED_DELETED);
+    t.merge_changes(changeset);
+    t.commit();
+
+    EXPECT_EQ(db.try_find(a, key1), monad::bytes32_t{});
+}
+
 TYPED_TEST(StateTest, commit_twice)
 {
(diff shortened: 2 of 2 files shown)


src/monad/state/test/state.cpp
```diff
@@ -466,6 +467,39 @@ TYPED_TEST(TrieDBTest, commit_storage_and_account_together_regression)
     t.commit();
 }
 
+TYPED_TEST(TrieDBTest, set_and_then_clear_storage_in_same_commit)
```
sheapwilliams, line 473: Not a `TrieDBTest`. I think this should just be a test in value_state.cpp?  Why did you add it here?

include/monad/state/value_state.hpp
```diff
@@ -166,7 +166,8 @@ struct ValueState<TValueDB>::ChangeSet : public ValueState<TValueDB>
 
         if (touched_.contains_key(a, key)) {
             if (touched_.storage_.at(a).at(key).updated != bytes32_t{}) {
-                touched_.storage_.at(a).at(key).updated = bytes32_t{};
+                assert(touched_.storage_.at(a).at(key).orig == bytes32_t{});
+                remove_touched_key(a, key);
```
sheapwilliams: Come to think of it, I'm not sure this is correct - I think we will return the wrong statuses here, which is also why we should probably add those assertions to the test below instead of `(void)`'ing them.

Once accessed, we need to keep the value, and not delete it from the changeset.  Let's chat tomorrow...

sheapwilliams: Ok, I've thought about it, and I think the statuses are just fine.  From delete, to something, then to delete, the statues start over, so removal is just fine.  Sorry for the chaff.

But I still think the test should be value_state.cpp, not state.cpp.  When refactoring the value_state.hpp code we should have the set of requirements in that unit test, not somewhere else.

zander-xyz: Also, the more I think about this the more I think that both of these lines should be brought one scope higher, right below line 167.

If there's nothing in merged, but touched contains the key, then the orig better be null, and the updated better be non-null. If we reach line 175, then I think we've hit a bug.

zander-xyz: So something like
```cpp
        if (touched_.contains_key(a, key)) {
            MONAD_DEBUG_ASSERT(touched_.storage_.at(a).at(key).orig == bytes32_t{});
            MONAD_DEBUG_ASSERT(touched_.storage.at(a).at(key).updated != bytes32_t{});
            remove_touched_key(a, key);
            return EVMC_STORAGE_ADDED_DELETED;
        }
```

zander-xyz: We can also further protect ourselves from bugs like this and general inefficiencies if, in `can_commit`, we also check that `orig != updated`.

src/monad/state/test/state.cpp
rgarc: Thanks! That is a good point. The test I started with was much larger and encompassed all of `account_state`, `code_state`, and `value_state`. However, now that I've reduced all the way, it is clearly just confined to `value_state`, so I will move it there.

rgarc: Wait on second thought. This test originally failing relies on an assertion that is hit inside the trie db implementation while calling `State::commit`

```
#4  0x00007ffff6858147 in __assert_fail () from /lib64/libc.so.6
#10 0x000000000075f482 in monad::trie::Trie<...>::process_updates(std::vector<std::variant<monad::trie::Upsert, monad::trie::Delete>, std::allocator<std::variant<monad::trie::Upsert, monad::trie::Delete> > > const&)::{lambda()#1}::operator()[abi:cxx11]() const (__closure=0x7fffffffc800) at /space/ssd_4tb_2/rgarcia/tmp/tmp.UxPSJ2OnPn/include/monad/trie/trie.hpp:106
#11 0x00000000007602bd in monad::trie::Trie<...>::process_updates (this=0x7fffffffdcf0, updates=std::vector of length 1, capacity 1 = {...}) at /space/ssd_4tb_2/rgarcia/tmp/tmp.UxPSJ2OnPn/include/monad/trie/trie.hpp:103
#12 0x0000000000726307 in monad::db::TrieDBInterface<...>::commit<monad::state::StateChanges> (this=0x7fffffffd720, obj=...) at /space/ssd_4tb_2/rgarcia/tmp/tmp.UxPSJ2OnPn/include/monad/db/trie_db_interface.hpp:189
#13 0x0000000000a35502 in monad::state::State<...>::commit (this=0x7fffffffd470) at /space/ssd_4tb_2/rgarcia/tmp/tmp.UxPSJ2OnPn/include/monad/state/state.hpp:287
#14 0x00000000009dbd00 in TrieDBTest_set_and_then_clear_storage_in_same_commit_Test<monad::db::detail::InMemoryTrieDB<monad::execution::SerialExecution> >::TestBody (this=0x602000004470) at /space/ssd_4tb_2/rgarcia/tmp/tmp.UxPSJ2OnPn/src/monad/state/test/state.cpp:494
```

I think the test needs to stay here or duplicated.

src/monad/state/test/state.cpp
```diff
@@ -466,6 +467,27 @@ TYPED_TEST(TrieDBTest, commit_storage_and_account_together_regression)
...
+    ValueState values{db};
+    code_db_t code_db{};
+    CodeState code{code_db};
+    State t{accounts, values, code, block_cache, db};
+
+    auto changeset = t.get_new_changeset(0u);
+    using namespace intx;
+    changeset.create_account(a);
+    (void)changeset.set_storage(a, key1, value1);
+    (void)changeset.set_storage(a, key1, null);
```
zander-xyz: Can we please check the return value here too?

PR #217, opened 2023-08-03
✨ Add compile-time enforced read-only database 

zander-xyz:
Problem:
- We would like to enforce that other processes (ie. RPC server) which
  depend on our persistent state access it in a read-only fashion.

Solution:
- Add in a read-only version of the database

### DIFF
--- a/include/monad/db/concepts.hpp
+++ b/include/monad/db/concepts.hpp
@@ -1,0 +1,25 @@
+#pragma once
+
+#include <monad/db/config.hpp>
+
+#include <concepts>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+struct ReadOnly
+{
+};
+struct ReadWrite
+{
+};
+
+template <typename T>
+concept Permission = std::same_as<T, ReadOnly> || std::same_as<T, ReadWrite>;
+
+template <typename T>
+concept Writable = std::same_as<T, ReadWrite>;
+
+template <typename T>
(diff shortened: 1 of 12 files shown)


include/monad/db/db_interface.hpp
```diff
@@ -4,21 +4,15 @@
...
+template <typename TDBImpl, typename TExecutor, Permission TPermission>
 struct DBInterface
 {
-    struct Updates
-    {
-        std::unordered_map<address_t, std::optional<Account>> accounts;
-        std::unordered_map<address_t, std::unordered_map<bytes32_t, bytes32_t>>
-            storage;
-    } updates{};
-
```
tongzhi1998, line 21: Thanks for removing this... I was about to remove it in my WIP PR too

include/monad/db/concepts.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <monad/db/config.hpp>
+
+#include <concepts>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+struct ReadOnly
```
tongzhi1998, line 9: Sorry I am a bit confused on this... How do you actually enforce `ReadOnly` then? To me it looks like user can declare a `ReadOnlyRocksDB` and write to it?

tongzhi1998: Okay I zoomed with Alex and now understand.

PR #219, opened 2023-08-03
:recycle: Refactor CodeState to take code_hash as key

tongzhi1998:


include/monad/state/code_state.hpp
```diff
@@ -4,6 +4,7 @@
 #include <monad/core/address.hpp>
 #include <monad/core/assert.h>
 #include <monad/core/byte_string.hpp>
+#include <monad/core/bytes.hpp>
```
sheapwilliams: Probably don't need address anymore...

include/monad/state/code_state.hpp
```diff
@@ -89,34 +90,34 @@ struct CodeState<TCodeDB>::ChangeSet : public CodeState<TCodeDB>
...
         }
-        auto const &[_, inserted] = code_.emplace(a, code);
+        auto const &[_, inserted] = code_.emplace(b, code);
         MONAD_DEBUG_ASSERT(inserted);
     }
 
-    // EVMC Host Interface
-    [[nodiscard]] size_t get_code_size(address_t const &a) const noexcept
+    // EVMC Host Interface (Part of)
+    [[nodiscard]] size_t get_code_size(bytes32_t const &b) const noexcept
```
sheapwilliams: "Part of" is implied everywhere, I don't think we need to add this here.

include/monad/state/state.hpp
```diff
@@ -162,28 +162,28 @@ struct State
...
-            return code_.copy_code(a, offset, buffer, size);
+            return code_.copy_code(get_code_hash(a), offset, buffer, size);
         }
 
         [[nodiscard]] byte_string_view
         get_code(address_t const &a) const noexcept
         {
-            return code_.code_at(a);
+            return code_.code_at(get_code_hash(a));
         }
```
sheapwilliams: I think we should change this function as well to take a hash parameter.  Let the caller deal with it, I think.

src/monad/state/test/state.cpp
```diff
@@ -116,16 +116,16 @@ TYPED_TEST(StateTest, get_code)
...
         .storage_changes = {}});
 
-    [[maybe_unused]] auto bs = as.get_new_changeset(0);
-
+    auto bs = as.get_new_changeset(0);
     bs.access_account(a);
-    auto const c = bs.get_code(a);
+    bs.set_code(a, contract);
 
+    auto const c = bs.get_code(a);
```
sheapwilliams: This test is named "get_code", meaning it is just testing that one function.  I don't think we should change the intention of the test to be setting an account in a changes set, but that was previously existing.

In fact, I think we should probably look to disallow this sort of thing - it is clearly a usage bug.

test/integration/evm_state_host/evm_state_host.cpp
```diff
@@ -74,7 +74,6 @@ TEST(EvmInterpStateHost, return_existing_storage)
         0x60, // PUSH1, 3 gas
         0x1f, // offset
         0xf3}; // RETURN
-    code_db.emplace(a, code);
```
sheapwilliams, line 77: Same as above - let's leave this test as it was, just fixed up for the API changes.

src/monad/state/test/state.cpp
```diff
@@ -116,16 +116,16 @@ TYPED_TEST(StateTest, get_code)
     CodeState code{code_db};
     State as{accounts, values, code, block_cache, db};
     byte_string const contract{0x60, 0x34, 0x00};
-    code_db.emplace(a, contract);
+
     db.commit(StateChanges{
         .account_changes = {{a, Account{.balance = 10'000}}},
```
sheapwilliams: BTW - this code is hash-specific agnostic, meaning it doesn't care about what hash (of even if a hash is used).  You can use any `bytes32_t` value you want.

include/monad/state/state.hpp
tongzhi1998: I am not sure if this is the best idea. It seems that every other function is `state.hpp` is taking address as parameter (and `EVMC` interface requires it anyways). And if we change this function, we probably need to change `set_code()` to take the hash too since it is also not `EVMC` interface.

src/monad/state/test/state.cpp
tongzhi1998: Hm... How exactly are we going to disallow this sort of behavior? Are we going to make it a private function and add some friend to other classes that access it?

include/monad/state/state.hpp
zander-xyz: The function `get_code_hash` eventually calls into `get_committed_storage` which could potentially induce a database read (which is expensive and something we want to avoid). It'd be nice if this function could take in the code hash directly and let the caller deal with saving the code hash when appropriate. Yes, it's true that every other function takes address as parameter, and consistency is nice, but in this case consistency is costing us performance.

Now if this function were part of the EVMC Host Interface we'd be forced to abide by their function signature.

tongzhi1998: Cool, I think I understand why we need this. I just realized that after this change, the `State's` interface changes so the old fake `State` doesn't work anymore (and even if it works, we should probably change its internal DS to be the same as real `State` (or rather, `CodeState`)'s ). Therefore I made a bunch of other changes... Sorry for the additional stuff that you guys need to review.

src/monad/state/test/state.cpp
zander-xyz: Assertions, probably

include/monad/execution/test/fakes.hpp
```diff
@@ -151,14 +154,23 @@ namespace fake
...
+                auto const code_hash = std::bit_cast<const monad::bytes32_t>(
+                    ethash::keccak256(c.data(), c.size()));
+
+                _code.insert({code_hash, c});
+                if (!account_exists(a)) {
+                    _accounts.emplace(a, Account{.code_hash = code_hash});
+                }
+                else {
+                    _accounts.at(a).code_hash = code_hash;
+                }
```
zander-xyz, line 166: why are we handling account creation here? the real set_code doesn't care so why should we care here?

tongzhi1998: It's a bit complicated. There are a lot of test cases where we use `fake::State` and we call `set_code(address, code)` before `get_code_hash(address)`. However, if `set_code` doesn't handle account creation, `get_code_hash` will not be able to find the respective account, creating `std::map` `invalid_access` error. 
So why does the real `State` work? Because in the real `State` we are using a real db backend for account storage, instead of the `unordered_map` simplification, and together with the real `access_account` function, will ensure that `invalid_access` would never happen. 
Feel free to Zoom me if I fail to explain any part.

src/monad/state/test/state.cpp
tongzhi1998: Issue #224

include/monad/execution/test/fakes.hpp
zander-xyz: we don't call access_account when using fakes? I see the function exists, we just don't do anything

tongzhi1998: No we don't. Yeah another solution would be changing the `access_account` function in `fake::State` to create non-existing account. I am not sure if this is better than the current option since I think it has greater impact.

PR #225, opened 2023-08-07
:new:Integrate DB to CodeState

tongzhi1998:
Problem:
- Currently, our code base uses simple stl unordered_map for storing account's code.
- However, some rpc method will query account's code, so we need a persistant storage.

Solution:
- Adding all 4 DB support for CodeState

By the way, I need to give more thoughts on how we handle empty byte_string. The currently version should work but not very performant / strict.

### DIFF
--- a/include/monad/db/db_interface.hpp
+++ b/include/monad/db/db_interface.hpp
@@ -3,6 +3,7 @@
 #include <monad/core/account.hpp>
 #include <monad/core/address.hpp>
 #include <monad/core/assert.h>
+#include <monad/core/byte_string.hpp>
 #include <monad/core/bytes.hpp>
 #include <monad/db/concepts.hpp>
 #include <monad/db/config.hpp>
@@ -26,6 +27,7 @@ struct DBInterface
     // Accessors
     ////////////////////////////////////////////////////////////////////
 
+    // Account
     [[nodiscard]] constexpr std::optional<Account> try_find(address_t const &a)
         requires Readable<TPermission>
     {
@@ -48,6 +50,7 @@ struct DBInterface
         return ret.value();
     }
 
+    // Storage
     [[nodiscard]] constexpr bytes32_t
     try_find(address_t const &a, bytes32_t const &k)
         requires Readable<TPermission>
@@ -69,6 +72,29 @@ struct DBInterface
     {
         auto const ret = try_find(a, k);
         MONAD_ASSERT(ret != bytes32_t{});
+        return ret;
+    }
+
+    // Code
+    [[nodiscard]] constexpr byte_string try_find(bytes32_t const &ch)
+        requires Readable<TPermission>
+    {
+        return TExecutor::execute(
+            [=, this]() { return self().try_find_impl(ch); });
+    }
+
+    [[nodiscard]] constexpr bool contains(bytes32_t const &ch)
+        requires Readable<TPermission>
+    {
+        return TExecutor::execute(
+            [=, this]() { return self().contains_impl(ch); });
+    }
+
+    [[nodiscard]] constexpr byte_string at(bytes32_t const &ch)
+        requires Readable<TPermission>
+    {
+        auto const ret = try_find(ch);
+        MONAD_ASSERT(ret != byte_string{});
         return ret;
     }
 
--- a/include/monad/db/in_memory_db.hpp
+++ b/include/monad/db/in_memory_db.hpp
@@ -1,5 +1,10 @@
 #pragma once
 
+#include <monad/core/account.hpp>
+#include <monad/core/address.hpp>
+#include <monad/core/byte_string.hpp>
+#include <monad/core/bytes.hpp>
+
 #include <monad/db/config.hpp>
 #include <monad/db/db_interface.hpp>
 #include <monad/execution/execution_model.hpp>
@@ -20,6 +25,7 @@ namespace detail
         std::unordered_map<address_t, Account> accounts;
         std::unordered_map<address_t, std::unordered_map<bytes32_t, bytes32_t>>
             storage;
+        std::unordered_map<bytes32_t, byte_string> code;
 
         ////////////////////////////////////////////////////////////////////
         // DBInterface implementations
@@ -35,6 +41,12 @@ namespace detail
             requires Readable<TPermission>
         {
             return storage.contains(a) && storage.at(a).contains(k);
+        }
+
+        [[nodiscard]] bool contains_impl(bytes32_t const &ch)
+            requires Readable<TPermission>
+        {
+            return code.contains(ch);
         }
 
         [[nodiscard]] std::optional<Account> try_find_impl(address_t const &a)
@@ -56,6 +68,15 @@ namespace detail
             return storage.at(a).at(k);
         }
 
+        [[nodiscard]] byte_string try_find_impl(bytes32_t const &ch)
+            requires Readable<TPermission>
+        {
+            if (code.contains(ch)) {
+                return code.at(ch);
+            }
+            return byte_string{};
+        }
+
         void commit(state::changeset auto const &obj)
             requires Writable<TPermission>
         {
@@ -72,6 +93,10 @@ namespace detail
                 if (storage[a].empty()) {
                     storage.erase(a);
                 }
+            }
+
+            for (auto const &[ch, c] : obj.code_changes) {
+                code[ch] = c;
             }
 
             for (auto const &[a, acct] : obj.account_changes) {
--- a/include/monad/db/in_memory_trie_db.hpp
+++ b/include/monad/db/in_memory_trie_db.hpp
@@ -7,6 +7,9 @@
 #include <monad/trie/in_memory_writer.hpp>
 #include <monad/trie/trie.hpp>
 
+#include <monad/state/concepts.hpp>
+#include <monad/state/state_changes.hpp>
+
 MONAD_DB_NAMESPACE_BEGIN
 
 namespace detail
@@ -17,6 +20,9 @@ namespace detail
         : public TrieDBInterface<
               InMemoryTrieDB<TExecutor, TPermission>, TExecutor, TPermission>
     {
+        using this_t = InMemoryTrieDB<TExecutor, TPermission>;
+        using base_t = TrieDBInterface<this_t, TExecutor, TPermission>;
+
         template <typename TComparator>
         struct Trie
         {
@@ -55,10 +61,27 @@ namespace detail
 
         Trie<trie::InMemoryPathComparator> accounts_trie;
         Trie<trie::InMemoryPrefixPathComparator> storage_trie;
+        std::unordered_map<bytes32_t, byte_string> code;
 
         ////////////////////////////////////////////////////////////////////
         // DBInterface implementations
         ////////////////////////////////////////////////////////////////////
+
+        [[nodiscard]] bool contains_impl(bytes32_t const &ch)
+            requires Readable<TPermission>
+        {
+            return code.contains(ch);
+        }
+
+        [[nodiscard]] byte_string try_find_impl(bytes32_t const &ch)
+            requires Readable<TPermission>
+        {
+            if (code.contains(ch)) {
+                return code.at(ch);
+            }
+            return byte_string{};
+        }
+
         constexpr void
         create_and_prune_block_history(uint64_t /* block_number */) const
             requires Writable<TPermission>
@@ -73,6 +96,14 @@ namespace detail
         auto &storage() { return storage_trie; }
         auto const &accounts() const { return accounts_trie; }
         auto const &storage() const { return storage_trie; }
+
+        void commit(state::changeset auto const &obj)
+        {
+            for (auto const &[ch, c] : obj.code_changes) {
+                code[ch] = c;
+            }
+            base_t::commit(obj);
+        }
     };
 }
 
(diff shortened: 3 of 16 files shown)


include/monad/state/code_state.hpp
```diff
@@ -117,7 +121,7 @@ struct CodeState<TCodeDB>::ChangeSet : public CodeState<TCodeDB>
         auto const code = code_at(b);
         assert(code.size() > offset);
         auto const bytes_to_copy = std::min(code.size() - offset, buffer_size);
-        std::memcpy(buffer, code.begin() + offset, bytes_to_copy);
+        std::memcpy(buffer, &(*code.begin()) + offset, bytes_to_copy);
```
tongzhi1998: For some reason, without doing this deref-ref thing, the code won't compile. I paste the error below:
```
 no viable conversion from '__normal_iterator<const unsigned char *, basic_string<unsigned char>>' to 'const void *'
 std::memcpy(buffer, code.begin() + offset, bytes_to_copy);
```
I am not sure why this happens. Anyone has ideas?

include/monad/db/db_interface.hpp
```diff
@@ -72,6 +75,26 @@ struct DBInterface
...
+        return TExecutor::execute(
+            [=, this]() { return self().contains_impl(b); });
+    }
+
+    [[nodiscard]] constexpr byte_string at(bytes32_t const &b)
+    {
+        auto const ret = try_find(b);
+        MONAD_ASSERT(ret != byte_string{});
+        return ret;
+    }
```
zander-xyz: these all need `requires Readable<TPermission>`

ditto everywhere else

include/monad/db/in_memory_db.hpp
```diff
@@ -74,6 +93,12 @@ namespace detail
                 }
             }
 
+            for (auto const &[ch, c] : obj.code_changes) {
+                if (c != byte_string{}) {
+                    code[ch] = c;
+                }
```
zander-xyz: I think we should interpret an empty byte_string as deletion. ditto here and everywhere else

include/monad/db/rocks_db.hpp
```diff
@@ -173,6 +185,26 @@ namespace detail
...
 
+        [[nodiscard]] byte_string try_find_impl(bytes32_t const &b)
+        {
+            rocksdb::PinnableSlice value;
+            auto const res =
+                db->Get(rocksdb::ReadOptions{}, code_cf(), to_slice(b), &value);
+            if (res.IsNotFound()) {
+                return byte_string{};
+            }
+            MONAD_ASSERT(res.ok());
```
zander-xyz: why not `MONAD_ROCKS_ASSERT` `res` itself? That'll give us a more detailed message

include/monad/db/rocks_db.hpp
```diff
@@ -207,6 +239,16 @@ namespace detail
                 }
             }
 
+            for (auto const &[ch, c] : obj.code_changes) {
+                if (c != byte_string{}) {
+                    auto const res = batch.Put(
+                        code_cf(),
+                        to_slice(ch),
+                        to_slice(rlp::encode_string(c)));
+                    MONAD_ROCKS_ASSERT(res);
```
zander-xyz: I don't think there's a real need to rlp encode here. Just store `c` itself. Same comment in RocksTrieDB

include/monad/db/rocks_trie_db.hpp
```diff
@@ -80,6 +80,7 @@ namespace detail
         std::shared_ptr<rocksdb::DB> db;
         Trie accounts_trie;
         Trie storage_trie;
+        rocksdb::WriteBatch batch;
```
zander-xyz: Not a huge deal, but can we make this
```cpp
[[no_unique_address]] std::conditional_t<Writable<TPermission>, rocksdb::WriteBatch, Empty> batch;
```
consistent with `block_history_size`?

This makes it so that `batch` doesn't occupy any space in the class if the database is not declared writable.

include/monad/db/rocks_trie_db.hpp
```diff
@@ -165,7 +167,9 @@ namespace detail
             }())
             , accounts_trie(db, cfs[1], cfs[2])
             , storage_trie(db, cfs[3], cfs[4])
+            , batch()
             , block_history_size(block_history_size)
+
```
zander-xyz, line 188: did you clang-format this file?

include/monad/db/rocks_trie_db.hpp
```diff
@@ -217,9 +245,39 @@ namespace detail
...
+            byte_string ret{};
+            auto const rest = rlp::decode_string(
+                ret,
+                {reinterpret_cast<byte_string_view::value_type const *>(
+                     value.data()),
+                 value.size()});
+            MONAD_ASSERT(rest.empty());
+            return ret;
+        }
```
zander-xyz: Yeah...because of how we structured our class hierarchy, now there's a bunch of duplicated code across the four classes. For the in memory stuff, I think this is less significant because the duplicated code is actually quite small. For rocksdb, there's more, and it's probably worth trying to consolidate the `commit_code`, `try_find_impl`, and `contains_impl`.

This isn't rigorous at all, but I suspect the following signatures for free standing functions would suffice
```cpp
// namespace monad::db::detail
void commit_code(rocksdb::WriteBatch&, rocksdb::ColumnFamilyHandle*, state::changeset auto const&);
byte_string try_find_impl(bytes32_t const&, std::shared_ptr<rocksdb::DB>, rocksdb::ColumnFamilyHandle*);
byte_string contains_impl(bytes32_t const&, std::sahred_ptr<rocksdb::DB>, rocksdb::ColumnFamilyHandle*);
``` 

And if RocksDB and RocksTrieDB both call into these functions, now they both pick up any changes that are made.

include/monad/state/code_state.hpp
```diff
@@ -66,16 +68,19 @@ struct CodeState
...
+    StateChanges::CodeChanges gather_changes() const
     {
         assert(can_commit());
+        StateChanges::CodeChanges code_changes;
 
-        for (auto &[a, code] : merged_) {
-            auto const &[_, inserted] = db_.emplace(a, std::move(code));
-            MONAD_DEBUG_ASSERT(inserted);
+        for (auto const &[code_hash, code] : merged_) {
+            code_changes.emplace_back(code_hash, std::move(code));
```
zander-xyz, line 77: `code` is const so you aren't actually moving anything. you need to make it non-const if you actually want the move to occur, otherwise this is just a copy

include/monad/state/code_state.hpp
zander-xyz:
```cpp
std::copy_n(std::next(code.begin(), offset), bytes_to_copy, buffer);
```

src/monad/db/test/db.cpp
```diff
@@ -15,8 +15,6 @@ using namespace monad::db;
```
zander-xyz: Can you add to the `HijackedExecutionDBTest` typed test please? Also I think we're missing code changes tests here. We should have one for code change creation, deletion, and we should verify the modification (non-empty code -> non-empty code) is disallowed through assertions.

src/monad/state/test/code_state.cpp
```diff
@@ -10,240 +18,286 @@
...
-        auto const total = t.copy_code(b, 0u, buffer, size);
+        auto const total = changeset.copy_code(code_hash2, 0u, buffer, size);
         EXPECT_EQ(total, size);
-        EXPECT_EQ(0, std::memcmp(buffer, c2.c_str(), total));
+        EXPECT_EQ(0, std::memcmp(buffer, code2.c_str(), total));
     }
 }
 
-TEST(CodeState, can_merge)
+// TODO: I feel like those 2 tests are a bit repetitive
```
zander-xyz: Agreed...i wouldn't be against merging the two

src/monad/state/test/code_state.cpp
```diff
@@ -10,240 +18,286 @@
...
+        EXPECT_TRUE(s.can_merge(changeset));
+        s.merge_changes(changeset);
     }
     EXPECT_TRUE(s.can_commit());
-    s.commit_all_merged();
 
-    EXPECT_EQ(s.code_at(a), c1);
-    EXPECT_EQ(s.code_at(b), c2);
-    EXPECT_EQ(s.code_at(c), c3);
+    auto _ = s.gather_changes();
```
zander-xyz: why this?

include/monad/db/in_memory_db.hpp
tongzhi1998: Resolved over Zoom.

include/monad/state/code_state.hpp
tongzhi1998: Cool

include/monad/db/in_memory_trie_db.hpp
```diff
@@ -73,6 +96,17 @@ namespace detail
         auto &storage() { return storage_trie; }
         auto const &accounts() const { return accounts_trie; }
         auto const &storage() const { return storage_trie; }
+
+        void commit(state::changeset auto const &obj)
+        {
+            for (auto const &[ch, c] : obj.code_changes) {
+                if (c != byte_string{}) {
+                    code[ch] = c;
+                }
```
zander-xyz: did you forget to make the same change `code[ch] = c` here as you did in the in_memory_db?

include/monad/db/rocks_db_helper.hpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+    };
+
+    template <Permission TPermission>
+    void rocks_db_commit_code(
+        state::changeset auto const &obj,
+        std::conditional_t<Writable<TPermission>, rocksdb::WriteBatch, Empty>
+            &batch,
+        rocksdb::ColumnFamilyHandle *cf)
+    {
```
zander-xyz: You can not ever commit code if the database is not writable, so I think we can just assume that the type of batch is WriteBatch here. The compiler will complain about the type mismatch if what we end up passing here is in fact not a WriteBatch.

Then we can do away with all the templating and Empty

include/monad/db/rocks_db_helper.hpp
```diff
@@ -0,0 +1,49 @@
...
+        for (auto const &[ch, c] : obj.code_changes) {
+            if (c != byte_string{}) {
+                auto const res = batch.Put(cf, to_slice(ch), to_slice(c));
+                MONAD_ROCKS_ASSERT(res);
+            }
```
zander-xyz: I thought we were getting rid of the if statement?

include/monad/db/rocks_db_helper.hpp
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    bool rocks_db_contains_impl(
+        bytes32_t const &b, std::shared_ptr<rocksdb::DB> db,
+        rocksdb::ColumnFamilyHandle *cf);
+
+    byte_string rocks_db_try_find_impl(
+        bytes32_t const &b, std::shared_ptr<rocksdb::DB> db,
+        rocksdb::ColumnFamilyHandle *cf);
```
zander-xyz: [[nodiscard]] please

include/monad/state/code_state.hpp
```diff
@@ -66,16 +68,19 @@ struct CodeState
             merged_, [&](auto const &a) { return db_.contains(a.first); });
     }
 
-    void commit_all_merged()
+    StateChanges::CodeChanges gather_changes() const
```
zander-xyz: [[nodiscard]] here too please

include/monad/state/code_state.hpp
```diff
@@ -117,7 +121,10 @@ struct CodeState<TCodeDB>::ChangeSet : public CodeState<TCodeDB>
         auto const code = code_at(b);
         assert(code.size() > offset);
         auto const bytes_to_copy = std::min(code.size() - offset, buffer_size);
-        std::memcpy(buffer, code.begin() + offset, bytes_to_copy);
+        std::copy_n(
+            std::next(code.begin(), static_cast<long>(offset)),
```
zander-xyz, line 125: I'm always paranoid about these types of casts, so can we also assert that the offset can be represented as a long?
```cpp
MONAD_DEBUG_ASSERT(offset >= std::numeric_limits<long>::min());
MONAD_DEBUG_ASSERT(offset <= std::numeric_limits<long>::max());
```

src/monad/db/rocks_db_helper.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <rocksdb/db.h>
+
+#include <memory>
+
+MONAD_DB_NAMESPACE_BEGIN
+
+namespace detail
+{
+    bool rocks_db_contains_impl(
+        bytes32_t const &b, std::shared_ptr<rocksdb::DB> db,
```
zander-xyz: Can we do `b` -> `hash` here and everywhere for the code functions please? It's much clearer

include/monad/db/in_memory_trie_db.hpp
tongzhi1998: Sorry I forgot. Good catch!

include/monad/state/code_state.hpp
tongzhi1998: Ignore these for now since it also creates compilation error.

src/monad/db/rocks_db_helper.cpp
tongzhi1998: I changed to ch instead to avoid possible confusing with other type of hashes.

PR #229, opened 2023-08-09
:sparkles: Rocksdb block history support

zander-xyz:


test/unit/common/include/monad/test/hijacked_db.hpp
```diff
@@ -39,4 +39,11 @@ as_string<test::hijacked::RocksTrieDB>() noexcept
     return "hijackedrockstriedb";
 }
 
+template <>
+[[nodiscard]] constexpr std::string_view
+as_string<test::hijacked::RocksDB>() noexcept
+{
+    return "hijackedrocksdb";
```
tongzhi1998, line 46: Should we also mark this with suffix sv to avoid implicit construction? (Same with the above triedb)

zander-xyz: the string_view constructor is constexpr so this should all be evaluated at compile time

PR #239, opened 2023-08-10
:new: Add support for DAO fork

tongzhi1998:
Problem:
- We don't support the DAO hard fork now

Solution:
- When block_number = dao_fork, transfer some accounts' balance to the withdraw account, per EIP-779


include/monad/execution/block_processor.hpp
```diff
@@ -28,6 +29,12 @@ struct AllTxnBlockProcessor
             b.transactions.size());
         MONAD_LOG_DEBUG(block_logger, "BlockHeader Fields: {}", b.header);
 
+        if (b.header.number == ethereum::dao_block_number) {
+            auto change_set = s.get_new_changeset(0u);
+            ethereum::transfer_balance_dao(change_set);
+            s.merge_changes(change_set);
+        }
```
zander-xyz: I would've thought that we would check both against the traits and the block number

```cpp
if constexpr (std::same_as<TTraits, DAO>) {
    if (b.header.number == TTraits::first_block_number) {
        // blah
    }
}
```

Mostly so that we're taking advantage of the fact that the fork traits have been predetermined at compile-time so as to not incur the cost of checking in other forks.

include/monad/execution/ethereum/dao.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <intx/intx.hpp>
+
+MONAD_EXECUTION_ETHEREUM_NAMESPACE_BEGIN
+
+inline constexpr auto dao_block_number = 1'920'000u;
+
+inline constexpr auto withdraw_account{
+    0xbf4ed7b27f1d666546e30d74d50d173d20bca754_address};
+
+inline constexpr address_t child_accounts[]{
```
zander-xyz: I would drop the inline (it's implicit with the constexpr)

there are many philosophies on this, but this is at least consistent with how it's done in the rest of this repo

include/monad/execution/ethereum/dao.hpp
```diff
@@ -0,0 +1,149 @@
...
+    0xd343b217de44030afaa275f54d31a9317c7f441e_address,
+    0x84ef4b2357079cd7a7c69fd7a37cd0609a679106_address,
+    0xda2fef9e4a3230988ff17df2165440f37e8b1708_address,
+    0xf4c64518ea10f995918a454158c6b61407ea345c_address,
+    0x7602b46df5390e432ef1c307d4f2c9ff6d65cc97_address,
+    0xbb9bc244d798123fde783fcc1c72d3bb8c189413_address,
+    0x807640a13483f8ac783c557fcdf27be11ea4ac7a_address};
+
+template <typename TState>
+void transfer_balance_dao(TState &s)
```
zander-xyz: Does this need to be templated? And can we change this to return the ChangeSet directly instead of taking it as an in parameter?

src/monad/execution/ethereum/test/dao.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+std::vector<std::pair<address_t, std::optional<Account>>> generate_accounts()
+{
+    std::vector<std::pair<address_t, std::optional<Account>>> v{};
+    for (auto const addr : ethereum::child_accounts) {
+        Account a{.balance = individual};
+        v.emplace_back(std::make_pair(addr, a));
+    }
+    return v;
+}
```
zander-xyz: This helper function feels uneccessary, I would just include this code directly below.

include/monad/execution/ethereum/dao.hpp
tongzhi1998: I might be wrong, but ChangeSet is not really the delta: it is the updated State. Without taking reference to the old State, I don't think it is possible to get a new State, so in a sense we have to take in the State object regardless.

src/monad/execution/ethereum/test/dao.cpp
tongzhi1998: Completely agree!

include/monad/execution/block_processor.hpp
tongzhi1998: That makes sense, but I am not so certain if adding an another fork is the best idea. The `dao_fork` is in the middle of homestead, so adding this might interrupt our current design a bit. Let me know if you want to Zoom to talk about this.

tongzhi1998: I am currently using the __builtin_expect macro to tell the compiler than this is unlikely... I am not sure if this is a good idea either.

zander-xyz: You're right, we would need to add a DAO struct and then modify fork_traits so that homestead points to it. But that's probably the right thing to do? Where are you thinking the design would change significantly?

zander-xyz: Discussed further on zoom. We should add a DAO fork and a `transfer_balance_dao` function to it. We'll also need to add to other forks (albeit empty).

include/monad/execution/block_processor.hpp
```diff
@@ -27,6 +29,14 @@ struct AllTxnBlockProcessor
...
             b.transactions.size());
         MONAD_LOG_DEBUG(block_logger, "BlockHeader Fields: {}", b.header);
+        if constexpr (std::same_as<TTraits, fork_traits::dao>) {
+            if (__builtin_expect(
+                    b.header.number == ethereum::dao::dao_block_number, 0)) {
+                auto change_set = s.get_new_changeset(0u);
+                TTraits::transfer_balance_dao(change_set);
+                s.merge_changes(change_set);
+            }
+        }
```
zander-xyz: A couple things:
- You may drop the constexpr since the empty functions should be optimized away.
- move the if statement into the `transfer_balance_dao`

include/monad/execution/block_processor.hpp
```diff
@@ -27,6 +29,14 @@ struct AllTxnBlockProcessor
             b.header.number,
             b.transactions.size());
         MONAD_LOG_DEBUG(block_logger, "BlockHeader Fields: {}", b.header);
+        if constexpr (std::same_as<TTraits, fork_traits::dao>) {
+            if (__builtin_expect(
+                    b.header.number == ethereum::dao::dao_block_number, 0)) {
+                auto change_set = s.get_new_changeset(0u);
```
zander-xyz: why the new change_set? can't we operate on `s` itself?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -251,9 +258,35 @@ namespace fork_traits
...
+        static constexpr void transfer_balance_dao(TState &)
+        {
+        }
+    };
+
+    struct dao : public homestead
+    {
+        using next_fork_t = spurious_dragon;
+        static constexpr auto last_block_number = 2'674'999u;
+        // EVMC revision for DAO should just be EVMC_HOMESTEAD
```
zander-xyz, line 276: 👍 good note

src/monad/execution/ethereum/test/dao.cpp
```diff
@@ -0,0 +1,58 @@
...
+    fork_traits::dao::transfer_balance_dao(change_set);
+
+    for (auto const &addr : ethereum::dao::child_accounts) {
+        EXPECT_EQ(intx::be::load<uint256_t>(change_set.get_balance(addr)), 0u);
+    }
+    EXPECT_EQ(
+        intx::be::load<uint256_t>(
+            change_set.get_balance(ethereum::dao::withdraw_account)),
+        total);
+}
```
zander-xyz: this test belongs in `fork_traits.cpp`

src/monad/execution/ethereum/test/replay_eth_block_db.cpp
```diff
@@ -471,11 +471,21 @@ TEST(ReplayFromBlockDb_Eth, frontier_to_spurious_dragon)
...
+    for (auto i = 11u; i < fork_traits::homestead::last_block_number -
+                               fork_traits::frontier::last_block_number + 11u;
+         ++i) {
         EXPECT_EQ(
             receipt_collector[i][0].status,
             fork_traits::homestead::last_block_number);
     }
+    for (auto i = fork_traits::homestead::last_block_number -
+                  fork_traits::frontier::last_block_number + 11u;
+         i < receipt_collector.size() - 9u;
```
zander-xyz: where did the 11 and 9 come from?

src/monad/execution/ethereum/test/dao.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <cstdlib>
+#include <optional>
+#include <vector>
+
+using namespace monad;
+using namespace monad::execution;
+
+using db_t = db::InMemoryDB;
+
+inline constexpr auto individual = 100u;
```
zander-xyz: drop the inline and put these two in an anonymous namespace please

src/monad/execution/test/block_processor.cpp
```diff
@@ -14,6 +27,17 @@ using namespace monad::execution;
...
+using block_db_t = db::BlockDb;
+using db_t = db::InMemoryDB;
+using account_state_t = state::AccountState<db_t>;
+using value_state_t = state::ValueState<db_t>;
+using code_state_t = state::CodeState<db_t>;
+using real_state_t = state::State<
+    account_state_t, value_state_t, code_state_t, block_db_t, db_t>;
+
+inline constexpr auto individual = 100u;
+inline constexpr auto total = individual * 116u;
```
zander-xyz: drop inline and anonymous namespace

src/monad/execution/test/block_processor.cpp
```diff
@@ -92,3 +116,41 @@ TEST(AllTxnBlockProcessor, execute_some_failed)
...
+    state::ValueState values{db};
+    state::CodeState codes{db};
+    real_state_t s{accounts, values, codes, blocks, db};
+
+    Block b{};
+    b.header.number = ethereum::dao::dao_block_number;
+
+    AllTxnBlockProcessor<BoostFiberExecution> bp{};
+    [[maybe_unused]] auto const r =
+        bp.execute<real_state_t, fork_traits::dao, fiber_data_t>(s, b);
```
zander-xyz, line 146: Why not assert the expected `r`?

include/monad/execution/block_processor.hpp
```diff
@@ -27,6 +29,14 @@ struct AllTxnBlockProcessor
             b.header.number,
             b.transactions.size());
         MONAD_LOG_DEBUG(block_logger, "BlockHeader Fields: {}", b.header);
+        if constexpr (std::same_as<TTraits, fork_traits::dao>) {
+            if (__builtin_expect(
```
rgarc: small nit: is it possible to use the macro for consistency https://github.com/monad-crypto/monad-core/blob/3ed56a5dfedbeab36a2f659c4d6249283f8747a2/include/monad/core/likely.h#L3

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -203,15 +205,20 @@ namespace fork_traits
         {
             s.add_txn_award(uint256_t{gas_used} * gas_price(t, base_gas_cost));
         }
+
```
rgarc, line 213: I'm not sure the empty definitions of `transfer_balance_dao` are necessary for the `frontier` fork, and the ones after `dao`.

include/monad/execution/block_processor.hpp
tongzhi1998: Unfortunately I don't think we can. State only has high-level functions, not able to access / change anything related to accounts.

src/monad/execution/ethereum/test/replay_eth_block_db.cpp
tongzhi1998: Because the start block number is `frontier::last_block_number - 10` and end block number is `dao::last_block_number + 10`

src/monad/execution/test/block_processor.cpp
tongzhi1998: `r` is actually completely empty in this case.

include/monad/execution/ethereum/fork_traits.hpp
tongzhi1998: I think we call `TTraits::transfer_balance_dao()` inside `BlockProcessor` (for now, might change later), so if I am understanding correctly, we need to define them for every fork. But I think we only need to define an empty function for frontier and the `fork` immediately after `dao` (which, for now, is `spurious_dragon`), I will remove the rest

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -253,7 +261,45 @@ namespace fork_traits
...
+                    change_set.set_balance(
+                        execution::ethereum::dao::withdraw_account,
+                        intx::be::load<uint256_t>(change_set.get_balance(
+                            execution::ethereum::dao::withdraw_account)) +
+                            intx::be::load<uint256_t>(
+                                change_set.get_balance(addr)));
+                    change_set.set_balance(addr, 0u);
+                }
+
+                // Workaround to get txn_id right for now. Ideally we want to
```
zander-xyz: nit: 
```suggestion
                // TODO: Workaround to get txn_id right for now. Ideally we want to
```
for easy grepping

src/monad/execution/test/block_processor.cpp
zander-xyz: Right, so `EXPECT_TRUE(r.empty());`?

src/monad/execution/test/block_processor.cpp
```diff
@@ -92,3 +119,41 @@ TEST(AllTxnBlockProcessor, execute_some_failed)
...
+
+    for (auto const &addr : ethereum::dao::child_accounts) {
+        change_set.access_account(addr);
+        EXPECT_EQ(intx::be::load<uint256_t>(change_set.get_balance(addr)), 0u);
+    }
+    change_set.access_account(ethereum::dao::withdraw_account);
+    EXPECT_EQ(
+        intx::be::load<uint256_t>(
+            change_set.get_balance(ethereum::dao::withdraw_account)),
+        total);
```
zander-xyz, line 158: why do you need the change set? you're only accessing the state, not writing to it right?

tongzhi1998: Unfortunately we still need a `ChangeSet` to access the state... The `State` object itself doesn't have any accessor method defined

Issue #244, opened 2023-08-11
Error in how we deal with OOG on code_deployment


tongzhi1998:
- Currently, if the smart contract executes successfully but we don't have enough gas for code deployment, we just charge all the remaining gas and set empty code
- However, what we should do instead is not charging the additional gas for code deployment and set empty code

Block Number: 49018

Change I made temporarily:
https://github.com/monad-crypto/monad/pull/242/commits/82dab5da5c6439e529f6d07617340a3776945dc0

```
@@ -173,10 +175,8 @@ namespace fork_traits
                 auto const deploy_cost =
                     static_cast<int64_t>(result.output_size) * 200u;
                 result.create_address = a;
-
                 if (result.gas_left < deploy_cost) {
                     s.set_code(a, {});
-                    result.gas_left = 0;
                 }
                 else {
                     result.gas_left -= deploy_cost;
```

Reference to Silkworm:
https://github.com/monad-crypto/silkworm/blob/master/core/silkworm/execution/evm.cpp#L181

Reference to Yellow Paper

PR #248, opened 2023-08-11
:bug: fixes to code deployment charge

zander-xyz:
Addresses #244

### DIFF
--- a/include/monad/execution/ethereum/fork_traits.hpp
+++ b/include/monad/execution/ethereum/fork_traits.hpp
@@ -174,10 +174,14 @@ namespace fork_traits
                 result.create_address = a;
 
                 if (result.gas_left < deploy_cost) {
+                    // From YP: "No code is deposited in the state if the gas
+                    // does not cover the additional per-byte contract deposit
+                    // fee, however, the value is still transferred and the
+                    // execution side- effects take place."
                     s.set_code(a, {});
-                    result.gas_left = 0;
                 }
                 else {
+                    s.set_code(a, {result.output_data, result.output_size});
                     result.gas_left -= deploy_cost;
                 }
             }
@@ -247,6 +251,7 @@ namespace fork_traits
                 else {
                     result.create_address = a;
                     result.gas_left -= deploy_cost;
+                    s.set_code(a, {result.output_data, result.output_size});
                 }
             }
             return result;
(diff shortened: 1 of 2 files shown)


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -174,8 +174,13 @@ namespace fork_traits
                 result.create_address = a;
 
                 if (result.gas_left < deploy_cost) {
+                    // From YP: "No code is deposited in the state if the gas
+                    // does not cover the additional per-byte contract deposit
+                    // fee, however, the value is still transferred and the
+                    // execution side- effects take place."
+                    //
+                    // One of the side effects is the gas refund
```
tongzhi1998: I think the sentence "One of the side effects is the gas refund" is a bit confusing. From what I understand, 'execution' here just means the EVMC execution, that is, without the code deployment (because that need to be handled by our code). So I interpret this as all the EVMC related changes are not reverted, but that's it. As a result, we should not charge the code deployment cost.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -228,26 +228,26 @@ namespace fork_traits
...
-            }
-
-            s.set_code(a, code);
-            return {.status_code = EVMC_SUCCESS, .gas_left = gas - deploy_cost};
-        }
-
         template <class TState>
         static evmc::Result finalize_contract_storage(
-            TState &, address_t const &a, evmc::Result result) noexcept
+            TState &s, address_t const &a, evmc::Result result) noexcept
```
tongzhi1998, line 237: It took my a while to notice that your `set_code` function is in a different PR. I am a bit hesitant to do this because this commit by itself will break additional things (There is no `set_code` being called at all). I think those 2 changes might be more appropriate if they are in the same PR/commit. Feel free to zoom me and discuss more about this!

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: I agree that the second comment is confusing. Our understandings seem to align. However, the YP has no concept of EVMC execution. I'll remove the second comment.

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: These are two separate bugs, belonging in two separate commits. I personally like it when PRs are as small as possible.

tongzhi1998: Make sense!

PR #249, opened 2023-08-11
:bug: code should be generated from executing init code

zander-xyz:
Problem:
- currently the code hash associated with an account is that of the
  initialization code
- the YP states that the code stored for an account should be the result
  of executing the init code

From YP: "Additionally, a contract creation transaction (regardless
whether legacy or EIP-2930) contains: init: An unlimited size byte array
specifying the EVM-code for the account initialisation procedure, formally
Ti. init is an EVM-code fragment; it returns the body, a second fragment
of code that executes each time the account receives a message call (either
through a transaction or due to the internal execution of code). init is
executed only once at account creation and gets discarded immediately
thereafter."

Solution:
- Use the body instead of the initialization code for the code hash

On top of #248

### DIFF
--- a/include/monad/execution/ethereum/fork_traits.hpp
+++ b/include/monad/execution/ethereum/fork_traits.hpp
@@ -170,7 +170,7 @@ namespace fork_traits
         {
             if (result.status_code == EVMC_SUCCESS) {
                 auto const deploy_cost =
-                    static_cast<int64_t>(s.get_code_size(a)) * 200;
+                    static_cast<int64_t>(result.output_size) * 200;
                 result.create_address = a;
 
                 if (result.gas_left < deploy_cost) {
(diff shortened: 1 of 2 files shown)


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -184,6 +184,8 @@ namespace fork_traits
                 }
                 else {
                     result.gas_left -= deploy_cost;
+                    s.set_code(
+                        a, byte_string{result.output_data, result.output_size});
                 }
```
tongzhi1998: I think you might forget to change deploy cost, which should be `deploy_cost = result.output_size * 200u`

zander-xyz: You are absolutely right.

PR #261, opened 2023-08-16
:bug: make initialization code temporal

zander-xyz:
Problem:
- Currently we store initialization code in state, which eventually gets committed. Because we initially did not distinguish between contract code and initialization code, we have both store_contract_code and finalize_contract_code. We only need one deploy code function and all checks for gas, illegal code, etc should be done on the deployed code rather than the initialization code

Solution:
- Consolidate store_contract_code and finalize_contract_code
- Refactor the interpreter such that it takes in the code directly instead of reading from state
- Perform checks on deployed code instead of initialization code

### DIFF
--- a/include/monad/execution/ethereum/fork_traits.hpp
+++ b/include/monad/execution/ethereum/fork_traits.hpp
@@ -159,34 +159,24 @@ namespace fork_traits
         }
 
         template <class TState>
-        [[nodiscard]] static evmc_result store_contract_code(
-            TState &s, address_t const &a, byte_string code,
-            int64_t gas) noexcept
-        {
-            s.set_code(a, code);
-            return {.status_code = EVMC_SUCCESS, .gas_left = gas};
-        }
-
-        template <class TState>
-        static evmc::Result finalize_contract_storage(
+        static evmc::Result deploy_contract_code(
             TState &s, address_t const &a, evmc::Result result) noexcept
         {
-            if (result.status_code == EVMC_SUCCESS) {
-                auto const deploy_cost =
-                    static_cast<int64_t>(result.output_size) * 200;
-                result.create_address = a;
+            MONAD_DEBUG_ASSERT(result.status_code == EVMC_SUCCESS);
+            auto const deploy_cost =
+                static_cast<int64_t>(result.output_size) * 200;
+            result.create_address = a;
 
-                if (result.gas_left < deploy_cost) {
-                    // From YP: "No code is deposited in the state if the gas
-                    // does not cover the additional per-byte contract deposit
-                    // fee, however, the value is still transferred and the
-                    // execution side- effects take place."
-                    s.set_code(a, {});
-                }
-                else {
-                    s.set_code(a, {result.output_data, result.output_size});
-                    result.gas_left -= deploy_cost;
-                }
+            if (result.gas_left < deploy_cost) {
+                // From YP: "No code is deposited in the state if the gas
+                // does not cover the additional per-byte contract deposit
+                // fee, however, the value is still transferred and the
+                // execution side- effects take place."
+                s.set_code(a, {});
+            }
+            else {
+                s.set_code(a, {result.output_data, result.output_size});
+                result.gas_left -= deploy_cost;
             }
             return result;
         }
@@ -241,29 +231,28 @@ namespace fork_traits
         }
 
         template <class TState>
-        static evmc::Result finalize_contract_storage(
+        static evmc::Result deploy_contract_code(
             TState &s, address_t const &a, evmc::Result result) noexcept
         {
-            if (result.status_code == EVMC_SUCCESS) {
-                auto const deploy_cost =
-                    static_cast<int64_t>(result.output_size) * 200;
+            MONAD_DEBUG_ASSERT(result.status_code == EVMC_SUCCESS);
+            auto const deploy_cost =
+                static_cast<int64_t>(result.output_size) * 200;
 
-                if (result.gas_left < deploy_cost) {
-                    // EIP-2: If contract creation does not have enough gas to
-                    // pay for the final gas fee for adding the contract code to
-                    // the state, the contract creation fails (ie. goes
-                    // out-of-gas) rather than leaving an empty contract.
-                    //
-                    // TODO: make sure old initialization code does not get
-                    // committed to the database
-                    result.status_code = EVMC_OUT_OF_GAS;
-                    result.gas_left = 0;
-                }
-                else {
-                    result.create_address = a;
-                    result.gas_left -= deploy_cost;
-                    s.set_code(a, {result.output_data, result.output_size});
-                }
+            if (result.gas_left < deploy_cost) {
+                // EIP-2: If contract creation does not have enough gas to
+                // pay for the final gas fee for adding the contract code to
+                // the state, the contract creation fails (ie. goes
+                // out-of-gas) rather than leaving an empty contract.
+                //
+                // TODO: make sure old initialization code does not get
+                // committed to the database
+                result.status_code = EVMC_OUT_OF_GAS;
+                result.gas_left = 0;
+            }
+            else {
+                result.create_address = a;
+                result.gas_left -= deploy_cost;
+                s.set_code(a, {result.output_data, result.output_size});
             }
             return result;
         }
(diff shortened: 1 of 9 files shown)


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -156,34 +156,24 @@ namespace fork_traits
...
-                    // does not cover the additional per-byte contract deposit
-                    // fee, however, the value is still transferred and the
-                    // execution side- effects take place."
-                    s.set_code(a, {});
-                }
-                else {
-                    s.set_code(a, {result.output_data, result.output_size});
-                    result.gas_left -= deploy_cost;
-                }
+            MONAD_DEBUG_ASSERT(result.status_code == EVMC_SUCCESS);
```
rgarc, line 165: Why did you trade the branch for a `MONAD_DEBUG_ASSERT`? I am unsure about the consequences if, in release mode, this function is called with `result.status_code != EVMC_SUCCESS`.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -233,29 +223,28 @@ namespace fork_traits
...
-                    // committed to the database
-                    result.status_code = EVMC_OUT_OF_GAS;
-                    result.gas_left = 0;
-                }
-                else {
-                    result.create_address = a;
-                    result.gas_left -= deploy_cost;
-                    s.set_code(a, {result.output_data, result.output_size});
-                }
+            MONAD_DEBUG_ASSERT(result.status_code == EVMC_SUCCESS);
```
rgarc, line 237: Same as above.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -284,16 +273,15 @@ namespace fork_traits
         }
 
         template <class TState>
-        [[nodiscard]] static evmc_result store_contract_code(
-            TState &s, address_t const &a, byte_string code,
-            int64_t gas) noexcept
+        [[nodiscard]] static evmc::Result deploy_contract_code(
+            TState &s, address_t const &a, evmc::Result result) noexcept
         {
+            MONAD_DEBUG_ASSERT(result.status_code == EVMC_SUCCESS);
```
rgarc, line 325: This one makes sense to me because we previously weren't even checking `result.status_code`.

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: the assert is there because I am assuming that it never happens (we check result before calling this function). If it does, then it's a bug that needs to be fixed.

include/monad/execution/ethereum/fork_traits.hpp
tongzhi1998: I might be wrong, but I think this will never happen because it is called inside this `if` statement
https://github.com/monad-crypto/monad/pull/261/commits/d1156adce4695a26fa3a600506f7adace41492c5#diff-e89801613b8eabcafb27aad6fd504338eb332972a633a47f7e57e501aac9d9a1R61

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -156,34 +156,24 @@ namespace fork_traits
...
+            auto const deploy_cost =
+                static_cast<int64_t>(result.output_size) * 200;
+            result.create_address = a;
+
+            if (result.gas_left < deploy_cost) {
+                // From YP: "No code is deposited in the state if the gas
+                // does not cover the additional per-byte contract deposit
+                // fee, however, the value is still transferred and the
+                // execution side- effects take place."
+                s.set_code(a, {});
```
sheapwilliams, line 175: Do we take the rest of the gas here, even if returning a success result?

include/monad/execution/test/fakes.hpp
```diff
@@ -183,6 +183,9 @@ namespace fake
             [[nodiscard]] byte_string_view
             get_code(bytes32_t const &b) const noexcept
             {
+                if (b == NULL_HASH) {
```
tongzhi1998, line 214: Do we need this because we move the `get_code_hash()` from `Interpreter` to `Evm` and some tests are calling with function with `NULL_HASH`?

test/integration/txn_proc_evm_state_host/txn_proc_evm_state_host.cpp
```diff
@@ -106,71 +106,76 @@ TEST(TxnProcEvmInterpStateHost, account_transfer_miner_ommer_award)
...
-    s.merge_changes(changeset);
-
-    fork_t::apply_block_award(s, b);
-
-    auto ws = s.get_new_changeset(1u);
-    ws.access_account(a);
-    ws.access_account(o);
-    EXPECT_EQ(ws.get_balance(a), bytes32_t{5'240'000'000'000'000'000});
-}
+// TODO: Fix this test. This test is commented out because state reversion
```
tongzhi1998, line 109: Not an important question, but what makes it passes previously but not now?

include/monad/execution/evmone_baseline_interpreter.hpp
```diff
@@ -24,21 +24,22 @@
...
+
 MONAD_EXECUTION_NAMESPACE_BEGIN
 
 template <class TState, concepts::fork_traits<TState> TTraits>
 struct EVMOneBaselineInterpreter
 {
     template <class TEvmHost>
     static evmc::Result
-    execute(TEvmHost *h, TState const &s, evmc_message const &m)
+    execute(TEvmHost *h, evmc_message const &m, std::span<std::byte const> code)
```
sheapwilliams: span instead of byte string view?  you just create the view below, why not just use that?

include/monad/execution/evm.hpp
```diff
@@ -32,33 +34,44 @@ struct Evm
...
+            if (auto const transfer_result =
+                    transfer_call_balances(state, m_call);
+                transfer_result.status_code != EVMC_SUCCESS) {
+                return evmc::Result{transfer_result};
+            }
+
+            return TInterpreter::execute(
+                host,
+                m_call,
+                std::as_bytes(std::span(m.input_data, m.input_size)));
```
sheapwilliams: Same here.  Seems like we should have a slight preference to consistently using our internal types since we've gone through the trouble to make them?

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: @tongzhi1998 - do you know the answer to this?  Do we take the rest of the gas if there isn't enough to store the code on Frontier?  Or do we return the remaining-but-insufficient?

tongzhi1998: We do the later. We don't take the remaining and don't deploy anything.

tongzhi1998: I think EIP-2 (homestead) changes it to be taking the remaining.

test/integration/txn_proc_evm_state_host/txn_proc_evm_state_host.cpp
sheapwilliams: The new code calls `revert` if the code failed to deploy on later forks.  And currently that is revoking the irrevocable changes...I suggested that he just comment them out and I'll work to make them work again with my changes.

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: I don't read either way based on the EIP...Seems unstated.  Have we run across a ~~contract~~ *txn* like this as of yet?

tongzhi1998: https://github.com/monad-crypto/monad/issues/244

Block 49018

Yeah I agree that Frontier yellow paper & EIP are unclear about this...

include/monad/execution/evm.hpp
zander-xyz: Yeah, you're probably right. I've been trying to coax our codebase towards the direction of span of bytes but also happy to just abandon that effort lol

include/monad/execution/ethereum/fork_traits.hpp
sheapwilliams: Great!  TY!

include/monad/execution/test/fakes.hpp
zander-xyz: That's right. See `call_evm` and `revert_call_evm`

include/monad/execution/evmone_baseline_interpreter.hpp
```diff
@@ -31,14 +31,13 @@ struct EVMOneBaselineInterpreter
 {
     template <class TEvmHost>
     static evmc::Result
-    execute(TEvmHost *h, TState const &s, evmc_message const &m)
+    execute(TEvmHost *h, evmc_message const &m, byte_string_view code)
```
tongzhi1998, line 34: Just realized that this Class might no longer need `TState` template... But I don't want to bother you changing all related stuff. Maybe a separate PR?

tongzhi1998: Sorry I am being dumb, disregard this plz

PR #264, opened 2023-08-17
:bug: Fix get_code_hash for accounts not in ChangeSet

tongzhi1998:
Problem:
- A smart contract could access code hash for any arbitrary account not in `ChangeSet` (it could even not exist!)
- Our code segfaults on this because we require any accessed account to be in `ChangeSet` first

Solution:
- When an account is not in `ChangeSet`, we call `AccountState.get_code_hash()`, which will look at `DB` and return `NULL_HASH` if not exist

Reference Block Number: 68460

### DIFF
--- a/include/monad/state/account_state.hpp
+++ b/include/monad/state/account_state.hpp
@@ -204,7 +204,11 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
     [[nodiscard]] bytes32_t
     get_code_hash(address_t const &address) const noexcept
     {
-        return changed_.at(address).updated.value_or(Account{}).code_hash;
+        if (changed_.contains(address)) {
+            return changed_.at(address).updated.value_or(Account{}).code_hash;
+        }
+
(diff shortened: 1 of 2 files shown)


include/monad/state/account_state.hpp
```diff
@@ -204,7 +204,11 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
     [[nodiscard]] bytes32_t
     get_code_hash(address_t const &address) const noexcept
     {
-        return changed_.at(address).updated.value_or(Account{}).code_hash;
+        if (changed_.contains(address)) {
+            return changed_.at(address).updated.value_or(Account{}).code_hash;
+        }
+
+        return AccountState::get_code_hash(address);
```
sheapwilliams, line 211: As discussed offline, I think if we don't capture the accessed account in a changeset, then there is a possibility that an optimistic execution changes what should have been returned here.  I think we need to make this part of the changeset instead of just returning a value from the database...

PR #267, opened 2023-08-18
✨ Add in support for tangerine whistle

zander-xyz:
Tangerine whistle contains EIP-150: https://eips.ethereum.org/EIPS/eip-150. These are all gas changes, which is handled by evmone.

https://eips.ethereum.org/EIPS/eip-608

### DIFF
--- a/src/monad/execution/ethereum/test/fork_traits.cpp
+++ b/src/monad/execution/ethereum/test/fork_traits.cpp
@@ -148,6 +148,41 @@ TEST(fork_traits, homestead)
     }
 }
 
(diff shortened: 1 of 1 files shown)


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -297,9 +298,21 @@ namespace fork_traits
         }
     };
 
-    // tangerine_whistle - 2'463'000
```
sheapwilliams, line 300: Was this comment incorrect? Or was it the starting block?

zander-xyz: Right, starting block

PR #268, opened 2023-08-18
:recycle: destruct touched dead accounts after suicides

zander-xyz:
Problem:
- EIP-161 states that destruction of touched empty accounts should happen after suicides.
- We flip the order

Solution:
- Destruct the touched empty accounts after suicides instead

### DIFF
--- a/include/monad/execution/transaction_processor.hpp
+++ b/include/monad/execution/transaction_processor.hpp
@@ -104,14 +104,11 @@ struct TransactionProcessor
             static_cast<uint64_t>(result.gas_refund));
 
         // finalize state, Eqn. 77-79
-        TTraits::destruct_touched_dead(s);
         s.destruct_suicides();
+        TTraits::destruct_touched_dead(s);
 
         TTraits::apply_txn_award(
-            s,
-            t,
-            base_fee_per_gas,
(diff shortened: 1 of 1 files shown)


include/monad/execution/transaction_processor.hpp
```diff
@@ -104,14 +104,11 @@ struct TransactionProcessor
             static_cast<uint64_t>(result.gas_refund));
 
         // finalize state, Eqn. 77-79
-        TTraits::destruct_touched_dead(s);
         s.destruct_suicides();
+        TTraits::destruct_touched_dead(s);
```
sheapwilliams, line 108: Technically this is true, but I recall flipping them for some reason...lemme try to recall.  

BTW - Are you seeing this affect anything specifically, or just finding yourself being pedantic?

zander-xyz: Mostly pedantic, and this is what other implementations do as well. None of the tests crashed

sheapwilliams: This ordering was originally changed due to the old `deleted_` vector that was separate from the changeset.  But that was changed in #200, so we don't rely on this ordering anymore.

Looks good to me!

PR #273, opened 2023-08-19
:bug: Fix CodeState can_merge logic

tongzhi1998:
Problem:
- Copy the same code to another address is allowed. However, since our code is stored based on hash, when we are replicating code, our current `can_merge` will fail because it assert that the same hash existed

Solution:
- Return false when key is the same but value is not the same instead
- Remove `MONAD_DEBUG_ASSERT()` in `merge_changes`

### DIFF
--- a/include/monad/state/code_state.hpp
+++ b/include/monad/state/code_state.hpp
@@ -47,8 +47,9 @@ struct CodeState
 
     [[nodiscard]] bool can_merge(ChangeSet const &w) const
     {
-        return std::ranges::none_of(w.code_, [&](auto const &a) {
-            return merged_.contains(a.first) || db_.contains(a.first);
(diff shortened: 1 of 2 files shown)


include/monad/state/code_state.hpp
```diff
@@ -48,7 +48,9 @@ struct CodeState
     [[nodiscard]] bool can_merge(ChangeSet const &w) const
     {
         return std::ranges::none_of(w.code_, [&](auto const &a) {
-            return merged_.contains(a.first) || db_.contains(a.first);
+            return (merged_.contains(a.first) &&
+                    (merged_.at(a.first) != a.second)) ||
+                   (db_.contains(a.first) && (db_.at(a.first) != a.second));
```
zander-xyz: `a.second != code_at(a.first)`?

PR #283, opened 2023-08-25
Serialize change sets to JSON post state.

rgarc:
Builds on #282 

Problem:
- We would like to output state mutations to JSON so we can build the
  post state after executing a transaction.

Solution:
- Implement change set JSON serialization and pass a callback into
  commit to log it via quill.

### DIFF
--- a/test/ethereum_test/src/ethereum_test.cpp
+++ b/test/ethereum_test/src/ethereum_test.cpp
@@ -13,7 +13,8 @@ using state_t = monad::state::State<
     monad::state::AccountState<in_memory_trie_db_t>,
     monad::state::ValueState<in_memory_trie_db_t>,
     monad::state::CodeState<in_memory_trie_db_t>,
-    monad::execution::fake::BlockDb, in_memory_trie_db_t>;
+    monad::execution::fake::BlockDb, in_memory_trie_db_t,
+    /*EnableChangeSetTracing=*/true>;
 using working_state_t = decltype(std::declval<state_t>().get_new_changeset(0u));
 
 template <typename TFork>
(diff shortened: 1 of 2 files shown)


test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -240,6 +241,48 @@ StateTransitionTest EthereumTests::load_state_test(
...
 
+nlohmann::json
+serialize_state_changes(monad::state::StateChanges const &state_changes)
+{
+    using namespace nlohmann;
+    nlohmann::json state = json::object();
+
+    for (auto const &[address, maybe_account] : state_changes.account_changes) {
+        if (!maybe_account) {
+            continue;
```
rgarc, line 252: If the account does not exist, then this is a deletion. There has to be logic for clearing accounts from the JSON blob.

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -296,7 +345,12 @@ void EthereumTests::run_state_test(
                 execute(fork_index, block_header, change_set, transaction);
             state.merge_changes(change_set);
             state.apply_block_reward(block_header.beneficiary, 0);
-            state.commit();
+            state.commit_with_callback(state_changes_callback);
+            nlohmann::json post_state;
+            for (auto const &mutation : serialized_mutations) {
+                post_state.update(mutation, true);
```
rgarc, line 351: calling `update` won't suffice here as mentioned above, we need to clear accounts that were deleted. right now if an account gets deleted, it will just stay in the state JSON object, frozen in time :')

PR #284, opened 2023-08-25
Add support for Shanghai fork

tongzhi1998:
Add support for Shanghai, which has the following EIPs:
- EIP-3651 (Warm Coinbase)
- EIP-3855 (PUSH 0) -- Nothing to do on our end
- EIP-3860 (Limit and Meter InitCode)
- EIP-4895 (Beacon chain push withdrawals as operations)
- EIP-6049 (Deprecate selfdestruct) -- Not a real change

### DIFF
--- a/include/monad/execution/ethereum/fork_traits.hpp
+++ b/include/monad/execution/ethereum/fork_traits.hpp
@@ -212,6 +212,12 @@ namespace fork_traits
         static constexpr void warm_coinbase(TState &, address_t const &)
         {
         }
+
+        template <typename TState>
+        static constexpr void process_withdrawal(
+            TState &, std::optional<std::vector<Withdrawal>> const &)
+        {
+        }
     };
 
     struct homestead : public frontier
@@ -585,6 +591,31 @@ namespace fork_traits
             }
             return g_txcreate(t) + 21'000u + g_data(t) +
                    g_access_and_storage(t) + g_extra_cost_init(t);
+        }
+
+        // EIP-4895
+        template <typename TState>
+        static constexpr void process_withdrawal(
+            TState &s,
+            std::optional<std::vector<Withdrawal>> const &withdrawals) noexcept
+        {
+            if (withdrawals.has_value()) {
+                auto change_set = s.get_new_changeset(0u);
+
+                for (auto const &withdrawal : withdrawals.value()) {
+                    change_set.set_balance(
+                        withdrawal.recipient,
+                        intx::be::load<uint256_t>(
+                            change_set.get_balance(withdrawal.recipient)) +
+                            uint256_t{withdrawal.amount} *
+                                uint256_t{1'000'000'000u});
+                }
+
+                MONAD_DEBUG_ASSERT(
+                    s.can_merge_changes(change_set) ==
+                    TState::MergeStatus::WILL_SUCCEED);
+                s.merge_changes(change_set);
+            }
         }
     };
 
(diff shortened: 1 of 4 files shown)


src/monad/rlp/decode_helpers.cpp
```diff
@@ -321,6 +321,41 @@ byte_string_view decode_receipt(Receipt &receipt, byte_string_view const enc)
     return decode_untyped_receipt(receipt, enc);
 }
 
+byte_string_view
+decode_withdrawal(Withdrawal &withdrawal, byte_string_view const enc)
+{
+    MONAD_ASSERT(enc.size() > 0);
```
tongzhi1998: For all of these asserts in rlp, should they be DEBUG_ASSERT or just ASSERT?

include/monad/state/account_state.hpp
```diff
@@ -173,6 +177,10 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
     // EVMC Host Interface
     evmc_access_status access_account(address_t const &a)
     {
+        if (AccountState::beneficiary.has_value() &&
+            AccountState::beneficiary.value() == a) {
+            return EVMC_ACCESS_WARM;
+        }
```
sheapwilliams: Why would we do this as a check?  Why not just add it at the beginning of the transaction to the accessed vector and be don't with it?

include/monad/core/withdrawal.hpp
```diff
@@ -0,0 +1,21 @@
...
+MONAD_NAMESPACE_BEGIN
+
+// TODO: Potential Optimization for padding??
+// EIP-4895
+struct Withdrawal
+{
+    uint64_t index{0};
+    uint64_t validator_index{};
+    address_t recipient{};
+    uint64_t amount{}; // this should never be 0
```
sheapwilliams: Is you reverse these two, then amount will be 64-bit aligned.

src/monad/rlp/decode_helpers.cpp
sheapwilliams: I don't think we want to run-time assert if we encounter abnormal situations - we need to be more robust than that.  Can it just be an early exit, perhaps a log?

src/monad/rlp/decode_helpers.cpp
```diff
@@ -321,6 +321,41 @@ byte_string_view decode_receipt(Receipt &receipt, byte_string_view const enc)
...
+    return rest_of_enc;
+}
+
+byte_string_view decode_withdrawal_list(
+    std::vector<Withdrawal> &withdrawal_list, byte_string_view const enc)
+{
+    byte_string_view payload{};
+    auto const rest_of_enc = parse_list_metadata(payload, enc);
+
+    withdrawal_list.reserve(payload.size() / 48);
```
sheapwilliams: 48 here is `sizeof(Withdrawl)`, right?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -40,6 +40,8 @@ namespace fork_traits
     struct istanbul;
     struct berlin;
     struct london;
+    struct paris;
+    struct shanghai;
 
     using no_next_fork_t = london;
```
rgarc: you might need to change this to the last fork.

tongzhi1998: Nice catch!

PR #285, opened 2023-08-25
:new: add in capability to specify fork when running ethereum tests

zander-xyz:
Problem:
- each ethereum test is ran for multiple forks, and we have no way of specifying which fork we care about

Solution:
- Add in a command line argument 'fork' which allows us to specify fork

This is implemented as a command line options as opposed to registering several tests with GTEST representing different forks to avoid the expensive json parsing at registration time, which gets executed every time the binary is ran .

### DIFF
--- a/test/ethereum_test/src/main.cpp
+++ b/test/ethereum_test/src/main.cpp
@@ -43,6 +43,7 @@ int main(int argc, char *argv[])
     monad::log::level_t evmone_baseline_interpreter_log_level =
         quill::LogLevel::None;
     monad::log::level_t state_log_level = quill::LogLevel::None;
+    std::optional<size_t> fork_index = std::nullopt;
 
     // The default test filter. To enable all tests use `--gtest_filter=*`.
     testing::FLAGS_gtest_filter =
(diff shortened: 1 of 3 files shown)


test/ethereum_test/src/main.cpp
```diff
@@ -88,6 +87,10 @@ int main(int argc, char *argv[])
     log_levels->add_option("--state", state_log_level, "Log level for state")
         ->transform(CLI::CheckedTransformer(log_levels_map, CLI::ignore_case));
 
+    app.add_option("--fork", fork_index, "Fork to run unit tests for")
+        ->transform(CLI::CheckedTransformer(
+            monad::test::fork_index_map, CLI::ignore_case));
+
```
sheapwilliams, line 93: The only observation is that it might be nice to print the map on `--help` to see what the available forks are.  Not for this PR, but perhaps later.

zander-xyz: It does print it!
```
monad ethereum tests runner
Usage: ./test/ethereum_test/monad-ethereum-test [OPTIONS] [SUBCOMMAND]

Options:
  -h,--help                   Print this help message and exit
  --fork UINT:value in {Berlin->8,Istanbul->7,Constantinople->6,Byzantium->5,EIP158->4,London->9,Homestead->1,Frontier->0} OR {8,7,6,5,4,9,1,0}
                              Fork to run unit tests for

Subcommands:
  log_levels                  level of logging
```

PR #289, opened 2023-08-28
🐛 do not execute transaction and verify during txn failture

### LINKED ISSUE #266: Handle invalid transactions in ethereum tests
When executing tests from the ethereum-tests suite, we must validate transactions that are parsed. We currently do not handle cases where validating the transaction fails, and just execute the transaction regardless.

zander-xyz:
closes #266 

Problem:
- Currently txn still executes when validation fails, which is erroenous

Solution:
- do not execute transaction if validation fails and validate that the
  ethereum test expects an error

### DIFF
--- a/include/monad/execution/test/fakes.hpp
+++ b/include/monad/execution/test/fakes.hpp
@@ -464,6 +464,12 @@ namespace fake
                 std::optional<std::vector<Withdrawal>> const &) noexcept
             {
             }
+
+            [[nodiscard]] static constexpr bool
+            access_list_valid(Transaction::AccessList const &)
+            {
+                return true;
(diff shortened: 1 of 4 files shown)


include/monad/execution/test/fakes.hpp
```diff
@@ -449,6 +449,12 @@ namespace fake
             transfer_balance_dao(TState &, block_num_t const)
             {
             }
+
+            [[nodiscard]] static constexpr bool
+            access_list_valid(Transaction::AccessList const &)
+            {
+                return false;
```
sheapwilliams: Prolly want the fake to return true...

PR #290, opened 2023-08-28
:bug: make ethereum tests open a unique persistent store on every run

zander-xyz:
Problem:
- The same persistent store is shared across runs and also between transactions within a single test

Solution:
- Start from clean database state by changing the directory structure to include epoch time and case index

Thanks to @rgarc for noticing.

### DIFF
--- a/test/ethereum_test/src/ethereum_test.cpp
+++ b/test/ethereum_test/src/ethereum_test.cpp
@@ -260,12 +260,14 @@ void EthereumTests::run_state_test(
             }();
 
             monad::execution::fake::BlockDb fake_block_db;
-            db_t db{
-                db::Writable{},
-                monad::test_resource::build_dir / "rocksdb" / suite_name /
-                    test_name / fork_name,
-                std::nullopt,
-                0};
+
+            static auto const time = fmt::format(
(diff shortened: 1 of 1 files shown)


test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -263,12 +263,14 @@ void EthereumTests::run_state_test(
...
             monad::execution::fake::BlockDb fake_block_db;
-            db_t db{
-                db::Writable{},
-                monad::test_resource::build_dir / "rocksdb" / suite_name /
-                    test_name / fork_name,
-                std::nullopt,
-                0};
+
+            static auto const time = fmt::format(
+                "{}", std::chrono::system_clock::now().time_since_epoch());
```
rgarc, line 265: If we want all tests to have the same epoch, we might wanna generate `std::chrono::system_clock::now().time_since_epoch()` once somewhere in `main()` and pipe it through all the way to this `run_state_test` function.

zander-xyz: time has static storage duration

rgarc: cool yeah I missed that, and verified one directory gets made per run.

PR #292, opened 2023-08-29
:new: add ability to specify txn index in ethereum tests

zander-xyz:
Problem:
- Currently there is no way to specify the specific transaction to run

Solution:
- Add command line option to specify transaction by index

Also requested by @rgarc for his scripting work.

### DIFF
--- a/test/ethereum_test/src/ethereum_test.cpp
+++ b/test/ethereum_test/src/ethereum_test.cpp
@@ -117,7 +117,7 @@ using execution_variant =
 
 void EthereumTests::register_test(
     std::string const &suite_name, std::filesystem::path const &file,
-    std::optional<size_t> fork_index)
+    std::optional<size_t> fork_index, std::optional<size_t> txn_index)
 {
     testing::RegisterTest(
         suite_name.c_str(),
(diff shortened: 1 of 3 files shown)


test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -215,11 +219,25 @@ StateTransitionTest EthereumTests::load_state_test(
...
-             expectations.get<std::vector<Case::Expectation>>()});
+
+        if (txn_index.has_value()) {
+            if (txn_index.value() < expectations.size()) {
+                test_cases.emplace_back(Case{
+                    .fork_index = maybe_fork_index.value(),
+                    .fork_name = revision_name,
+                    .expectations = {expectations.at(txn_index.value())
+                                         .get<Case::Expectation>()}});
+            }
```
sheapwilliams, line 230: Is this a silent dropping of an incorrect txn id?

zander-xyz: It'll just skip the test if it doesn't contain a transaction with a specific index. It's clear when you run gtest what's happening (SKIPPING will appear in all caps). So not really "silently dropping", but some version of it I suppose. This is the same behavior as fork.

PR #293, opened 2023-08-29
Serialize Database state to JSON.

rgarc:
This PR implements state serialization to JSON for the four database variants.

Example command
```
$ ./cmake-build-debug/test/ethereum_test/monad-ethereum-test \
        --gtest_catch_exceptions=0 \
        --gtest_filter=VMTests/vmArithmeticTest.add \
        --dump_state \
        --fork Berlin \
        --txn 0 \
        log_levels \
        --state debug \
        --change_set debug \
        --trie_db debug \
        --evmone debug \
        --ethereum_test debug\
        | grep post_state | cut -c 105- | jq
```
Example output
```json
{
  "0x03601462093b5945d1676df093446790fd31b20e7b12a2e8e5e09d068109616b": {
    "balance": "0xba1a9ce0b9aa781",
    "code": "0x",
    "nonce": "0x1"
  },
  "0x227a737497210f7cc2f464e3bfffadefa9806193ccdf873203cd91c8d3eab518": {
    "balance": "0xba1a9ce0ba1a9ce",
    "code": "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0160005500",
    "nonce": "0x0",
    "storage": {
      "0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563": "0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe",
      "original_account_address": "0x0000000000000000000000000000000000000100"
    }
  },
  "0x4599828688a5c37132b6fc04e35760b4753ce68708a7b7d4d97b940047557fdb": {
    "balance": "0xba1a9ce0ba1a9ce",
    "code": "0x60047fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0160005500",
    "nonce": "0x0"
  },
  "0x4c933a84259efbd4fb5d1522b5255e6118da186a2c71ec5efaa5c203067690b7": {
    "balance": "0xba1a9ce0ba1a9ce",
    "code": "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff60010160005500",
    "nonce": "0x0"
  },
  "0x9d860e7bb7e6b09b87ab7406933ef2980c19d7d0192d8939cf6dc6908a03305f": {
    "balance": "0x7024c",
    "code": "0x",
    "nonce": "0x0"
  },
  "0xa17eacbc25cda025e81db9c5c62868822c73ce097cee2a63e33a2e41268358a1": {
    "balance": "0xba1a9ce0ba1a9ce",
    "code": "0x60017fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0160005500",
    "nonce": "0x0"
  },
  "0xa5cc446814c4e9060f2ecb3be03085683a83230981ca8f19d35a4438f8c2d277": {
    "balance": "0xba1a9ce0ba1a9ce",
    "code": "0x600060000160005500",
    "nonce": "0x0"
  },
  "0xf057b39b049c7df5dfa86c4b0869abe798cef059571a5a1e5bbf5168cf6c097b": {
    "balance": "0xba1a9ce0ba1a9cf",
    "code": "0x600060006000600060006004356101000162fffffff100",
    "nonce": "0x0"
  }
}
```

### DIFF
--- a/test/ethereum_test/src/ethereum_test.cpp
+++ b/test/ethereum_test/src/ethereum_test.cpp
@@ -1,6 +1,7 @@
 #include <ethereum_test.hpp>
 #include <monad/db/rocks_trie_db.hpp>
 #include <monad/logging/monad_log.hpp>
+#include <monad/test/dump_state_from_db.hpp>
 #include <test_resource_data.h>
 
 #include <gtest/gtest.h>
@@ -315,6 +316,10 @@ void EthereumTests::run_state_test(
             state.apply_block_reward(block_header.beneficiary, 0);
             state.commit();
 
+            MONAD_LOG_INFO(
+                logger,
+                "post_state: {}",
+                monad::test::dump_state_from_db(state.db_).dump());
             MONAD_LOG_INFO(
                 logger,
                 "finished transaction index: {} revision: {}, state_root: {}",


test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
         .cases = std::move(test_cases)};
 }
 
+nlohmann::json dump_accounts_from_rocks_db(db_t &rocks_db)
```
rgarc: This file is starting to get chunky. Is it worth consolidating serialization code somewhere else?

zander-xyz: Discussed in person, but the intuition is definitely right. It'd be great to have this in a separate file, probably within monad/test/unit/common so that it's obvious to future developers that this function should only be used for debugging purposes.

Also, this function probably deserves a unit test just to make sure we are outputting the right thing.

Side note: we probably also want support for the other three variants of databases (InMemoryTrieDb, InMemoryDb, RocksDb)

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,6 +244,158 @@ StateTransitionTest EthereumTests::load_state_test(
...
+        auto const key_nibbles = monad::trie::deserialize_nibbles(it->key());
+
+        auto leaf_cursor = rocks_db.accounts_trie.make_leaf_cursor();
+        auto trie_cursor = rocks_db.accounts_trie.make_trie_cursor();
+
+        auto const [key, exists] =
+            monad::trie::get_trie_key_of_leaf(key_nibbles, leaf_cursor);
+
+        // we are walking this, it better exist >:(
+        MONAD_ASSERT(exists);
```
zander-xyz: 👍 great assert

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+            rlp::decode_account(account, _, std::get<trie::Leaf>(node).value);
+        MONAD_ASSERT(rest.empty());
+
+        auto const keccaked_account_hex = [&]() {
+            monad::bytes32_t res;
+            auto key_slice = it->key();
+            key_slice.remove_prefix(1);
+            std::memcpy(&res, key_slice.data(), sizeof(monad::bytes32_t));
+            return fmt::format("0x{}", evmc::hex(res));
+        }();
```
zander-xyz: Why not just use `key_nibbles` directly?

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+        }
+
+        if (account.nonce) {
+            state[keccaked_account_hex]["nonce"] =
+                "0x" + intx::to_string(monad::uint256_t{account.nonce});
+        }
+
+        if (account.code_hash != monad::NULL_HASH) {
+            auto code = rocks_db.read_code(account.code_hash);
+            state[keccaked_account_hex]["code"] = "0x" + evmc::hex(code);
```
zander-xyz: Why not just write the balance, nonce, and code hash regardless, and let the python parsing script handle the details of how it does the diff to geth?

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+
+    auto handle = rocks_db.db;
+    std::unique_ptr<rocksdb::Iterator> it{
+        handle->NewIterator({}, rocks_db.accounts_trie.leaves_cursor.cf_)};
+
+    for (it->SeekToFirst(); it->Valid(); it->Next()) {
+
+        auto const key_nibbles = monad::trie::deserialize_nibbles(it->key());
+
+        auto leaf_cursor = rocks_db.accounts_trie.make_leaf_cursor();
```
zander-xyz: We can just create this once and reuse it for `get_trie_key_of_leaf`

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+        auto trie_cursor = rocks_db.accounts_trie.make_trie_cursor();
```
zander-xyz: Same here, we don't need to recreate the cursor every time.

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+
+        MONAD_ASSERT(*key_slice.data() == 2 * sizeof(monad::bytes32_t));
+        key_slice.remove_prefix(1);
+
+        monad::bytes32_t keccaked_storage_key;
+        std::memcpy(
+            &keccaked_storage_key, key_slice.data(), sizeof(monad::bytes32_t));
+
+        auto leaf_cursor = rocks_db.storage_trie.make_leaf_cursor();
+        auto trie_cursor = rocks_db.storage_trie.make_trie_cursor();
```
zander-xyz: same comment about reusing cursors

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+
+        auto const keccaked_storage_key_nibbles =
+            monad::trie::deserialize_nibbles(key_slice);
+
+        MONAD_ASSERT(*key_slice.data() == 2 * sizeof(monad::bytes32_t));
+        key_slice.remove_prefix(1);
+
+        monad::bytes32_t keccaked_storage_key;
+        std::memcpy(
+            &keccaked_storage_key, key_slice.data(), sizeof(monad::bytes32_t));
```
zander-xyz: So this is just used for the fmt? Why not use `keccaked_storage_key_nibbles ` since you already have it?

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+        std::copy_n(
+            zeroless.data(),
+            zeroless.size(),
+            std::next(
+                std::begin(storage_value.bytes),
+                static_cast<uint8_t>(sizeof(bytes32_t) - zeroless.size())));
+        MONAD_ASSERT(storage_value != bytes32_t{});
+
+        auto const keccaked_account_address = std::bit_cast<monad::bytes32_t>(
+            ethash::keccak256(account_address.bytes, sizeof(monad::address_t)));
```
zander-xyz: yeah...probably a good call to do this so that the account and storage json matches up. Since we have the unhashed address as well, do you think it's worth writing that as a separate key at the same level as `storage`? Might help with debugging.

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+        if (account.code_hash != monad::NULL_HASH) {
+            auto code = rocks_db.read_code(account.code_hash);
+            state[keccaked_account_hex]["code"] = "0x" + evmc::hex(code);
+        }
+    }
+
+    return state;
+}
+
+nlohmann::json dump_storage_from_rocks_db(db_t &rocks_db)
```
zander-xyz: We should try to use `trie_db_read_storage` to avoid as much duplicate parsing code as possible.

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -244,9 +248,162 @@ StateTransitionTest EthereumTests::load_state_test(
...
+        MONAD_ASSERT(value.has_value());
+
+        auto const node = trie::deserialize_node(key, *value);
+        MONAD_ASSERT(std::holds_alternative<trie::Leaf>(node));
+
+        Account account;
+        bytes32_t _;
+        auto const rest =
+            rlp::decode_account(account, _, std::get<trie::Leaf>(node).value);
+        MONAD_ASSERT(rest.empty());
```
zander-xyz: Much of this logic is identical to `trie_db_read_account`, so we should just use that. Though we'll need to add in an overload to take in the keccak instead of the 20 byte address

test/ethereum_test/src/ethereum_test.cpp
rgarc: I think that's a good idea.

include/monad/db/trie_db_read_storage.hpp
```diff
@@ -57,4 +55,19 @@ template <typename TCursor>
     return ret;
 }
 
-MONAD_NAMESPACE_END
\ No newline at end of file
+template <typename TCursor>
+[[nodiscard]] bytes32_t trie_db_read_storage(
+    address_t const &a, uint64_t _, bytes32_t const &k, TCursor &&leaves_cursor,
```
zander-xyz: Eh, yeah we don't use incarnation here but we do pass it on to another function, so maybe let's just `%s/_/incarnation` here so that I don't have to go to the call site to figure out what it is if we do decide to care in the future

test/ethereum_test/src/main.cpp
```diff
@@ -93,6 +94,8 @@ int main(int argc, char *argv[])
             monad::test::fork_index_map, CLI::ignore_case));
 
     app.add_option("--txn", txn_index, "Index of transaction to run");
+    app.add_flag(
+        "--dump_state", dump_state, "Toggle dumping state from database");
```
zander-xyz: We control other logs by logging them at the appropriate level and then specifying the level on the command line to control whether or not we see them. Any reason why you opted to go the boolean approach here?

test/unit/common/CMakeLists.txt
```diff
@@ -10,3 +10,17 @@ target_include_directories(
...
+  monad-dump-state-from-rocks-db
+  SOURCES
+  ${PROJECT_SOURCE_DIR}/test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
+  ${PROJECT_SOURCE_DIR}/test/unit/common/src/test/dump_state_from_rocks_db.cpp
+  LIBRARIES
+  monad
+  nlohmann_json::nlohmann_json)
+target_include_directories(
+  monad-dump-state-from-rocks-db
+  PUBLIC ${PROJECT_SOURCE_DIR}/test/unit/common/include)
```
zander-xyz: - This belongs in `test/unit/common/src/test/CMakeLists.txt`.
- Specify `monad_unit_test_common` library in `add_unit_test`. That will link and do all the proper set up for you. 
- rename unit test to monad-unittestcommontests

Putting this all together...
```cpp
add_unit_test(
  TARGET
  monad-unittestcommontests
  SOURCES
  ${PROJECT_SOURCE_DIR}/test/unit/common/src/test/dump_state_from_rocks_db.cpp
  LIBRARIES
  monad_unit_test_common
  nlohmann_json::nlohmann_json)
```

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
+#include <monad/test/config.hpp>
+
+#include <monad/db/rocks_trie_db.hpp>
+
+#include <monad/logging/formatter.hpp>
+
+MONAD_TEST_NAMESPACE_BEGIN
+
+nlohmann::json dump_accounts_from_rocks_db(monad::db::RocksTrieDB &rocks_db)
```
zander-xyz: Can you please also add support for the other three database variants? It should be fairly simple, give me a call if you want to discuss.

also `inline` please

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+    for (it->SeekToFirst(); it->Valid(); it->Next()) {
+        auto const key_nibbles = monad::trie::deserialize_nibbles(it->key());
+
+        auto const key_nibbles_substring = key_nibbles.rep.substr(1);
+        monad::bytes32_t keccaked_account_address;
+        std::copy_n(
+            key_nibbles_substring.data(),
+            sizeof(monad::bytes32_t),
+            keccaked_account_address.bytes);
```
zander-xyz: Relying on the `rep` here is a bit dangerous - it is an implementation detail which can change at any time. This makes me think that the right overload probably uses `Nibbles` instead of `bytes32_t`. If you look at the overload we added, we convert straight into `Nibbles` anyways.

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+        auto const maybe_account = monad::trie_db_read_account(
+            keccaked_account_address, leaf_cursor, trie_cursor);
+
+        MONAD_ASSERT(maybe_account.has_value());
+        auto const &account = maybe_account.value();
+
+        auto const keccaked_account_hex =
+            fmt::format("0x{}", evmc::hex(key_nibbles.rep.substr(1)));
```
zander-xyz: We can just format `key_nibbles` directly, no?

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+        state[keccaked_account_hex]["balance"] =
+            "0x" + intx::to_string(account.balance, 16);
+
+        state[keccaked_account_hex]["nonce"] =
+            "0x" + intx::to_string(monad::uint256_t{account.nonce});
```
zander-xyz: format supports the conversion directly `fmt::format("0x{:x}", account.nonce);`

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+        auto code = rocks_db.read_code(account.code_hash);
```
zander-xyz:
```suggestion
        auto const code = rocks_db.read_code(account.code_hash);
```

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+        state[keccaked_account_hex]["code"] = "0x" + evmc::hex(code);
+    }
+
+    return state;
+}
+
+nlohmann::json dump_storage_from_rocks_db(monad::db::RocksTrieDB &rocks_db)
```
zander-xyz: `inline`

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+        auto const keccaked_storage_key_nibbles =
+            monad::trie::deserialize_nibbles(key_slice);
+
+        auto const keccaked_storage_key_nibbles_substring =
+            keccaked_storage_key_nibbles.rep.substr(1);
+        monad::bytes32_t keccaked_storage_key;
+        std::copy_n(
+            keccaked_storage_key_nibbles_substring.data(),
+            sizeof(monad::bytes32_t),
+            keccaked_storage_key.bytes);
```
zander-xyz: Same comment as with account

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+                trie_cursor);
+        auto const keccaked_account_address = fmt::format(
+            "{}",
+            std::bit_cast<monad::bytes32_t>(ethash::keccak256(
+                account_address.bytes, sizeof(monad::address_t))));
+        state[keccaked_account_address]["storage"]["original_account_address"] =
+            fmt::format("{}", account_address);
+
+        auto const storage_key = fmt::format(
+            "0x{}", evmc::hex(keccaked_storage_key_nibbles.rep.substr(1)));
```
zander-xyz: format keccaked_storage_key_nibbles directly

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
```diff
@@ -0,0 +1,115 @@
...
+        state[keccaked_account_address]["storage"][storage_key] =
+            "0x" + evmc::hex(storage_value);
```
zander-xyz: storage_value is directly fmtable

test/unit/common/src/test/dump_state_from_rocks_db.cpp
```diff
@@ -0,0 +1,161 @@
...
+     --gtest_filter="VMTests/vmArithmeticTest.add" \
+     --dump_state \
+     --fork Berlin --txn 0  \
+     log_levels \
+     --state debug \
+     --change_set debug \
+     --trie_db debug \
+     --evmone debug \
+     --ethereum_test debug
+     */
```
zander-xyz: What's this?

test/unit/common/src/test/dump_state_from_rocks_db.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <monad/execution/evmone_baseline_interpreter.hpp>
+#include <monad/execution/test/fakes.hpp>
+#include <monad/execution/transaction_processor.hpp>
+#include <monad/state/account_state.hpp>
+#include <monad/state/code_state.hpp>
+#include <monad/state/state.hpp>
+#include <monad/state/value_state.hpp>
+#include <monad/test/config.hpp>
+
+TEST(StateSerialization, serialize_add)
```
zander-xyz: Once we add in support for the other databases, this can be a TYPED_TEST

test/unit/common/src/test/dump_state_from_rocks_db.cpp
```diff
@@ -0,0 +1,161 @@
...
+        state.merge_changes(change_set);
+        state.apply_block_reward(beneficiary, 0);
+        state.commit();
+    }
+
+    auto const actual_payload =
+        monad::test::dump_state_from_rocks_db(state.db_);
+    // performs a deep comparison:
+    // https://github.com/nlohmann/json/blob/6cc0eaf88f14493719c440dc5c343ee70aa5551f/include/nlohmann/json.hpp#L3687-L3698
+    EXPECT_EQ(expected_payload, actual_payload);
```
zander-xyz: I do not think this unit test needs to involve the `State`. This is purely a `db` thing, so we should just interact with that.

test/ethereum_test/src/main.cpp
rgarc: I thought it would be a good feature to have granular control of printing out state rather than tying it to a logger/log-level.

test/unit/common/src/test/dump_state_from_rocks_db.cpp
rgarc: It's the command I used to generate the test code below (all the change_set mutations). I have a script that parses the logs and converts it to C++ that I can then paste into a unit test.

test/unit/common/include/monad/test/dump_state_from_rocks_db.hpp
rgarc: Ah I see. I'm used to assuming that any public member variable is fair game to use which is problematic because we use a lot of structs lol. I factored this code away to use `trie::Nibbles` directly.

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -315,6 +321,12 @@ void EthereumTests::run_state_test(
             state.apply_block_reward(block_header.beneficiary, 0);
             state.commit();
 
+            if (dump_state) {
+                MONAD_LOG_INFO(
+                    logger,
+                    "post_state: {}",
+                    monad::test::dump_state_from_db(state.db_).dump());
+            }
```
zander-xyz: I observe that this only dumps state if we specify both `--dump_state` and `log_levels <blah> INFO` on the command line. What are your thoughts on just relying on the log level itself to display these logs? That might be more ergonomic.

test/unit/common/include/monad/test/dump_state_from_db.hpp
```diff
@@ -0,0 +1,251 @@
...
+#include <monad/test/config.hpp>
+
+MONAD_TEST_NAMESPACE_BEGIN
+
+template <typename THasBytes>
+[[nodiscard]] inline bytes32_t hash(THasBytes const &hashable)
+{
+    return std::bit_cast<bytes32_t>(
+        ethash::keccak256(hashable.bytes, sizeof(hashable.bytes)));
+}
```
zander-xyz: Can you put this (and all helper functions in this file) inside of a `detail` namespace please?

test/unit/common/include/monad/test/dump_state_from_db.hpp
```diff
@@ -0,0 +1,251 @@
...
+void dump_storage_from_trie(
+    nlohmann::json &state, byte_string_view key_slice, TCursor &leaf_cursor,
+    TCursor &trie_cursor)
+{
+    monad::address_t account_address;
+    std::memcpy(&account_address, key_slice.data(), sizeof(monad::address_t));
+    key_slice.remove_prefix(sizeof(monad::address_t));
+
+    auto const [keccaked_storage_key_nibbles, _] =
+        monad::trie::deserialize_nibbles(key_slice);
```
zander-xyz: We should also assert that `num_bytes` (`_`, here) is `key_slice.size()`

test/unit/common/include/monad/test/dump_state_from_db.hpp
```diff
@@ -0,0 +1,251 @@
...
+    state[keccaked_account_address]["storage"]["original_account_address"] =
+        fmt::format("{}", account_address);
+
+    auto const storage_key = fmt::format("{}", keccaked_storage_key_nibbles);
+    state[keccaked_account_address]["storage"][storage_key] =
+        fmt::format("{}", storage_value);
+}
+
+template <typename TDatabase>
+void dump_accounts_from_memory(
```
zander-xyz: this function seems inaccurate - it's used not only for the InMemoryDb but also the RocksDb

test/unit/common/include/monad/test/dump_state_from_db.hpp
```diff
@@ -0,0 +1,251 @@
...
+    TDatabase &db, nlohmann::json &state, address_t address,
+    Account const &account)
+{
+    auto const keccaked_address_hex = fmt::format("{}", hash(address));
```
zander-xyz: I'm impartial, but wouldn't it be helpful for your script to also publish `address` as well?

test/ethereum_test/src/ethereum_test.cpp
rgarc: Oh yeah that's a good point. I'll log it on INFO and revert all the CLI changes. Can always raise the log level later.

test/unit/common/include/monad/test/dump_state_from_db.hpp
rgarc: ~Not sure where the unhashed storage key would go.~

rgarc: It's a little confusing, but the unhashed address is being published when we serialize storage.

zander-xyz: What if the account has no storage?

rgarc: If the account has no storage, and we're serializing a trie-building database then the output will be 
```json
  "0x03601462093b5945d1676df093446790fd31b20e7b12a2e8e5e09d068109616b": {
    "balance": "0xba1a9ce0b9aa781",
    "code": "0x",
    "nonce": "0x1"
  }
```
we won't have enough information to recover the original address. what you're saying will only be helpful when serializing the non-trie-building databases in which case we'd get
```json
  "0x03601462093b5945d1676df093446790fd31b20e7b12a2e8e5e09d068109616b": {
    "balance": "0xba1a9ce0b9aa781",
    "code": "0x",
    "nonce": "0x1",
    "original_account_address": "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b"
  }
```

zander-xyz: right, is that helpful to you?

PR #302, opened 2023-08-30
:new: Receipt trie building

tongzhi1998:
- Add Receipt Trie building & storing mechanisms

### DIFF
--- a/include/monad/db/in_memory_trie_db.hpp
+++ b/include/monad/db/in_memory_trie_db.hpp
@@ -96,6 +96,7 @@ struct InMemoryTrieDB : public Db
         for (auto const &[ch, c] : obj.code_changes) {
             code[ch] = c;
         }
+
         trie_db_process_changes(
             obj, accounts_trie, storage_trie, receipts_trie);
 
--- a/include/monad/db/rocks_trie_db.hpp
+++ b/include/monad/db/rocks_trie_db.hpp
@@ -107,8 +107,9 @@ struct RocksTrieDB : public Db
             rocksdb::ColumnFamilyHandle *lc, rocksdb::ColumnFamilyHandle *tc)
             : leaves_cursor(db, lc)
             , trie_cursor(db, tc)
-            , leaves_writer(trie::RocksWriter{.batch = batch, .cf = lc})
-            , trie_writer(trie::RocksWriter{.batch = batch, .cf = tc})
+            , leaves_writer(
+                  trie::RocksWriter{.db = db, .batch = batch, .cf = lc})
+            , trie_writer(trie::RocksWriter{.db = db, .batch = batch, .cf = tc})
             , trie(leaves_cursor, trie_cursor, leaves_writer, trie_writer)
         {
         }
@@ -117,6 +118,12 @@ struct RocksTrieDB : public Db
         {
             leaves_cursor.reset();
             trie_cursor.reset();
+        }
+
+        void reset_writer()
+        {
+            leaves_writer.del_prefix({});
+            trie_writer.del_prefix({});
         }
 
         [[nodiscard]] trie::RocksCursor make_leaf_cursor() const
@@ -232,6 +239,9 @@ struct RocksTrieDB : public Db
     void commit(state::StateChanges const &obj) override
     {
         detail::rocks_db_commit_code_to_batch(batch, obj, code_cf());
+
+        receipts_trie.reset_cursor();
+        receipts_trie.reset_writer();
 
         trie_db_process_changes(
             obj, accounts_trie, storage_trie, receipts_trie);
--- a/include/monad/db/trie_db_process_changes.hpp
+++ b/include/monad/db/trie_db_process_changes.hpp
@@ -17,8 +17,6 @@ void trie_db_process_changes(
     state::StateChanges const &obj, TAccountTrie &account_trie,
     TStorageTrie &storage_trie, TReceiptTrie &receipt_trie)
 {
-    // TODO: Not sure if this is the best place to put this assertion
-    MONAD_DEBUG_ASSERT(receipt_trie.trie.leaves_cursor_.empty());
     std::unordered_map<address_t, bytes32_t> updated_storage_roots;
     std::vector<trie::Update> storage_trie_updates;
     std::vector<trie::Update> account_trie_updates;
@@ -127,7 +125,7 @@ void trie_db_process_changes(
         std::ranges::sort(
             receipt_trie_updates, std::less<>{}, trie::get_update_key);
 
-        receipt_trie.trie.process_updates(receipt_trie_updates);
+        receipt_trie.trie.process_updates_from_scratch(receipt_trie_updates);
     }
 }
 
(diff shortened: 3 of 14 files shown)


include/monad/trie/rocks_writer.hpp
```diff
@@ -55,6 +58,18 @@ struct RocksWriter
...
     }
+
+    void clear()
+    {
+        std::unique_ptr<rocksdb::Iterator> it(
+            db->NewIterator(rocksdb::ReadOptions{}, cf));
+        rocksdb::Status res;
+
+        for (it->SeekToFirst(); it->Valid(); it->Next()) {
+            res = db->Delete(rocksdb::WriteOptions{}, cf, it->key());
```
tongzhi1998: I know this is not ideal, but using batch will can't achieve the purpose of getting rid of the entry in the DB immediately, and that creates a circular dependency between {clear(), Write() and root_hash()}.

include/monad/trie/key_buffer.hpp
```diff
@@ -20,6 +20,13 @@ struct KeyBuffer
     size_t buf_size = 0;
     size_t prefix_size = 0;
 
+    void clear()
+    {
+        buf_size = 0;
+        prefix_size = 0;
+        memset(raw.data(), 0, sizeof(raw));
```
rgarc: In case you find memset error-prone, I think you could also do
```C++
raw.fill(0);
```

include/monad/db/in_memory_trie_db.hpp
```diff
@@ -41,6 +41,22 @@ struct InMemoryTrieDB : public Db
         {
         }
 
+        void reset_cursor()
+        {
+            leaves_cursor.reset();
+            trie_cursor.reset();
+            leaves_storage.clear();
+            trie_storage.clear();
+        }
```
zander-xyz: This is doing two separate things
- resetting the cursor (which is a noop so these calls are not necessary)
- resetting the storage

This function ought to be named more accurately

include/monad/db/in_memory_trie_db.hpp
```diff
@@ -41,6 +41,22 @@ struct InMemoryTrieDB : public Db
...
+
+        void clear_writer()
+        {
+            trie_writer.clear();
+            leaves_writer.clear();
+        }
+
+        void clear_buffer() { trie.clear_buffer(); }
```
zander-xyz: why are these needed?

include/monad/db/rocks_trie_db.hpp
```diff
@@ -114,6 +120,14 @@ struct RocksTrieDB : public Db
             trie_cursor.reset();
         }
 
+        void clear_writer()
+        {
+            leaves_writer.clear();
+            trie_writer.clear();
+        }
+
+        void clear_buffer() { trie.clear_buffer(); }
```
zander-xyz: ditto question - why is this needed?

include/monad/db/trie_db_process_changes.hpp
```diff
@@ -110,6 +116,19 @@ void trie_db_process_changes(
...
+
+    // Receipts Trie Building
+    if (!obj.receipts.empty()) {
+        for (auto i = 0u; i < obj.receipts.size(); ++i) {
+            receipt_trie_updates.emplace_back(trie::Upsert{
+                .key = trie::Nibbles{rlp::encode_unsigned(i)},
+                .value = rlp::encode_receipt(obj.receipts[i])});
+        }
+        std::ranges::sort(
+            receipt_trie_updates, std::less<>{}, trie::get_update_key);
```
zander-xyz, line 126: I have not looked too in depth into how we generate receipts - is it necessary to sort (ie. does it come presorted)?

include/monad/state/state.hpp
```diff
@@ -306,15 +307,22 @@ struct State
                code_.can_commit();
     }
 
+    void add_receipts(std::vector<Receipt> &receipts)
+    {
+        receipts_ = std::move(receipts);
+    }
```
zander-xyz: if the intention is to transfer ownership, just take in `receipts` by rvalue reference to make it clear that this is happening (otherwise risk dangling pointer errors)

include/monad/execution/replay_block_db.hpp
```diff
@@ -158,18 +157,10 @@ class ReplayFromBlockDb
...
-
-                auto const transactions_root = transaction_trie.root_hash();
-                auto const receipts_root = receipt_trie.root_hash();
-
                 if (!verify_root_hash(
                         block.header,
-                        transactions_root,
-                        receipts_root,
+                        NULL_ROOT, // txn root
+                        state.get_receipt_hash(),
```
zander-xyz, line 163: I don't see us using `add_receipt` anywhere except for int he unit tests. So this `verify` will always just log a mismatch. Can we add this in?

include/monad/trie/in_memory_writer.hpp
```diff
@@ -67,6 +67,14 @@ struct InMemoryWriter
         prefix_deletions.clear();
         deletions.clear();
     }
+
+    void clear()
+    {
+        storage.clear();
+        upserts.clear();
+        prefix_deletions.clear();
+        deletions.clear();
```
zander-xyz: This function should just be `prefix_deletions.emplace({})`, and then let `write` actually do the deletion when that gets called

include/monad/trie/key_buffer.hpp
```diff
@@ -20,6 +20,13 @@ struct KeyBuffer
     size_t buf_size = 0;
     size_t prefix_size = 0;
 
+    void clear()
+    {
+        buf_size = 0;
+        prefix_size = 0;
+        raw.fill(0);
+    }
```
zander-xyz: why is this needed?

include/monad/trie/key_buffer.hpp
```diff
@@ -31,7 +38,13 @@ struct KeyBuffer
...
     constexpr void set_prefix(byte_string_view prefix) noexcept
     {
-        MONAD_DEBUG_ASSERT(prefix.size() == sizeof(address_t));
+        MONAD_DEBUG_ASSERT(
+            prefix.size() == sizeof(address_t) || prefix.size() == 0);
+        if (prefix.size() == 0) {
+            prefix_size = 0;
+            buf_size = 0;
+            return;
+        }
```
zander-xyz, line 40: what is this?

include/monad/trie/rocks_cursor.hpp
```diff
@@ -26,6 +26,8 @@ struct RocksCursor
     rocksdb::ReadOptions read_opts_;
     mutable KeyBuffer buf_;
 
+    bool force_reset = false;
```
zander-xyz: I don't think these are the changes that I meant, let's discuss in person.

include/monad/trie/trie.hpp
```diff
@@ -57,6 +57,8 @@ struct Trie
     {
     }
 
+    constexpr void clear_buffer() { buf_.clear(); }
```
zander-xyz: why?

include/monad/execution/replay_block_db.hpp
tongzhi1998: Discussed on Zoom, decided to punt on this until state2 comes out

PR #306, opened 2023-08-31
:sparkles: Add Code storage to the BlockState

sheapwilliams:
Problem:
- We need to capture the code used by a transaction, and merge it back into the block state

Solution:
- Add Code storage to the BlockState
- Provide merge function to merge that back into the BlockState

### DIFF
--- a/include/monad/state2/state.hpp
+++ b/include/monad/state2/state.hpp
@@ -4,6 +4,7 @@
 #include <monad/core/account.hpp>
 #include <monad/core/address.hpp>
 #include <monad/core/bytes.hpp>
+#include <monad/core/byte_string.hpp>
 
 #include <ankerl/unordered_dense.h>
 
@@ -39,14 +40,23 @@ using State = ankerl::unordered_dense::segmented_map<address_t, AccountState>;
 static_assert(sizeof(State) == 64);
 static_assert(alignof(State) == 8);
 
+using Code =
+    ankerl::unordered_dense::segmented_map<bytes32_t, byte_string>;
+
+static_assert(sizeof(Code) == 64);
+static_assert(alignof(Code) == 8);
+
 template <class Mutex>
 struct BlockState
 {
     Mutex mutex;
     State state;
+    Code code;
 };
 
 bool can_merge(State &to, State const &from);
 void merge(State &to, State const &from);
+
+void merge(Code &to, Code &from);
 
 MONAD_NAMESPACE_END
--- a/src/monad/state2/state.cpp
+++ b/src/monad/state2/state.cpp
@@ -61,4 +61,12 @@ void merge(State &s1, State const &s2)
     });
 }
 
+void merge(Code &c1, Code &c2)
+{
+    merge_f(c1, c2, [](auto &d1, auto &d2) {
+        if (d1.empty())
+            d1 = std::move(d2);
+        });
+}
+
 MONAD_NAMESPACE_END
--- a/src/monad/state2/state_ops.cpp
+++ b/src/monad/state2/state_ops.cpp
@@ -100,11 +100,50 @@ delta_t<bytes32_t> &read_storage(
     return it2->second;
 }
 
+template <class Mutex>
+byte_string &read_code(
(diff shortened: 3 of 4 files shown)


src/monad/state2/state.cpp
```diff
@@ -61,4 +61,16 @@ void merge(State &s1, State const &s2)
     });
 }
 
+bool can_merge(CodeStore const &c1, CodeStore const &c2)
+{
+    return subset_f(c1, c2, [](auto const &d1, auto const &d2) {
+        return d1 == d2 || d1.empty() || d2.empty();
```
sheapwilliams: I was wondering if the right thing here is:

```suggestion
        return d1 == d2 || (d1.empty() ^ d2.empty());
```

...discuss.

include/monad/state2/state.hpp
```diff
@@ -39,11 +40,18 @@ using State = ankerl::unordered_dense::segmented_map<address_t, AccountState>;
...
+
 template <class Mutex>
 struct BlockState
 {
     Mutex mutex;
     State state;
+    CodeStore code;
 };
 
 bool can_merge(State &to, State const &from);
```
sheapwilliams, line 57: Need declarations of new merge fcns...

src/monad/state2/state.cpp
```diff
@@ -61,4 +61,24 @@ void merge(State &s1, State const &s2)
...
 
+bool can_merge(CodeDeltas const &c1, CodeDeltas const &c2)
+{
+    return subset_f(c1, c2, [](auto const &d1, auto const &d2) {
+        MONAD_DEBUG_ASSERT(!(d2.first == true && d2.first == false));
+        if (d1.first && d2.first) {
+            MONAD_DEBUG_ASSERT(d1 == d2);
+            return true;
+        }
+        return d1.second.empty() && !d2.second.empty();
```
sheapwilliams: This guy isn't right - it won't allow {empty, empty}, which is an ok thing I think.

src/monad/state2/state_ops.cpp
```diff
@@ -100,11 +100,50 @@ delta_t<bytes32_t> &read_storage(
...
+            auto const &result = it->second;
+            auto const [it, _] = code.try_emplace(hash, result);
+            return it->second.second;
+        }
+    }
+    // database
+    auto result = db.read_code(hash);
+    {
+        std::lock_guard<Mutex> const lock{block_state.mutex};
+        auto const [it, inserted] = block_state.code.try_emplace(hash, result.empty(), result);
```
sheapwilliams: The bool is super weird - because we don't have any other interface with the db, it just means "is the byte_string empty", and we already have that information...

include/monad/state2/state.hpp
```diff
@@ -39,14 +40,24 @@ using State = ankerl::unordered_dense::segmented_map<address_t, AccountState>;
...
     Mutex mutex;
     State state;
+    Code code;
 };
 
 bool can_merge(State &to, State const &from);
 void merge(State &to, State const &from);
 
+bool can_merge(Code &to, Code const &from);
+void merge(Code &to, Code const &from);
```
sheapwilliams: Wrong `const` here... in work.

src/monad/state2/state_ops.cpp
```diff
@@ -100,11 +100,50 @@ delta_t<bytes32_t> &read_storage(
...
+    auto result = db.read_code(hash);
+    {
+        std::lock_guard<Mutex> const lock{block_state.mutex};
+        auto const [it, inserted] = block_state.code.try_emplace(hash, result);
+        if (MONAD_UNLIKELY(!inserted)) {
+            result = it->second;
+        }
+    }
+    auto const [it, _] = code.try_emplace(hash, result);
+    return it->second;
```
rgarc, line 136: The `byte_string&` references in an `ankerl::unordered_dense::segmented_map<bytes32_t, byte_string>` are not stable. Is this on purpose?

sheapwilliams: Yea - the only place this is called is in the getter function of the EVMC Host Interface, and they immediately copy it out, or set the byte_string to something new if on creation.

PR #308, opened 2023-08-31
:sparkles: Add EVMC compatible state2

sheapwilliams:
Just for early review and feedback...

I've programmed a compatible EVMC state object based on state2.  I'm currently missing:

- mergeability tests (merge model is slightly different, so will require some adaptation)
- audit of existing *_state tests to make sure everything is accounted for
- in-order merging uses txn id currently, and that needs to evolve to the promise/future based approach.

It will require some adaptation in the other execution objects to work, but this is mostly around how a new revertible context is started/ended.

Lemme know your thoughts!

### DIFF
--- a/include/monad/state2/state.hpp
+++ b/include/monad/state2/state.hpp
@@ -1,0 +1,397 @@
+#pragma once
+
+#include <monad/core/account.hpp>
+#include <monad/core/address.hpp>
+#include <monad/core/assert.h>
+#include <monad/core/bytes.hpp>
+#include <monad/core/likely.h>
+#include <monad/core/receipt.hpp>
+
+#include <monad/db/db.hpp>
+
+#include <monad/logging/monad_log.hpp>
+
+#include <monad/state/config.hpp>
+#include <monad/state2/block_state.hpp>
+#include <monad/state2/block_state_ops.hpp>
+#include <monad/state2/state.hpp>
+
+#include <ankerl/unordered_dense.h>
+
+#include <ethash/keccak.hpp>
+
+MONAD_STATE_NAMESPACE_BEGIN
+
+// EVMC state object
+template <class Mutex, class TBlockCache>
+struct State
+{
+    BlockState<Mutex> &bs_;
+    Db &db_;
+    TBlockCache &block_cache_;
+    StateDeltas state_;
+    Code code_;
+    ankerl::unordered_dense::set<address_t> accessed_;
+    ankerl::unordered_dense::map<
+        address_t, ankerl::unordered_dense::set<bytes32_t>>
+        accessed_storage_;
+    unsigned total_selfdestructs_;
+    uint256_t gas_award_;
+    std::vector<Receipt::Log> logs_;
+
+    decltype(monad::log::logger_t::get_logger()) logger_ =
+        monad::log::logger_t::get_logger("state_logger");
+
+    explicit State(BlockState<Mutex> &b, Db &d, TBlockCache &cache)
+        : bs_{b}
+        , db_{d}
+        , block_cache_{cache}
+        , state_{}
+        , code_{}
+        , accessed_{}
+        , accessed_storage_{}
+        , total_selfdestructs_{}
+        , gas_award_{}
+        , logs_{}
+    {
+    }
+
+    // EVMC Host Interface
+    evmc_access_status access_account(address_t const &address)
+    {
+        MONAD_LOG_DEBUG(logger_, "access_account: {}", address);
+
+        auto const [_, inserted] = accessed_.insert(address);
+        if (inserted) {
+            return EVMC_ACCESS_COLD;
+        }
+        return EVMC_ACCESS_WARM;
+    }
+
+    // EVMC Host Interface
+    [[nodiscard]] bool account_exists(address_t const &address)
+    {
+        MONAD_LOG_DEBUG(logger_, "account_exists: {}", address);
+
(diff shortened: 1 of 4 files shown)


include/monad/state2/state.hpp
```diff
@@ -1,52 +1,403 @@
...
+
+    void merge(State &new_state)
+    {
+        MONAD_DEBUG_ASSERT(can_merge(*this, new_state));
+
+        state::merge(*this, new_state);
+        accessed_ = std::move(new_state.accessed_);
+        accessed_storage_ = std::move(new_state.accessed_storage_);
+        total_selfdestructs_ = new_state.total_selfdestructs_;
+        gas_award_ = new_state.gas_award_;
```
sheapwilliams: Ignore this guy - totally wrong and currently in work...

include/monad/state2/state.hpp
```diff
@@ -1,52 +1,403 @@
...
+        return account.has_value();
+    };
+
+    void create_account(address_t const &address)
+    {
+        MONAD_LOG_DEBUG(logger, "create_account: {}", address);
+
+        s.try_emplace(
+            address,
+            AccountState{.account = {std::nullopt, Account{}}, .storage = {}});
```
sheapwilliams: This guy isn't right - the account won't be in the BlockState, and subsequent storage accesses will fail.  But we don't want to fetch via `read_acount` - if there is a creation race and someone else wants to create the same account.

So we want to indicate a create explicitly, I think.  How do we either explicitly signal creation, or signal an abort if we detect somewhere we are going to collide?

include/monad/state2/state.hpp
```diff
@@ -0,0 +1,393 @@
...
+        auto const previous_balance = account.value().balance;
+        MONAD_LOG_DEBUG(
+            logger,
+            "set_balance: {} = {}, ({}{})",
+            address,
+            intx::to_string(new_balance, 16),
+            new_balance >= previous_balance ? "+" : "-",
+            new_balance >= previous_balance
+                ? intx::to_string(new_balance - previous_balance, 16)
+                : intx::to_string(previous_balance - new_balance, 16));
```
zander-xyz: This is okay too, but we do have formatters for uint* values (see `monad/logging/formatter.hpp`)

include/monad/state2/state.hpp
```diff
@@ -0,0 +1,393 @@
...
+        }
+        return false;
+    }
+
+    [[nodiscard]] unsigned total_selfdestructs() const noexcept
+    {
+        return total_selfdestructs_;
+    }
+
+    void destruct_suicides() const noexcept {}
```
zander-xyz, line 186: I noticed that we also don't do anything in `state`, are we supposed to do something? Does this need an issue (and subsequent comment)?

include/monad/state2/state.hpp
```diff
@@ -0,0 +1,393 @@
...
+
+    void merge(State &new_state)
+    {
+        s = std::move(new_state.s);
+        code_ = std::move(new_state.code_);
+        accessed_ = std::move(new_state.accessed_);
+        accessed_storage_ = std::move(new_state.accessed_storage_);
+        total_selfdestructs_ = new_state.total_selfdestructs_;
+        gas_award_ = new_state.gas_award_;
+        logs_ = std::move(new_state.logs_);
```
zander-xyz: Why are these moved as opposed to merged? Some detail of the ethereum spec?

src/monad/state2/test/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_unit_test(
+    # ENABLE_LOGGING
+    TARGET
+    monad-state2tests
+    SOURCES
+    #account_state.cpp
+    #code_state.cpp
```
zander-xyz: should these commented out files be removed completely?

include/monad/state2/state.hpp
```diff
@@ -0,0 +1,393 @@
...
+
+MONAD_STATE_NAMESPACE_BEGIN
+
+// EVMC state object
+template <class Mutex, class TBlockCache>
+struct State
+{
+    BlockState<Mutex> &bs;
+    Db &db;
+    TBlockCache &block_cache_;
```
zander-xyz, line 31: When is the block_cache_ supposed to be written to? (i don't see it happening here)

include/monad/state2/state.hpp
zander-xyz: okay I think I understand this part now...sorry

include/monad/state2/state.hpp
sheapwilliams: Yea - this is an open issue.  We are supposed to only destruct suicides at the end of the txn...lemme find reference:

YP, Pg. 38 Virtual Machine Specification, Section f0s System Operations: 

> Halt execution and register account for later deletion.

Register for later deletion is the key, as the account still exists for purposes of gathering reward, etc.  But when deleted, anything left over is blackholed.  Turning into an issue.

include/monad/state2/state.hpp
sheapwilliams: It gets written to internally, if you request a valid block header hash but haven't read it yet (cause you started up not from the genesis block), then it will record it so that it is cached.

In normal operation, the cache will be populated as blocks are written, and the state won't ever have a non-const interface to this block.

We can make this const by making this populate all the block header hashes on startup, and then the only non-const usage will be in sync with the block processor...

PR #315, opened 2023-09-05
🐛 do not spawn empty beneficiary from suprious dragon and beyond 

zander-xyz:
Problem:
- In ethereum test and future forks, the block award applied is zero.
  This can result in a an empty account being committed to state, which
  post spurious dragon should be destructed. Currently we commit the
  empty account to state.

Solution:
- in spurious dragon and beyond, do not apply the reward if it is zero,
  to avoid creating the empty account

Passes `./test/ethereum_test/monad-ethereum-test --fork eip158 --gtest_filter=stExample.basefeeExample`

### DIFF
--- a/scripts/ethereum-test-filter/filter
+++ b/scripts/ethereum-test-filter/filter
@@ -1,2 +1,3 @@
 frontier -stRefundTest.refundResetFrontier
 homestead -VMTests/vmIOandFlowOperations.jumpToPush
+eip158 -VMTests/vmIOandFlowOperations.jumpToPush


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -53,19 +53,26 @@ namespace fork_traits
...
     static constexpr void apply_mining_award(
         TState &s, Block const &b, uint256_t const &reward,
         uint256_t const &ommer_reward)
     {
         // reward block beneficiary, YP Eqn. 172
-        uint256_t const miner_reward = reward + ommer_reward * b.ommers.size();
-        s.apply_block_reward(b.header.beneficiary, miner_reward);
+        auto const miner_reward =
+            reward + ommer_reward * b.ommers.size() + s.gas_award();
+        if (miner_reward || TTraits::create_empty_account()) {
```
sheapwilliams: You are already in the TTraits class...I think you can just invoke the function simply:
```suggestion
        if (miner_reward || create_empty_account()) {
```

Without adding another template parameter.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -53,19 +53,26 @@ namespace fork_traits
...
+            s.apply_reward(b.header.beneficiary, miner_reward);
+        }
 
         // reward ommers, YP Eqn. 175
         for (auto &i : b.ommers) {
             auto const subtrahend = ((b.header.number - i.number) * reward) / 8;
-            s.apply_ommer_reward(i.beneficiary, reward - subtrahend);
+            auto const ommer_reward = reward - subtrahend;
+
+            if (ommer_reward || TTraits::create_empty_account()) {
```
sheapwilliams: Same...

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -323,6 +336,19 @@ namespace fork_traits
...
+        [[nodiscard]] static constexpr bool create_empty_account()
+        {
+            return false;
+        }
+
+        template <class TState>
+        static constexpr void apply_block_award(TState &s, Block const &b)
+        {
+            apply_mining_award<TState, spurious_dragon>(
+                s, b, block_reward, additional_ommer_reward);
```
sheapwilliams: I see - then you'd have to either re-define apply_mining_award here or pass something else like a bool.... 🤔 

Maybe that is it?  Don't define a new function, but change the impl of apply_mining_award when we need to change it.  E.g. no `if`'s in the frontier version of the `apply_mining_award`, and add them in the spurious dragon version.  Thoughts?

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -351,7 +377,8 @@ namespace fork_traits
         template <class TState>
         static constexpr void apply_block_award(TState &s, Block const &b)
         {
-            apply_mining_award(s, b, block_reward, additional_ommer_reward);
+            apply_mining_award<TState, byzantium>(
+                s, b, block_reward, additional_ommer_reward);
```
sheapwilliams: We don't need to redefine these - I think this is overkill...

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -371,7 +398,8 @@ namespace fork_traits
         template <class TState>
         static constexpr void apply_block_award(TState &s, Block const &b)
         {
-            apply_mining_award(s, b, block_reward, additional_ommer_reward);
+            apply_mining_award<TState, constantinople_and_petersburg>(
+                s, b, block_reward, additional_ommer_reward);
```
sheapwilliams: Same...

PR #318, opened 2023-09-05
Run ethereum tests in CI with a filter.

rgarc:
Problem:
- We want some ethereum tests to run in CI to catch regressions, but not run the entire suite since there would be too much noise from failing tests.

Solution:
- Run tests in CI by specifying a fork and a custom filter.

### DIFF
--- a/scripts/ethereum-test-filter/filter
+++ b/scripts/ethereum-test-filter/filter
@@ -1,0 +1,2 @@
+frontier -stRefundTest.refundResetFrontier
+homestead -VMTests/vmIOandFlowOperations.jumpToPush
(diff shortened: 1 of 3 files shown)


.github/workflows/ethereum-tests-fillter.yaml
```diff
@@ -0,0 +1,54 @@
+name: Monad Ethereum Tests Filter
```
zander-xyz: can you please fix the spelling of this file? also any reason why this can not take advantage of the existing workflow? I see some replication happening between the two

test/ethereum_test/src/main.cpp
```diff
@@ -46,8 +46,13 @@ int main(int argc, char *argv[])
     std::optional<size_t> fork_index = std::nullopt;
     std::optional<size_t> txn_index = std::nullopt;
 
+    testing::InitGoogleTest(&argc, argv); // Process GoogleTest flags.
+
     // The default test filter. To enable all tests use `--gtest_filter=*`.
-    testing::FLAGS_gtest_filter =
+    if (!testing::FLAGS_gtest_filter.empty()) {
+        testing::FLAGS_gtest_filter += ":";
+    }
```
zander-xyz: I don't think you need this if statement, should work regardless of `empty` (as long as you prepend with : as you have done here).

scripts/ethereum-test-filter/filter
```diff
@@ -0,0 +1 @@
+frontier -stRefundTest.refundResetFrontier
```
zander-xyz, line 1: mind adding the homestead tests in here too? thanks!

PR #322, opened 2023-09-06
:bug: replace minus sign with underscore to allow

zander-xyz:
Problem:
- gtest uses minus sign as exclusion in the filter so test names with minus signs in them mess with gtest

Solution:
- replace all minus signs with underscores before registering

### DIFF
--- a/test/ethereum_test/src/ethereum_test.cpp
+++ b/test/ethereum_test/src/ethereum_test.cpp
@@ -6,6 +6,7 @@
 
 #include <gtest/gtest.h>
 
+#include <algorithm>
 #include <iostream>
 
 MONAD_TEST_NAMESPACE_BEGIN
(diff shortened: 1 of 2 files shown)


test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -127,9 +128,15 @@ void EthereumTests::register_test(
...
     std::optional<size_t> fork_index, std::optional<size_t> txn_index)
 {
+    // Normalize the test name so that gtest_filter can recognize it. ie.
+    // --gtest_filter=stZeroKnowledge2.ecmul_0-3_5616_21000_128 will not work
+    // because gtest interprets the minus sign as exclusion
+    auto test_name = file.stem().string();
+    std::ranges::replace(test_name, '-', '_');
+
     testing::RegisterTest(
         suite_name.c_str(),
```
rgarc, line 139: You have to do the same thing to the suite name unfortunately.
See https://github.com/ethereum/tests/tree/7e5aa011f63a9119baec27ddcc18ea18235a6047/GeneralStateTests/Shanghai/stEIP3855-push0

zander-xyz: thanks!

PR #323, opened 2023-09-06
:bug: Create beneficiary account before transferring to in selfdestruct.

rgarc:
Note:
- This is already fixed in the `state2` implementation.

Problem:
- Currently, when calling selfdestruct with a beneficiary account that doesn't exist, an assertion trips and the program crashes.

Solution:
- Before transferring the destructed account's balance to the beneficiary, we create the beneficiary account.

### DIFF
--- a/include/monad/state/account_state.hpp
+++ b/include/monad/state/account_state.hpp
@@ -245,6 +245,9 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
     selfdestruct(address_t const &a, address_t const &beneficiary) noexcept
     {
         if (get_account(a).has_value()) {
+            if (!account_exists(beneficiary)) {
+                create_account(beneficiary);
+            }
             set_balance(
                 beneficiary,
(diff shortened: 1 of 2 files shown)


include/monad/state/account_state.hpp
```diff
@@ -245,6 +245,9 @@ struct AccountState<TAccountDB>::ChangeSet : public AccountState<TAccountDB>
     selfdestruct(address_t const &a, address_t const &beneficiary) noexcept
     {
         if (get_account(a).has_value()) {
+            if (!account_exists(beneficiary)) {
+                create_account(beneficiary);
+            }
```
zander-xyz, line 250: What if the balance of a is zero and beneficiary is non-existent? Post spurious dragon this account is not supposed to exist.

sheapwilliams: It should be cleaned up when we run the `destroy_empty_touched` (or whatever we call it), right?  I think this will work correctly.

zander-xyz: Ah, that's right. I missed that this happens on the ChangeSet

PR #324, opened 2023-09-06
🐛 copy_code should return 0 if offset is larger than code size 

zander-xyz:
Problem:
- When copy_code is called for an address without code, the offset ends
  up being larger than the code size (which is 0) and one of our
  assertions gets tripped

Solution:
- Return 0 as expected and add ethereum test to filter

### DIFF
--- a/test/ethereum_test/src/ethereum_test.cpp
+++ b/test/ethereum_test/src/ethereum_test.cpp
@@ -334,7 +334,7 @@ void EthereumTests::run_state_test(
                 state.db_.state_root());
 
             auto const msg = fmt::format(
-                "fork_name: {}, transaction_index: {}", fork_name, case_index);
+                "fork_name: {}, case_index: {}", fork_name, case_index);
 
             EXPECT_EQ(state.db_.state_root(), expected.state_hash) << msg;
             EXPECT_EQ(maybe_receipt.has_value(), !expected.exception) << msg;


scripts/ethereum-test-filter/filter
```diff
@@ -2,4 +2,5 @@ frontier -stRefundTest.refundResetFrontier
 homestead -VMTests/vmIOandFlowOperations.jumpToPush
 eip158 -VMTests/vmIOandFlowOperations.jumpToPush
 byzantium -VMTests/vmIOandFlowOperations.jumpToPush
-constantinoplefix -VMTests/vmIOandFlowOperations.jumpToPush
\ No newline at end of file
+constantinoplefix -VMTests/vmIOandFlowOperations.jumpToPush
+istanbul stArgsZeroOneBalance.extcodecopyNonConst
```
zander-xyz, line 6: this one will be removed when we get all istanbul tests passing and replaced with `istanbul *`

include/monad/state/code_state.hpp
```diff
@@ -117,7 +117,11 @@ struct CodeState<TCodeDB>::ChangeSet : public CodeState<TCodeDB>
         size_t buffer_size) const
     {
         auto const code = code_at(b);
-        assert(code.size() > offset);
+
+        if (offset >= code.size()) {
+            return 0;
+        }
+
```
sheapwilliams, line 124: Looks fine, just needs a test I think here: https://github.com/monad-crypto/monad/blob/main/src/monad/state/test/code_state.cpp#L106

PR #336, opened 2023-09-08
:recycle: State2 integration

sheapwilliams:
Temporarily disable `replay_ethereum` to unblock further Ethereum test development.

### DIFF
--- a/include/monad/execution/evm.hpp
+++ b/include/monad/execution/evm.hpp
@@ -11,6 +11,7 @@
 
 #include <tl/expected.hpp>
 
+#include <limits>
 #include <optional>
 
 MONAD_EXECUTION_NAMESPACE_BEGIN
@@ -135,13 +136,13 @@ struct Evm
     [[nodiscard]] static std::optional<evmc_result>
     increment_sender_nonce(TState &s, evmc_message const &m) noexcept
     {
-        auto const n = s.get_nonce(m.sender) + 1;
-        if (s.get_nonce(m.sender) > n) {
+        auto const n = s.get_nonce(m.sender);
+        if (n == std::numeric_limits<decltype(n)>::max()) {
             // Match geth behavior - don't overflow nonce
             return evmc_result{
                 .status_code = EVMC_ARGUMENT_OUT_OF_RANGE, .gas_left = m.gas};
         }
-        s.set_nonce(m.sender, n);
+        s.set_nonce(m.sender, n + 1);
         return std::nullopt;
     }
 


test/integration/evm_state_host/evm_state_host.cpp
```diff
@@ -88,21 +85,20 @@ TEST(EvmInterpStateHost, return_existing_storage)
...
-    using state_t = decltype(changeset);
+    using state_t = decltype(s);
 
     // Prep per transaction processor
-    changeset.access_account(to);
-    changeset.access_account(from);
-    changeset.access_account(a);
+    s.access_account(to);
+    s.access_account(from);
+    s.access_account(a);
```
tongzhi1998: I might be wrong, but I think code_address might not be automatically added to the `access_list`? If this is just for testing purpose it's probably just fine tho.

test/integration/evm_state_host/evm_state_host.cpp
```diff
@@ -155,21 +148,20 @@ TEST(EvmInterpStateHost, store_then_return_storage)
...
```
tongzhi1998: Similarly here.

test/integration/evm_state_host/evm_state_host.cpp
sheapwilliams: I think you are correct.  I'll fix it.  Tests should mimic reality.

include/monad/execution/evm.hpp
```diff
@@ -74,21 +83,37 @@ struct Evm
...
         }
+
+        auto const clean_up_result = [&](auto &&result) {
+            if (result.status_code != EVMC_SUCCESS) {
+                result.gas_refund = 0;
+                if (result.status_code != EVMC_REVERT) {
+                    result.gas_left = 0;
+                }
+            }
+            return std::optional<evmc::Result>(std::move(result));
```
zander-xyz: We discussed this a while ago over zoom, but I don't think the explicit setting of refund and gas_left is necessary.

https://github.com/monad-crypto/evmone/blob/1c0aabf3a770029a27d46dad9532ac1a3fcdd81d/lib/evmone/baseline.cpp#L351-L352

include/monad/execution/evmc_host.hpp
```diff
@@ -25,6 +25,13 @@ struct EvmcHost : public evmc::Host
 
     using uint256be = evmc::uint256be;
 
+    EvmcHost(EvmcHost const &e, TState &s)
+        : block_header_{e.block_header_}
+        , transaction_{e.transaction_}
+        , state_{s}
+    {
+    }
```
zander-xyz, line 33: QQ: does this render the other constructor(s) as deprecated? If so, should we mark it as so through compiler directives and tracking?

src/monad/execution/test/evm.cpp
```diff
@@ -105,13 +132,26 @@ TEST(Evm, create_with_insufficient)
     EXPECT_EQ(result.status_code, EVMC_INSUFFICIENT_BALANCE);
 }
 
+/* This will have to work on the create_contract_account fcn, not
```
zander-xyz: Is this difficult to fix in this PR?

include/monad/execution/evm.hpp
sheapwilliams: I'll follow-up with that if that is ok.  We should have an integration test for that I think.

include/monad/execution/evmc_host.hpp
sheapwilliams: No, still used for the first one.  Can't copy an EvmcHost without having one in the first place.

src/monad/execution/test/evm.cpp
sheapwilliams: I spaced that.  I'll take a look - it is easy.

include/monad/execution/evm.hpp
sheapwilliams: I see this tested elsewhere....removed.

PR #337, opened 2023-09-08
Fix commit bug

tongzhi1998:
Problem:
- Current code pushes storage changes for deleted account into trie, which will be interpreted as deletion, causing one of the assertions inside trie to fail

Solution:
- If account.updates = null, don't push to trie

### DIFF
--- a/src/monad/db/test/db.cpp
+++ b/src/monad/db/test/db.cpp
@@ -1,6 +1,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include <monad/core/bytes.hpp>
 #include <monad/db/in_memory_db.hpp>
 #include <monad/db/in_memory_trie_db.hpp>
 #include <monad/db/rocks_db.hpp>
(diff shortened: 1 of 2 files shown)


src/monad/db/test/db.cpp
```diff
@@ -195,6 +195,43 @@ TYPED_TEST(TrieDBTest, ModifyStorageOfAccount)
...
 }
 
+TYPED_TEST(TrieDBTest, touch_without_modify_regression)
+{
+    auto db = test::make_db<TypeParam>();
+    db.commit(
+        StateDeltas{{a, StateDelta{.account = {std::nullopt, std::nullopt}}}},
+        Code{});
+
+    EXPECT_EQ(db.read_account(a), std::nullopt);
```
zander-xyz, line 206: also expect that the root hash is null, probably

src/monad/db/test/db.cpp
```diff
@@ -195,6 +195,43 @@ TYPED_TEST(TrieDBTest, ModifyStorageOfAccount)
...
+        StateDeltas{
+            {a,
+             StateDelta{
+                 .account = {acct, std::nullopt},
+                 .storage =
+                     {{key1, {value1, value2}}, {key2, {value2, value1}}}}}},
+        Code{});
+
+    EXPECT_EQ(db.read_account(a), std::nullopt);
+    EXPECT_EQ(db.read_storage(a, 0u, key1), bytes32_t{});
```
zander-xyz, line 234: same here?

PR #341, opened 2023-09-11
remove rocksdb db, not needed

jhunsaker:


include/monad/db/util.hpp
```diff
@@ -22,9 +21,6 @@ template <typename TDB>
     if constexpr (std::same_as<TDB, db::RocksTrieDB>) {
```
sheapwilliams: Do we even need the if, or rather, do we even need this fcn anymore?

src/monad/db/test/db.cpp
```diff
@@ -34,7 +33,7 @@ template <typename TDB>
 struct DBTest : public testing::Test
 {
 };
-using DBTypes = ::testing::Types<RocksDB, InMemoryTrieDB, RocksTrieDB>;
+using DBTypes = ::testing::Types<InMemoryTrieDB, RocksTrieDB>;
 TYPED_TEST_SUITE(DBTest, DBTypes);
```
tongzhi1998, line 37: Now since DBTypes is essentially TrieDBTypes, we can probably replace all TrieDBTest with just DBTest. Also maybe we could make RocksDBTest no longer a typed test 🤔

PR #342, opened 2023-09-11
:recycle: Simplfy transaction processor fiber with new state model

sheapwilliams:
Problem:
- The existing transaction processor fiber is built with a bunch of spin-loops, and uses a push model for state changes.
- We have a new pull state model.

Solution:
- Simplify the new transaction processor fiber by using the new state model
- Eliminate pushing changes, instead make them available to be pulled.

~~NOTE: pending #336~~ Merged.


src/monad/execution/test/transaction_processor_data.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include <gtest/gtest.h>
+
+#include <shared_mutex>
+
+using namespace monad;
+using namespace monad::execution;
+
+static struct fakeBlockCache
+{
+    [[nodiscard]] bytes32_t get_block_hash(int64_t) const noexcept
```
tongzhi1998: Not a big deal, but `get_block_hash` in `State2.block_cache` actually takes in a `uint64_t`.  Is it possible to change it here to match the interface?

src/monad/execution/test/transaction_processor_data.cpp
```diff
@@ -0,0 +1,143 @@
...
+    {
+        return static_cast<Status>(fake_status);
+    }
+};
+
+using tp_t = fakeGlobalStatusTP<state_t, traits_t>;
+
+TEST(TransactionProcessorFiberData, successful)
+{
+    using tp_t = fakeGlobalStatusTP<state_t, traits_t>;
```
tongzhi1998: This line might be extra? Line 79 is the same

src/monad/execution/test/transaction_processor_data.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+TEST(TransactionProcessorFiberData, failed_validation)
+{
+    auto db = test::make_db<db_t>();
+    BlockState<mutex_t> bs;
+    state_t s{bs, db, block_cache};
+    static BlockHeader const b{};
+    static Transaction t{.gas_limit = 15'000};
+
+    {
```
tongzhi1998, line 108: I understand why there are multiple almost identical tests, but it feels a bit weird (but I don't have strong opinion on this, looks fine to me as it is). Also have we decided on exactly what to do for `LATER_NONCE`?

src/monad/execution/test/transaction_processor_data.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+using tp_t = fakeGlobalStatusTP<state_t, traits_t>;
+
+TEST(TransactionProcessorFiberData, successful)
+{
+    using tp_t = fakeGlobalStatusTP<state_t, traits_t>;
+    auto db = test::make_db<db_t>();
+    BlockState<mutex_t> bs;
+    state_t s{bs, db, block_cache};
+    static BlockHeader const b{};
```
tongzhi1998, line 88: Any specific reason for both of these variables to be static?

sheapwilliams: No, not really.  Just general test hygiene.

src/monad/execution/test/transaction_processor_data.cpp
sheapwilliams: Yup.  I'll do that

src/monad/execution/test/transaction_processor_data.cpp
sheapwilliams: Yup.  Removed locally,  I'll push that change.

src/monad/execution/test/transaction_processor_data.cpp
sheapwilliams: We can remove that one if we want.  No use for it now.

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -5,123 +5,91 @@
...
         if (status == txn_processor_status_t::SUCCESS) {
-            return TxnReadyStatus::WILL_SUCCEED;
-        }
-        else if (status == txn_processor_status_t::LATER_NONCE) {
-            return TxnReadyStatus::POSSIBLY_SUCCEED;
+            return true;
         }
-        return TxnReadyStatus::ERROR;
+        return false;
     }
```
zander-xyz, line 50: any reason to keep this function around now that we've done away with the TxnReadyStatus?

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -5,123 +5,91 @@
...
 
     Receipt get_receipt() const noexcept { return result_; }
+    state_t &get_state() noexcept { return state_; }
```
zander-xyz, line 53: why did you need this?

include/monad/execution/transaction_processor_data.hpp
```diff
@@ -5,123 +5,91 @@
...
-
-                s_.merge_changes(changeset);
-                return;
-            }
-            MONAD_LOG_INFO(txn_logger, "Transaction {} rescheduled", id_);
-            TExecution::yield();
+        auto const valid = is_valid(
+            p.validate(state_, txn_, bh_.base_fee_per_gas.value_or(0)));
+        if (!valid) {
+            MONAD_LOG_INFO(txn_logger, "Transaction {} invalid!", id_);
```
zander-xyz: can we also log the status?

include/monad/execution/transaction_processor_data.hpp
tongzhi1998: I might be wrong, but I think since we are using the 'pulling' model now, at some point the `BP` will need to pull the `state` from `TPFiberData` to validate / merge?

sheapwilliams: This is the reason.  Block processor uses it to test for mergability and then merge if successful.

include/monad/execution/transaction_processor_data.hpp
sheapwilliams: Sure!

include/monad/execution/transaction_processor_data.hpp
sheapwilliams: Txn processor status is still there...we can collapse this to a bool if we want.

include/monad/execution/transaction_processor_data.hpp
sheapwilliams: A reason not to collapse to bool just yet?

PR #354, opened 2023-09-13
propagate invalid gas price error

rgarc:
Problem:
- Currently, the return type of `gas_price` is insufficient for
  propagating an error. This causes us to fail the `stEIP1559.lowFeeCap`
  test with an assertion.

Solution:
- Add a new `validate_transaction` function at the fork level.


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -160,7 +160,9 @@ namespace fork_traits
             TState &s, Transaction const &t, uint64_t base_gas_cost,
             uint64_t gas_used)
         {
-            s.add_txn_award(uint256_t{gas_used} * gas_price(t, base_gas_cost));
+            auto const maybe_gas_price = gas_price(t, base_gas_cost);
+            MONAD_ASSERT(maybe_gas_price.has_value());
+            s.add_txn_award(uint256_t{gas_used} * maybe_gas_price.value());
```
sheapwilliams: In frontier, the base_gas_cost will always be zero - we don't need this value here.  I think this is:

```
...
            TState &s, Transaction const &t, uint64_t /*base_gas_cost*/,
            uint64_t gas_used)
        {
            auto const maybe_gas_price = gas_price(t, 0);
            MONAD_ASSERT(maybe_gas_price.has_value());
            s.add_txn_award(uint256_t{gas_used} * maybe_gas_price.value());
...
```

Thoughts?

include/monad/execution/transaction_processor.hpp
```diff
@@ -116,6 +116,12 @@ struct TransactionProcessor
     Status
     validate(TState &state, Transaction const &t, uint64_t base_fee_per_gas)
     {
+        auto const transaction_validation_result =
+            TTraits::validate_transaction(t, base_fee_per_gas);
+        if (transaction_validation_result !=
+            monad::TransactionValidationResult::Ok) {
+            return Status::INVALID_GAS_LIMIT;
+        }
```
sheapwilliams: Could collapse the call into the if structure as a small simplification....

include/monad/execution/test/fakes.hpp
```diff
@@ -426,6 +426,11 @@ namespace fake
                 }
                 return r;
             }
+            static constexpr monad::TransactionValidationResult
+            validate_transaction(Transaction const &, uint64_t)
+            {
+                return {};
```
zander-xyz: please spell out Ok explicitly

PR #356, opened 2023-09-14
:recycle: Move genesis logic from ReplayBlockDb to replay_ethereum

tongzhi1998:
Problem:
- In BP, we have a conditional checking if block_number = 0, then don't apply block reward

Solution:
- Move genesis logic from ReplayBlockDb to replay_ethereum (which means block number always starts > 0 now)

### DIFF
--- a/src/monad/execution/ethereum/replay_ethereum.cpp
+++ b/src/monad/execution/ethereum/replay_ethereum.cpp
@@ -59,6 +59,7 @@ int main(int argc, char *argv[])
 
     std::filesystem::path block_db_path{};
     std::filesystem::path state_db_path{};
+    std::filesystem::path genesis_file_path{};
     uint64_t block_history_size = 1u;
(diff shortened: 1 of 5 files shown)


src/monad/execution/ethereum/replay_ethereum.cpp
```diff
@@ -53,6 +53,22 @@ class fakeEmptyReceiptTrie
 
 MONAD_NAMESPACE_END
 
+template <class TBlockDB, class TStateDB>
+void handle_genesis(TBlockDB &block_db, TStateDB &db)
+{
+    // TODO: Do we need to support functionality to specify genesis
+    // file location?
+    auto const genesis_file_path =
+        test_resource::ethereum_genesis_dir / "mainnet.json";
```
sheapwilliams: Yes, please have this file take it as a parameter.

tongzhi1998: Cool, will do

PR #357, opened 2023-09-14
🐛 create account if does not exist when transfering balance 

zander-xyz:
Problem:
- selfdestruct followed by a transfer on the same account currently
  yields an assertion error

Solution:
- create account if it does not exist


include/monad/execution/evm.hpp
```diff
@@ -94,14 +94,14 @@ struct Evm
...
-                                  })
-                                  .value();
+        evmc::Result result =
+            check_call_precompile<TTraits>(m)
+                .or_else([&] {
+                    auto const code = state.get_code(m.code_address);
+                    return std::optional<evmc::Result>(
+                        TInterpreter::execute(&new_host, m, code));
+                })
+                .value();
```
zander-xyz, line 104: there's a bug in this code which an upcoming PR will fix...

PR #358, opened 2023-09-14
Refactor txn & block reward functions

tongzhi1998:
Problem:
- apply_txn_award and apply_block_award assuems that State is shared between different txns
- However, our new State model is per txn, so the previous functions don't make much sense anymore

Solution:
- Refactor apply_block_reward function to take in BlockState instead
- Move apply_txn_award out of TP, restructure it to be just calculating txn_award
- TODO: BlockProcessor now needs to call calculate_txn_award function directly

### DIFF
--- a/include/monad/state2/state.hpp
+++ b/include/monad/state2/state.hpp
@@ -107,6 +107,10 @@ struct State
             account = Account{};
         }
 
+        MONAD_DEBUG_ASSERT(
+            std::numeric_limits<uint256_t>::max() - delta >=
+            account.value().balance);
+
         MONAD_LOG_DEBUG(
             logger_,
             "add_to_balance {} = {} + {}",
@@ -123,9 +127,10 @@ struct State
     {
         auto &account = read_account<Mutex>(address, state_, bs_, db_);
         if (!account.has_value()) {
-            MONAD_DEBUG_ASSERT(delta == 0);
             account = Account{};
         }
+
+        MONAD_DEBUG_ASSERT(delta <= account.value().balance);
 
         MONAD_LOG_DEBUG(
             logger_,


include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -340,12 +353,24 @@ namespace fork_traits
...
         {
-            apply_block_award_impl(s, b, block_reward, additional_ommer_reward);
+            apply_block_award_impl(
+                bs,
+                db,
+                block_cache,
+                b,
+                block_reward,
+                additional_ommer_reward,
+                gas_award);
```
sheapwilliams, line 376: I don't think we need this 2x dispatch structure anymore, especially here.  We should drop tie `_impl`.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -382,10 +416,19 @@ namespace fork_traits
...
```
sheapwilliams, line 434: ...same here.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -520,10 +561,19 @@ namespace fork_traits
...
+            Block const &b, uint256_t const &gas_award)
+        {
+            apply_block_award_impl(
+                bs,
+                db,
+                block_cache,
+                b,
+                block_reward,
+                additional_ommer_reward,
+                gas_award);
```
sheapwilliams, line 592: ...and same...

include/monad/core/concepts.hpp
```diff
@@ -36,15 +36,16 @@ namespace concepts
...
         { T::access_list_valid(t.access_list) } -> std::convertible_to<bool>;
 
-        // TODO: These 2 functions would require more template params for the TTraits concept
+        // TODO: These 4 functions would require more template params for the TTraits concept
         //       Comment back once decided what the new concept should be like
         // { T::transfer_balance_dao(s, block_number) } -> std::convertible_to<void>;
         // { T::process_withdrawal(s, w) } -> std::convertible_to<void>;
+        // { T::apply_block_award(s, b) } -> std::convertible_to<void>;
+        // { T::apply_txn_award(s, t, base_gas_price, gas_used) } -> std::convertible_to<void>;
+
```
sheapwilliams: Let's just remove the ones that require new template parameters, and fixup the ones that can be fixed.

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -112,8 +112,16 @@ using execution_variant =
...
-                TTraits::apply_block_award_impl(
-                    state, monad::Block{.header = block_header}, 0, 0);
+                auto const gas_used = maybe_receipt.has_value()
+                                          ? maybe_receipt.value().gas_used
+                                          : 0;
+                auto const gas_award = TTraits::calculate_txn_award(
+                    transaction,
+                    block_header.base_fee_per_gas.value_or(0),
+                    gas_used);
+                if (fork_index < 4 || gas_award != 0) { // EIP-158
```
sheapwilliams: Use the name instead of '4'

include/monad/execution/ethereum/fork_traits.hpp
tongzhi1998: I just realized that `apply_block_reward_impl` is hard to get rid of, because we have fork specific `block_reward` and `additional_ommer_reward` . The only way to remedy this is to let `apply_block_reward` taking in 7 params, but I think that's making it looking uglier when getting called :thinking_face:

test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -112,8 +112,17 @@ using execution_variant =
...
+                                          ? maybe_receipt.value().gas_used
+                                          : 0;
+                auto const gas_award = TTraits::calculate_txn_award(
+                    transaction,
+                    block_header.base_fee_per_gas.value_or(0),
+                    gas_used);
+                if (fork_index < fork_index_map.at("EIP158") ||
+                    gas_award != 0) {
+                    state.add_to_balance(block_header.beneficiary, gas_award);
+                }
```
zander-xyz, line 125: The previous block award application matches the evmone and geth impls. I understand why we need to do this right now - is the plan in the future to refactor this to use apply_block_award again?

include/monad/execution/ethereum/fork_traits.hpp
zander-xyz: Yeah when I wrote this I had the exact same thoughts...in this order lol

test/ethereum_test/src/ethereum_test.cpp
tongzhi1998: Yeah, we can refactor this in the future!

include/monad/execution/ethereum/fork_traits.hpp
tongzhi1998: Great minds think alike 💯

PR #361, opened 2023-09-15
:recycle: Use promise/future to merge result into block state

sheapwilliams:
Problem:
- Currently, when joining a fiber, the txn processor data is destroyed
- The new state model pulls results rather than having them pushed into the block state.

Solution:
- Track a set of futures that represents the txn results.
- Merge them in order, and re-run if the result is unmergable.

Note: move semantics, goto, and fibers...Oh my!  The `goto`s might seem pretty ugly at first, but I couldn't think of a different way to minimize the critical section and continue on.  Also, there is a lot of moving because promises are move-only.  Open to feedback.

- [ ] Write test for the re-processing logic

### DIFF
--- a/include/monad/execution/block_processor.hpp
+++ b/include/monad/execution/block_processor.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <monad/core/assert.h>
 #include <monad/core/block.hpp>
 #include <monad/core/receipt.hpp>
 #include <monad/core/transaction.hpp>
@@ -10,17 +11,46 @@
 
 #include <monad/logging/monad_log.hpp>
 
+#include <monad/state2/block_state.hpp>
+#include <monad/state2/state.hpp>
+#include <monad/state2/state_deltas.hpp>
+
 #include <chrono>
 #include <vector>
 
+#include <boost/fiber/all.hpp>
+
 MONAD_EXECUTION_NAMESPACE_BEGIN
 
-template <class TExecution>
+template <class TResult, class TFnObject>
+struct PromiseReturningInvocable
+{
+    boost::fibers::promise<TResult> p_;
+    TFnObject obj_;
+
+    PromiseReturningInvocable(
+        boost::fibers::promise<TResult> &&p, TFnObject &&obj)
+        : p_{std::move(p)}
+        , obj_{std::move(obj)}
+    {
+    }
+
+    void operator()()
+    {
+        obj_();
+        p_.set_value(std::move(obj_.get_result()));
+    }
+};
+
 struct AllTxnBlockProcessor
 {
-    template <class TState, class TTraits, class TFiberData>
-    [[nodiscard]] std::vector<Receipt> execute(TState &s, Block &b)
+    template <class TMutex, class TTraits, class TxnProcData, class TBlockCache>
+    [[nodiscard]] std::vector<Receipt>
+    execute(Block &b, Db &db, TBlockCache &block_cache)
     {
+        using result_t = typename TxnProcData::result_t;
+        using obj_task_t = PromiseReturningInvocable<result_t, TxnProcData>;
+
         auto *block_logger = log::logger_t::get_logger("block_logger");
         auto const start_time = std::chrono::steady_clock::now();
         MONAD_LOG_INFO(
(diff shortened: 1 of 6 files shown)


include/monad/execution/block_processor.hpp
```diff
@@ -30,36 +60,89 @@ struct AllTxnBlockProcessor
...
-        std::vector<Receipt> r{};
-        r.reserve(b.transactions.size());
-        for (auto &fiber : fibers) {
-            fiber.join();
-        }
-        for (auto &d : data) {
-            r.push_back(d.get_receipt());
+        // Merge in order, or re-run if necessary
+        for (unsigned index = 0u; auto &future : futures) {
+            future.wait();
```
rgarc: This looks to me like this `future.wait()` call blocks such that we are just executing each task serially. If that's the case, then why are we bothering with fibers? What am I missing?

sheapwilliams: In our tests this is true - everything in order.  But we can write a work-stealing fiber schedule over threads, and it will schedule any awaiting fibers to run.  So this will likely some of the fibers, but many/most will have completed on other threads and the future will be instantly readable.

Make sense?

tongzhi1998: Following up on this: I think `future.get() `is a blocking call itself? Why do we need `future.wait()` before future.get()?

include/monad/execution/block_processor.hpp
```diff
@@ -10,17 +11,46 @@
...
+        boost::fibers::promise<TResult> &&p, TFnObject &&obj)
+        : p_{std::move(p)}
+        , obj_{std::move(obj)}
+    {
+    }
+
+    void operator()()
+    {
+        obj_();
+        p_.set_value(std::move(obj_.get_result()));
```
zander-xyz, line 41: seems a bit dangerous - we're stealing the result straight from underneath of obj_
might be safer/more clear if we have get_result return an rvalue reference to release ownership?

include/monad/execution/block_processor.hpp
```diff
@@ -30,36 +60,88 @@ struct AllTxnBlockProcessor
...
 
-        std::vector<Receipt> r{};
-        r.reserve(b.transactions.size());
-        for (auto &fiber : fibers) {
-            fiber.join();
-        }
-        for (auto &d : data) {
-            r.push_back(d.get_receipt());
+        // Merge in order, or re-run if necessary
+        for (unsigned index = 0u; auto &future : futures) {
```
tongzhi1998, line 89: More like a personal preference than anything. Can we just do `for(index = 0; index < size; ++index)` and get future from `futures[index]`? Then we don't need to explicitly manage `index++` at other places.

include/monad/execution/block_processor.hpp
```diff
@@ -30,36 +60,88 @@ struct AllTxnBlockProcessor
...
-        TTraits::process_withdrawal(s, b.withdrawals);
+        // Process withdrawls
+        TTraits::process_withdrawal(block_state, db, block_cache, b.withdrawals);
 
-        if (b.header.number != 0u) {
-            TTraits::apply_block_award(s, b);
+        { // Apply block reward to beneficiary
+            state::State state{block_state, db, block_cache};
+
+            if (b.header.number != 0u) {
```
tongzhi1998, line 140: Maybe I could do it, but maybe consider removing this conditional logic now that the genesis refactor is already in?

src/monad/execution/test/block_processor.cpp
```diff
@@ -1,19 +1,14 @@
-#include <monad/db/block_db.hpp>
 #include <monad/db/in_memory_trie_db.hpp>
 
 #include <monad/execution/block_processor.hpp>
-#include <monad/execution/config.hpp>
-#include <monad/execution/execution_model.hpp>
```
tongzhi1998, line 6: Are we still using the `execution_model` anywhere? I saw it is included in `replay_block_db` but I don't think it is actually used (sorry I think this was my bad). Maybe consider removing the file either in this PR or the next one? Ignore what I just said if you have already done this.

sheapwilliams: Y - we can remove it at this point.

include/monad/execution/block_processor.hpp
sheapwilliams: Sure we should definitely remove.

include/monad/execution/block_processor.hpp
```diff
@@ -30,36 +60,88 @@ struct AllTxnBlockProcessor
...
+            {
+                std::shared_lock<TMutex> const lock{block_state.mutex};
+                MONAD_DEBUG_ASSERT(
+                    can_merge(block_state.state, new_result.second.state_));
+                merge(block_state.state, new_result.second.state_);
+                merge(block_state.code, new_result.second.code_);
+            }
+            r.push_back(new_result.first);
+            ++index;
+            fibers[index].join();
```
tongzhi1998, line 130: Maybe I am missing something, but why do we ++index and then join? I thought we should do the opposite (And I believe this is the reason for segmentation fault).

include/monad/execution/block_processor.hpp
zander-xyz, line 41: the std::move isn't necessary

include/monad/execution/block_processor.hpp
sheapwilliams: Totally correct - I'll fix this.

PR #362, opened 2023-09-15
call get_code, track touched accounts

zander-xyz:
🐛 call get_code on new_state in call_evm

Problem:
- account must be read before its storage is accessed
- this invariant, captured as an assertion, was failing because
  get_code was being invoked on the old state rather than the new state

Solution:
- call get_code on the new state instead

🐛 track touched accounts separately from StateDeltas

Problem:
- Delegate calls do not result in an account being touched
- We erroneously destruct empty accounts if it is delegate called to

Solution:
- Keep track of touched accounts separately


include/monad/state2/state.hpp
```diff
@@ -201,10 +205,11 @@ struct State
...
 
-        for (auto &it : state_) {
-            if (it.second.account.second.has_value() &&
-                it.second.account.second.value() == Account{})
-                it.second.account.second.reset();
+        for (auto const &touched : touched_) {
+            auto &account = read_account<Mutex>(touched, state_, bs_, db_);
+            if (account.has_value() && account.value() == Account{}) {
+                account.reset();
+            }
```
sheapwilliams, line 212: Might just be ignorance, but is it not enough to go through the accessed accounts instead of tracking a new map?

PR #365, opened 2023-09-18
:recycle: add balance assertions to state

zander-xyz:


include/monad/state2/state.hpp
```diff
@@ -123,10 +127,11 @@ struct State
     {
         auto &account = read_account<Mutex>(address, state_, bs_, db_);
         if (!account.has_value()) {
-            MONAD_DEBUG_ASSERT(delta == 0);
```
tongzhi1998, line 126: I am a little confused how this shows up in the diff tho...

https://github.com/monad-crypto/monad/blob/main/include/monad/state2/state.hpp#L106

PR #373, opened 2023-09-20
:sparkles: Add custom formatter for StateDelta and CodeDelta

tongzhi1998:


include/monad/logging/formatter.hpp
```diff
@@ -301,6 +329,82 @@ struct formatter<monad::Transaction::Type> : public monad::log::basic_formatter
...
+    auto format(monad::delta_t<T> const &delta, FormatContext &ctx) const
+    {
+        fmt::format_to(ctx.out(), "{{");
+        fmt::format_to(
+            ctx.out(), "Original: {}, Updated: {}", delta.orig, delta.updated);
+        fmt::format_to(ctx.out(), "}}");
+
+        return ctx.out();
+    }
+};
```
zander-xyz: are pairs not automatically formattable?

include/monad/logging/formatter.hpp
```diff
@@ -13,6 +13,9 @@
```
zander-xyz, line 13: can you please also add the associated log statements?

tongzhi1998: I kinda intentionally removed them since we don't have a refactored BP yet. So I am not sure how to integrate these logs in. But I verified those logs are working fine in my own hacky branch rebased on the previous version of BP

include/monad/logging/formatter.hpp
tongzhi1998: It is automatically formattable. But it should just be (a,b), I kind of want to have Original & Updated explicitly stated, but I don't have any strong opinion on this. I can change it to just auto format.

include/monad/logging/formatter.hpp
tongzhi1998: After cleaning up the pair logger to be default, it would look like:

`State Deltas: {Address: 0x42e6723a0c884e922240e56d7b618bec96f35800, State Delta: {Account Delta: (optional(Account{balance=3940254a5da0c50c, code_hash=0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470, nonce=766}), optional(Account{balance=3904bc419953390b, code_hash=0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470, nonce=767})) Storage Deltas: {}} Address: 0x1194e966965418c7d73a42cceeb254d875860356, State Delta: {Account Delta: (optional(Account{balance=5a7, code_hash=0x32e7a6a214206e5822d71def806a2f764c50fb8b612e1caafc11674040ff45d5, nonce=0}), optional(Account{balance=5a8, code_hash=0x32e7a6a214206e5822d71def806a2f764c50fb8b612e1caafc11674040ff45d5, nonce=0})) Storage Deltas: {Key: 0x1539952ab901838d9e97d5f26d23efb41e5a91903626d60f9899fc3ec9007dfe, Storage Delta: (0x0000000000000000000000000000000000000000000000000000000000000001, 0x0000000000000000000000000000000000000000000000000000000000000001)}}`

include/monad/logging/formatter.hpp
zander-xyz: I think it's reasonable to do that, but delta_t is just an alias so this hijacks the formatter for all pairs. if you want this specialized type of logging, this belongs directly in the StateDelta formatter itself

include/monad/logging/formatter.hpp
zander-xyz: Gotcha, this is an ok start, but we probably want to look into quill's built-in json structured logging for parsing (and tbh, readability) at some point. 

https://quillcpp.readthedocs.io/en/latest/tutorial.html#json-log

tongzhi1998: Makes sense! I will look into this and add it once the BP & replay_ethereum are ready (otherwise I don't know where to put it) 👍

zander-xyz: Maybe you can try it on your branch locally? Might be helpful for you even now 🤷 . but only you will know

PR #374, opened 2023-09-20
:recycle: make fee names and types consistent

zander-xyz:


include/monad/core/transaction.hpp
```diff
@@ -40,15 +40,15 @@ struct Transaction
 
     SignatureAndChain sc{};
     uint64_t nonce{};
-    uint64_t gas_price{}; // max per gas fee
+    uint256_t max_fee_per_gas{};
```
sheapwilliams: Probably need to invert the comment - for the majority of txns, this is now just the gas_price.

include/monad/execution/ethereum/fork_traits.hpp
```diff
@@ -125,16 +125,16 @@ namespace fork_traits
...
         }
 
-        static constexpr uint64_t
-        gas_price(Transaction const &t, uint64_t const /*base_gas_price*/)
+        static constexpr uint256_t
+        gas_price(Transaction const &t, uint256_t const & /*base_fee_per_gas*/)
         {
-            return t.gas_price;
+            return t.max_fee_per_gas;
         }
```
sheapwilliams: We should at least be consistent about naming unused parameters...either leave them commented or remove them?

PR #375, opened 2023-09-20
Fiber aware shared mutex

rgarc:
This PR copies the `std::shared_mutex` from the [standard library](https://github.com/gcc-mirror/gcc/blob/36eec7995b4d53083c3ee7824bd765b5eba8b1a1/libstdc%2B%2B-v3/include/std/shared_mutex#L265-L408), but uses synchronization primitives that are fiber aware (`boost::fibers::mutex` and `boost::fibers::condition_variable` instead of the `std` variants).

Problem:
- We would like to use a reader-writer lock in the context of a boost
  fiber.
- Boost.Fiber does not provide one.

Solution:
- Copy the `shared_mutex` implementation from the standard library but
  replace the blocking synchronization primitives with fiber-aware ones
  that yield the current fiber instead of blocking.

### DIFF
--- a/src/monad/fiber/test_shared_mutex.cpp
+++ b/src/monad/fiber/test_shared_mutex.cpp
@@ -35,8 +35,18 @@ class Counter
 public:
     [[nodiscard]] uint64_t get() const
     {
-        std::shared_lock<TMutex> lock{mutex_};
-        return counter_;
+        if constexpr (requires(TMutex m) {
+                          {
+                              m.lock_shared()
+                          } -> std::convertible_to<void>;
+                      }) {
+            std::shared_lock<TMutex> lock{mutex_};
+            return counter_;
+        }
+        else {
+            std::unique_lock<TMutex> lock{mutex_};
+            return counter_;
+        }
     }
 
     [[nodiscard]] uint64_t increment()
@@ -65,8 +75,9 @@ struct SharedMutex : public testing::Test
 {
 };
 
-using MutexTypes =
-    ::testing::Types<NoLockingMutex, monad::shared_mutex, std::shared_mutex>;
+using MutexTypes = ::testing::Types<
+    NoLockingMutex, monad::shared_mutex, boost::fibers::mutex,
+    std::shared_mutex, std::mutex, std::timed_mutex, std::shared_timed_mutex>;
 
 TYPED_TEST_SUITE(SharedMutex, MutexTypes);
 


include/monad/fiber/shared_mutex.hpp
```diff
@@ -0,0 +1,112 @@
...
+            ++state_;
+            return true;
+        }
+        return false;
+    }
+
+    void unlock_shared()
+    {
+        std::lock_guard<boost::fibers::mutex> guard(mutex_);
+        MONAD_ASSERT(readers() > 0);
```
sheapwilliams, line 97: This call should only be called paired with `lock_shared()`, so I understand that this assert really protects against being called when paired with `lock()`, yea?

If so, then is this just protecting against programming errors?  Is `MONAD_ASSERT` the right thing for detecting programming errors?

src/monad/fiber/test_shared_mutex.cpp
```diff
@@ -0,0 +1,139 @@
...
+struct SharedMutex : public testing::Test
+{
+};
+
+using MutexTypes =
+    ::testing::Types<DumbMutex, monad::shared_mutex, std::shared_mutex>;
+
+TYPED_TEST_SUITE(SharedMutex, MutexTypes);
+
+TYPED_TEST(SharedMutex, DISABLED_simple_bench)
```
sheapwilliams: Why disabled?  What is the follow-on plan?

src/monad/fiber/test_shared_mutex.cpp
```diff
@@ -0,0 +1,139 @@
...
+                    }
+                }
+                for (auto &fiber : fibers) {
+                    fiber.join();
+                }
+            });
+        }
+    }
+
+    EXPECT_EQ(counter.get(), 34);
```
sheapwilliams, line 146: Curious about getting better coverage on this...but this is probably fine for now.

include/monad/fiber/shared_mutex.hpp
rgarc: I see. After talking about asserts yesterday in the run loop meeting, it's clear this should be a debug assert.

src/monad/fiber/test_shared_mutex.cpp
rgarc: I'm actually not sure what we want to do for benchmark tests. Motivation here was to not hog up CI, but I was also not sure if this warranted being a separate binary.

src/monad/fiber/test_shared_mutex.cpp
rgarc: I was also puzzled at coming up with a reasonable testing strategy for this since all the interesting edge cases are more likely to arise from using schedulers other than the default round robin.

PR #379, opened 2023-09-22
🐛 failed contract creation should return no data 

zander-xyz:
Problem:
- eip211 introduces RETURNDATASIZE and RETURNDATACOPY and defines that
  contract creation failure due to a call frame not being instantiated
  should return an empty buffer
- we currently do not return an empty buffer

Solution:
- clear the output buffer in such cases, with the exception of reversion
  (see eip-140)


include/monad/execution/evm.hpp
```diff
@@ -43,46 +43,35 @@ struct Evm
...
+        }
 
-        auto deploy_result = deploy_code(execute_init_code());
+        if (result.status_code == EVMC_SUCCESS) {
+            result = TTraits::deploy_contract_code(
+                new_state, contract_address.value(), std::move(result));
+        }
 
-        if (deploy_result.status_code == EVMC_SUCCESS) {
+        if (result.status_code == EVMC_SUCCESS) {
```
sheapwilliams, line 70: Collapse ifs?

zander-xyz: the result gets reassigned so the if's need to exist, unless you were thinking of something else?

PR #384, opened 2023-09-27
🐛 destruct suicides at the end of txn 

zander-xyz:
Problem:
- YP states that throughout the txn execution, self destructed accounts
  are accrued as substate and destructed at the end
- Current impl immediately destructs accounts

Solution:
- Add self destructed to set and clean up at the end of the transaction


src/monad/state2/test/state.cpp
```diff
@@ -223,13 +223,11 @@ TYPED_TEST(StateTest, selfdestruct)
     s.add_to_balance(b, 28'000);
 
     EXPECT_TRUE(s.selfdestruct(a, c));
-    EXPECT_EQ(s.total_selfdestructs(), 1u);
```
tongzhi1998, line 226: Not a big deal, but instead of deleting it, can we change it to `EXPECT_EQ(s.destruct.size(), 1)`?

PR #385, opened 2023-09-27
:sparkles: Add incarnation

tongzhi1998:
Problem:
- If an account is deleted and created again in the same block, we need to remove the old storage completely
- Our current code doesn't support for that

Solution:
- Add incarnation to Account 
- If it is a new account, then remove old storage from the db

### DIFF
--- a/test/ethereum_test/exclude/berlin.cmake
+++ b/test/ethereum_test/exclude/berlin.cmake
@@ -1,9 +1,4 @@
 set(berlin_excluded_tests
-    "BlockchainTests.GeneralStateTests/stCreate2/RevertInCreateInInitCreate2.json" # Incarnation
-    "BlockchainTests.GeneralStateTests/stCreate2/create2collisionStorage.json" # Incarnation
-    "BlockchainTests.GeneralStateTests/stExtCodeHash/dynamicAccountOverwriteEmpty.json" # Incarnation
-    "BlockchainTests.GeneralStateTests/stRevertTest/RevertInCreateInInit.json" # Incarantion
-    "BlockchainTests.GeneralStateTests/stSStoreTest/InitCollision.json" # Incarnation
     "BlockchainTests.InvalidBlocks/bcInvalidHeaderTest/DifficultyIsZero.json" # Difficulty (Pre-merge)
     "BlockchainTests.InvalidBlocks/bcInvalidHeaderTest/timeDiff0.json" # ParentHeader
     "BlockchainTests.InvalidBlocks/bcInvalidHeaderTest/wrongCoinbase.json" # StateRoot
@@ -54,10 +49,6 @@ set(berlin_excluded_tests
     "BlockchainTests.ValidBlocks/bcMultiChainTest/ChainAtoChainB_difficultyB.json" # Multichain
     "BlockchainTests.ValidBlocks/bcMultiChainTest/ChainAtoChainBtoChainA.json" # Multichain
     "BlockchainTests.ValidBlocks/bcMultiChainTest/ChainAtoChainBtoChainAtoChainB.json" # Multichain
-    "BlockchainTests.ValidBlocks/bcStateTests/suicideStorageCheck.json" # Incarnation
-    "BlockchainTests.ValidBlocks/bcStateTests/suicideStorageCheckVCreate.json" # Incarnation
-    "BlockchainTests.ValidBlocks/bcStateTests/suicideStorageCheckVCreate2.json" # Incarnation
-    "BlockchainTests.ValidBlocks/bcStateTests/suicideThenCheckBalance.json" # Incarnation
     "BlockchainTests.ValidBlocks/bcTotalDifficultyTest/lotsOfBranchesOverrideAtTheEnd.json" # Difficulty (Pre-merge)
     "BlockchainTests.ValidBlocks/bcTotalDifficultyTest/lotsOfBranchesOverrideAtTheMiddle.json" # Difficulty (Pre-merge)
     "BlockchainTests.ValidBlocks/bcTotalDifficultyTest/lotsOfLeafs.json" # Difficulty (Pre-merge)
--- a/test/ethereum_test/exclude/istanbul.cmake
+++ b/test/ethereum_test/exclude/istanbul.cmake
@@ -1,9 +1,4 @@
 set(istanbul_excluded_tests
-    "BlockchainTests.GeneralStateTests/stCreate2/RevertInCreateInInitCreate2.json" # Incarnation
-    "BlockchainTests.GeneralStateTests/stCreate2/create2collisionStorage.json" # Incarnation
-    "BlockchainTests.GeneralStateTests/stExtCodeHash/dynamicAccountOverwriteEmpty.json" # Incarnation
-    "BlockchainTests.GeneralStateTests/stRevertTest/RevertInCreateInInit.json" # Incarantion
-    "BlockchainTests.GeneralStateTests/stSStoreTest/InitCollision.json" # Incarnation
     "BlockchainTests.InvalidBlocks/bcInvalidHeaderTest/DifficultyIsZero.json" # Difficulty (Pre-merge)
     "BlockchainTests.InvalidBlocks/bcInvalidHeaderTest/timeDiff0.json" # ParentHeader
     "BlockchainTests.InvalidBlocks/bcInvalidHeaderTest/wrongCoinbase.json" # StateRoot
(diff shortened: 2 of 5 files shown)


include/monad/db/in_memory_trie_db.hpp
```diff
@@ -74,6 +74,9 @@ struct InMemoryTrieDB : public Db
         address_t const &a, uint64_t incarnation,
         bytes32_t const &key) const override
     {
+        if (MONAD_UNLIKELY(incarnation != 0)) {
+            return bytes32_t{};
+        }
```
zander-xyz: i would have thought that reads for `incarnation != 0` never makes it to the db layer

include/monad/db/rocks_trie_db.hpp
```diff
@@ -215,6 +216,9 @@ struct RocksTrieDB : public Db
         address_t const &a, uint64_t incarnation,
         bytes32_t const &key) const override
     {
+        if (MONAD_UNLIKELY(incarnation != 0)) {
+            return bytes32_t{};
+        }
```
zander-xyz: I think we should instead assert that incarnation is zero. We should never call into the DB if incarnation is zero - that should be handled inside of read_*

include/monad/db/trie_db_process_changes.hpp
```diff
@@ -129,6 +129,9 @@ void trie_db_process_changes(
         // process storage updates first
         storage_trie_updates.clear();
         if (account_delta.second.has_value()) {
+            if (MONAD_UNLIKELY(account_delta.second.value().incarnation) != 0) {
```
zander-xyz: this is not necessarily unlikely

also, I think we should also check against `orig.has_value()`. practically it should not matter but it doesn't make sense to try to clear the storage trie if it never existed to start.

did you test what happens if, in the same set of transactions, the account gets destructed and recreated multiple times? does this logic break down?

include/monad/state2/state.hpp
```diff
@@ -209,13 +216,19 @@ struct State
 
     void destruct_suicides() const noexcept {}
 
+    bool is_dead(Account const &account)
+    {
+        return account.balance == 0 && account.nonce == 0 &&
+               account.code_hash == NULL_HASH;
+    }
```
zander-xyz: can you please rebase on top of main? my changes add this function

include/monad/state2/block_state.hpp
```diff
@@ -12,6 +12,7 @@ struct BlockState
     Mutex mutex;
     StateDeltas state;
     Code code;
+    Incarnations incarnations;
```
zander-xyz: why is this needed?

include/monad/trie/rocks_writer.hpp
```diff
@@ -54,6 +55,15 @@ struct RocksWriter
         auto const res = batch.DeleteRange(
             cf, monad::db::to_slice(buf.view()), monad::db::to_slice(end));
         MONAD_ROCKS_ASSERT(res);
+
+        write();
```
zander-xyz: It's fine to immediately write the del prefix, but can we do this at the call-site instead of right here? That seems to match the current usage a bit more intuitively

include/monad/db/in_memory_trie_db.hpp
tongzhi1998: That makes sense. However, I think we still want to push `bytes32_t{}` to both blockstate and state, so the current logic in `read_storage` seems appropriate. Moving this to the read_storage() end might introduce some redundent code.

include/monad/db/trie_db_process_changes.hpp
tongzhi1998: Cool, I can do that. I will also add test case for multiple deletions & creations.

include/monad/state2/block_state.hpp
tongzhi1998: Because we need to record how many times an account is being destructed. Upon self-destruct, account is blockstate is being reset so it loses the incarnation value that it holds

include/monad/db/in_memory_trie_db.hpp
zander-xyz: why is it redundant?

include/monad/state2/block_state.hpp
zander-xyz: why do you need to keep track how many times?

include/monad/db/in_memory_trie_db.hpp
tongzhi1998: Sorry I was misunderstanding this before. Fixed now!

include/monad/state2/block_state.hpp
tongzhi1998: Might be a bit complicated to explain. Maybe we can zoom to discuss.

zander-xyz: Can you present a simple scenario where you need to keep track of the number of times the account has reincarnated?

include/monad/db/trie_db_process_changes.hpp
```diff
@@ -129,6 +129,11 @@ void trie_db_process_changes(
...
```
zander-xyz: sorry, I think my comment was confusing
> this is not necessarily unlikely

this was suggesting that we get rid of MONAD_UNLIKELY

> also, I think we should also check against orig.has_value(). practically it should not matter but it doesn't make sense to try to clear the storage trie if it never existed to start.

did this get addressed?

include/monad/state2/state.hpp
```diff
@@ -77,6 +77,7 @@ struct State
         return account.has_value();
     };
 
+    // TODO: Lock Issue
```
zander-xyz: what's this?

include/monad/state2/block_state.hpp
tongzhi1998: I can try. Let's say a block has 5 txns (in order):
1: destroy account A
2: re-create account A (with storage)
3: destroy account A again
4: re-create account A again (with diff storage)
5. Access account A's storage

In optimistic execution, let's say 5 run after 2, followed by 3 & 4. Without incarnation number, 5 will be able to merge, whereas in reality, 5 can't be merged.

zander-xyz: Incarnation number is on the account, so would this not fail to merge (that is done in order)? In your example above, for each transaction, these would be the incarnations.

| transaction | `A` incarnation (before, after) |
-----|--------
| 1     |  (0, 0)        |
| 2     |  (0, 1)        |
| 5     |  (1, 1)        |
| 3     |  (1, 1)        |
| 4     |  (1, 2)        |

but then we merge in order,

| transaction | `A` incarnation (before, after) |
-----|--------
| 1     |  (0, 0)        |
| 2     |  (0, 1)        |
| 3     |  (1, 1)        |
| 4     |  (1, 2)        |
| 5     |  (1, 1)        |

when trying to merge the state deltas for 5 into 4 (which includes 1, 2, 3, 4), we should see that the orig for 5 != updated for 4, and fail to merge.

tongzhi1998: Incarnation number is not on the account (because of reset). Incarnation bool can be on the account, but the above illustrates why a boolean will not be enough.

tongzhi1998: Only by adding this incarnation map can incarnation number be on the account.

zander-xyz: Sorry, can you link me to the reset you are referring to?

include/monad/state2/state.hpp
tongzhi1998: This function need to be protected by a lock since we are accessing blockstate directly here. Also I think even w/o the current implementation of accessing blockstate, we need a lock to protect the whole function. 
From my understanding, when EVMOne will cause create_account when it thinks an account doesn't exist, but if we don't lock this function, upon executing read_account, the account might actually be in the blockstate already.

include/monad/state2/block_state.hpp
tongzhi1998: Account is resetted upon self-destruct, and merged into blockstate. Upon merging, in blockstate, account.current = std::nullopt, and therefore lose the incarnation that it previous has

include/monad/db/trie_db_process_changes.hpp
tongzhi1998: Sorry I think I forget to address this one. I will do it in a sec.

include/monad/state2/block_state.hpp
tongzhi1998: The map is there so that every time on reset, it records the previous incarnation that an account has so that we don't lose track of it.

include/monad/db/trie_db_process_changes.hpp
tongzhi1998: I think 2 conditions combined will make it unlikely. So I will keep it for now.

include/monad/db/in_memory_trie_db.hpp
```diff
@@ -51,6 +51,23 @@ struct InMemoryTrieDB : public Db
...
+            trie_cursor.reset();
+        }
+
+        void clear_and_reset_prefix(address_t const &addr)
+        {
+            trie.clear();
+            trie.leaves_writer_.write();
+            trie.trie_writer_.write();
+
+            reset_cursor();
```
zander-xyz: delete this

src/monad/state2/state_deltas.cpp
```diff
@@ -57,9 +57,14 @@ void merge(StateDeltas &s1, StateDeltas const &s2)
...
-        merge_f(d1.storage, d2.storage, [](auto &st1, auto const &st2) {
-            st1.second = st2.second;
-        });
+        if (d2.account.second.has_value()) {
+            merge_f(d1.storage, d2.storage, [](auto &st1, auto const &st2) {
+                st1.second = st2.second;
+            });
+        }
+        else {
+            d1.storage.clear();
```
zander-xyz, line 66: now that we do this, we should probably `MONAD_DEBUG_ASSERT(!updated_storage_roots.contains(addr))` in `trie_db_process_changes` instead of `updated_storage_roots.erase(addr);`

include/monad/db/in_memory_trie_db.hpp
tongzhi1998: Makes sense.

PR #399, opened 2023-10-04
Logging cleanups

tongzhi1998:
- Remove unused log formatter
- Add more logging statement where needed
- Change log levels where appropriate

### DIFF
--- a/include/monad/state2/state.hpp
+++ b/include/monad/state2/state.hpp
@@ -57,7 +57,7 @@ struct State
     // EVMC Host Interface
     evmc_access_status access_account(address_t const &address)
     {
-        LOG_DEBUG("access_account: {}", address);
+        LOG_TRACE_L1("access_account: {}", address);
 
         auto const [_, inserted] = accessed_.insert(address);
         if (inserted) {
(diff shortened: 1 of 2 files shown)


include/monad/state2/state.hpp
```diff
@@ -381,6 +389,8 @@ struct State
     // EVMC Host Interface
     [[nodiscard]] bytes32_t get_block_hash(int64_t const number) const noexcept
     {
+        LOG_DEBUG("get_code_size: {}", number);
```
zander-xyz: this log is wrong

include/monad/state2/state.hpp
```diff
@@ -201,7 +205,7 @@ struct State
 
     void destruct_suicides() noexcept
     {
-        LOG_DEBUG("destruct_suicides");
+        LOG_DEBUG("{}", "destruct_suicides");
```
zander-xyz: why this change?

include/monad/state2/state.hpp
tongzhi1998: It is indeed. Thanks for catching that!

include/monad/state2/state.hpp
tongzhi1998: I was intending to match it with https://github.com/monad-crypto/monad/blob/main/include/monad/state2/state.hpp#L213, but I realized that I should change the other one inside

PR #404, opened 2023-10-05
🐛 fail to validate if nonce is equal to 2^64-1

zander-xyz:


include/monad/execution/transaction_processor.hpp
```diff
@@ -141,12 +142,22 @@ struct TransactionProcessor
         if (state.get_code_hash(*t.from) != NULL_HASH) {
             return TransactionStatus::DEPLOYED_CODE;
         }
+
+        MONAD_DEBUG_ASSERT(t.from.has_value());
+        auto const sender_nonce = state.get_nonce(t.from.value());
+
+        // eip-2681
+        if (sender_nonce >= std::numeric_limits<uint64_t>::max()) {
```
tongzhi1998: Just checking my understanding here: this is the same as == right?

zander-xyz: yes

PR #408, opened 2023-10-06
🐛 populate mix_hash and difficulty correctly in eth tests 

zander-xyz:
Problem:
- difficulty should be set to zero post merge
- mix_hash should be set to currentRandom
- DIFFICULTY opcode is returning the wrong value

Solution:
- set difficulty to zero post merge and populate mix_hash with
  currentRandom


include/monad/core/block.hpp
```diff
@@ -24,7 +24,7 @@ struct BlockHeader
     bytes32_t state_root{};
     bytes32_t transactions_root{};
     bytes32_t receipts_root{};
-    bytes32_t mix_hash{};
+    bytes32_t prev_randao{}; // aka mix_hash
```
tongzhi1998, line 27: If I recall correctly, it's called `mixed_hash` before the merge and called `prev_randao` after. I don't mind changing the name but just pointing out here.

zander-xyz: that's correct

PR #410, opened 2023-10-06
:recycle: decouple block number from fork traits

zander-xyz:


include/monad/core/configuration.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <monad/config.hpp>
+
+#include <optional>
+
+MONAD_NAMESPACE_BEGIN
+
+struct Configuration
+{
+    std::optional<uint64_t> homestead_block;
```
rgarc, line 11: why are all these optional?

zander-xyz: i imagine when we go live, we probably won't configure these and only configure the latest fork

rgarc: makes sense

PR #411, opened 2023-10-10
:recycle: Refactor chain_id  & v type; Split validate function

tongzhi1998:


test/ethereum_test/src/ethereum_test.cpp
```diff
@@ -285,16 +288,17 @@ void EthereumTests::run_state_test(
...
                     .sc =
                         SignatureAndChain{
                             .r = {},
                             .s = {},
                             // Only supporting mainnet for now
-                            .chain_id =
-                                fork_index >= to_fork_index("EIP158").value()
-                                    ? 1
-                                    : 0},
+                            .chain_id = chain_id},
```
tongzhi1998, line 298: We should not set chain_id to 0 before SPURIOUS_DRAGON, but leave it as std::nullopt

PR #419, opened 2023-10-13
✨ add in support for blockchain tests

zander-xyz:


test/ethereum_test/src/general_state_test.cpp
```diff
@@ -14,6 +14,33 @@ using namespace monad;
 
 namespace
 {
+    [[nodiscard]] constexpr BlockHeader
+    block_header_from_env(nlohmann::json const &json)
```
rgarc, line 18: Why was this moved from `from_json.hpp` ?

zander-xyz: a block header in general state tests is different from a block header in blockchain tests. because of how it's labeled in general state tests, i separated the impl

PR #489, opened 2023-11-13
Add parent header

tongzhi1998:


include/monad/execution/buffer.hpp
```diff
@@ -0,0 +1,59 @@
...
+    bytes32_t block_hash_buffer_[block_hash_buffer_size];
+    uint64_t last_block_number_;
+
+    BlockHeader parent_header_;
+    bytes32_t parent_hash_;
+
+public:
+    Buffer();
+
+    void set_block_hash(uint64_t const block_number, bytes32_t const &hash)
```
rgarc, line 25: Can you move these definitions into the buffer.cpp file?

PR #606, opened 2024-01-22
optimistic execution

jhunsaker:


src/monad/execution/execute_block.cpp
```diff
@@ -73,54 +77,65 @@ constexpr Receipt::Bloom compute_bloom(std::vector<Receipt> const &receipts)
...
+                    promises,
+                    transaction,
+                    header,
+                    block_hash_buffer,
+                    block_state);
+            });
+    }
+
+    auto const last = static_cast<std::ptrdiff_t>(block.transactions.size());
+    promises[last].get_future().wait();
```
klemens-morgenstern, line 127: Are we only waiting for the last execution to complete? Is it a guarantee that they complete in order, and if so, how?

jhunsaker: they each wait on the previous via the array of promises

PR #642, opened 2024-02-15
incarnation race condition

jhunsaker:


libs/execution/src/monad/core/fmt/account_fmt.hpp
```diff
@@ -22,12 +22,13 @@ struct fmt::formatter<monad::Account> : public monad::BasicFormatter
...
             "}}",
             a.balance,
             a.code_hash,
             a.nonce,
-            a.incarnation);
+            a.incarnation.get_block(),
+            a.incarnation.get_tx());
         return ctx.out();
     }
 };
```
magedm, line 34: I tried to log some incarnations without the rest of the account and it looks like we need a formatter.
```
template <>
struct fmt::formatter<monad::Incarnation> : public monad::BasicFormatter
{
    template <typename FormatContext>
    auto format(monad::Incarnation const &a, FormatContext &ctx) const
    {
        fmt::format_to(
            ctx.out(),
            "({}, {})",
            a.get_block(),
            a.get_tx());
        return ctx.out();
    }
};
```

jhunsaker: will add separate pr

PR #680, opened 2024-03-28
`replay_ethereum` now loads into memory the top five levels of the latest block

ned14:
... so the memory cache is fully "warm" and benchmarks are more representative of real world performance.

Many thanks to Vicky for the assist.

### DIFF
--- a/cmd/replay_ethereum.cpp
+++ b/cmd/replay_ethereum.cpp
@@ -151,6 +151,17 @@ int main(int const argc, char const *argv[])
 
     if (load_snapshot.empty()) {
         init_block_number = db.current_block_number();
+        LOG_INFO("Loading current root into memory");
+        auto const start_time = std::chrono::steady_clock::now();
+        auto const nodes_loaded = db.prefetch_current_root();
+        auto const finish_time = std::chrono::steady_clock::now();
+        auto const elapsed = std::chrono::duration_cast<std::chrono::seconds>(
+            finish_time - start_time);
+        LOG_INFO(
+            "Finish loading current root into memory, time_elapsed = {}, "
+            "nodes_loaded = {}",
+            elapsed,
+            nodes_loaded);
     }
     if (init_block_number == 0) {
         MONAD_ASSERT(*has_genesis_file);
--- a/db/src/monad/mpt/db.cpp
+++ b/db/src/monad/mpt/db.cpp
@@ -28,6 +28,7 @@
 #include <cerrno>
 #include <chrono>
 #include <condition_variable>
+#include <cstddef>
 #include <cstdint>
 #include <filesystem>
 #include <iterator>
@@ -51,29 +52,22 @@ MONAD_MPT_NAMESPACE_BEGIN
 
 struct Db::Impl
 {
+    bool upsert_running_{false};
+    bool prefetch_running_{false};
+
     virtual ~Impl() = default;
 
     virtual Node::UniquePtr &root() = 0;
     virtual UpdateAux<> &aux() = 0;
-    virtual void upsert(
+    virtual void upsert_fiber_blocking(
         UpdateList &&, uint64_t, bool enable_compaction,
         bool can_write_to_fast) = 0;
 
     virtual find_result_type
-    find(NodeCursor const &root, NibblesView const &key)
-    {
-        return find_blocking(aux(), root, key);
-    }
-
-    virtual bool is_latest() const
-    {
-        MONAD_ASSERT(false);
-    };
-
-    virtual void load_latest()
-    {
-        MONAD_ASSERT(false);
-    };
+    find_fiber_blocking(NodeCursor const &root, NibblesView const &key) = 0;
+    virtual bool is_latest() const = 0;
+    virtual void load_latest_fiber_blocking() = 0;
+    virtual size_t prefetch_fiber_blocking(uint64_t latest_block_id) = 0;
 };
 
 struct Db::ROOnDisk final : public Db::Impl
@@ -86,7 +80,7 @@ struct Db::ROOnDisk final : public Db::Impl
     chunk_offset_t last_loaded_offset_;
     Node::UniquePtr root_;
 
-    explicit ROOnDisk(ReadOnlyOnDiskDbConfig const &options)
+    ROOnDisk(ReadOnlyOnDiskDbConfig const &options)
         : pool_{[&] -> async::storage_pool {
             async::storage_pool::creation_flags pool_options;
             pool_options.open_read_only = true;
@@ -130,9 +124,16 @@ struct Db::ROOnDisk final : public Db::Impl
         return aux_;
     }
 
-    virtual void upsert(UpdateList &&, uint64_t, bool, bool) override
+    virtual void
+    upsert_fiber_blocking(UpdateList &&, uint64_t, bool, bool) override
     {
         MONAD_ASSERT(false);
+    }
+
+    virtual find_result_type
+    find_fiber_blocking(NodeCursor const &root, NibblesView const &key) override
+    {
+        return find_blocking(aux(), root, key);
     }
 
     virtual bool is_latest() const override
@@ -140,10 +141,18 @@ struct Db::ROOnDisk final : public Db::Impl
         return last_loaded_offset_ == aux_.get_root_offset();
     }
 
-    virtual void load_latest() override
+    virtual void load_latest_fiber_blocking() override
     {
+        if (last_loaded_offset_ == aux_.get_root_offset()) {
+            return;
+        }
         last_loaded_offset_ = aux_.get_root_offset();
         root_.reset(read_node_blocking(pool_, last_loaded_offset_));
+    }
+
+    virtual size_t prefetch_fiber_blocking(uint64_t) override
+    {
+        MONAD_ASSERT(false);
     }
 };
 
@@ -169,11 +178,29 @@ struct Db::InMemory final : public Db::Impl
         return aux_;
     }
 
-    virtual void
-    upsert(UpdateList &&list, uint64_t block_id, bool, bool) override
+    virtual void upsert_fiber_blocking(
+        UpdateList &&list, uint64_t block_id, bool, bool) override
     {
         root_ = aux_.do_update(
             std::move(root_), machine_, std::move(list), block_id, false);
+    }
+
+    virtual find_result_type
+    find_fiber_blocking(NodeCursor const &root, NibblesView const &key) override
+    {
+        return find_blocking(aux(), root, key);
+    }
+
+    virtual bool is_latest() const override
+    {
+        return true;
+    }
+
+    virtual void load_latest_fiber_blocking() override {}
+
+    virtual size_t prefetch_fiber_blocking(uint64_t) override
+    {
+        return 0;
     }
 };
 
@@ -190,8 +217,16 @@ struct Db::RWOnDisk final : public Db::Impl
         bool const can_write_to_fast;
     };
 
-    using Comms =
-        std::variant<std::monostate, fiber_find_request_t, FiberUpsertRequest>;
+    struct FiberLoadAllFromBlockRequest
+    {
+        threadsafe_boost_fibers_promise<size_t> *promise;
+        NodeCursor root;
+        StateMachine &sm;
+    };
+
+    using Comms = std::variant<
+        std::monostate, fiber_find_request_t, FiberUpsertRequest,
+        FiberLoadAllFromBlockRequest>;
     ::moodycamel::ConcurrentQueue<Comms> comms_;
 
     std::mutex lock_;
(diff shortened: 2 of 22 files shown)


db/src/monad/mpt/trie.cpp
```diff
@@ -133,6 +164,105 @@ Node::UniquePtr upsert(
...
+                sm.up(1 + nv.nibble_size());
+            }
+        }
+    }
+};
+
+size_t load_all(UpdateAuxImpl &aux, StateMachine &sm, NodeCursor root)
+{
+    load_all_impl_ impl(aux);
+    impl.process(root.node, sm);
```
Chen-Yifan: process() has to take `NodeCursor` input because root we start loading cache is leaf of block number trie, and `NodeCursor::prefix_index` is the index current state machine is at and is where we should start `process()`, I remember I changed it in the earlier commit, any reason why we changed it back?

db/src/monad/mpt/db.cpp
```diff
@@ -126,20 +118,33 @@ struct Db::ROOnDisk final : public Db::Impl
...
-    virtual void load_latest() override
+    virtual size_t load_latest_fiber_blocking(bool load_all) override
     {
+        if (last_loaded_offset_ == aux_.get_root_offset()) {
+            return 0;
+        }
         last_loaded_offset_ = aux_.get_root_offset();
         root_.reset(read_node_blocking(pool_, last_loaded_offset_));
+        if (load_all) {
+            return 1 + mpt::load_all(aux_, machine_, *root_);
```
Chen-Yifan: should set the `NodeCursor` to leaf of latest block id before calling `load_all()`, otherwise when trie has multiple versions/blocks, memory will blow up loading all versions of the same nodes. Also for execution purpose we only care about caching the latest version.

Chen-Yifan: other thing we shall change is `OnDiskMachine::cache()`, we don't need to cache any receipts

db/src/monad/mpt/trie.cpp
ned14: Fixed

db/src/monad/mpt/db.cpp
ned14: Did you have code before solving this which I failed to merge?

Chen-Yifan: yes, I pushed to `vicky-warm-cache-replay-ethereum-unrebased`, hasn't rebased to latest main yet. Here's the `OnDiskMachine::cache()` impl 
```
    virtual bool cache() const override
    {
        return depth <= cache_depth && trie_section != TrieType::Receipt;
    }
```

src/monad/db/trie_db.cpp
```diff
@@ -833,6 +833,11 @@ nlohmann::json TrieDb::to_json()
     return traverse.json;
 }
 
+size_t TrieDb::load_all_of_current_root()
+{
+    return db_.load_latest(true);
```
Chen-Yifan: can assert `MONAD_ASSERT(machine_->trie_section == Machine::TrieType::Prefix);`  before return

db/src/monad/mpt/db.cpp
ned14: Ok should be fixed now

db/src/monad/mpt/db.cpp
```diff
@@ -128,20 +149,34 @@ struct Db::ROOnDisk final : public Db::Impl
...
+
     virtual bool is_latest() const override
     {
         return last_loaded_offset_ == aux_.get_root_offset();
     }
 
-    virtual void load_latest() override
+    virtual size_t load_latest_fiber_blocking(bool load_all) override
     {
+        if (last_loaded_offset_ == aux_.get_root_offset()) {
```
Chen-Yifan, line 146: for `ROOnDisk` is constructed with `latest_loaded_offset_ = aux_.get_root_offset()`, if then db has no change, and user invokes `load_latest_fiber_blocking(true)`, it doesn't seem to load any nodes as intended.

db/src/monad/mpt/db.cpp
```diff
@@ -53,25 +53,41 @@ struct Db::Impl
...
-    virtual bool is_latest() const
+    explicit constexpr Impl(Db *parent)
+        : parent_(parent)
     {
-        MONAD_ASSERT(false);
-    };
+    }
 
-    virtual void load_latest()
+    size_t load_all_(UpdateAuxImpl &aux, StateMachine &sm, NodeCursor root)
```
Chen-Yifan: we're loading the latest block, so might be better to rename `load_all()` to `load_latest_block_()`, so as `FiberLoadAllRequest` -> `FiberLoadLatestBlockRequest`

src/monad/db/trie_db.cpp
ned14: Done.

db/src/monad/mpt/db.cpp
ned14: Fixed.

db/src/monad/mpt/db.cpp
ned14: Done.

db/src/monad/mpt/db.cpp
Chen-Yifan: it now traverse the whole in memory trie if root offset hasn't changed from last time? rpc server will probably call `load_latest_fiber_blocking()` pretty frequently, not sure if they will invoke with `load_all`, but if they do, seems it'd be quite expensive.

db/src/monad/mpt/db.cpp
Chen-Yifan: forgot `FiberLoadAllRequest`

ned14: Done

db/src/monad/mpt/trie.cpp
```diff
@@ -133,6 +164,109 @@ Node::UniquePtr upsert(
...
+            impl->process(NodeCursor{*node}, *sm);
+        }
+    };
+
+    explicit constexpr load_all_impl_(UpdateAuxImpl &aux)
+        : aux(aux)
+    {
+    }
+
+    void process(NodeCursor node_cursor, StateMachine &sm)
```
Chen-Yifan: nit: `const node_cursor`

db/src/monad/mpt/trie.cpp
```diff
@@ -133,6 +164,109 @@ Node::UniquePtr upsert(
...
+    {
+        Node *node = node_cursor.node;
```
Chen-Yifan: `Node *const node`

db/src/monad/mpt/db.cpp
```diff
@@ -128,20 +150,37 @@ struct Db::ROOnDisk final : public Db::Impl
...
+        return find_blocking(aux(), root, key);
+    }
+
     virtual bool is_latest() const override
     {
         return last_loaded_offset_ == aux_.get_root_offset();
     }
 
-    virtual void load_latest() override
+    virtual size_t load_latest_fiber_blocking(bool load_all) override
```
Chen-Yifan: `bool const load_all`

db/src/monad/mpt/db.cpp
ned14: Fixed.

db/src/monad/mpt/trie.cpp
```diff
@@ -133,6 +164,109 @@ Node::UniquePtr upsert(
...
+                        process(NodeCursor{*next}, sm);
+                    }
+                }
+                sm.up(1 + nv.nibble_size());
+            }
+        }
+    }
+};
+
+size_t load_all(UpdateAuxImpl &aux, StateMachine &sm, NodeCursor root)
```
Chen-Yifan: `const root`

db/src/monad/mpt/db.cpp
```diff
@@ -305,6 +374,15 @@ struct Db::RWOnDisk final : public Db::Impl
                             compaction && req->enable_compaction,
                             req->can_write_to_fast));
                     }
+                    else if (auto *req = std::get_if<3>(&request.front());
+                             req != nullptr) {
+                        // Ditto to above
+                        load_all_promises.emplace_back(
+                            std::move(*req->promise));
+                        req->promise = &load_all_promises.back();
+                        req->promise->set_value(parent->load_latest_block_(
```
Chen-Yifan: seems problematic as `load_latest_block_()` invokes `parent_->get(root, block_nv), which blocks fiber, and these are all calling at the worker thread.

We ought to move the logic that sets cursor to latest block to `load_latest_fiber_blocking()` on main thread

db/src/monad/mpt/trie.cpp
ned14: Fixed.

db/src/monad/mpt/db.cpp
ned14: You can't make a virtual function argument const without it impinging on all subclasses. This is a big reason why I personally speaking don't think function arguments should ever be const, unless you are applying const to what is pointed at or referenced at. I'm fine with stack declared variables being const, there I think it adds value, but I don't with function arguments.

db/src/monad/mpt/trie.cpp
ned14: Fixed.

db/src/monad/mpt/trie.cpp
ned14: Fixed.

db/src/monad/mpt/db.cpp
ned14: Hopefully new approach in this PR fixes the issue.

db/src/monad/mpt/db.cpp
```diff
@@ -84,9 +78,10 @@ struct Db::ROOnDisk final : public Db::Impl
     async::AsyncIO io_;
     UpdateAux<> aux_;
     chunk_offset_t last_loaded_offset_;
+    chunk_offset_t last_loaded_all_offset_;
```
Chen-Yifan: saw this not used anymore, can remove

cmd/replay_ethereum.cpp
```diff
@@ -151,6 +151,17 @@ int main(int const argc, char const *argv[])
 
     if (load_snapshot.empty()) {
         init_block_number = db.current_block_number();
+        LOG_INFO("Loading current root into memory");
```
Chen-Yifan, line 154: I saw we assert(false) for InMemory Db, so if we run in-memory replay would it assert? Can either only prefetch for on disk trie, or change the InMemory Impl to simply return 0.

db/src/monad/mpt/db.cpp
```diff
@@ -572,7 +654,22 @@ bool Db::is_latest() const
...
+{
+    MONAD_ASSERT(impl_);
+    MONAD_ASSERT(!impl_->prefetch_running_);
+    auto const latest_block_id = get_latest_block_id();
+    if (!latest_block_id.has_value()) {
+        return 0;
+    }
+    impl_->prefetch_running_ = true;
+    size_t const nodes_loaded =
+        impl_->load_all_fiber_blocking(latest_block_id.value());
```
Chen-Yifan: nit: can `load_all_fiber_blocking()` -> `prefetch_fiber_blocking()` to match name elsewhere.

db/src/monad/mpt/trie.cpp
```diff
@@ -133,6 +164,109 @@ Node::UniquePtr upsert(
...
+                        process(NodeCursor{*next}, sm);
+                    }
+                }
+                sm.up(1 + nv.nibble_size());
+            }
+        }
+    }
+};
+
+size_t load_all(UpdateAuxImpl &aux, StateMachine &sm, NodeCursor const root)
```
Chen-Yifan, line 262: also here `load_all()` to `prefetch_root()`, and `load_all_impl_` -> `prefetch_impl_()`

db/src/monad/mpt/db.cpp
ned14: Fixed.

cmd/replay_ethereum.cpp
ned14: Fixed.

db/src/monad/mpt/db.cpp
ned14: Fixed.

db/src/monad/mpt/trie.cpp
ned14: This I didn't fix, because it is the state machine which decides where a load all actually loads all, or not. The function `load_all()` does actually load all if allowed, so its name is right.

The current _users_ of `load_all()` do use it to implement prefetch, and so naming them prefetch makes sense.

cmd/replay_ethereum.cpp
Chen-Yifan: saw you changed `ROOnDisk::prefetch_fiber_blocking()` to return 0, are you mean to change it for `InMemory`?

ned14: Fixed.

db/src/monad/mpt/db.cpp
```diff
@@ -572,7 +652,22 @@ bool Db::is_latest() const
 void Db::load_latest()
 {
     MONAD_ASSERT(impl_);
-    impl_->load_latest();
+    // always latest
```
Chen-Yifan: ah, missed `impl_->load_latest_fiber_blocker()`, this should resolve ci failure

ned14: Fixed

PR #686, opened 2024-04-03
To `monad_mpt` tool add `--chunk-increasing` and `--debug`. Fix bug in archive and restoration which breaks compaction.

ned14:
Also archive and restore exact insertion counts, otherwise compaction
gets broken.

### DIFF
--- a/db/src/monad/mpt/cli_tool_impl.cpp
+++ b/db/src/monad/mpt/cli_tool_impl.cpp
@@ -378,6 +378,8 @@ struct impl_t
     bool create_database = false;
     bool truncate_database = false;
     bool create_empty_database = false;
+    bool create_chunk_increasing = false;
+    bool debug_printing = false;
     std::filesystem::path archive_database;
     std::filesystem::path restore_database;
     std::vector<std::filesystem::path> storage_paths;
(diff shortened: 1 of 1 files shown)


db/src/monad/mpt/cli_tool_impl.cpp
```diff
@@ -795,26 +807,61 @@ struct impl_t
...
+                                    fast_list_base_insertion_count);
+                        }
+                        else {
+                            return (a.metadata.insertion_count() -
+                                    slow_list_base_insertion_count) <
+                                   (b.metadata.insertion_count() -
+                                    slow_list_base_insertion_count);
+                        }
+                    }
                 }
```
Chen-Yifan, line 847: I think the changes in this part won't make a difference in the ordering unless we later on deal with insertion count overflow problem.

ned14: That code is specifically to handle insertion count wrapping. Is it incorrect?

Chen-Yifan: yeah, I'm just making comment because we currently assert on insertion count wrapping. It doesn't have any effect for now but good to have it covered for the future.

PR #692, opened 2024-04-11
spmc stream

zander-xyz:


c++/monad/core/test/test_spmc_queue.cpp
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    SpmcQueue *const queue = spmc_queue_create(path.c_str(), true);
+    std::thread c1(consumer);
+    std::thread c2(consumer);
+    std::thread p([&] {
+        ASSERT_NE(queue, nullptr);
+        for (uint32_t i = 0; i < N; ++i) {
+            // sleep to give consumers enough time
+            std::this_thread::sleep_for(std::chrono::microseconds(1));
```
magedm: Looks like this test may fail sometimes if somehow a consumers ends up running slow.
I suggest that we either avoid tests with delays or have a clear process of identifying some tests as potentially flaky (maybe retry and if it keeps failing give a warning).
Down the road with more test automation, flaky tests tend to make people routinely ignore test failures.

core/c++/monad/core/spmc_queue.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+// Lockless shared memory single producer multiple consumer queue. It is the
+// consumers responsibility to keep up with the producer. Sequence numbers start
+// at 2 and increase by 2 on every message, unless a gap occurred.
```
magedm: This data structure is different from a conventional spmc queue. Each item is consumed by every consumer not just one. This looks more like a multi-consumer stream. 
Suggestions:
- Please consider a name that doesn't conflict with a common term for a different data structure.
- If you keep the name, please consider clarifying that items are expected to be consumed by all consumers and not just one.

core/c++/monad/core/spmc_queue.c
```diff
@@ -0,0 +1,191 @@
...
+        0);
+    if (shm == MAP_FAILED) {
+        return NULL;
+    }
+
+    if (is_producer) {
+        for (size_t i = 0;
+             i < (sizeof(SpmcQueueRegion) / sizeof(SpmcQueueElement));
+             ++i) {
+            atomic_init(&shm->buf[i].seq, 1);
```
magedm: Zero 'seq' is fine too and more consistent with normal conditions: The 'seq' for a not-yet-written item is even.

magedm: Correction: After DMs w/ Alex. I think element 0 needs to be set to 1.

Also, came up the issue of seq wraparound after 2^31 ops. Looks like this can cause double consumption of buffer items when seq wraps around to 0.

PR #714, opened 2024-05-10
Separate in-memory flat caches using incarnation

magedm:
- Replaces the hierarchical structure of the combined accounts-storage cache with separate LRU caches for accounts and storage.
- The storage cache is one big concurrent hash map (used to be distributed over many per-account storage concurrent hash maps).
- Uses incarnation to avoid reading deleted storage values.

- Slightly improved throughput and significantly improved memory usage compared to main:
```
12M to 12.1M
Before: tps 8578   rss 6.0 GB
After:  tps 8736   rss 4.5 GB

12M to 12.15M
Before: tps 7995   rss 7.1 GB
After:  tps 8158   rss 4.9 GB
```


libs/execution/src/monad/db/db.hpp
```diff
@@ -21,6 +21,9 @@ struct Db
 
     virtual bytes32_t read_storage(Address const &, bytes32_t const &key) = 0;
 
+    virtual bytes32_t
+    read_storage(Address const &, Incarnation const, bytes32_t const &key) = 0;
```
Chen-Yifan: can we simply replace the old `read_storage()` by this new api? any specific reason that the old api has to remain?

libs/execution/src/monad/db/trie_db.hpp
```diff
@@ -32,6 +32,8 @@ class TrieDb final : public ::monad::Db
     std::list<byte_string> bytes_alloc_;
     uint64_t block_number_;
     Mode mode_;
+    std::atomic<uint64_t> n_storage_from_db_no_value_{0};
```
Chen-Yifan: might be better to use a macro that enable/disable the stats like the one in `lru_cache`

libs/core/src/monad/lru/lru_cache.hpp
```diff
@@ -15,21 +15,22 @@
...
-    using HashMap = tbb::concurrent_hash_map<Key, HashMapValue>;
+    using HashMap = tbb::concurrent_hash_map<Key, HashMapValue, KeyHashCompare>;
     using HashMapKeyValue = std::pair<Key, HashMapValue>;
     using Accessor = HashMap::accessor;
     using Mutex = SpinLock;
     using Pool = BatchMemPool<ListNode>;
 
     /// CONSTANTS
-    static constexpr size_t slack = 16;
+    static constexpr size_t k_slack = 16;
```
Chen-Yifan: why we are adding `k_` prefix to constants?

.gitmodules
```diff
@@ -39,3 +39,6 @@
 [submodule "core/third_party/unordered_dense"]
 	path = libs/core/third_party/unordered_dense
 	url = https://github.com/martinus/unordered_dense.git
+[submodule "core/third_party/komihash"]
```
Chen-Yifan: we have two komihash submodules in two folders, can we get rid of one?

libs/core/src/monad/lru/lru_cache.hpp
magedm: I'll remove it if that's the Monad convention. Personally, I think it's helpful to differentiate constants from local variables.

.gitmodules
magedm: Fixing it.

magedm: Fixed

libs/execution/src/monad/db/db.hpp
magedm: Done

libs/execution/src/monad/db/trie_db.hpp
magedm: Done

libs/core/src/monad/lru/lru_cache.hpp
Chen-Yifan: global constants are currently capitalized, and in class constants are in snake case. `k_` seems like a C convention. I think we can keep the constants naming as they were, but also I don't really have strong opinions here, open to other thoughts.

libs/core/src/monad/lru/lru_cache.hpp
```diff
@@ -187,26 +190,28 @@ class LruCache
...
         {
-            return (cur_time() - lru_time_) >= lru_update_period;
+            return (cur_time() - lru_time_.load(std::memory_order_acquire)) >=
+                   k_lru_update_period;
         }
 
         uint64_t cur_time() const
         {
-            return std::chrono::duration_cast<std::chrono::nanoseconds>(
+            return (uint64_t)
```
Chen-Yifan: use `int64_t` for all time related like `lru_time_`.

Issue #136, opened 2023-06-27
UBSan warning in TrieDB


sheapwilliams:
When running locally (not with ctest), I observe the following UBSan warning in completely unrelated code:

```
[----------] 28 tests from AccountStoreTest/2, where TypeParam = monad::db::impl::TrieDBImpl<monad::db::impl::InMemoryTrieSetup, monad::execution::BoostFiberExecution>
[ RUN      ] AccountStoreTest/2.account_exists
/usr/include/c++/13/bits/char_traits.h:275:23: runtime error: null pointer passed as argument 2, which is declared to never be null
...
```

_Originally posted by @sheapwilliams in https://github.com/monad-crypto/monad/issues/135#issuecomment-1610375639_

PR #722, opened 2024-05-21
Receipt decode fix and add contract address on create

brett-monad:
Fix for receipt decoding (issue #177) and addition of contract address to receipts on create (issue #136).

### DIFF
--- a/libs/db/src/monad/mpt/compute.hpp
+++ b/libs/db/src/monad/mpt/compute.hpp
@@ -279,7 +279,6 @@ struct VarLenMerkleComputeBase : Compute
         if (node->number_of_children() == 0) {
             MONAD_ASSERT(node->has_value());
             return encode_two_pieces(
-                // buffer, node->path_nibble_view(), node->value(), true);
                 buffer, node->path_nibble_view(), TComputeLeafData::compute(node->value()), true);
         }
         // Ethereum extension: there is non-empty path
@@ -373,7 +372,7 @@ private:
                    concat(single_child.branch, node->path_nibble_view()),
                    /* second: branch hash or leaf value */
                    node->mask ? (node->bitpacked.data_len ? node->data()
-                                                          : [&] -> byte_string {
+                                                : [&] -> byte_string {
                        MONAD_ASSERT(!node->has_path());
                        unsigned char branch_hash[KECCAK256_SIZE];
                        return {branch_hash, compute_branch(branch_hash, node)};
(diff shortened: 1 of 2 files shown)


libs/execution/src/monad/core/rlp/receipt_rlp.cpp
```diff
@@ -112,13 +119,6 @@ Result<std::vector<Receipt::Log>> decode_logs(byte_string_view &enc)
 {
     std::vector<Receipt::Log> logs;
     BOOST_OUTCOME_TRY(auto payload, parse_list_metadata(enc));
-    constexpr size_t approx_data_size = 32;
-    constexpr size_t approx_num_topics = 10;
-    // 20 bytes for address, 33 bytes per topic
-    constexpr auto log_size_approx =
-        20 + approx_data_size + 33 * approx_num_topics;
-    auto const list_space = payload.size();
-    logs.resize(list_space / log_size_approx);
```
Chen-Yifan, line 121: the intended might be `logs.reserve()`

libs/db/src/monad/mpt/compute.hpp
```diff
@@ -214,7 +214,13 @@ in the middle of a variable length trie.
 
 TODO for vicky: consolidate VarLenMerkleCompute and MerkleCompute into one.
 */
-struct VarLenMerkleCompute : Compute
+
+template <typename T>
+concept var_len_compute_leaf_data = requires {
+    { T::compute(std::declval<byte_string_view const>()) } -> std::same_as<byte_string>; };
+
+template <var_len_compute_leaf_data TComputeLeafData>
```
Chen-Yifan, line 222: can we rename `TComputeLeafData` to `TVarLenComputeLeafData`, just to differentiate from the one in `MerkleComputeBase`

libs/execution/src/monad/core/test/test_receipt_rlp.cpp
```diff
@@ -192,6 +192,57 @@ TEST(Rlp_Receipt, DecodeEncodeEip155Receipt)
     }
 }
 
+TEST(Rlp_Receipt, DecodeEncodeReceiptLogSize)
```
Chen-Yifan, line 195: can we split the rlp receipt changes (impl and test) to a separate commit?

libs/execution/src/monad/core/rlp/receipt_rlp.cpp
tongzhi1998: Yes indeed, but we saw yesterday that this is far from being accurate (it often allocates way more space in vector then it needs because of the long-tail distribution. e.g. Some logs can have 1k topics)

PR #724, opened 2024-05-22
Removed resize code that was causing receipt rlp decode to not match …

brett-monad:
…encode (issue #177)

### DIFF
--- a/libs/execution/src/monad/core/test/test_receipt_rlp.cpp
+++ b/libs/execution/src/monad/core/test/test_receipt_rlp.cpp
@@ -52,11 +52,11 @@ TEST(Rlp_Receipt, DecodeEncodeLog)
         EXPECT_EQ(encoded, rlp_log);
 
         byte_string_view encoded_log_view{encoded};
-        auto const decoded_log_opt = decode_log(encoded_log_view);
-        ASSERT_FALSE(decoded_log_opt.has_error());
+        auto const decoded_log_outcome = decode_log(encoded_log_view);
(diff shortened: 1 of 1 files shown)


libs/execution/src/monad/core/test/test_receipt_rlp.cpp
```diff
@@ -192,6 +192,57 @@ TEST(Rlp_Receipt, DecodeEncodeEip155Receipt)
     }
 }
 
+TEST(Rlp_Receipt, DecodeEncodeReceiptLogSize)
+{
+    using namespace intx;
```
tongzhi1998: I know you probably copied this from above. I don't think these are actually needed. Can you remove this and the ones in the previous test case?

libs/execution/src/monad/core/test/test_receipt_rlp.cpp
```diff
@@ -52,16 +52,17 @@ TEST(Rlp_Receipt, DecodeEncodeLog)
         EXPECT_EQ(encoded, rlp_log);
 
         byte_string_view encoded_log_view{encoded};
-        auto const decoded_log = decode_log(encoded_log_view);
-        ASSERT_FALSE(decoded_log.has_error());
+        auto const decoded_log_opt = decode_log(encoded_log_view);
+        ASSERT_FALSE(decoded_log_opt.has_error());
```
tongzhi1998: I like this change. Just to clarify that the return value is not a `std::optional` but a `boost::outcome`. So maybe we should change the name of the variable (and maybe the commit message too) to avoid confusion.

PR #732, opened 2024-05-29
Relaxed merge of account nonce and balance

magedm:
The relaxed merge applies relative (instead of absolute) checks on the nonce and balance of accounts accessed by transactions.

This PR reduces the number of retries on 12M 10K blocks, 4 threads, 128 fibers:
```
Before: 868765
After:  307027
```

### DIFF
--- a/libs/execution/src/monad/state2/block_state.cpp
+++ b/libs/execution/src/monad/state2/block_state.cpp
@@ -121,9 +121,13 @@ bool BlockState::can_merge(State const &state)
         auto const &account = account_state.account_;
         auto const &storage = account_state.storage_;
         StateDeltas::const_accessor it{};
-        MONAD_ASSERT(state_.find(it, address));
-        if (account != it->second.account.second) {
-            return false;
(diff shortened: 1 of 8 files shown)


libs/execution/src/monad/state2/block_state.cpp
```diff
@@ -121,9 +121,13 @@ bool BlockState::can_merge(State const &state)
         auto const &account = account_state.account_;
         auto const &storage = account_state.storage_;
         StateDeltas::const_accessor it{};
-        MONAD_ASSERT(state_.find(it, address));
-        if (account != it->second.account.second) {
-            return false;
+        [[maybe_unused]] bool const found = state_.find(it, address);
+        MONAD_ASSERT(found);
```
tongzhi1998, line 125: Just out of curiosity, why can't we still use the original `MONAD_ASSERT(state_.find(it, address));`

magedm: It isn't necessary to make this change. But I find it confusing that an essential action is in an assertion. A common view of assertions is that they should be treated as optional.

PR #735, opened 2024-05-31
to_binary writes a trie db to disk in a binary format (issue #197)

brett-monad:


cmd/replay_ethereum.cpp
```diff
@@ -223,7 +223,7 @@ int main(int const argc, char const *argv[])
 
     if (!dump_snapshot.empty()) {
         LOG_INFO("Dump db of block: {}", last_block_number);
-        write_to_file(db.to_json(), dump_snapshot, last_block_number);
+        db.to_binary(dump_snapshot, last_block_number);
```
kkuehlz, line 226: Should we plumb a corresponding a command line option so users can choose binary or JSON format? Only because the JSON can be helpful for debugging stuff or getting a readable sample of data.

PR #741, opened 2024-05-31
Add c++ wrapper for OpenSSL keccak256

tongzhi1998:


libs/core/src/monad/core/keccak.cpp
```diff
@@ -0,0 +1,12 @@
+#include <monad/core/keccak.h>
+#include <monad/core/keccak.hpp>
+
+#include <ethash/hash_types.hpp>
+
+ethash::hash256
+keccak256(unsigned char const *const in, unsigned long const len)
```
kkuehlz: I think you can consolidate this into a single keccak file. The `.h` file is not needed and just adds a public definition for something you are replacing. I'd say bring the stuff from `keccak.c` into the anonymous namespace in this file and remove `keccak.h` entirely.

tongzhi1998: Sorry for the late reply. I replied something under the execution channel. The old c-style keccak function is still called in some places throughout the codebase.

PR #747, opened 2024-06-07
Merge replay_etheruem into monad and remove replay_ethereum

andreaslyn:
Remove the `replay_ethereum` executable and merge its functionality into the `monad` executable. Closes https://github.com/monad-crypto/monad-internal/issues/234

I have added two new flags

- Flag `--chain` to select chain implementation. Currently one can select between `EthereumMainnet` and the new `MonadDevnet`. No default value at the moment, so user must provide this flag.
- Flag `--block_compression` is used to select the block compression method used for blocks in `block_db`. Select between Brotli compression and none (no compression). The default is Brotli at the moment.

### DIFF
--- a/cmd/monad.cpp
+++ b/cmd/monad.cpp
@@ -68,7 +68,7 @@ static TrieDb make_db(
             "directory you pass contains the block number of the snapshot "
             "in its path and includes files 'accounts' and 'code'.");
     }
-    uint64_t init_block_number = std::stoul(snapshot.stem());
+    uint64_t init_block_number = std::stoull(snapshot.stem());
     MONAD_ASSERT(fs::exists(snapshot / "code"));
     LOG_INFO("Loading from binary checkpoint in {}", snapshot);
     std::ifstream accounts(snapshot / "accounts");


cmd/monad.cpp
```diff
@@ -123,30 +236,49 @@ int main(int const argc, char const *argv[])
...
         ->transform(CLI::CheckedTransformer(log_level_map, CLI::ignore_case));
+    auto *has_genesis_file = cli.add_option(
+        "--genesis_file", genesis_file_path, "genesis file directory")
+        ->check(CLI::ExistingFile);
+    auto *has_nblocks = cli.add_option(
+            "--nblocks", nblocks,
+            "number of blocks to execute (unbounded when unspecified)");
     cli.add_option("--nthreads", nthreads, "number of threads");
     cli.add_option("--nfibers", nfibers, "number of fibers");
+    cli.add_flag("--no_compaction", no_compaction, "disable compaction");
```
andreaslyn, line 360: The corresponding flag is called `--no-compaction` in `replay_ethereum`. I renamed it `--no_compaction` to be consistent with other flags.

cmd/monad.cpp
```diff
@@ -100,20 +191,42 @@ void run_monad(
...
         LOG_DEBUG("generated receipts {}", receipts.assume_value());
         block_state.log_debug();
         block_state.commit(receipts.assume_value());
 
-        LOG_INFO(
+        LOG_DEBUG(
             "finished executing {} txns in block {}, time elasped={}",
             block.transactions.size(),
             block.header.number,
             std::chrono::steady_clock::now() - before);
```
andreaslyn: I changed this to log level debug, because logging this for every block seemed excessive.

libs/execution/src/monad/db/block_db.cpp
```diff
@@ -46,7 +46,9 @@ bool BlockDb::get(uint64_t const num, Block &block) const
     byte_string_view view2{brotli_buffer};
 
     auto const decoded_block = rlp::decode_block(view2);
-    MONAD_ASSERT(!decoded_block.has_error());
+    if (decoded_block.has_error()) {
+        return false;
+    }
```
andreaslyn: Change to `return false` instead of assertion error to accommodate the case where a block just arrived and only partially written. Should be consistent with the previous behavior of `monad`.

cmd/monad.cpp
```diff
@@ -123,30 +236,49 @@ int main(int const argc, char const *argv[])
...
-        ->required()
-        ->check(CLI::ExistingFile);
     cli.add_option("--log_level", log_level, "level of logging")
         ->transform(CLI::CheckedTransformer(log_level_map, CLI::ignore_case));
+    auto *has_genesis_file = cli.add_option(
+        "--genesis_file", genesis_file_path, "genesis file directory")
+        ->check(CLI::ExistingFile);
+    auto *has_nblocks = cli.add_option(
+            "--nblocks", nblocks,
+            "number of blocks to execute (unbounded when unspecified)");
```
andreaslyn: The `--nblocks` flag has changed slightly when compared to `replay_ethereum`. When the `--nblocks` flag is not specified, that means execute an unbounded number of blocks (until interrupted). In `replay_ethereum` the default was to execute zero blocks, in this case when `--nblocks` not specified.

libs/execution/src/monad/db/block_db.cpp
```diff
@@ -16,10 +17,46 @@
 #include <string>
 #include <string_view>
 
+using namespace monad;
+
+static bool
```
andreaslyn, line 22: I implemented the Brotli decompession in a more robust way, using `BrotliDecoderDecompressStream` to avoid guessing an upper bound on decompressed data

libs/execution/src/monad/chain/monad_devnet.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+evmc_revision MonadDevnet::get_revision(BlockHeader const &) const
+{
+    return EVMC_SHANGHAI;
+}
+
+Result<void>
+MonadDevnet::static_validate_header(BlockHeader const &header) const
+{
+    // EIP-779
```
andreaslyn, line 33: Copy pasted from `ethereum_mainnet.cpp`

cmd/monad.cpp
```diff
@@ -33,44 +34,130 @@ quill::Logger *tracer = nullptr;
...
+    }
+    if (!(fs::is_directory(snapshot) &&
+            fs::exists(snapshot / "accounts") &&
+            fs::exists(snapshot / "code"))) {
+        throw std::runtime_error(
+            "Invalid snapshot folder provided. Please ensure that the "
+            "directory you pass contains the block number of the snapshot "
+            "in its path and includes files 'accounts' and 'code'.");
+    }
+    uint64_t init_block_number = std::stoul(snapshot.stem());
```
tongzhi1998: Maybe we can change this to `std::stoull` to avoid 32/64 bit system indiscrepency.

cmd/monad.cpp
```diff
@@ -33,44 +34,130 @@ quill::Logger *tracer = nullptr;
...
-
-        std::ifstream istream(path);
-        std::ostringstream buf;
-        buf << istream.rdbuf();
-        auto view = byte_string_view{
-            (unsigned char *)buf.view().data(), buf.view().size()};
-        auto block_result = rlp::decode_block(view);
-        if (block_result.has_error()) {
-            continue;
+    // TODO: replace with monad specfiic mainnet
```
tongzhi1998: Small typo "specific".

cmd/monad.cpp
tongzhi1998: I think we could use the log_tps() lambda from replay_block_db to record stats, etc.

andreaslyn: Right, below I have already integrated the `log_tps` into the run-loop. Perhaps we can remove this `LOG_DEBUG`?

tongzhi1998: Sorry I overlooked. Yeah I think removing this is fine.

PR #748, opened 2024-06-07
Speed up traverse by async i/o

Chen-Yifan:
On top of https://github.com/monad-crypto/monad/pull/745 

Made two APIs for traverse: `Db::traverse_blocking()` and `Db::traverse()`, with
 the latter implemented as parallel traverse. We can remove `traverse_blocking()` 
later if needed.

**Note on `Db::traverse()`:**
It does parallel traversal through async i/o. Note that RWDb impl waits 
on a fiber future, therefore any parallel traverse run on RWDb should not
do any blocking i/o because that will block the fiber and hang forever. If you
have to do blocking i/o during the traversal on RWDb, use the
`traverse_blocking` api.

**Benchmark `to_json()` at 3M**
- parallel takes 34s (without using sqpoll kernel thread)
``` 
16:02:08.066194436 [2014192] replay_ethereum.cpp:225      LOG_INFO      root         Dump db of block: 3000000
16:02:42.791833847 [2014192] util.cpp:37                  LOG_INFO      root         Finished dumping to json file at block = 3000000, time elapsed = 3594ms
```
- parallel traverse with sqpoll thread: 21s

- blocking takes 191s
```
16:05:58.307858808 [2019042] replay_ethereum.cpp:225      LOG_INFO      root         Dump db of block: 3000000
16:08:37.589583008 [2019042] util.cpp:37                  LOG_INFO      root         Finished dumping to json file at block = 3000000, time elapsed = 3023ms
```
generated state.json are identical

See discussion on 12M state traversal in the linked issue https://github.com/monad-crypto/monad-internal/issues/222

### DIFF
--- a/libs/db/src/monad/mpt/traverse.hpp
+++ b/libs/db/src/monad/mpt/traverse.hpp
@@ -90,19 +90,20 @@ namespace detail
         {
             static constexpr bool lifetime_managed_internally = true;
             preorder_traverse_impl *impl;
-            unsigned char const branch;
             std::unique_ptr<TraverseMachine> traverse;
             chunk_offset_t rd_offset{0, 0};
             unsigned bytes_to_read;
             uint16_t buffer_off;
+            unsigned char const branch;
 
             receiver_t(
                 preorder_traverse_impl *impl, unsigned char const branch,
                 chunk_offset_t const offset,
                 std::unique_ptr<TraverseMachine> traverse)
                 : impl(impl)
-                , branch(branch)
                 , traverse(std::move(traverse))
+                , branch(branch)
+
             {
                 auto const num_pages_to_load_node =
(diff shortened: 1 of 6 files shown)


libs/db/src/monad/mpt/traverse.hpp
```diff
@@ -1,20 +1,209 @@
...
+            }
+        };
+
+        void process(
+            Node const &node, unsigned char const branch,
+            TraverseMachine &traverse)
+        {
+            traverse.down(branch, node);
+            for (unsigned char i = 0; i < 16; ++i) {
+                if (node.mask & (1u << i)) {
```
ned14: Variable bit shifts are slow :)

Chen-Yifan: fixed, also fixed a similar case in `load_all_impl_`

PR #765, opened 2024-06-24
construct TrieDb from mpt::Db

zander-xyz:
mostly moving code around

new db construction/load pattern
1. construct mpt::Db
2. load mpt::Db (ie. with binary checkpoint, or with statesync in the future)
3. construct TrieDb with fully initialized mpt::Db

### DIFF
--- a/libs/execution/src/monad/db/trie_db.cpp
+++ b/libs/execution/src/monad/db/trie_db.cpp
@@ -6,30 +6,25 @@
 #include <monad/core/bytes.hpp>
 #include <monad/core/fmt/bytes_fmt.hpp> // NOLINT
 #include <monad/core/fmt/int_fmt.hpp> // NOLINT
-#include <monad/core/int.hpp>
+#include <monad/core/keccak.h>
 #include <monad/core/keccak.hpp>
-#include <monad/core/likely.h>
 #include <monad/core/receipt.hpp>
 #include <monad/core/rlp/bytes_rlp.hpp>
 #include <monad/core/rlp/int_rlp.hpp>
 #include <monad/core/rlp/receipt_rlp.hpp>
-#include <monad/core/unaligned.hpp>
 #include <monad/db/trie_db.hpp>
+#include <monad/db/util.hpp>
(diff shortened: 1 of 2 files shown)


libs/db/src/monad/mpt/db.cpp
```diff
@@ -974,6 +975,16 @@ size_t Db::poll(bool const blocking, size_t const count)
...
 }
 
+bool Db::is_on_disk() const
+{
+    return impl_->aux().is_on_disk();
+}
+
+bool Db::is_read_only() const
+{
+    return impl_->aux().io->is_read_only();
```
Chen-Yifan: `io` can be `nullptr` for in memory db

libs/execution/src/monad/db/trie_db.cpp
```diff
@@ -47,555 +35,27 @@
...
-        if (config.has_value()) {
-            return std::make_unique<OnDiskMachine>();
-        }
-        return std::make_unique<InMemoryMachine>();
-    }()}
-    , db_{config.has_value() ? mpt::Db{*machine_, config.value()}
-                             : mpt::Db{*machine_}}
-    , block_number_{[&] {
-        if (config.has_value()) {
-            if (config->start_block_id.has_value()) {
```
Chen-Yifan, line 573: It looks like we don't allow user to pass in `start_block_id` anymore, when resume it always starts from the latest block number if on disk, so I think it's time to remove `OnDiskConfig::start_block_id` field.

libs/execution/src/monad/db/trie_db.cpp
```diff
@@ -770,7 +228,7 @@ void TrieDb::commit(
 
 void TrieDb::increment_block_number()
 {
-    if (mode_ == Mode::OnDisk) {
+    if (db_.is_on_disk()) {
         ++block_number_;
     }
```
Chen-Yifan, line 258: can probably add this assertion so that it's more clear in memory block_number_ is always 0:
```else{ MONAD_ASSERT(block_number_ == 0)}```

PR #766, opened 2024-06-25
Remove block number prefix 

kkuehlz:
DB Changes
----------
Change single root_offset to be a ring buffer of versions

  * Replace `root_offset` in `db_metadata` with a thread safe ring
    buffer of versions. The assumption is that the min and max versions
    are _always_ updated in lockstep with calls to
    `advance_offsets_to_()`. That assumption holds in replay ethereum,
    however, would be good if both of those calls could be conjoined
    into a single function. Doing that right now is out of scope since
    it would require nontrivial changes to API boundaries.
  * The storage offset for a block is mapped by subtracting the version
    from the maximum version in the metadata. Users can't query specific
    indices on the root buffer; only offsets relative to the root are
    allowed. This way, even when the ring buffer is not full everything
  * Get all existing tests passing with these new changes.

Add RWDb test for FiberLoadRootVersionRequest coverage

Remove find_min_key_blocking() and find_max_key_blocking(). No longer
needed under new design.

Remove BLOCK_NUM_* constants, verify version after deserialization in async get

Testing
-------
Improve read_only_db_single_thread_async

  * query historical version
  * will make IO actually async in followup

`DbTest.read_only_db_traverse_concurrent`,
  * now actually does traverse as name suggests

 edited some trie tests, add `StateMachineMerkleWithPrefix` and
`StateMachineVarLenTrieWithPrefix`

Public API changes
-------------------
Remove `is_latest()` and `load_latest()` APIs

  * Db layer: Remove is_latest() and load_latest() APIs
  * TrieDb layer: Remove is_latest() and load_latest() APIs
  * The RODb will check the version and load only the correct root as needed.
  * Loading the same version is a no-op, and won't re-read the root from
    disk.

Core Changes
------------
Add simple single threaded LRU cache for storing async root nodes

It is no longer valid to call UpdateAux().max_version_in_db_history() or
UpdateAux().min_version_in_db_history() from in memory database. In
memory database can just keep track of the first and last block ids
upserted. There is no value in providing a helper for that.

Co-authored-by: Vicky Chen <yifanc3@illinois.edu>


libs/db/src/monad/mpt/update_aux.cpp
```diff
@@ -581,19 +562,7 @@ Node::UniquePtr UpdateAuxImpl::move_subtrie(
...
-    auto root = copy_node(
-        *this,
-        std::move(prev_root),
-        src_block_number_prefix,
-        serialize_as_big_endian<BLOCK_NUM_BYTES>(dest));
-    Update u = make_erase(src_block_number_prefix);
-    UpdateList updates;
-    updates.push_front(u);
-    root = upsert(*this, sm, std::move(root), std::move(updates));
     update_ondisk_db_history_metadata(dest, dest);
```
Chen-Yifan: does this update the root offset of dest?

libs/db/src/monad/mpt/db.cpp
```diff
@@ -130,7 +131,7 @@ struct Db::ROOnDisk final : public Db::Impl
                                        ->max_db_history_version.load(
                                            std::memory_order_acquire)}
         , last_loaded_offset_{[&] {
-            auto root_offset = aux_.get_root_offset();
+            auto root_offset = aux_.get_latest_root_offset();
```
Chen-Yifan: seems no need to init RODb with latest version root. Can just init the `last_loaded_offset_` INVALID_OFFSET. I think removing this throw will also lead to a unit test being removed

libs/db/src/monad/mpt/update_aux.cpp
kkuehlz: We talked about this. The problem is the version bump and root offset bump are at different API layers. They _should_ be updated in lockstep, but they aren't, and plumbing the version through to the freestanding `upsert` function would be a big refactor.

* [Advance offsets](https://github.com/monad-crypto/monad/blob/ae15efb1be0359bb028284566002611efb4662eb/libs/db/src/monad/mpt/trie.cpp#L1377-L1380) (in `trie.cpp`)
* [Bump version](https://github.com/monad-crypto/monad/blob/ae15efb1be0359bb028284566002611efb4662eb/libs/db/src/monad/mpt/update_aux.cpp#L566) (in `update_aux.cpp`)

Chen-Yifan: but you can get root offset from the location at `src` in the ring buffer. Can we just copy it to the slot for dest?

libs/db/src/monad/mpt/db.cpp
```diff
@@ -566,13 +587,24 @@ struct Db::RWOnDisk final : public Db::Impl
...
+                        upsert_promises.emplace_back(std::move(
+                            *req->promise)); // TODO: switch this to its own vec
                         req->promise = &upsert_promises.back();
-                        req->promise->set_value(aux.move_subtrie(
-                            std::move(req->prev_root),
-                            req->sm,
-                            req->src,
-                            req->dest));
+                        auto root = aux.contains_version(req->version)
+                                        ? Node::UniquePtr{read_node_blocking(
```
Chen-Yifan, line 601: blocking read will make fiber hang, need it async. 
I think we don't have to have an additional request, we can make `find_fiber_blocking()` take an optional `NodeCursor`, and handle the case of loading the root in the beginning `FindFiberRequest`.

kkuehlz: This isn't in a fiber, it's in the trie db worker thread

Chen-Yifan: ah sorry, mb. Ok then I have no strong opinion on changing it, execution doesn't really go through this path to read history states

libs/db/src/monad/mpt/update_aux.cpp
kkuehlz: I don't think we need to do that since there's only a single version. When you call dest on the ring buffer for dest, it should return the offset that was at source. That's why the unit test you have passes.

libs/db/src/monad/mpt/db.cpp
kkuehlz: We would need to change `last_loaded_max_version_` to be `std::optional` too then. You think it is worth it?

Chen-Yifan: they don't seem to be relevant, `root_` and `last_loaded_offset_` are associated. Now I think about it, you may also want a `uint64_t current_version_` to represent the version your `root_` is on, so to save some redundant reading of roots of the same version.

kkuehlz: setting this to INVALID_OFFSET causes lots of tests to fail. I think we should do both of these in followups. there are several other parts of this code that assume a root is already loaded.

Chen-Yifan: let's add the `current_version_` thing, I'm okay with initializing `last_loaded_offset_` on construction

kkuehlz: done

PR #771, opened 2024-07-02
Include raw address in account leaf

kkuehlz:
* Add address to account leaf
* Refactor storage leaf to be an rlp encoded list instead of adjacent

### DIFF
--- a/libs/execution/src/monad/db/trie_db.cpp
+++ b/libs/execution/src/monad/db/trie_db.cpp
@@ -4,6 +4,7 @@
 #include <monad/core/assert.h>
 #include <monad/core/byte_string.hpp>
 #include <monad/core/bytes.hpp>
+#include <monad/core/fmt/address_fmt.hpp> // NOLINT
 #include <monad/core/fmt/bytes_fmt.hpp> // NOLINT
 #include <monad/core/fmt/int_fmt.hpp> // NOLINT
 #include <monad/core/keccak.h>
@@ -76,9 +77,8 @@ std::optional<Account> TrieDb::read_account(Address const &addr)
     }
 
     auto encoded_account = value.value();
-    auto acct = decode_account_db(encoded_account);
+    auto const acct = decode_account_db_ignore_address(encoded_account);
     MONAD_DEBUG_ASSERT(!acct.has_error());
-    MONAD_DEBUG_ASSERT(encoded_account.empty());
     return acct.value();
 }
 
@@ -106,15 +106,9 @@ TrieDb::read_storage(Address const &addr, Incarnation, bytes32_t const &key)
     }
     STATS_STORAGE_VALUE();
     auto encoded_storage = value.value();
-    auto const storage = decode_storage_value_only(encoded_storage);
+    auto const storage = decode_storage_db_ignore_slot(encoded_storage);
     MONAD_ASSERT(!storage.has_error());
-
-    bytes32_t output;
-    std::copy_n(
-        storage.value().begin(),
-        storage.value().size(),
-        output.bytes + sizeof(bytes32_t) - storage.value().size());
-    return output;
+    return to_bytes(storage.value());
 };
 
 std::pair<bytes32_t, bytes32_t>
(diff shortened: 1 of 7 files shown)


libs/execution/src/monad/core/account.hpp
```diff
@@ -14,16 +15,17 @@ struct Account
...
     bytes32_t code_hash{NULL_HASH}; // sigma[a]_c
+    Address address{};
     uint64_t nonce{0}; // sigma[a]_n
     Incarnation incarnation{0, 0};
 
     friend bool operator==(Account const &, Account const &) = default;
 };
 
-static_assert(sizeof(Account) == 80);
+static_assert(sizeof(Account) == 104);
```
kkuehlz: note: this adds 4 bytes of padding.

libs/execution/src/monad/db/trie_db.cpp
```diff
@@ -655,24 +679,7 @@ std::optional<Account> TrieDb::read_account(Address const &addr)
...
-    }
-    STATS_STORAGE_VALUE();
-    auto encoded_storage = value.value();
-    auto const storage = rlp::decode_bytes32_compact(encoded_storage);
-    MONAD_ASSERT(!storage.has_error());
-    MONAD_ASSERT(
-        encoded_storage.size() >= 1 &&
-        encoded_storage.size() <= sizeof(bytes32_t) + 1);
-    return storage.value();
+    return read_storage_and_slot(addr, key).second;
```
Chen-Yifan: wouldn't this add one more memcpy?

libs/execution/src/monad/db/trie_db.cpp
```diff
@@ -97,6 +101,7 @@ namespace
         BOOST_OUTCOME_TRY(acct.nonce, rlp::decode_unsigned<uint64_t>(payload));
         BOOST_OUTCOME_TRY(
             acct.balance, rlp::decode_unsigned<uint256_t>(payload));
+        BOOST_OUTCOME_TRY(auto const address, rlp::decode_address(payload));
```
Chen-Yifan: The db commit hot path `ComputeAccountLeaf` invokes this func but address is ignored, we can optimize the memcpy away if replace it by `parse_string_metadata(payload)`. I noticed that this func also used elsewhere, so can just create a separate function in `ComputeAccountLeaf` that does the optimized decode which ignores address

libs/execution/src/monad/db/trie_db.cpp
kkuehlz: i thought the point of this refactor was readability

Chen-Yifan: I think it was to not invent a non-standard encoding method

libs/execution/src/monad/db/trie_db.cpp
kkuehlz: fwiw, i measured this branch against main on 12M+10K and don't see any perf regressions (if you want to verify, I can sync over the 12M checkpoint for this branch).

jhunsaker: with in-memory test?

kkuehlz: @jhunsaker was doing tests using disk - my mistake. i updated this branch with @Chen-Yifan's feedback already before you commented.

I think GitHub keeps force push history, so i will compare the before and after.

Chen-Yifan: would it be less code duplication if we encode address in the beginning?

kkuehlz: @Chen-Yifan latest commit is an idea i had to deduplicate the code. let me know your thoughts...

kkuehlz: @jhunsaker just re-ran benchmarks with this branch (including the latest commit) and latest `main`

```
branch,storage,checkpoint,nblocks,time(s),tps
main,mem,3M,60000,18,25264
main,mem,12M,25000,366,13570
main,disk,3M,60000,40,11368
main,disk,12M,25000,665,7468
PR771,mem,3M,60000,17,26750
PR771,mem,12M,25000,364,13644
PR771,disk,3M,60000,36,12632
PR771,disk,12M,25000,663,7491
```

Chen-Yifan: it's a smart try, but I'm not sure how we feel about the templated return. Especially after the recent commit to main, we had to put those to header file as they are templated. I was more thinking about having two separate functions one return `Result<Account>` one returns `Result<std::pair<>>`, and a common helper function decoding the rest of the payload after parsing the first element in list (the address), either ignore it or construct an `Address`.

kkuehlz: @Chen-Yifan updated and rebased

PR #773, opened 2024-07-03
Alee/reserve

zander-xyz:
- execution will validate that there is enough reserve balance to pay for the intrinsic gas
- execution will still charge the carriage cost if the balance is insufficient to pay for the value transfer + gas limit

### DIFF
--- a/libs/execution/src/monad/chain/chain.hpp
+++ b/libs/execution/src/monad/chain/chain.hpp
@@ -1,16 +1,22 @@
 #pragma once
 
 #include <monad/config.hpp>
+#include <monad/core/address.hpp>
 #include <monad/core/bytes.hpp>
 #include <monad/core/int.hpp>
 #include <monad/core/result.hpp>
 
 #include <evmc/evmc.h>
+#include <optional>
 
 MONAD_NAMESPACE_BEGIN
 
+struct Account;
+class BlockHashBuffer;
 struct BlockHeader;
 struct Receipt;
+class State;
+struct Transaction;
 
 struct Chain
 {
@@ -29,6 +35,20 @@ struct Chain
         evmc_revision, BlockHeader const &, bytes32_t const &state_root,
         bytes32_t const &receipts_root, bytes32_t const &transactions_root,
         std::optional<bytes32_t> const &withdrawals_root) const = 0;
+
+    virtual Result<void> validate_transaction(
+        evmc_revision, Transaction const &, Address const &,
+        std::optional<Account> const &) const = 0;
+
+    virtual evmc::Result execute_impl_no_validation(
+        evmc_revision, BlockHashBuffer const &, BlockHeader const &, State &,
+        Transaction const &, Address const &sender,
+        std::optional<Account> const &) = 0;
+
+    virtual Receipt execute_final(
+        evmc_revision, State &, Transaction const &, Address const &sender,
+        uint256_t const &base_fee_per_gas, evmc::Result const &,
+        Address const &beneficiary) = 0;
 };
 
 MONAD_NAMESPACE_END
--- a/libs/execution/src/monad/chain/monad_chain.cpp
+++ b/libs/execution/src/monad/chain/monad_chain.cpp
@@ -1,9 +1,44 @@
 #include <monad/chain/monad_chain.hpp>
 #include <monad/config.hpp>
+#include <monad/core/block.hpp>
+#include <monad/execution/execute_transaction.hpp>
+#include <monad/execution/intrinsic_gas_buffer.hpp>
+#include <monad/execution/transaction_gas.hpp>
+#include <monad/execution/validate_transaction.hpp>
+#include <monad/state3/state.hpp>
 
 MONAD_NAMESPACE_BEGIN
 
 using BOOST_OUTCOME_V2_NAMESPACE::success;
+
+namespace
+{
+    uint256_t effective_reserve(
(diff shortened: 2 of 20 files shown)


libs/execution/src/monad/chain/monad_chain.cpp
```diff
@@ -1,10 +1,46 @@
...
+#include <monad/state3/state.hpp>
 
 MONAD_NAMESPACE_BEGIN
 
 using BOOST_OUTCOME_V2_NAMESPACE::success;
 
+namespace
+{
+    // TODO
+    constexpr uint256_t DEFAULT_MAX_RESERVE = 200'000;
```
mikety: The more practical value for max_reserve to pass validation is E17 - this is the number used in testing on devnet1.

libs/execution/src/monad/chain/monad_chain.cpp
```diff
@@ -1,10 +1,46 @@
...
+
+    uint256_t effective_reserve(
+        IntrinsicGasBuffer const &buf, uint256_t const &max_reserve,
+        Address const &addr, uint256_t const &balance)
+    {
+        if (balance < max_reserve) {
+            return balance;
```
mikety, line 21: Can you please clarify why not balance - gas ?

libs/execution/src/monad/execution/intrinsic_gas_buffer.hpp
```diff
@@ -0,0 +1,62 @@
...
+#pragma GCC diagnostic pop
+
+#include <cstdint>
+
+MONAD_NAMESPACE_BEGIN
+
+class IntrinsicGasBuffer
+{
+public:
+    static constexpr auto N = 4;
```
mikety: Can you please clarify why N = 4? I assume this is the degree of parallelism - i.e number of blocks being processed in parallel.

libs/execution/src/monad/chain/monad_chain.cpp
```diff
@@ -20,4 +56,60 @@ bool MonadChain::validate_root(
...
+    State &state, Transaction const &tx, Address const &sender,
+    std::optional<Account> const &acct)
+{
+    // Note: Intrinsic gas charged from the reserve balance
+    uint512_t const v0 =
+        tx.value +
+        max_gas_cost(tx.gas_limit - intrinsic_gas(rev, tx), tx.max_fee_per_gas);
+    auto const balance = acct.has_value() ? acct->balance : uint256_t{0};
+    if (effective_balance(DEFAULT_MAX_RESERVE, balance) < v0) {
+        return evmc::Result{EVMC_REJECTED};
```
mikety, line 89: Will this error be propagated to the receipt?

libs/execution/src/monad/chain/monad_chain.cpp
```diff
@@ -20,4 +56,60 @@ bool MonadChain::validate_root(
...
+    auto res = ::monad::validate_transaction(tx, acc);
+    if (res.has_error() &&
+        res.error() != TransactionError::InsufficientBalance) {
+        return res;
+    }
+
+    uint256_t const balance = acc.has_value() ? acc->balance : uint256_t{0};
+    if (max_gas_cost(intrinsic_gas(rev, tx), tx.max_fee_per_gas) >
+        effective_reserve(buf_, DEFAULT_MAX_RESERVE, sender, balance)) {
+        return TransactionError::InsufficientReserveBalance;
```
mikety, line 72: Will this error be propagated to the receipt?

libs/execution/src/monad/chain/monad_chain.cpp
zander-xyz: yes

libs/execution/src/monad/chain/monad_chain.cpp
zander-xyz: no

libs/execution/src/monad/execution/intrinsic_gas_buffer.hpp
zander-xyz: it should be set to whatever the consensus delay is

libs/execution/src/monad/chain/monad_chain.cpp
zander-xyz: that would be double charging

libs/execution/src/monad/execution/intrinsic_gas_buffer.hpp
mikety: in consensus execution_delay setting is 10. 
Please update. Nit: please add comment that N must match the consensus execution_delay .

libs/execution/src/monad/chain/monad_chain.cpp
mikety: in consensus max_reserve_balance is 100_000_000_000_000_000,
Please update. Nit: please add comment that DEFAULT_MAX_RESERVE must match the consensus max_reserve_balance.

libs/execution/src/monad/chain/chain.hpp
```diff
@@ -1,16 +1,23 @@
...
 
 #include <evmc/evmc.h>
+#include <optional>
 
 MONAD_NAMESPACE_BEGIN
 
+struct Account;
+class BlockHashBuffer;
 struct BlockHeader;
+class BlockState;
```
tongzhi1998: We probably don't need BlockState.

jhunsaker: removed

PR #780, opened 2024-07-10
Example of a different C style, not a real PR, for exposition only

kjcamann:
To understand what this is, read this comment: https://github.com/monad-crypto/monad-internal/issues/339#issuecomment-2219692344

### DIFF
--- a/libs/runloop/src/monad/async/context_switcher.h
+++ b/libs/runloop/src/monad/async/context_switcher.h
@@ -1,6 +1,26 @@
 #pragma once
 
+#ifdef __cplusplus
+#define MONAD_NODISCARD [[nodiscard]]
+#define MONAD_STD std::
+#define MONAD_ATOMIC(X) std::atomic<X>
+#else
+#define MONAD_NODISCARD __attribute__((warn_unused_result))
+#define MONAD_STD
+#define MONAD_ATOMIC(X) _Atomic X
+#endif
+
+#ifndef MONAD_PCONST
+    #ifdef MONAD_SOURCE
+        #define MONAD_PCONST
+    #else
+        #define MONAD_PCONST const
+    #endif
+#endif
+
 #include "config.h"
+
+#include <sys/queue.h>
 
 #include <stdatomic.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -1,7 +1,27 @@
...
 
+#ifdef __cplusplus
+#define MONAD_NODISCARD [[nodiscard]]
+#define MONAD_STD std::
+#define MONAD_ATOMIC(X) std::atomic<X>
+#else
+#define MONAD_NODISCARD __attribute__((warn_unused_result))
+#define MONAD_STD
+#define MONAD_ATOMIC(X) _Atomic X
+#endif
```
kjcamann, line 11: These macros would not actually be defined in this file, but in a common header. I put them here so I could check that this all compiles in `-std=c23` mode.

`MONAD_NODISCARD` is just a shorter spelling of the old `MONAD_ASYNC_NODISCARD` version.

`MONAD_STD` replaces the old code like

```{.cpp}
#ifdef __cplusplus
std::
#end
size_t atomic_uint; // std::atomic_uint in C++, atomic_uint in C
```

by making `MONAD_STD` into `std::` in C++ and nothing in C, then writing qualified names as `MONAD_STD atomic_uint`.

A similar thing for making things `std::atomic<T>` vs `_Atomic T`

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -1,7 +1,27 @@
...
+
+#ifndef MONAD_PCONST
+    #ifdef MONAD_SOURCE
+        #define MONAD_PCONST
+    #else
+        #define MONAD_PCONST const
+    #endif
+#endif
```
kjcamann, line 19: I actually am not sure what `MONAD_ASYNC_PUBLIC_CONST` was for. Was it to make some things constant externally, but not internally? In that case my change might not make sense as a "top-level" `MONAD_` thing, but would still need to be `MONAD_ASYNC_` to treat the library implementation differently than external code.

Normally I don't mind long identifiers as much as some people, but _lots_ of declarations had `MONAD_ASYNC_PUBLIC_CONST` in front in another header, it was pretty painful to look at IMO. Although `PCONST` is not very descriptive, if it's going to appear a lot, we might as well learn to accept it as a common abbreviation.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -1,7 +1,27 @@
...
+
 #include "config.h"
 
+#include <sys/queue.h>
```
kjcamann, line 23: Maybe we shouldn't use `<sys/queue.h>`, but it's probably not good to rewrite it entirely either.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
 {
 #endif
 
-typedef struct monad_async_task_head *monad_async_task;
-
+struct monad_async_context;
+struct monad_async_context_switcher;
+struct monad_async_task;
```
kjcamann, line 48: I was confused by the `_head` suffix, and originally thought it meant that it contained an intrusive list head for linking other objects.

Reading further in code it becomes clear that it's a "header" structure type, representing common fields for more specific types, where it will be the first field. It's like the "base class" of the actual implementation.

I suppose it's fine to call it anything, as long as we pick a common name for such a pattern and get used to it. `_head`, `_header`, `_hdr`?

One big question is, should there be typedefs at all here, and what rules will we follow? As we were talking about before, I also removed the implicit `*`.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+struct monad_async_asan_info {
+    union
+    {
+        void *fake_stack_save;
+        unsigned valgrind_stack_id;
+    };
+
+    void const *bottom;
+    size_t size;
+};
```
kjcamann, line 62: This was originally an anonymous structure inside `monad_async_context`. Here it has been pulled up to file scope and given it a name. There's two reasons:

1. Both the name and the associated doxygen declaration comment give the reader a better idea of what it's for
2. It's harder to "see" what all the members of the `monad_async_context` are if there are longer nested structures defined directly in the field declaration. In general it's fine if it's only two or three lines, as it doesn't disturb the flow of reading. There's an example below (the context tracker object), that I left alone.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+
+/// Defines the switcher interface
+///
+/// Multiple user-space context switcher backends are available
+/// (setjmp/longjmp-based, Boost.Context-based, C++20 coroutine-based, etc.).
+/// Each implementation provides a table of function pointers to implement
+/// the interface.
+struct monad_async_context_switcher_ops {
```
kjcamann, line 70: This is a pattern I borrowed from the original UNIX C code, and which is present in a lot of systems programs. If an object has regular data fields but is also a "virtual" interface, then the interface stuff is factored in a structure containing only function pointers.

Such a thing is called an "operations" structure, and here it ends in `_ops`.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+    /// Create a switchable context for a task
```
kjcamann, line 71: I got rid of the `//! \brief` everywhere.

In modern doxygen markup, I think it always extracts the first sentence as the "brief" text, so you no longer need the explicit command. Even if it doesn't always do it, I know there is at least a setting in the `Doxyfile` to do it.

And just as a pure stylistic thing, I vote for `///` over `//!` as the short-form doxygen comment introducer.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+    monad_async_result (*create_context)(
+        struct monad_async_context **context,
+        struct monad_async_context_switcher *switcher,
+        struct monad_async_task *task,
+        const struct monad_async_task_attr *attr);
+
+    /// Destroys a switchable context created by create_context
+    monad_async_result (*destroy_context)(struct monad_async_context *context);
+
```
kjcamann, line 80: I put some vertical spacing between the operations. It seems easier to read, but that's just my opinion.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+        struct monad_async_context_switcher *switcher,
         monad_async_result (*resumed)(
             void *user_ptr,
-            monad_async_context current_context_to_use_when_resuming),
+            struct monad_async_context *current_context_to_use_when_resuming),
         void *user_ptr);
 
+    /// Destroys the switcher object; the factory that creates the switcher
+    /// hangs its cleanup routine here
+    monad_async_result (*destroy_self)(
```
kjcamann, line 106: This used to be called `self_destroy`, now it's `destroy_self`. In general they all follow a pattern of `<verb>_<noun>` now. AFAIK this is the most common style, e.g., the pthread API `pthread_create_mutex` and not `pthread_mutex_create`.

I also removed some of the comment for `resume_many`. Generally long comments are good, but they can disrupt the flow of reading/understanding the interface if it's interrupted too much. I am not 100% sure, but I think you can put less important comment on the definition redeclaration and the two will be merged by Doxygen? This is a personal opinion but I think that's better: you get the gist of what it does from the header, and the details in both the source file and the documentation.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+{
+    void *user_ptr; ///< Opaque user data, passed into the resume_many callback
+
+    /*
+     * The following fields are not user modifiable
+     */
+
+    MONAD_PCONST MONAD_STD atomic_uint contexts; ///< Number of contexts
+    const struct monad_async_context_switcher_ops
+        *switcher_ops; ///< Switcher backend we're using
```
kjcamann, line 122: This uses the "trailing documentation" style, where the documentation of a field is written right after the field. Also note that the function pointers in the `_ops` structure are not const-qualified anymore, but now the whole structure itself is.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+
+/// Object which manages the low-level details of switching between two
+/// user-space contexts
+struct monad_async_context_switcher
+{
+    void *user_ptr; ///< Opaque user data, passed into the resume_many callback
+
+    /*
+     * The following fields are not user modifiable
+     */
```
kjcamann, line 118: This kind of comment:

```
/*
 * Hello
 */
```

takes more space, and it's generally to call out separate groups of things, or to highlight something important to make sure no-one misses it. Above we're telling them about a whole block of things they shouldn't modify.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
     {
-        mtx_t lock;
-        monad_async_context front, back;
-        size_t count;
-    } contexts_list;
+        mtx_t lock;                                ///< Protects contexts list
+        LIST_HEAD(, monad_async_context) contexts; ///< All contexts using switcher
+        size_t count;                              ///< Size of contexts list
+    } context_tracker;
 #endif
```
kjcamann: Because it's shorter, this inline anonymous struct was left alone and not factored out like the ASAN one. Also it uses `<sys/queue.h>` and the trailing documentation style.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
-    //! executor creates one of these per executor.
-    monad_async_result (*const create)(monad_async_context_switcher *switcher);
-} monad_async_context_switcher_impl;
-
-typedef struct monad_async_context_head
+    /// Create a switcher of contexts; this is called by the executor; the
+    /// corresponding destroy routine is part of the operations structure,
+    /// (see destroy_self)
+    monad_async_result (*const create)(struct monad_async_context_switcher **switcher);
+} monad_async_context_switcher_factory;
```
kjcamann: I don't love the name `_factory` (this isn't Java after all) but I changed it because `_impl` generally means "you are not supposed to use this", no? Are you supposed to use it, or does only the executor?

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
-            void *fake_stack_save;
-            unsigned valgrind_stack_id;
-        };
-
-        void const *bottom;
-        size_t size;
-    } sanitizer;
-} *monad_async_context;
+    struct monad_async_asan_info asan_stack_info;
+};
```
kjcamann: Look how much smaller the context is now. It's easier to see what it contains: a runnable state, a pointer to the switcher operations, linkage for the list of all contexts owned by the switcher type, and the ASAN stuff.

It's missing a nice comment though.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+/**
+ * Creates a none context switcher which can't suspend-resume. Useful
+ * for threadpool implementation.
+ *
+ * As this context switcher never suspends and resumes, it is safe to use a
+ * single instance of this across multiple threads. In fact, the current
+ * implementation always returns a static instance, and destruction does
+ * nothing. You may therefore find
+ * `monad_async_context_switcher_none_instance()` more useful.
+ */
```
kjcamann, line 193: This is a stylistic nit, but having a `*` reappear down every column makes it clear that this is a comment even in the absence of syntax highlighting.

Someone told me that at Bloomberg years ago and I thought it was just a silly nitpick, but it is really is jarring to not see a strong visual hint that you're looking at a giant block comment, it you're used to it. Right now, in this very PR user interface, we have very subtle highlighting because the strong colors are being used by the background diff.

As a person who _doesn't_ use this style, do you find the overzealous `*` similarly as jarring as I find its absence? I didn't use to do the `*`, but it didn't annoy me either (it's oddly satisfying!) which is why I went with the Bloomberg guy's suggestion and never looked back.

libs/runloop/src/monad/async/context_switcher.h
```diff
@@ -23,151 +43,174 @@ extern "C"
...
+    /// Create a switcher of contexts; this is called by the executor; the
+    /// corresponding destroy routine is part of the operations structure,
+    /// (see destroy_self)
+    monad_async_result (*const create)(struct monad_async_context_switcher **switcher);
+} monad_async_context_switcher_factory;
+
+enum monad_async_run_state {
+    Running,
+    Suspended
+};
```
kjcamann, line 146: In general "enums over flags" seems cleaner. I think I'm maybe missing something though, what if running and suspending are both false? `Inactive`? `Defunct`?

ned14: This puts `Running` and `Suspended` into the global namespace. I wouldn't recommend that.

PR #785, opened 2024-07-16
Triedb: Improve async read test and add inflight map for roots

kkuehlz:
Improve `OnDiskDbWithFileFixture.read_only_db_single_thread_async` to test
async reads. Previously, the test would poll between each read. The new test 
uncovers a bug that has been present on main since async was added. 
`find_get_sender()` was not passed an inflight map, so during async reads, 
the sender will issue duplicate reads and override existing nodes.
    
In 715c99d we added support for async root nodes. Similar to the
find_request_sender, add an inflight map for pending root nodes.
    
Reduce the dependency of async APIs on the DB class. The `AsyncContext`
struct has all the data needed for an async read without cluttering the
Db class with all these implementation details.  For example, we are
able to remove `db.trie_root_cache()` getter from the Db interface.

### DIFF
--- a/libs/db/src/monad/mpt/db.hpp
+++ b/libs/db/src/monad/mpt/db.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <memory>
+
 #include <monad/async/concepts.hpp>
 #include <monad/async/io.hpp>
 #include <monad/async/storage_pool.hpp>
@@ -18,19 +20,13 @@ struct OnDiskDbConfig;
 struct ReadOnlyOnDiskDbConfig;
 struct StateMachine;
 struct TraverseMachine;
-
-namespace detail
-{
-    template <class T>
-    struct DbGetSender;
-}
+struct AsyncContext;
 
 class Db
 {
-    template <class T>
-    friend struct detail::DbGetSender;
-
 private:
+    friend struct AsyncContext;
+
     struct Impl;
     struct RWOnDisk;
     struct ROOnDisk;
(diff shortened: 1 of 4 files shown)


libs/db/src/monad/mpt/db.hpp
```diff
@@ -92,71 +87,86 @@ class Db
...
+    };
 
-    async::result<void>
-    operator()(async::erased_connected_operation *io_state) noexcept;
+}
 
-    result_type completed(
-        async::erased_connected_operation *, async::result<void> res) noexcept;
-};
+AsyncContext *make_async_context(Db &db);
```
ned14: Should return a unique_ptr, not be manually lifetime managed.

kkuehlz: Totally agree in general but I think this will need to be created from the C API

Chen-Yifan: Sounds like we would need to have the async context stuff in a separate C header file so C code in bft side can actually include and use it. 
Also we prefer snake case as our C naming convention, and with `monad_` prefix (mostly learn from statesync library), so a few renaming suggestions:
`AsyncContext` -> `monad_triedb_async_context`
`make_async_context()` -> `monad_triedb_async_context_create()`
`free_async_context()` -> `monad_triedb_async_context_destroy()`

ned14: > Totally agree in general but I think this will need to be created from the C API

If in C++, you are in C++. Extra functions can be added to release a C pointer from a C++ unique ptr.

Not sure why you're caring about C from BFT side in a C++ header. If we want a C DB interface, the entire C++ API should be replaced with a C one and all C++ purged from `db.h`.

Chen-Yifan: the user of async rodb needs to create async context before any `triedb_async_read()` https://github.com/monad-crypto/monad-bft/blob/master/monad-triedb/triedb-driver/include/triedb.h#L28, either we add async context to the `struct triedb` or as a separate arg for `triedb_async_read()`. The latter will require it to be in a separate C interface

ned14: Me personally I'd put the C wrapper stuff into that triedb-driver, and not pollute the C++ codebase with C concerns.

If you'd like, I can further extend that triedb-driver with new C APIs so it becomes a stateful rather than stateless API?

kkuehlz: yeah, i don't know what i was thinking :)

this can all be done inside the `triedb-driver` - you are correct.

Updated the `AsyncContext` to be a `unique_ptr`.

ned14: It seems weird for there to be a `unique_ptr<AsyncContext>` when `AsyncContext` contains nothing but another unique ptr. Why have double pointers here?

kkuehlz: Compiler wasn't happy with my forward declared struct and `unique_ptr`. It was either this or pull all of `AsyncContext` into the header. If there's a better way to do this I'd like to know, because what i did is pretty ugly.

Chen-Yifan: would it be better if we have `AsyncContext` inherits from `Db`, and keeps a reference to `monad::Db& ro`. Then we no longer need the `AsyncContextImpl` or `make_async_context()`. 
triedb-driver would need a C struct wrapper for `AsyncContext` and `async_context_create()` and `destroy()` , `triedb_async_read()` will take the async context raw pointer handle instead of `triedb *` if that makes sense

kkuehlz: Not sure another level of inheritance is preferable here

ned14: > If there's a better way to do this I'd like to know, because what i did is pretty ugly.

What's wrong with a custom deleter for the unique ptr?

Chen-Yifan: > Not sure another level of inheritance is preferable here

I personally think that would set us up for a cleaner C interface for async. 
The way it is now is a bit interleaving that the context has `UpdateAux` but nothing else of rodb, and the triedb driver code would need `triedb*` to poll, but wrapper of `AsyncContext` for `triedb_async_read()`. 
Also I saw we are moving some but not all of the async functions out of Db::Impl to AsyncContext, the missing one is `Db::Impl::poll()`, the inheritance will help with that too.

kkuehlz: It just adds a lot of code to implement all the Db methods. Would this not suffice?

```    
explicit triedb(std::vector<std::filesystem::path> dbname_paths)
        : db_{monad::mpt::ReadOnlyOnDiskDbConfig{
              .disable_mismatching_storage_pool_check = true,
              .dbname_paths = std::move(dbname_paths)}}
        , async_ctx_{async_context_create(db_)}
    {
    }
    monad::mpt::Db db_;
    std::unique_ptr<monad::mpt::AsyncContext> async_ctx_;
```

Chen-Yifan: that seems fine too. At the end of the day probably rodb all going to switch to using async read.

kkuehlz: I'll switch to Niall's suggestion to simplify the `AsyncContext`.

kkuehlz: updated

PR #790, opened 2024-07-24
Now `monad_trie_test` `--append` is always to the latest block in db, and bug fix for in memory run

Chen-Yifan:
No functional changes in execution core logic, mainly bug fixes for `monad_trie_test` 
* remove `--block_id` option, now append is always onto the latest block in history
* fix `--in-memory` bug
* fix missed state_nibbles in `--random-read-benchmark`

### DIFF
--- a/libs/db/src/monad/mpt/test/monad_trie_test.cpp
+++ b/libs/db/src/monad/mpt/test/monad_trie_test.cpp
@@ -152,14 +152,17 @@ Node::UniquePtr batch_upsert_commit(
     fprintf(
         stdout,
         "next_key_id: %lu, nkeys upserted: %lu, upsert+commit in RAM: %f "
-        "/s, total_t %.4f s, max creads %u\n=====\n",
+        "/s, total_t %.4f s",
         (key_offset + vec_idx + nkeys) % MAX_NUM_KEYS,
         nkeys,
         (double)nkeys / tm_ram,
-        tm_ram,
-        aux.io->max_reads_in_flight());
+        tm_ram);
+    if (aux.is_on_disk()) {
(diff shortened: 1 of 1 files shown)


libs/db/src/monad/mpt/test/monad_trie_test.cpp
```diff
@@ -506,20 +507,17 @@ int main(int argc, char *argv[])
...
 
             Node::UniquePtr root{};
             if (append) {
                 root.reset(read_node_blocking(
                     io.storage_pool(), aux.get_latest_root_offset()));
             }
-            if (block_id == INVALID_BLOCK_ID) {
-                block_id = aux.db_history_max_version() + 1;
-            }
+            auto block_id = in_memory ? 0 : aux.db_history_max_version() + 1;
```
ned14: Brackets

libs/db/src/monad/mpt/test/monad_trie_test.cpp
```diff
@@ -506,20 +507,17 @@ int main(int argc, char *argv[])
                 }
             };
 
-            UpdateAux<> aux{};
+            UpdateAux<> aux = [&] {
```
ned14: Seems a little overwraught

Chen-Yifan: Ideally I'd like to make `UpdateAux::set_io()` private, this is the effort towards it, but `load_db` lambda in `random_read_benchmark` section still relies on it.

kkuehlz: This being in a lambda seems a tad excessive

PR #792, opened 2024-07-29
When fixing up dirty metadata, keep the destination dirty bit set during the memory copy.

ned14:


libs/db/src/monad/mpt/detail/db_metadata.hpp
```diff
@@ -513,6 +549,29 @@ namespace detail
...
+        std::byte buffer[up_to_dirty_bit_inclusive];
+        memcpy(buffer, src, up_to_dirty_bit_inclusive);
+        auto &dirty_bit = *start_lifetime_as<std::atomic<uint8_t>>(
+            &buffer[up_to_dirty_bit_inclusive - 1]);
+        MONAD_ASSERT(dirty_bit.load(std::memory_order_acquire) == false);
+        dirty_bit.store(true, std::memory_order_release);
+        auto g = dest->hold_dirty();
+        memcpy((void *)dest, buffer, up_to_dirty_bit_inclusive);
+        std::atomic_thread_fence(std::memory_order_release);
+        memcpy(
```
kkuehlz: `ASSERT` at the start of this function that `bytes >= up_to_dirty_bit_inclusive`.

ned14: Fixed

libs/db/src/monad/mpt/detail/db_metadata.hpp
```diff
@@ -1,17 +1,23 @@
 #pragma once
 
 #include <monad/mpt/config.hpp>
+
```
Chen-Yifan, line 4: nit: remove empty line

ned14: The empty line is deliberate to force alphabetical ordering of includes to not move that include below the others. It's on the "there should always be a principle include" principle.

If you feel strongly that include ordering should not matter, I can remove it.

PR #796, opened 2024-07-31
Add test that covers the ASSERT hit in set_io when RODb open with dirty bit set

kkuehlz:


libs/db/src/monad/mpt/test/update_aux_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+    monad::mpt::UpdateAux aux_writer{};
+    std::atomic<bool> io_set = false;
+    std::jthread rw_asyncio([&](std::stop_token token) {
+        monad::io::Ring ring1;
+        monad::io::Ring ring2;
+        monad::io::Buffers testbuf =
+            monad::io::make_buffers_for_segregated_read_write(
+                ring1,
+                ring2,
+                1024,
```
Chen-Yifan: nit: 2 read buffers should be enough

libs/db/src/monad/mpt/test/update_aux_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+            // bit.
+            write_aux.modify_metadata([](monad::mpt::detail::db_metadata *m) {
+                m->is_dirty().store(0, std::memory_order_release);
+            });
+        }
+    };
+
+    monad::io::Ring ring;
+    monad::io::Buffers testrobuf = monad::io::make_buffers_for_read_only(
+        ring, 1024, monad::async::AsyncIO::MONAD_IO_BUFFERS_READ_SIZE);
```
Chen-Yifan: same as above.

libs/db/src/monad/mpt/test/update_aux_test.cpp
kkuehlz: updated

PR #799, opened 2024-08-01
Add read only db stress test executable 

kkuehlz:
Includes some fixes discovered while testing. In addition, 715c99deffa518edbffbc59a2fe5a938575520b6, a buggy commit, fails with this test.

### Stress Test

A DB stress test for uncovering nondeterministic concurrency bugs in RODb

##### Overview

Run multiple instances of RODb performing various read ops in parallel. RWDb runs with compaction on, so we expect RODb to gracefully handle node compaction. The test will run until `ctrl+c` or until a timeout if `--timeout` is passed in. It should be fairly easy to hook up this test to CI and have it run for a minute or two.

The test will consists of:

* Main thread upserts new blocks very quickly. Each block is a single node. Leaf data is `[keccak(version), version]` so it can be independently validated by reader threads. DB grows unbounded.
* 2 threads performing random lookups within the version history, biased towards earlier in the range. Validates version for each lookup.
* 2 threads performing random traversals. Validates expected version at each leaf.
* 1 thread opening and closes the DB, reading the earliest version. 

The test runs cleanly with TSAN enabled.

### DIFF
--- a/libs/db/src/monad/mpt/test/read_only_db_stress_test.cpp
+++ b/libs/db/src/monad/mpt/test/read_only_db_stress_test.cpp
@@ -28,7 +28,8 @@
 using namespace monad::mpt;
 using namespace monad::test;
 
-std::atomic<bool> g_done{false};
+sig_atomic_t volatile g_done = 0;
+
 static_assert(std::atomic<bool>::is_always_lock_free); // async signal safe
 
 static monad::hash256 to_key(uint64_t const key)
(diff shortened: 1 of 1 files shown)


libs/db/src/monad/mpt/test/read_only_db_stress_test.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+#include <signal.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+using namespace monad::mpt;
+using namespace monad::test;
+
+std::atomic<bool> g_done{false};
```
ned14: Why is this extern?

Also, POSIX has a typedef for a type guaranteed async signal safe.

ned14: I found it - it is `sig_atomic_t`

PR #806, opened 2024-08-05
Increase db history length (which is also compaction length)

Chen-Yifan:
History length was 1000 blocks which is too small for devnet and state sync. 
This pr increases history length (default to 1000, same as before) and allows user to initialize a db history length that is <= root offsets ring buffer capacity (65536).
It's the first step towards maintaining a dynamic history length that will be helpful for compaction.

Perf result:

> TPS Results:
> Replay 15000000->15100001: 7144
> https://jenkins.devcore4.com/job/replay_ethereum_manual/33/
> TPS Results:
> Replay 15000000->15100001: 6294
> https://jenkins.devcore4.com/job/replay_ethereum_manual/34/

both runs are on pr 806,
run 1 history_len = 1000, same as before, performance also remains the same
run 2 history_len = 20000. slower, could be of compaction variance. [Update: there are other benchmark results showing that more spread out the data and less free disk space impedes the performance.]

Without TRIM on peach9
20000: 8283 tps
1000: 8067 tps

### DIFF
--- a/libs/db/src/monad/mpt/cli_tool_impl.cpp
+++ b/libs/db/src/monad/mpt/cli_tool_impl.cpp
@@ -825,6 +825,7 @@ public:
                         &old_metadata->root_offsets.arr_,
                         sizeof(metadata->root_offsets.arr_));
                     metadata->slow_fast_ratio = old_metadata->slow_fast_ratio;
+                    metadata->history_length = old_metadata->history_length;
                 });
                 fast_list_base_insertion_count =
                     old_metadata->fast_list_begin()->insertion_count();
--- a/libs/db/src/monad/mpt/db.cpp
+++ b/libs/db/src/monad/mpt/db.cpp
@@ -587,6 +587,10 @@ struct Db::RWOnDisk final : public Db::Impl
                        : Node::UniquePtr{};
         }())
     {
+        if (aux_.db_history_max_version() == INVALID_BLOCK_ID) {
+            // set history length on empty db
+            aux_.update_history_length_metadata(options.history_length);
+        }
     }
 
     ~RWOnDisk()
@@ -894,6 +898,11 @@ bool Db::is_read_only() const
 {
     MONAD_ASSERT(impl_);
     return is_on_disk() && impl_->aux().io->is_read_only();
+}
+
+uint64_t Db::get_history_length() const
+{
+    return is_on_disk() ? impl_->aux().version_history_length() : 1;
 }
 
 AsyncContext::AsyncContext(Db &db, size_t lru_size)
--- a/libs/db/src/monad/mpt/detail/db_metadata.hpp
+++ b/libs/db/src/monad/mpt/detail/db_metadata.hpp
@@ -45,7 +45,7 @@ namespace detail
     // For the memory map of the first conventional chunk
     struct db_metadata
     {
-        static constexpr char const *MAGIC = "MND5";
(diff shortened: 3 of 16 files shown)


libs/db/src/monad/mpt/trie.hpp
```diff
@@ -331,18 +331,12 @@ class UpdateAuxImpl
...
-    // root_offsets_ring_t capacity, rather can be less than
-    static constexpr auto VERSION_HISTORY_LEN = 1000;
-    static_assert(
-        VERSION_HISTORY_LEN <=
-        detail::db_metadata::root_offsets_ring_t::capacity());
-
-    UpdateAuxImpl(MONAD_ASYNC_NAMESPACE::AsyncIO *io_ = nullptr)
+    UpdateAuxImpl(
+        MONAD_ASYNC_NAMESPACE::AsyncIO *io_ = nullptr,
+        uint64_t const history_len = 20000)
```
kkuehlz: Make this a constant somewhere - it’s used both here and in replay as default.

libs/db/src/monad/mpt/detail/db_metadata.hpp
```diff
@@ -546,6 +547,13 @@ namespace detail
             }
         }
 
+        void set_history_length_(uint64_t history_len) noexcept
+        {
+            auto g = hold_dirty();
+            MONAD_ASSERT(history_len <= root_offsets_ring_t::capacity());
+            history_length = history_len;
```
kkuehlz: Does RODb ever query this?

libs/db/src/monad/mpt/trie.hpp
Chen-Yifan: Might be better I remove the default option here. 
Update: added a constant and default to `MAX_HISTORY_LEN`

libs/db/src/monad/mpt/detail/db_metadata.hpp
Chen-Yifan: yes, I need to make it atomic, thanks for the catch

libs/db/src/monad/mpt/cli_tool_impl.cpp
```diff
@@ -825,6 +825,7 @@ struct impl_t
                         &old_metadata->root_offsets.arr_,
                         sizeof(metadata->root_offsets.arr_));
                     metadata->slow_fast_ratio = old_metadata->slow_fast_ratio;
+                    metadata->history_length = old_metadata->history_length;
```
kkuehlz, line 828: i know it doesn't matter, but why not just use `set_history_length` here?

Chen-Yifan: it doesn't really matter if we use atomic or not, also `set_history_length_()` is private in `db_metadata`

libs/db/src/monad/mpt/db.cpp
```diff
@@ -586,6 +586,10 @@ struct Db::RWOnDisk final : public Db::Impl
                         worker_->pool, aux_.get_latest_root_offset())}
                   : Node::UniquePtr{})
     {
+        if (aux_.db_history_max_version() == INVALID_BLOCK_ID) {
```
kkuehlz, line 590: unit test this

Chen-Yifan: good idea, added it

PR #808, opened 2024-08-06
remove OnDiskDbConfig::append

illiashkirko:
removing `OnDiskDbConfig::append` and `OnDiskDbConfig::file_size_db`
reasoning: in tests we would like to initialize file for `storage_pool` manually before creating a db and always observe behavior as if `append = true`

### DIFF
--- a/libs/db/src/monad/mpt/db.cpp
+++ b/libs/db/src/monad/mpt/db.cpp
@@ -539,7 +539,7 @@ struct Db::RWOnDisk final : public Db::Impl
     Node::UniquePtr root_; // owned by worker thread
     Node::UniquePtr reader_root_; // lifetime for reads on a different block
 
-    RWOnDisk(OnDiskDbConfig const &options, StateMachine &machine)
+    RWOnDisk(StateMachine &machine, OnDiskDbConfig const &options)
         : worker_thread_([&] {
             {
                 std::unique_lock const g(lock_);
@@ -739,7 +739,7 @@ Db::Db(StateMachine &machine)
 }
 
 Db::Db(StateMachine &machine, OnDiskDbConfig const &config)
-    : impl_{std::make_unique<RWOnDisk>(config, machine)}
+    : impl_{std::make_unique<RWOnDisk>(machine, config)}
 {
     MONAD_DEBUG_ASSERT(impl_->aux().is_on_disk());
 }


libs/async/src/monad/async/storage_pool.cpp
```diff
@@ -382,13 +382,6 @@ storage_pool::device storage_pool::make_device_(
             buffer + bytesread - sizeof(device::metadata_t));
         if (memcmp(metadata_footer->magic, "MND0", 4) != 0 ||
             op == mode::truncate) {
-            // Uninitialised
```
Chen-Yifan: why remove this?

libs/async/src/monad/async/util.hpp
```diff
@@ -42,10 +46,13 @@ inline constexpr chunk_offset_t round_down_align(chunk_offset_t x) noexcept
...
 //! Returns a temporary directory in which `O_DIRECT` files definitely work
-extern const std::filesystem::path &working_temporary_directory();
+extern std::filesystem::path const &working_temporary_directory();
 
 //! Creates already deleted file so no need to clean it up
 //! after
 extern int make_temporary_inode() noexcept;
 
+//! For tests only
+extern std::filesystem::path initialize_db_file_for_test(std::string testname);
```
Chen-Yifan: the util function should take a created file, set up the storage pool.

libs/db/src/monad/mpt/test/db_test.cpp
```diff
@@ -467,14 +458,12 @@ TEST_F(OnDiskDbFixture, ro_historical)
 
 TEST(ReadOnlyDbTest, open_empty_rodb)
 {
-    std::filesystem::path const dbname{
```
Chen-Yifan, line 546: we should keep this, and call the util to set up storage pool

libs/async/src/monad/async/storage_pool.cpp
illiashkirko: will bring it back

libs/db/src/monad/mpt/test/db_test.cpp
```diff
@@ -558,10 +547,6 @@ TEST(ReadOnlyDbTest, read_only_db_concurrent)
 
     // construct RWDb
```
Chen-Yifan: comment outdated, can remove

PR #814, opened 2024-08-08
Log million gas per second as additional metric

tongzhi1998:


cmd/replay_ethereum.cpp
```diff
@@ -225,13 +225,16 @@ int main(int const argc, char const *argv[])
             "Finish running, finish(stopped) block number = {}, "
             "number of blocks run = {}, time_elapsed = {}, num transactions = "
             "{}, "
-            "tps = {}",
+            "tps = {}, mgps = {} M",
```
andreaslyn: This is a minor detail. It looks like we are logging *million gas per second = {} million*, perhaps it is more accurate to log *gas per second = {} million*, i.e. `gps = {} M`.

libs/execution/src/monad/execution/replay_block_db.hpp
```diff
@@ -86,18 +88,21 @@ class ReplayFromBlockDb
                     .count();
             uint64_t const tps =
                 (batch_num_txs) * 1'000'000 / static_cast<uint64_t>(elapsed);
+            uint64_t mgps = batch_gas / static_cast<uint64_t>(elapsed);
 
             LOG_INFO(
                 "Run {:4d} blocks to {:8d}, number of transactions {:6d}, "
-                "tps = {:5d}, rss = {:8d} MB",
+                "tps = {:5d}, mgps = {:4d} M, rss = {:6d} MB",
```
andreaslyn: Same minor detail here. I think that `gps = {:4d} M` is more accurate.

cmd/replay_ethereum.cpp
tongzhi1998: Thanks for the suggestion. Modified.

PR #816, opened 2024-08-09
Use mmap to load binary snapshots

tongzhi1998:


libs/execution/src/monad/db/util.cpp
```diff
@@ -56,179 +60,165 @@ namespace
...
-
-            fwrite(std::move(updates));
+            struct stat file_stat;
+            MONAD_ASSERT(fstat(fd, &file_stat) != -1);
+            size_t file_size = static_cast<size_t>(file_stat.st_size);
+            void *mmaped_content = mmap(
+                nullptr,
+                file_size,
+                PROT_READ,
+                MAP_PRIVATE | MAP_NORESERVE,
```
ned14: I'm not sure a copy on write read only map makes much sense.

`MAP_SHARED` makes most sense with `PROT_READ`.

tongzhi1998: Thanks. Sorry just saw this comment

libs/execution/src/monad/db/util.cpp
```diff
@@ -57,179 +61,165 @@ namespace
...
-
-            fwrite(std::move(updates));
+            struct stat file_stat;
+            MONAD_ASSERT(fstat(fd, &file_stat) != -1);
+            size_t file_size = static_cast<size_t>(file_stat.st_size);
+            void *mmaped_content = mmap(
+                nullptr,
+                file_size,
+                PROT_READ,
+                MAP_SHARED | MAP_NORESERVE,
```
ned14: You shouldn't use `MAP_NORESERVE` unless it is combined with `PROT_NONE`.

PR #835, opened 2024-08-15
Tune compaction: now we've run through ethereum history 

Chen-Yifan:
This is not the final version of what compaction would look like, it is simply an 
improvement on what we currently have.

**Result:**
- Before: we would run out of disk on replay etheruem, and experience unstable 
performance when slow list compaction starts.
- After: able to run from 12M to 19.5M (all we have in block_db) with 3000 history 
- versions to start with on a 1.8TB drive with steady performance. Disk usage ends 
at ~55%. During the test I forced it to start compacting slow ring earlier when 
disk usage hits 45%. 

**Compaction algo update:**
On compacting nodes on slow ring, we now copy nodes written after the 
compaction offset (`compact_offset_slow`) to fast ring instead of slow ring, 
which are the ancestors of the garbage collected nodes, so to keep slow ring 
data growth under control.

**Compaction offset update rule:**
 The fast ring is compacted at a steady pace based on the average disk growth
 observed over recent blocks. We define two disk usage thresholds:
`usage_limit_start_compact_slow` and `usage_limit`. When disk usage reaches
`usage_limit_start_compact_slow`, slow ring compaction begins, guided by the
slow ring garbage collection ratio from the last block. If disk usage exceeds 
`usage_limit`, the system will start shortening the history until disk usage is 
brought back within the threshold.

nit:
also renamed some stats variables, and improved stats printing format.

**Future work:** 
- make db version history K based on disk usage, including initial K before first compaction starts, and increasing and decreasing of history length.
- larger disk should initialize with large root ring buffer capacity, thus can hold longer history
- compact more when update load is low, compact less when load is high. (More for the slow ring compaction)

### DIFF
--- a/libs/db/src/monad/mpt/update_aux.cpp
+++ b/libs/db/src/monad/mpt/update_aux.cpp
@@ -17,7 +17,9 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
 #include <cstring>
+#include <random>
 #include <span>
 #include <stdexcept>
 #include <thread>
@@ -184,6 +186,7 @@ void UpdateAuxImpl::update_slow_fast_ratio_metadata() noexcept
 void UpdateAuxImpl::update_history_length_metadata(
     uint64_t const history_len) noexcept
 {
+    std::cout << "Update db history length to " << history_len << std::endl;
     MONAD_ASSERT(is_on_disk());
     auto do_ = [&](detail::db_metadata *m) {
         m->set_history_length_(history_len);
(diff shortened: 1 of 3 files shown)


libs/db/src/monad/mpt/detail/db_metadata.hpp
```diff
@@ -45,7 +45,7 @@ namespace detail
     // For the memory map of the first conventional chunk
     struct db_metadata
     {
-        static constexpr char const *MAGIC = "MND6";
+        static constexpr char const *MAGIC = "MND5";
```
ned14: Going backwards?

libs/db/src/monad/mpt/update_aux.cpp
```diff
@@ -184,6 +185,7 @@ void UpdateAuxImpl::update_slow_fast_ratio_metadata() noexcept
 void UpdateAuxImpl::update_history_length_metadata(
     uint64_t const history_len) noexcept
 {
+    std::cout << "Update db history length to " << history_len << std::endl;
```
ned14, line 189: Intentional to retain printing?

libs/db/src/monad/mpt/update_aux.cpp
```diff
@@ -632,52 +674,70 @@ void UpdateAuxImpl::advance_compact_offsets()
...
+    if (used_chunks_ratio() > usage_limit) {
+        // Shorten the history length
+        while (used_chunks_ratio() > usage_limit) {
+            version_to_erase = db_history_min_valid_version();
+            MONAD_ASSERT(version_to_erase != INVALID_BLOCK_ID);
+            if (version_to_erase == max_version) {
+                // the only version left, no more to erase
+                break;
+            }
+            std::tie(
```
ned14, line 734: Structured binding instead?

libs/db/src/monad/mpt/update_aux.cpp
Chen-Yifan: yes, I think it's useful to know current history length, it also infers disk usage is getting high.

libs/db/src/monad/mpt/update_aux.cpp
Chen-Yifan: can't do that for existing variables can we?

libs/db/src/monad/mpt/detail/db_metadata.hpp
Chen-Yifan: oops mb, changed this for testing where I started at version 5, forgot to change back, updated

libs/db/src/monad/mpt/update_aux.cpp
ned14: I think I was thinking of array members. Yes you're right about existing variables.

PR #846, opened 2024-08-20
[core] Enhance the spinlock type, used extensively in the fiber implementation

kjcamann:
This PR contains debugging enhancements to the existing `spinlock_t` type, which is renamed to `monad_spinlock_t`. This is an isolated change, because the spinlock was not currently used by any other code. To simplify the fiber implementation, there is no fancy lock-free programming in the subsequent PRs; the all major objects (the fibers, the synchronization primitives, and the scheduler) are simple implementations covered by a single spinlock. The locks are held for very short periods, which can be verified by enabling the locking statistics option. The other feature this adds is source location tracking for the location where the lock was last taken, so it adds a formal source location object for C, similar to `std::source_location`, and a C++ conversion function between the two.

### DIFF
--- a/libs/core/src/monad/core/spinlock.h
+++ b/libs/core/src/monad/core/spinlock.h
@@ -24,12 +24,18 @@ static_assert(ATOMIC_INT_LOCK_FREE == 2);
 
 typedef struct monad_spinlock monad_spinlock_t;
 
-#if MONAD_SPINLOCK_TRACK_STATS_ATOMIC
-typedef atomic_ulong monad_spinstat_t;
-    #define MONAD_SPINSTAT_INC(X) atomic_fetch_add(&(X), 1)
+#if MONAD_SPINLOCK_TRACK_STATS
+    #if MONAD_SPINLOCK_TRACK_STATS_ATOMIC
+        typedef atomic_ulong monad_spinstat_t;
+        #define MONAD_SPINSTAT_INC(X)                                          \
+            atomic_fetch_add_explicit(&(X), 1, memory_order_relaxed)
+    #else
+        typedef unsigned long monad_spinstat_t;
+        #define MONAD_SPINSTAT_INC(X) ++(X)
+    #endif
 #else
-typedef unsigned long monad_spinstat_t;
-    #define MONAD_SPINSTAT_INC(X) ++(X)
+    typedef unsigned long monad_spinstat_t;
+    #define MONAD_SPINSTAT_INC(X)
 #endif
 
 // TODO(ken): remove this workaround when we have clang-19
@@ -65,7 +71,7 @@ struct monad_spinlock
 #endif
 };
 
-static inline bool monad_spinlock_is_owned(monad_spinlock_t *const lock)
+static inline bool monad_spinlock_is_self_owned(monad_spinlock_t *const lock)
 {
     return atomic_load_explicit(&lock->owner_tid, memory_order_acquire) ==
            get_tl_tid();
@@ -94,31 +100,29 @@ static inline void monad_spinlock_init(monad_spinlock_t *const lock)
 static inline bool monad_spinlock_try_lock(monad_spinlock_t *const lock)
 {
     int expected = 0;
-    int const desired = get_tl_tid();
-    bool const is_locked = atomic_compare_exchange_weak_explicit(
+    MONAD_SPINSTAT_INC(lock->stats.total_try_locks);
+    bool const is_locked = atomic_compare_exchange_strong_explicit(
         &lock->owner_tid,
         &expected,
-        desired,
-        memory_order_acquire,
+        get_tl_tid(),
+        memory_order_acq_rel,
         memory_order_relaxed);
-#if MONAD_SPINLOCK_TRACK_STATS
-    MONAD_SPINSTAT_INC(lock->stats.total_try_locks);
     if (MONAD_UNLIKELY(!is_locked)) {
         MONAD_SPINSTAT_INC(lock->stats.total_try_lock_fail);
     }
-#endif
     return is_locked;
 }
 
 static inline void monad_spinlock_lock(monad_spinlock_t *const lock)
 {
+    MONAD_SPINSTAT_INC(lock->stats.total_locks);
     int const desired = get_tl_tid();
     int expected;
     bool owned;
     [[maybe_unused]] unsigned long tries = 0;
     [[maybe_unused]] unsigned histo_bucket;
 
-    MONAD_DEBUG_ASSERT(!monad_spinlock_is_owned(lock));
+    MONAD_DEBUG_ASSERT(!monad_spinlock_is_self_owned(lock));
 
 TryAgain:
     expected = 0;
@@ -126,7 +130,7 @@ TryAgain:
         &lock->owner_tid,
         &expected,
         desired,
-        memory_order_acquire,
+        memory_order_acq_rel,
         memory_order_relaxed);
     if (MONAD_UNLIKELY(!owned)) {
         monad_spinloop_hint();
@@ -137,7 +141,6 @@ TryAgain:
     }
 
 #if MONAD_SPINLOCK_TRACK_STATS
-    MONAD_SPINSTAT_INC(lock->stats.total_locks);
     if (MONAD_LIKELY(tries > 1)) {
         MONAD_SPINSTAT_INC(lock->stats.total_lock_init_fail);
     }
@@ -151,7 +154,7 @@ TryAgain:
 
 static inline void monad_spinlock_unlock(monad_spinlock_t *const lock)
 {
-    MONAD_DEBUG_ASSERT(monad_spinlock_is_owned(lock));
+    MONAD_DEBUG_ASSERT(monad_spinlock_is_self_owned(lock));
     atomic_store_explicit(&lock->owner_tid, 0, memory_order_release);
 }
 
--- a/libs/core/src/monad/core/spinlock.h
+++ b/libs/core/src/monad/core/spinlock.h
@@ -16,6 +16,10 @@
 #include <stddef.h>
 #include <string.h>
 
+#ifdef __cplusplus
+    #include <new>
+#endif
+
 static_assert(ATOMIC_INT_LOCK_FREE == 2);
 
 typedef struct monad_spinlock monad_spinlock_t;
--- a/libs/core/src/monad/core/spinlock.h
+++ b/libs/core/src/monad/core/spinlock.h
@@ -1,63 +1,185 @@
 #pragma once
 
-#include <monad/core/cpu_relax.h>
+#include <monad/core/assert.h>
 #include <monad/core/likely.h>
+#include <monad/core/spinloop.h>
 #include <monad/core/tl_tid.h>
+
+#if MONAD_SPINLOCK_TRACK_OWNER_INFO
+    #include <monad/core/srcloc.h>
+#endif
 
 #include <assert.h>
 #include <stdatomic.h>
+#include <stdbit.h>
 #include <stdbool.h>
+#include <stddef.h>
+#include <string.h>
 
 static_assert(ATOMIC_INT_LOCK_FREE == 2);
 
-typedef atomic_int spinlock_t;
+typedef struct monad_spinlock monad_spinlock_t;
 
-static inline void spinlock_init(spinlock_t *const lock)
+#if MONAD_SPINLOCK_TRACK_STATS_ATOMIC
+typedef atomic_ulong monad_spinstat_t;
+    #define MONAD_SPINSTAT_INC(X) atomic_fetch_add(&(X), 1)
+#else
+typedef unsigned long monad_spinstat_t;
+    #define MONAD_SPINSTAT_INC(X) ++(X)
+#endif
+
+// TODO(ken): remove this workaround when we have clang-19
+#if defined(__clang__) && __clang_major__ < 19
+    #define MONAD_SPINLOCK_HIST_BUCKETS 15
+#else
+constexpr size_t MONAD_SPINLOCK_HIST_BUCKETS = 15;
+#endif
+
+// clang-format off
+/// Lock statistics; these may not be 100% accurate, since we may not be
+/// atomically incrementing them (some data could be lost)
+struct monad_spinlock_stats
 {
-    atomic_init(lock, 0);
+    monad_spinstat_t total_try_locks;      ///< # times try_lock is called
+    monad_spinstat_t total_try_lock_fail;  ///< # times try_lock failed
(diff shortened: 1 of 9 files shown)


libs/core/src/monad/core/spinlock.h
```diff
@@ -1,34 +1,100 @@
...
+typedef atomic_ulong monad_spinstat_t;
+#define MONAD_SPINSTAT_INC(X) atomic_fetch_add(&(X), 1)
+#else
+typedef unsigned long monad_spinstat_t;
+#define MONAD_SPINSTAT_INC(X) ++(X)
+#endif
+
+// TODO(ken): remove this workaround when we have clang-19
+#if defined(__clang__) && __clang_major__ < 19
+#define MONAD_SPINLOCK_HIST_BUCKETS 15
```
ned14, line 41: You can do `static const` in C and it'll "do the right thing".

I'm not personally sure `constexpr` in C has much value outside of fairly niche use cases. I wouldn't be upgrading a toolchain for it anyway.

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,34 +1,100 @@
...
+#else
+constexpr size_t MONAD_SPINLOCK_HIST_BUCKETS = 15;
+#endif
+
+/// Lock statistics; these may not be 100% accurate, since we may not be
+/// atomically incrementing them (some data could be lost)
```
ned14, line 45: Problem is this will cause TSAN to fail.

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,34 +1,100 @@
...
 }
 
-static inline bool spinlock_try_lock(spinlock_t *const lock)
+static inline bool monad_spinlock_try_lock(monad_spinlock_t *const lock)
 {
     int expected = 0;
     int const desired = get_tl_tid();
-    return atomic_compare_exchange_weak_explicit(
-        lock, &expected, desired, memory_order_acquire, memory_order_relaxed);
+    const bool is_locked = atomic_compare_exchange_weak_explicit(
```
ned14: A speculative load beforehand avoids dirtying the cache line.

ned14 (commented):
Firstly, obligatory https://www.reddit.com/r/programming/comments/ejyr3r/mutexes_are_faster_than_spinlocks/

Secondly, can't say I like source location for this at all. It's nearly useless. We already have a stack backtrace capture and print functionality in Core. I'd suggest that instead.

Finally, me personally, I'd never use spinlocks around any code which isn't wait free. I'd even go so far that if a spin count takes too long in debug, we abort the process and force the dev to use a mutex. That's probably considered excessive, but I've seen many a codebase ruined by too many spinlocks because "they're faster", and they're only faster if the code they protect is extremely short and well bounded. Mutexes already have a short spin count in them, and the ones in libgcc are "magical" in that the compiler does clever tricks with them.

libs/core/src/monad/core/spinlock.h
kjcamann: But later I use it to declare the size of an array

```c
monad_spinstat_t tries_histogram[MONAD_SPINLOCK_HIST_BUCKETS + 1];
```

I am pretty sure that requires it to be an integer constant "for real"

libs/core/src/monad/core/spinlock.h
kjcamann: I am not too familiar with TSAN, what exactly would happen here?

libs/core/src/monad/core/spinlock.h
kjcamann: I didn't change this code at all (except to keep stats) but I'm curious what you mean here: what would we be speculatively loading?

libs/core/src/monad/core/spinlock.h
kjcamann: Discussed this with Niall offline. He thought TSAN would flag this (deliberately) un-atomic read-modify-write operation. I was able to get the clang (compiler-rt-based) TSAN working, but not the gcc TSAN. The clang one did not flag this, and I'm not sure what gcc would do here. Irritatingly it was failing during the compilation of unrelated Boost.Fiber stuff, which will eventually by removed after this long sequence of PRs. So when that happens I'll check it again.

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,63 +1,189 @@
...
+#if MONAD_SPINLOCK_TRACK_STATS
+    struct monad_spinlock_stats stats; ///< Lock contention stats
+#endif
+#if MONAD_SPINLOCK_TRACK_OWNER_INFO
+    monad_source_location_t srcloc; ///< Location in code where lock taken
+#endif
+};
 
-static inline void spinlock_init(spinlock_t *const lock)
+static inline bool monad_spinlock_is_owned(monad_spinlock_t *const lock)
```
ned14: is owned by this thread?

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,63 +1,189 @@
...
 {
     int expected = 0;
     int const desired = get_tl_tid();
-    return atomic_compare_exchange_weak_explicit(
-        lock, &expected, desired, memory_order_acquire, memory_order_relaxed);
+    bool const is_locked = atomic_compare_exchange_weak_explicit(
+        &lock->owner_tid,
+        &expected,
+        desired,
+        memory_order_acquire,
```
ned14: Technically acq_rel

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,63 +1,189 @@
...
+#endif
+}
+
+static inline bool monad_spinlock_try_lock(monad_spinlock_t *const lock)
 {
     int expected = 0;
     int const desired = get_tl_tid();
-    return atomic_compare_exchange_weak_explicit(
-        lock, &expected, desired, memory_order_acquire, memory_order_relaxed);
+    bool const is_locked = atomic_compare_exchange_weak_explicit(
```
ned14: strong is better here

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,63 +1,189 @@
...
-            break;
-        }
+    int expected;
+    bool owned;
+    [[maybe_unused]] unsigned long tries = 0;
+    [[maybe_unused]] unsigned histo_bucket;
+
+    MONAD_DEBUG_ASSERT(!monad_spinlock_is_owned(lock));
+
+TryAgain:
```
ned14, line 122: I'd repeat my previous comment that a speculative load before the CMPXCHG reduces cache coherency invalidation traffic.

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,63 +1,189 @@
...
+    expected = 0;
+    owned = atomic_compare_exchange_weak_explicit(
+        &lock->owner_tid,
+        &expected,
+        desired,
+        memory_order_acquire,
```
ned14: acq_rel

libs/core/src/monad/core/spinlock.h
kjcamann: Yeah I have often wondered if I should change this to make that clear. Maybe `monad_spinlock_is_self_owned?`

libs/core/src/monad/core/spinlock.h
kjcamann: Can you point to something explaining why? IIRC when I wrote this, I had recently read some advice about when to use strong vs. weak that seemed reasonable. The advice was that if intermittent failure is something you expect to happen anyway in the normal course of what you're doing (i.e., if you're in a loop, fighting for the lock), then use "weak". But if you are doing something where you don't want it to ever be spurious, then use strong. An example would be if you aren't just going to loop anyway (like you're going to try, and then exit if it fails because you're "too late") then use "strong." For example I used "strong" that way in the "CMA allocator" PR, for the thread-safe "bump allocator", where you are allowed to deallocate only if you are freeing the most recent allocation, otherwise it succeeds but doesn't give back the memory.
So I'm following that advice here.

Since you're suggesting something different here, can you explain why?

libs/core/src/monad/core/spinlock.h
kjcamann: Oh wow, I did not write this myself and was carrying over the old code, and did not notice this was not acq_rel! Good catch.

libs/core/src/monad/core/spinlock.h
jhunsaker: why? other impls have acquire not acquire_release. that seems to make sense because this thread needs to see the changes made by the thread which unlocked it but not necessarily the other way around

libs/core/src/monad/core/spinlock.h
kjcamann: Initially I thought you were right about the `acq_rel`, and the original was a mistake. I believed (incorrectly) that `acq_rel` was required (or implied anyway) by read-modify-write operations. But apparently I was wrong about this, and looking at the glibc `pthread_spinlock_t` I see that it's the same as this is (the lock side is compare_exchange acquire-only).

Now I am confused, and I don't want to spend too much time doing the perfect thing here because the memory model stuff is a really complex topic and none of it matters on x64 anyway. But for the sake of closing out this discussion, is there a simple reason why you think this should be `acq_rel`?

libs/core/src/monad/core/spinlock.h
ned14: If your lock implementation is not inlined, and you guarantee you don't touch locked state until you have the lock, you don't need release there.

BUT if code using the lock DOES touch locked state without having the lock (and who knows if that's the case when your lock implementation inlines and the compiler optimises), then the lack of release on compare-exchange permits the CPU to reorder loads and stores preceding the lock acquisition to after the compare-exchange success. If code touches locked state without having the lock, it may see "leakage" of state changes. I would therefore tend to err on the side of caution for inlined lock implementations.

`acquire` vs `acq_rel` doesn't matter on Intel, but on AArch64 it matters: https://godbolt.org/z/94PaGMn5q.

Having debugged weak memory model bugs in the past, I tend to err severely on the side of caution in this stuff. Especially on Intel, which hides mistakes very well and then you get weeks of dev time on hard to reproduce bugs after the ARM port.

libs/core/src/monad/core/spinlock.h
ned14: This is the "try lock" implementation yes?

If it's within a tight loop, you use weak.

If it might not be within a tight loop, you use strong which can on some CPU architectures use an implementation less prone to spurious failure. For example, on AArch64 you might use the linked load-store functions for the weak implementation, and the dedicated CAS opcode for the strong implementation.

Agreed on x64 this has no difference.

libs/core/src/monad/core/spinlock.h
ned14: Hopefully explained above. I guess I'm biased from past bitter experience tracking down weak memory model bugs. I'd tend to write as strong as possible initially, especially on Intel which is strong anyway, and only consider weakening when porting elsewhere.

libs/core/src/monad/core/spinlock.h
kjcamann: I do not understand this explanation because

> BUT if code using the lock DOES touch locked state without having the lock

Unless I've misunderstood this (and I probably have), this sounds like a programming error on behalf of the programmer rather than something that we should try to fix.

But I've come to the conclusion that making this perfect and optimizing it to death is a much harder task than I thought. So my idea is to table any discussion of making this picture perfect and can revisit later. I need to start landing PRs which are downstream of this one soon.

And since it doesn't matter anyway on Intel, I just made this `acq_rel` for now.

libs/core/src/monad/core/spinlock.h
```diff
@@ -1,63 +1,192 @@
...
+        typedef unsigned long monad_spinstat_t;
+        #define MONAD_SPINSTAT_INC(X) ++(X)
+    #endif
+#else
+    typedef unsigned long monad_spinstat_t;
+    #define MONAD_SPINSTAT_INC(X)
+#endif
+
+// TODO(ken): remove this workaround when we have clang-19
+#if defined(__clang__) && __clang_major__ < 19
```
ned14: I'd either just use the macro, or if you really really really want the non-macro approach then the old enum trick has worked since C89.

libs/core/src/monad/core/srcloc.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+typedef struct monad_source_location monad_source_location_t;
+
+/// A pure C structure compatible with C++20's std::source_location
```
ned14: Do we static assert that this is true somewhere?

libs/core/test/spinlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    std::vector<std::thread> threads;
+    for (unsigned i = 0; i < NUM_THREADS; ++i) {
+        threads.emplace_back(
+            try_lock_thread_function, std::ref(s), i, std::ref(latch));
+    }
+    pin_threads_to_cores(threads);
+
+    latch.arrive_and_wait();
+    std::atomic_thread_fence(std::memory_order::seq_cst);
```
ned14: Are these to ensure time measurement doesn't get reordered?

They should be redundant as an extern function call will trigger escape analysis i.e. no reordering across the extern function call.

I suppose that the CPU would also reorder without these, but if you're relying on that, I think the test needs redesigning.

ned14 (dismissed):
Minor nitpicks only

libs/core/src/monad/core/spinlock.h
kjcamann: Done

libs/core/src/monad/core/srcloc.h
kjcamann: Ah, I think my comment is misleading there. I meant to say "similar to", not "compatible with" as in for example `std::is_layout_compatible`. It's suppose to make you think of the idea of `std::source_location`, but not going as far as to `std::bit_cast` between them. That might work, but it would a terrible idea, as I think nothing about the layout is mandated per the standard. I change update the comment.

libs/core/test/spinlock.cpp
kjcamann: If by "these" you mean the barriers, they were paranoia, to make sure that `std::chrono::system_clock::now()` isn't moved around. This was kind of in the spirit of P0342 (which originally was about clocks, and became `pessimize_hint`). I removed them though, since that's stupid for two reasons:

1. The top one isn't needed since the latch barrier is immediately above it
2. The bottom one was in the wrong place anyway, and it isn't needed because the standard library implementers can be trusted to suppress whatever optimization might be needed so it "does the right thing." Since that's the case IMO, it's just a weird distraction.

kjcamann:
Closing a defunct PR

PR #887, opened 2024-09-03
Make precompiles backend flexible

kkuehlz:
Make precompile backend flexible

Before this PR, monad's precompile backend was dictated entirely by the
silkpre API. Normally, this wouldn't be an issue. However, silkpre is no
longer maintained, and Dencun support requires us to add a new
precompiled contract type.

Remove the `kSilkpreContracts` (defined in their headers) with a lut of
our own. For now, all of our dispatch functions wrap silkpre, but we can
modify existing functions and add new ones without having to modify and
third party code.

One quirk of this design, in the interim, is that execute functions will
need to allocate output buffers and return those to users. This can be
de-duplicated once we move off of silkpre. For the time being, however,
we need to do this, as the silkpre `run` functions allocate output
buffers.

### Benchmarks

This Pr:
* Replay 3000000->3100000 in-memory: tps=16373, gps=559 M
* Replay 12000000->12100000 in-memory: tps=9790, gps=607 M
* Replay 15000000->15100000 on-disk:

Baseline:
* Replay 3000000->3100000 in-memory: tps=16373, gps=559 M
* Replay 12000000->12100000 in-memory: tps=9938, gps=616 M
* Replay 15000000->15100000 on-disk:

### DIFF
--- a/libs/execution/src/monad/execution/precompiles.cpp
+++ b/libs/execution/src/monad/execution/precompiles.cpp
@@ -68,18 +68,20 @@ struct PrecompiledContract
     precompiled_execute_fn *execute_func;
 };
 
-std::array<PrecompiledContract, num_precompiles(EVMC_SHANGHAI) + 1> dispatch{{
-    {}, // precompiles start at address 0x1
-    {ecrecover_gas_cost, ecrecover_execute},
-    {sha256_gas_cost, sha256_execute},
-    {ripemd160_gas_cost, ripemd160_execute},
-    {identity_gas_cost, identity_execute},
-    {expmod_gas_cost, expmod_execute},
-    {ecadd_gas_cost, ecadd_execute},
-    {ecmul_gas_cost, ecmul_execute},
-    {snarkv_gas_cost, snarkv_execute},
-    {blake2bf_gas_cost, blake2bf_execute},
-}};
+inline constexpr std::array<
+    PrecompiledContract, num_precompiles(EVMC_SHANGHAI) + 1>
+    dispatch{{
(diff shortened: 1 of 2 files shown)


libs/execution/src/monad/execution/precompiles.hpp
```diff
@@ -18,4 +19,31 @@ bool is_precompile(Address const &) noexcept;
...
 std::optional<evmc::Result> check_call_precompile(evmc_message const &);
 
+struct ExecutionResult
+{
+    evmc_status_code status_code;
+    uint8_t *obuf;
+    size_t output_size;
+};
+
+uint64_t ecrecover_requirement(byte_string_view, evmc_revision);
```
kjcamann: A humble suggestion:

Looking through this, I did not immediately understand what `_requirement` meant. I think this would be a bit clearer as:

```.cpp
/*
 * Gas cost functions
 */

using precompiled_gas_cost_fn = uint64_t(byte_string_view, emvc_revision);

uint64_t sha256_gas_cost(byte_string_view, evmc_revision);
uint64_t ripemd160_gas_cost(byte_string_view, emvc_revision);
...
```

Then instead of `decltype`:

```.hpp
struct PrecompiledContract
{
    precompiled_gas_cost_fn *gas_cost;
    precompiled_exec_fn *execute;
};
```

These are not perfect names either, and my names tend to be too long / verbose. But it would be nice for `gas` and/or `cost` to be in there somewhere, since that's the kind of requirement that they are.

kkuehlz: updated

PR #895, opened 2024-09-06
[core][cmake] some minor improvements to the build system

kjcamann:
Various improvements to the build system that were part of #892, support for C binaries removed

### DIFF
--- a/libs/core/CMakeLists.txt
+++ b/libs/core/CMakeLists.txt
@@ -69,26 +69,14 @@ endif()
 # libs
 # ##############################################################################
 
-# monad_core_c is a library containing source code that is compiled in C23 mode
-# and contains no C++ source files; it may be linked by downstream libraries
-# which only need C and do not want the C++ runtime as a dependency
-add_library(
-  monad_core_c OBJECT
-  "src/monad/core/assert.c"
-  "src/monad/core/assert.h")
-monad_compile_options(monad_core_c)
-if(MONAD_CORE_FORCE_DEBUG_ASSERT)
-  target_compile_definitions(monad_core_c
-                             PUBLIC "MONAD_CORE_FORCE_DEBUG_ASSERT")
-endif()
-target_include_directories(monad_core_c PUBLIC "src")
-
 add_library(
   monad_core STATIC
   # config
   "src/monad/config.hpp"
   # core
   "src/monad/core/array.hpp"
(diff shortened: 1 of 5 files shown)


libs/core/test/CMakeLists.txt
```diff
@@ -18,3 +18,8 @@ monad_add_test(literal_test "literal_test.cpp")
 monad_add_test(priority_pool_test "priority_pool_test.cpp")
 set_target_properties(priority_pool_test PROPERTIES RUN_SERIAL ON)
 monad_add_test(unordered_map_test "unordered_map.cpp")
+
+# This target checks that we can compile a C program without any (direct)
+# C++ dependencies
+add_executable(check_c_only check_c_only.c)
+target_link_libraries(check_c_only PRIVATE monad_core_c monad_core_crt)
```
tongzhi1998: Very minor point: can we add a line break at the end of the file?

libs/core/CMakeLists.txt
```diff
@@ -64,14 +67,30 @@ target_include_directories(openssl_sha3_asm INTERFACE "third_party/openssl")
 # libs
 # ##############################################################################
 
+# monad_core_c is a library containing source code that is compiled in C23 mode
+# and contains no C++ source files; it may be linked by downstream libraries
+# which only need C and do not want the C++ runtime as a dependency
+add_library(monad_core_c OBJECT
```
tongzhi1998: Probably a dumb question: why not moving other c code here?

kjcamann: This is not a dumb question at all. I thought I had a written something explaining why I didn't do this, apparently it was lost to time.

Rather than move all the C stuff into here at once, I figured I would move stuff over incrementally, as a pure C binaries actually need them. For example, in later PRs, the fiber stuff begins to need `tl_tid.{h,c}` in a pure C program for the first time, so it won't compile unless they are moved to this new target. So I moved them over in that PR. Over time, the library contains only the things that are strictly needed to be in C because other C dependencies use them.

As files get moved, sometimes they need something fixed also, e.g., headers that forgot to add extern "C" , etc. I didn't want to make a potentially large patch full of changes to the code that may not even be needed by anything. It seemed more reasonable to me to make a smaller PR that expresses the intention.

tongzhi1998: Thanks for the explanation.

PR #905, opened 2024-09-10
[core] add a MONAD_ABORT macro, allow MONAD_ASSERT to accept a printf-style message

kjcamann:
Occasionally we flow to a place in the code where we must die, where there is no obvious choice for the exact expression to assert is false. There are currently about 20 calls to `MONAD_ASSERT(false)` when this happens, which is not very informative.

This adds a `MONAD_ABORT` macro. The advantage over the `abort()` function in <stdlib.h> is that it may include a printf-style message that explains why we cannot continue. It includes the same source location and stack backtrace machinery as the assertion failure function.

The also changes `MONAD_ASSERT` to accept an optional, printf-style message.

### DIFF
--- a/libs/core/src/monad/core/assert.c
+++ b/libs/core/src/monad/core/assert.c
@@ -43,12 +43,12 @@ void monad_assertion_failed(
             line,
             function);
     }
-    if ((size_t)written >= sizeof(buffer)) {
-        written = (int)(sizeof(buffer) - 1);
+    if (written < 0 || (size_t)written >= sizeof(buffer)) {
+        written = (ssize_t)sizeof(buffer) - 1;
     }
-    if (msg != nullptr && written < (int)(sizeof buffer - 1)) {
+    if (msg != nullptr) {
         written += (ssize_t)strlcpy(
-            buffer + written, msg, sizeof buffer - (size_t)written);
+            buffer + written, msg, sizeof(buffer) - (size_t)written);
         if ((size_t)written >= sizeof(buffer)) {
             written = (int)(sizeof(buffer) - 1);
         }
@@ -57,10 +57,8 @@ void monad_assertion_failed(
         }
     }
     // abort() is async signal safe in glibc
-    written = write(2 /*stderr*/, buffer, (size_t)written);
-    if (written == -1) {
-        // This is to shut up the warning
-        abort();
+    if (write(STDERR_FILENO, buffer, (size_t)written) == -1) {
+        abort(); // Needed because of -Werror=unused-result
     }
     abort();
 }
--- a/libs/core/src/monad/core/assert.h
+++ b/libs/core/src/monad/core/assert.h
@@ -13,10 +13,9 @@ extern "C"
 
 #define MONAD_ASSERTION_FAILED_WITH_MSG(expr, msg)                             \
     /* Ensure msg is a static string at a fixed address; we do this because */ \
-    /* MONAD_ASSERT can be used within a signal handler, and we don't want */  \
-    /* a "double fault" occurring if dereferencing an unknown pointer could */ \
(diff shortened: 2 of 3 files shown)


libs/core/src/monad/core/assert.c
```diff
@@ -9,27 +10,52 @@ extern char const *__progname;
 
 extern void monad_stack_backtrace_capture_and_print(
     char *buffer, size_t size, int fd, unsigned indent,
-    _Bool print_async_unsafe_info);
+    bool print_async_unsafe_info);
 
-void monad_assertion_failed(
+void monad_assertion_failed_vprintf(
```
ned14: The existing code is _probably_ async signal safe most of the time. The new code has variable stack requirements, so wouldn't be alternate signal stack compatible.

If I were you, I'd implement this differently: the original macro has two forms, one is async signal safe, the other allows variable args. I'd have that macro format those variable args into a buffer, then call the async signal safe function with that buffer. Then all the guarantees and semantics are preserved.

kjcamann: Thanks for the review Niall! I don't think I understand what you're saying with respect to what is incompatible with alternate signal stacks. Could you clarify?

Also, doesn't that only mean that you couldn't reliably call `MONAD_ASSERT` or `MONAD_ABORT` from inside a signal handler? It seems pretty reasonable to disallow that by policy anyway, since they're both (potentially) going to directly _call_ `abort(3)` themselves. So I'd imagine in your carefully written signal handler you would avoid them, as they themselves raise signals.

ned14: `abort()` is async signal safe. I tried reasonably hard to ensure so would be our `MONAD_ASSERT` function, as it can get called from within async signals. The bit which isn't async signal safe is the stack backtrace print, but it should have printed at least the failure before it crashes/hangs.

Where things get tricky is if the user supplies arbitrary parameters on the stack. What if the abort/assert is failing due to stack corruption? What if the data you are trying to print in the custom message is corrupted? This is why I'd favour a two level ABORT/ASSERT design, with the base not taking any non-static parameters at all from outside and thus being impervious to program corruption.

I put a 16Kb text rendering buffer in there because on Linux the alt signal stack size is I believe 32Kb. Linux is unusually generous here, many platforms give you one page and some give you 512 bytes. I don't know what Mac OS does, but generally they have far lower defaults than Linux.

As I mentioned, if the outer macro renders the custom string into a buffer first, these issues go away. Signal handling code needs to call the async safe functions only (which it already does). My concern was making an async signal safe function no longer async signal safe - it breaks the contract promised.

kjcamann: Ah, I see your point about the message or something in the format string being corrupt while handling the assertion. In fact that seems likely to happen e.g.,


```MONAD_ASSERT(v != nullptr, "v was nullptr while handling item %d", x->i);```

Where `x` being `nullptr` is maybe more likely if `v` is (either because the stack is corrupt or some kind of logic error where they're both reading from a pile of zero-initialized memory).

I will fix tomorrow!

Comment (kjcamann) [reply libs/core/src/monad/core/assert.c]: Let me know what you think of the new version.

File: libs/core/src/monad/core/assert.c [hunk 2]
```diff
@@ -43,12 +43,12 @@ void monad_assertion_failed(
             line,
             function);
     }
-    if ((size_t)written >= sizeof(buffer)) {
-        written = (int)(sizeof(buffer) - 1);
+    if (written < 0 || (size_t)written >= sizeof(buffer)) {
+        written = (ssize_t)sizeof(buffer) - 1;
     }
-    if (msg != nullptr && written < (int)(sizeof buffer - 1)) {
+    if (msg != nullptr) {
```
kjcamann, line 49: The removed part of the guard is unnecessary since it can't possibly be false given what happened above

PR #907, opened 2024-09-10
[core] add an interface and implementations for platform dependent thread functionality

kjcamann:
Platform-dependent functionality is declared in `core/thread.h` and implemented in a platform-specific file, e.g, `core/thread-linux.c`. The two things added so far are:

1. Ability to set a thread's name
2. Ability to return a thread's system id
 
The system id has type `monad_tid_t`, which is `long`. For Linux, the system thread id is `pid_t` which is `int` but on nearly every other platform the thread id is `long` or `unsigned long`, even when `pid_t` is `int`.

The original users of the system id were transitioned to use the new interface and `tl_tid.h` and `tl_tid.c` were removed.

### DIFF
--- a/libs/core/src/monad/core/thread-darwin.c
+++ b/libs/core/src/monad/core/thread-darwin.c
@@ -22,3 +22,13 @@ int monad_thread_set_name(char const *name)
     pthread_setname_np(name);
     return 0;
 }
+
+int monad_thread_get_stack(pthread_t thread, void **stack_addr, size_t *stack_size)
+{
+    if (stack_addr == nullptr || stack_size == nullptr) {
+        return EFAULT;
+    }
+    *stack_addr = pthread_get_stackaddr_np(thread);
+    *stack_size = pthread_get_stacksize_np(thread);
+    return 0;
+}
--- a/libs/core/src/monad/core/thread-linux.c
+++ b/libs/core/src/monad/core/thread-linux.c
@@ -1,3 +1,4 @@
+#include <errno.h>
 #include <pthread.h>
 #include <unistd.h>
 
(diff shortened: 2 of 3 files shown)


libs/core/src/monad/core/spinlock.h
```diff
@@ -2,15 +2,15 @@
...
 
 #include <assert.h>
 #include <stdatomic.h>
 #include <stdbool.h>
 
-static_assert(ATOMIC_INT_LOCK_FREE == 2);
+static_assert(ATOMIC_LONG_LOCK_FREE == 2);
 
-typedef atomic_int spinlock_t;
+typedef atomic_long spinlock_t;
```
ned14, line 13: This raised an eyebrow. Intel does a 32 bit CAS faster than 64 bit, so one would need good reason to bump this to 64 bit.

If you need to store a 64 bit tid, you could CAS 32 bits of it and then store the other 32 bits after.

libs/core/src/monad/core/thread-darwin.c
```diff
@@ -0,0 +1,24 @@
+#include <err.h>
+#include <pthread.h>
+#include <stdint.h>
+
+#include <monad/core/assert.h>
+#include <monad/core/thread.h>
+
+static_assert(sizeof(monad_tid_t) >= sizeof(uint64_t));
+
+__thread monad_tid_t _monad_tl_tid = 0;
```
ned14: Not `thread_local`?

libs/core/src/monad/core/spinlock.h
kjcamann: While I definitely believe that's true for IA32, I think it's impossible to know this without testing a particular CPU model.

In a later PR I did some manual optimization of the spin lock. I started out by checking the DPDK source code for a spin lock, figuring since it's an Intel-led bleeding-edge perf project, it would contain the best known code in existence for x64. It used `lock xchg` instead of `cmpxchg`.

Curious why this was faster, I searched around the Internet to find "that used to be true, but it no longer is." A lot of this assembly stuff is like that.

The cmpxhcg is going to muck with the entire cache line. And x64 full size register operations are generally not slower than their half-size counterparts. I'm skeptical.

I know "double wide" operations _are_ still slower. I wanted to do a `CMPXCHG16` to sneak some extra information in there. But that _is_ much slower, like I would believe `CMPXCHG8` is on IA32.

How about I will leave this for now, but test out your suggestion in the spin lock PR #846 , where I have a microbenchmark?

libs/core/src/monad/core/thread-darwin.c
kjcamann: Yeah I think I can probably do `thread_local` and be fine. There is some difference between the two on Darwin, but only in C++. So if you link C and C++, you may need `constinit`. I encountered something about that in a later PR:

https://github.com/monad-crypto/monad/pull/928/commits/af52cff2b71d20239621ab81cc1ef2c27f9aa21e#diff-ebcbd0f628c7f0adc7488ecf8fcd9172f2261123eae2a46cb786e3468cc26294R63

kjcamann: That erased my line number marker, but search for `constinit` and read the comment.

kjcamann: OK I did this same thing here and checked that it doesn't emit any extra guard variables

PR #908, opened 2024-09-10
Relaxed merge validation of account nonce and balance

magedm:
Relaxed merge validation of account nonce and balance.

Replaces, extends, and simplifies [732](https://github.com/monad-crypto/monad/pull/732) and [864](https://github.com/monad-crypto/monad/pull/864). Instead of disabling relaxed validation when encountering (1) invalid transactions, (2) account creation, (3) failed calls, this PR allows relaxed validation with special validation of the affected accounts.

Replaced in part by [PR#1451](https://github.com/category-labs/monad/pull/1451).

### DIFF
--- a/category/execution/ethereum/state3/account_state.hpp
+++ b/category/execution/ethereum/state3/account_state.hpp
@@ -20,7 +20,7 @@
 
 MONAD_NAMESPACE_BEGIN
 
-class AccountState final : public AccountSubstate
+class AccountState : public AccountSubstate
 {
 public: // TODO
     template <class Key, class T>
@@ -83,6 +83,58 @@ public:
     void set_transient_storage(bytes32_t const &key, bytes32_t const &value)
     {
         transient_storage_[key] = value;
+    }
+};
+
+class OriginalAccountState final : public AccountState
+{
+    bool validate_exact_nonce_{false};
+    bool validate_exact_balance_{false};
+    uint256_t min_balance_{0};
+
+public:
+    explicit OriginalAccountState(std::optional<Account> &&account)
+        : AccountState(std::move(account))
+    {
+    }
+
+    explicit OriginalAccountState(std::optional<Account> const &account)
+        : AccountState{account}
+    {
+    }
+
+    [[nodiscard]] bool validate_exact_nonce() const
+    {
+        return validate_exact_nonce_;
+    }
+
+    [[nodiscard]] bool validate_exact_balance() const
+    {
+        return validate_exact_balance_;
+    }
+
+    [[nodiscard]] uint256_t const &min_balance() const
+    {
+        return min_balance_;
+    }
+
+    void set_validate_exact_nonce()
+    {
+        validate_exact_nonce_ = true;
+    }
+
+    void set_validate_exact_balance()
+    {
+        validate_exact_balance_ = true;
+    }
+
+    void set_min_balance(uint256_t const &value)
+    {
+        MONAD_DEBUG_ASSERT(account_.has_value());
+        MONAD_DEBUG_ASSERT(account_->balance >= value);
+        if (value > min_balance_) {
+            min_balance_ = value;
+        }
     }
 };
 
--- a/category/execution/ethereum/state3/state.hpp
+++ b/category/execution/ethereum/state3/state.hpp
@@ -42,7 +42,7 @@ class State
 
     Incarnation const incarnation_;
 
-    Map<Address, AccountState> original_{};
+    Map<Address, OriginalAccountState> original_{};
 
     Map<Address, VersionStack<AccountState>> current_{};
 
@@ -52,7 +52,10 @@ class State
 
     unsigned version_{0};
 
-    AccountState &original_account_state(Address const &address)
+    bool relaxed_validation_{false};
+
+public:
+    OriginalAccountState &original_account_state(Address const &address)
     {
         auto it = original_.find(address);
         if (it == original_.end()) {
@@ -63,6 +66,7 @@ class State
         return it->second;
     }
 
+private:
     AccountState const &recent_account_state(Address const &address)
     {
         // current
@@ -98,6 +102,16 @@ public:
         : block_state_{block_state}
         , incarnation_{incarnation}
     {
+    }
+
+    State(
+        BlockState &block_state, Incarnation const incarnation,
+        Address const &sender, uint64_t const nonce)
+        : block_state_{block_state}
+        , incarnation_{incarnation}
+        , relaxed_validation_{true}
+    {
+        set_original_nonce(sender, nonce);
     }
 
     State(State &&) = delete;
@@ -145,6 +159,11 @@ public:
         --version_;
     }
 
+    [[nodiscard]] bool relaxed_validation() const
+    {
+        return relaxed_validation_;
+    }
+
     ////////////////////////////////////////
 
     vm::VM &vm()
@@ -165,6 +184,7 @@ public:
             account = Account{};
         }
         account->nonce = nonce;
+        account_state.set_validate_exact_nonce();
     }
 
     ////////////////////////////////////////
@@ -182,6 +202,7 @@ public:
     uint64_t get_nonce(Address const &address)
     {
         auto const &account = recent_account(address);
+        original_account_state(address).set_validate_exact_nonce();
         if (MONAD_LIKELY(account.has_value())) {
             return account.value().nonce;
         }
@@ -191,6 +212,7 @@ public:
     bytes32_t get_balance(Address const &address)
     {
         auto const &account = recent_account(address);
+        original_account_state(address).set_validate_exact_balance();
         if (MONAD_LIKELY(account.has_value())) {
             return intx::be::store<bytes32_t>(account.value().balance);
         }
@@ -265,6 +287,21 @@ public:
     }
 
     ////////////////////////////////////////
+
+    bool increment_nonce(Address const &address)
+    {
+        auto &account = current_account(address);
+        auto const nonce = account.has_value() ? account->nonce : 0;
+        if (nonce == std::numeric_limits<decltype(nonce)>::max()) {
+            // Match geth behavior - don't overflow nonce
+            return false;
+        }
+        if (MONAD_UNLIKELY(!account.has_value())) {
+            account = Account{.incarnation = incarnation_};
+        }
+        account->nonce = nonce + 1;
+        return true;
+    }
 
     void set_nonce(Address const &address, uint64_t const nonce)
     {
@@ -377,12 +414,14 @@ public:
         if constexpr (rev < EVMC_CANCUN) {
             add_to_balance(beneficiary, account.value().balance);
             account.value().balance = 0;
+            original_account_state(address).set_validate_exact_balance();
         }
         else {
             if (address != beneficiary ||
                 account->incarnation == incarnation_) {
                 add_to_balance(beneficiary, account.value().balance);
                 account.value().balance = 0;
+                original_account_state(address).set_validate_exact_balance();
             }
         }
 
@@ -558,6 +597,64 @@ public:
             account = Account{.incarnation = incarnation_};
         }
         account.value().incarnation = incarnation_;
+    }
+
+    bool try_fix_account_mismatch(
+        Address const &address, OriginalAccountState &original_state,
+        std::optional<Account> const &actual)
+    {
+        auto &original = original_state.account_;
+        if (is_dead(original)) {
+            return false;
+        }
+        if (is_dead(actual)) {
+            return false;
+        }
+        if (original->code_hash != actual->code_hash) {
+            return false;
+        }
+        if (original->incarnation != actual->incarnation) {
+            return false;
+        }
+        bool const nonce_mismatch = original->nonce != actual->nonce;
+        if (nonce_mismatch) {
+            if (!relaxed_validation()) {
+                return false;
+            }
+            if (original_state.validate_exact_nonce()) {
+                return false;
+            }
+        }
+        bool const balance_mismatch = original->balance != actual->balance;
+        if (balance_mismatch) {
+            if (!relaxed_validation()) {
+                return false;
+            }
+            if (original_state.validate_exact_balance()) {
+                return false;
+            }
+            if (actual->balance < original_state.min_balance()) {
+                return false;
+            }
+        }
+        MONAD_DEBUG_ASSERT(nonce_mismatch || balance_mismatch);
+        auto it = current_.find(address);
+        if (it != current_.end()) {
+            MONAD_DEBUG_ASSERT(it->second.size() == 1);
+            auto &recent_state = it->second.recent();
+            auto &recent = recent_state.account_;
+            if (!recent) {
+                return false;
+            }
+            MONAD_DEBUG_ASSERT(original->nonce <= actual->nonce);
+            recent->nonce += actual->nonce - original->nonce;
+            // Modular arithmetic properly handles both cases: whether
+            // actual->balance or original->balance is larger than the other.
+            recent->balance += actual->balance - original->balance;
+        }
+        original->nonce = actual->nonce;
+        original->balance = actual->balance;
+        return true;
     }
 };
 
--- a/category/execution/ethereum/evm.cpp
+++ b/category/execution/ethereum/evm.cpp
@@ -32,8 +32,22 @@ MONAD_NAMESPACE_BEGIN
 bool sender_has_balance(State &state, evmc_message const &msg) noexcept
 {
     auto const value = intx::be::load<uint256_t>(msg.value);
-    auto const balance =
-        intx::be::load<uint256_t>(state.get_balance(msg.sender));
+    auto const &account = state.recent_account(msg.sender);
+    auto const balance = account.has_value() ? account->balance : 0;
+    auto &original_state = state.original_account_state(msg.sender);
+    if (balance >= value) {
+        auto const diff = balance - value;
+        auto const &original = original_state.account_;
+        auto const original_balance =
+            original.has_value() ? original->balance : 0;
+        if (original_balance > diff) {
+            auto const min_balance = original_balance - diff;
+            original_state.set_min_balance(min_balance);
+        }
+    }
+    else {
+        original_state.set_validate_exact_balance();
+    }
     return balance >= value;
 }
 
@@ -156,18 +170,17 @@ evmc::Result create(
         return result;
     }
 
-    auto const nonce = state.get_nonce(msg.sender);
-    if (nonce == std::numeric_limits<decltype(nonce)>::max()) {
+    if (!state.increment_nonce(msg.sender)) {
         // overflow
         evmc::Result result{EVMC_ARGUMENT_OUT_OF_RANGE, msg.gas};
         call_tracer.on_exit(result);
         return result;
     }
-    state.set_nonce(msg.sender, nonce + 1);
 
     Address const contract_address = [&] {
         if (msg.kind == EVMC_CREATE) {
-            return create_contract_address(msg.sender, nonce); // YP Eqn. 85
+            return create_contract_address(
+                msg.sender, state.get_nonce(msg.sender) - 1); // YP Eqn. 85
         }
         else { // msg.kind == EVMC_CREATE2
             auto const code_hash = keccak256({msg.input_data, msg.input_size});
--- a/category/execution/ethereum/execute_transaction.cpp
+++ b/category/execution/ethereum/execute_transaction.cpp
@@ -42,8 +42,8 @@ constexpr void irrevocable_change(
     uint256_t const &base_fee_per_gas)
 {
     if (tx.to) { // EVM will increment if new contract
-        auto const nonce = state.get_nonce(sender);
-        state.set_nonce(sender, nonce + 1);
+        bool const success = state.increment_nonce(sender);
+        MONAD_ASSERT(success);
     }
 
     auto const upfront_cost =
@@ -196,7 +196,16 @@ Result<evmc::Result> execute_impl2(
     BlockHashBuffer const &block_hash_buffer, State &state)
 {
     auto const sender_account = state.recent_account(sender);
-    BOOST_OUTCOME_TRY(validate_transaction(tx, sender_account));
+    auto const validate_txn = [&tx, &sender_account, &state, &sender]() {
+        auto result = validate_transaction(tx, sender_account);
+        if (!result) {
+            auto &account_state = state.original_account_state(sender);
+            account_state.set_validate_exact_nonce();
+            account_state.set_validate_exact_balance();
+        }
+        return result;
+    };
+    BOOST_OUTCOME_TRY(validate_txn());
 
     size_t const max_code_size =
         chain.get_max_code_size(hdr.number, hdr.timestamp);
@@ -234,8 +243,8 @@ Result<ExecutionResult> execute(
     {
         TRACE_TXN_EVENT(StartExecution);
 
-        State state{block_state, Incarnation{hdr.number, i + 1}};
-        state.set_original_nonce(sender, tx.nonce);
+        State state{
+            block_state, Incarnation{hdr.number, i + 1}, sender, tx.nonce};
 
 #ifdef ENABLE_CALL_TRACING
         CallTracer call_tracer{tx};
--- a/category/execution/ethereum/state2/block_state.cpp
+++ b/category/execution/ethereum/state2/block_state.cpp
@@ -120,15 +120,17 @@ vm::SharedVarcode BlockState::read_code(bytes32_t const &code_hash)
     }
 }
 
-bool BlockState::can_merge(State const &state)
+bool BlockState::can_merge(State &state) const
 {
     MONAD_ASSERT(state_);
-    for (auto const &[address, account_state] : state.original_) {
+    for (auto &[address, account_state] : state.original_) {
         auto const &account = account_state.account_;
         auto const &storage = account_state.storage_;
         StateDeltas::const_accessor it{};
         MONAD_ASSERT(state_->find(it, address));
-        if (account != it->second.account.second) {
+        auto const &actual = it->second.account.second;
+        if ((account != actual) &&
+            !state.try_fix_account_mismatch(address, account_state, actual)) {
             return false;
         }
         // TODO account.has_value()???
--- a/category/execution/ethereum/state2/block_state.hpp
+++ b/category/execution/ethereum/state2/block_state.hpp
@@ -39,7 +39,7 @@ public:
 
     vm::SharedVarcode read_code(bytes32_t const &);
 
-    bool can_merge(State const &);
+    bool can_merge(State &) const;
 
     void merge(State const &);
 


libs/execution/src/monad/state2/block_state.cpp
```diff
@@ -115,14 +115,16 @@ std::shared_ptr<CodeAnalysis> BlockState::read_code(bytes32_t const &code_hash)
     }
 }
 
-bool BlockState::can_merge(State const &state)
+bool BlockState::can_merge(State &state) const
```
aa755: because this function now changes its argument (potentially), we may now what to tweak its name to reflect that, e.g. `try_make_mergeable`

libs/execution/src/monad/state3/state.hpp
```diff
@@ -52,6 +52,9 @@ class State
 
     unsigned version_{0};
 
+    bool relaxed_validation_{false};
```
aa755: does `relaxed_validation` convey any information not already conveyed by `validate_exact_nonce_` and `validte_exact_balance_`. is `relaxed_validation` always equal to `!(validate_exact_nonce_ && validate_exact_balance_)`? if so, can we drop this field?

libs/execution/src/monad/state3/version_stack.hpp
```diff
@@ -43,6 +43,13 @@ class VersionStack
         return stack_.back().second;
     }
 
+    T &recent()
```
aa755: intuitively, what does this function return?

libs/execution/src/monad/execution/evm.cpp
```diff
@@ -34,11 +34,23 @@ std::optional<evmc::Result>
...
         return evmc::Result{EVMC_INSUFFICIENT_BALANCE, msg.gas};
     }
+    else {
+        auto const diff = balance - value;
+        auto const &original = original_state.account_;
+        auto const original_balance =
+            original.has_value() ? original->balance : 0;
+        if (original_balance > diff) {
+            auto const min_balance = original_balance - diff;
+            original_state.set_min_balance(min_balance);
```
aa755: this seems potentially buggy to me.
if a contract does 2 calls, transfer 200ETH in the first call, and 100ETH in second, after the second call, the min_balance would be set to 200, not 300, whereas, it seems can_merge should check that the balance at the start of the transaction is at least 300.

libs/execution/src/monad/state2/block_state.cpp
magedm: Sounds good. This will affect additional files. I'll change it in a separate commit.

libs/execution/src/monad/state3/state.hpp
magedm: `relaxed_validation_` applies to the whole transaction, whereas `validate_exact_nonce_` etc apply to a specific account.
The logic is that if `relaxed_validation_` is `false` (the default), then there is no relaxed validation at all, for any account.
If `relaxed_validation_` is `true` and `validate_exact_nonce_` is `false` then nonce validation is relaxed for the associated account. 
If `relaxed_validation_` is `true` and `validate_exact_nonce_` is `true` then nonce validation is strict specifically for the associated account.

libs/execution/src/monad/state3/version_stack.hpp
magedm: It is just a non-const version of the existing const `recent` function. 
It returns a non-const reference to the latest version (i.e., the top of the stack of versions).

libs/execution/src/monad/execution/evm.cpp
magedm: Can you please provide a counter example of how this can be buggy?

Say the original balance is 500.
As a result of the first call, we set `min_balance` to 200.
The call happens and the balance is decremented by 200 to 300.
Now as we check for the second call, we set `min_balance` to 300. Why? Because we calculate `min_balance = original_balance - diff`, i.e., 500 - 200 = 300.

libs/execution/src/monad/state3/state.hpp
aa755: Thanks for the clarification? Do you mind including descriptions like these as comments in code? I can do that if you agree.

Is it true that : `relaxed_validation_` is true iff `validate_exact_nonce_` or `validate_exact_balance` is false for some account? (if so `relaxed_validation_` would logically be redundant but may be useful for computational efficiency)

magedm: I'd be happy to add comments here and in general, but I wanted to be consistent with the style of the surrounding code (without comments). 
I'll edit the PR description to add detailed comments, and if it is OK with @jhunsaker I'll add brief comments to the code.

magedm: > Is it true that : relaxed_validation_ is true iff validate_exact_nonce_ or validate_exact_balance is false for some account? (if so relaxed_validation_ would logically be redundant but may be useful for computational efficiency)

No. I don't think we can be certain that it is impossible for `relaxed_validation_` to be `true` and all accounts accessed in the transaction have both `validate_exact_nonce_` and `validate_exact_balance_` be `true`.

libs/execution/src/monad/execution/evm.cpp
aa755: Thanks for that explanation. I misunderstood what the code was doing and your code seems correct actually. 
We need to ensure  `actual_current_balance >= new_debit`, which  (after rearranging) is equivalent to the check you are doing:

```
actual_current_balance >= new_debit
<-> actual_original_balance + change_this_transaction_did_sofar >= new_debit
<-> actual_original_balance + (cur_bal-speculated_original_balance) >= new_debit
<-> actual_original_balance >= new_debit  - (cur_bal- speculated_original_balance)
<-> actual_original_balance >= speculated_original_balance - (cur_bal - new_debit) 
<-> actual_original_balance >= speculated_original_balance - diff
```
`actual_current_balance` stands for the balance at this point in the sequential execution of transactions.
`actual_original_balance` stands for the balance just before this transaction in the sequential execution.

when computing `speculated_original_balance - (cur_bal - new_debit)` before every subtraction, you check that the LHS>=RHS, so there is no overflow/undeflow issue

libs/execution/src/monad/state3/version_stack.hpp
aa755: minor nit (feel free to ignore this or any other comment): it would be nice to name it as `latest()` as that name is more precise than `recent()`.

libs/execution/src/monad/state3/account_state.hpp
```diff
@@ -28,6 +28,9 @@ class AccountState final : public AccountSubstate
 
     std::optional<Account> account_{};
     Map<bytes32_t, bytes32_t> storage_{};
+    bool validate_exact_nonce_{false};
+    bool validate_exact_balance_{false};
```
aa755:
```suggestion
    bool validate_exact_balance_{false}; // once this value is set to true for a transaction, it can never be set to false for that transaction, even if the transaction does a CALL and as a result, check_sender_balance sets a minumum balance. In general, the constraints to be checked at the end cannot be weakened.
```

libs/execution/src/monad/execution/evm.cpp
```diff
@@ -34,11 +34,23 @@ std::optional<evmc::Result>
...
-    auto const balance =
-        intx::be::load<uint256_t>(state.get_balance(msg.sender));
+    auto const &account = state.recent_account(msg.sender);
+    auto const balance = account.has_value() ? account->balance : 0;
+    auto &original_state = state.original_account_state(msg.sender);
     if (balance < value) {
+        original_state.set_validate_exact_balance();
         return evmc::Result{EVMC_INSUFFICIENT_BALANCE, msg.gas};
     }
+    else {
```
aa755: the current code seems correct but the following should also work (not sure which one is better for performance, feel free to ignore)
```suggestion
    else  if (! original_state.validate_exact_balance()) {
```

libs/execution/src/monad/execution/execute_transaction.cpp
```diff
@@ -208,7 +208,14 @@ Result<Receipt> execute_impl(
         TRACE_TXN_EVENT(StartExecution);
 
         State state{block_state, Incarnation{hdr.number, i + 1}};
-        state.set_original_nonce(sender, tx.nonce);
+        state.set_relaxed_validation();
```
aa755: as soon as we start a transaction, we set its `relaxed_validation_` to true. I do not see any place where `relaxed_validation_` is set to false. I am struggling to see why this variable carries any useful information.

libs/execution/src/monad/state3/state.hpp
```diff
@@ -187,6 +208,7 @@ class State
     {
         auto const &account = recent_account(address);
         if (MONAD_LIKELY(account.has_value())) {
+            original_account_state(address).set_validate_exact_balance();
             return intx::be::store<bytes32_t>(account.value().balance);
```
aa755: Just a question: will this return the balance of the account at the beginning of the block? or the balance that has already been updated with the changes of all the committed transactions in the current block?

libs/execution/src/monad/execution/execute_transaction.cpp
magedm: Relaxed validation is not always set to true. There are several cases, where it remains false.
https://github.com/monad-crypto/monad/blob/main/libs/execution/src/monad/execution/execute_transaction.cpp#L239
https://github.com/monad-crypto/monad/blob/main/libs/execution/src/monad/execution/execute_block.cpp#L74
https://github.com/monad-crypto/monad/blob/main/libs/execution/src/monad/execution/execute_block.cpp#L190

libs/execution/src/monad/state3/state.hpp
magedm: In sequential execution, `State::get_balance` returns the balance of the account already updated with the changes of all prior transactions (with transaction indexes less than the current transaction) in the block and the changes of the current transaction so far.
In optimistic execution, it may miss some changes by prior transactions.

libs/execution/src/monad/execution/evm.cpp
```diff
@@ -142,8 +155,12 @@ std::optional<evmc::Result> pre_create_contract_account(
...
 
     // Prevent overwriting contracts - EIP-684
-    if (state.get_nonce(contract_address) != 0 ||
-        state.get_code_hash(contract_address) != NULL_HASH) {
+    if (state.get_code_hash(contract_address) != NULL_HASH) {
+        return evmc::Result{EVMC_INVALID_INSTRUCTION};
+    }
+    if (state.get_nonce(contract_address) != 0) {
+        state.original_account_state(contract_address)
+            .set_validate_exact_nonce();
```
aa755: the following change also seems correct and will likely reduce the number of retries:
```suggestion
        if (msg.kind == EVMC_CREATE)
             state.original_account_state(contract_address)
                 .set_validate_exact_nonce();
```

In case `msg.kind == EVMC_CREATE2`, the address of the new contract is independent of the nonce (IIUC) of the sender so even if the nonce was different, the overwriting (due to address conflict) would have happened even in sequential execution.

magedm: I haven't seen this case occur in replay, so it doesn't contribute to retries. At some point I had the validation check for nonzero nonce but because it is a rare case (if ever) I opted for exact validation for simplicity.

Also I am not sure I follow the logic. In the original code this check happens whether the call is for CREATE or CREATE2. Is it conceivable that the contract_address happens to be used in the same transaction for CREATE and CREATE2?

aa755: Do you have a script to measure the retry count of a replay test?

magedm: I have patches that count retries among other stats.

andreaslyn: I think we always need to validate exact nonce, because non-zero nonce indicates that the given contract address is already used.
```suggestion
    state.original_account_state(contract_address)
            .set_validate_exact_nonce();
    if (state.get_nonce(contract_address) != 0) {
```
In general we need to do something when there is a potential branch on a speculated value.
* In case `state.get_nonce(contract_address) == 0` at this point, we need to re-execute if `state.get_nonce(contract_address) != 0` when checking for mergeability.
* Otherwise if `state.get_nonce(contract_address) != 0` at this point, we need to re-execute if `state.get_nonce(contract_address) == 0` when checking for mergeability.

libs/execution/src/monad/state2/block_state.cpp
```diff
@@ -202,4 +204,74 @@ void BlockState::log_debug()
...
+    MONAD_DEBUG_ASSERT(nonce_mismatch || balance_mismatch);
+    auto it = state.current_.find(address);
+    if (it != state.current_.end()) {
+        MONAD_ASSERT(it->second.size() == 1);
+        auto &local_state = it->second.recent();
+        auto &local = local_state.account_;
+        if (nonce_mismatch) {
+            MONAD_ASSERT(original->nonce < actual->nonce);
+            auto const diff = actual->nonce - original->nonce;
+            local->nonce += diff;
```
andreaslyn: Does this hold?
```suggestion
            local->nonce += diff;
            MONAD_ASSERT(actual->nonce == local->nonce);
```

libs/execution/src/monad/execution/evm.cpp
magedm: Will do.

libs/execution/src/monad/state2/block_state.cpp
magedm: The assertion doesn't always hold because local includes changes during the current transaction, whereas actual is the state at the start of the transaction.

libs/execution/src/monad/execution/evm.cpp
magedm: Thank you. Changed the code to add one line to current main code:
```
    // Prevent overwriting contracts - EIP-684
    state.original_account_state(contract_address).set_validate_exact_nonce();
    if (state.get_nonce(contract_address) != 0 ||
        state.get_code_hash(contract_address) != NULL_HASH) {
        return evmc::Result{EVMC_INVALID_INSTRUCTION};
    }
```

aa755: If you think there is a bug, can you post a concrete set of transactions for a block in which the code in this MR can produce incorrect results.

andreaslyn: 1. Start speculative execute of transaction T1 where sender = `0x10` and `state.get_nonce(0x10) == 0`.
2. Start speculative execution of transaction T2.
3. Transaction T2 calls `CREATE` with `contract_address = 0x10`, so that `state.get_nonce(contract_address) == 0`.
4. Transaction T1 increments nonce of sender = `0x10`.
5. Transaction T2 waits for T1 to finish.
6. Transaction T1 finishes successfully by merging its final state. In particular, `state.get_nonce(0x10) == 1` at this point.
7. Transaction T2 finishes successfully by merging its final state.

Transaction T1 has updated the nonce of sender = `0x10` to `state.get_nonce(0x10) == 1`, therefore T1 should have failed with `EVMC_INVALID_INSTRUCTION`, due to the incremented nonce
```
state.get_nonce(contract_address) == state.get_nonce(0x10) == 1
```

libs/execution/src/monad/state2/block_state.cpp
```diff
@@ -228,4 +230,74 @@ void BlockState::log_debug()
...
+        if (actual->balance < original_state.min_balance()) {
+            return false;
+        }
+    }
+    MONAD_DEBUG_ASSERT(nonce_mismatch || balance_mismatch);
+    auto it = state.current_.find(address);
+    if (it != state.current_.end()) {
+        MONAD_ASSERT(it->second.size() == 1);
+        auto &local_state = it->second.recent();
+        auto &local = local_state.account_;
```
technovision99: Question I had when reading the code, is it guaranteed that `local` is not `std::nullopt` here?

libs/execution/src/monad/execution/execute_transaction.cpp
msetrin, line 1: Does any of this logic break with account abstraction?  I'm specifically thinking of the max_debit calculation.

magedm: The `max_debit` calculation is intended to make the transaction pass the balance check in [`validate_transaction`](https://github.com/category-labs/monad/blob/main/libs/execution/src/monad/execution/validate_transaction.cpp#L131-L159).
The calculation of `max_debit` follows the calculation of `v0` in `validate_transaction`.
If the latter calculation changes or goes away in the future, this optimization should adapt accordingly.

libs/execution/src/monad/state3/account_state.hpp
```diff
@@ -86,4 +86,54 @@ class AccountState final : public AccountSubstate
...
+    {
+        validate_exact_nonce_ = true;
+    }
+
+    void set_validate_exact_balance()
+    {
+        validate_exact_balance_ = true;
+    }
+
+    void set_min_balance(uint256_t const &value)
```
aa755:
```suggestion
   /** \pre ` has_value (this -> account_) && this->account.value().balance  >= value` */
    void set_min_balance(uint256_t const &value)
```
I think `this->account.value().balance  >= value` is needed because in `try_fix_account_mismatch`, we do not check this condition if there was no mismatch in balance. I think it is appropriate to consider `this->account.value().balance  >= this->min_balance_` as a class invariant, so the `\pre` is needed to ensure the class invariant holds at the end of this function call.
It may make sense to add the `\pre` as an assert for now (at least for debug builds) until all the code is proven. it seems all callers of the function already satisfy this precondition but the assert may catch future violations.

libs/execution/src/monad/state3/state.hpp
```diff
@@ -559,6 +598,79 @@ class State
...
+                return false;
+            }
+            if (original_state.validate_exact_balance()) {
+                return false;
+            }
+            if (actual->balance < original_state.min_balance()) {
+                return false;
+            }
+        }
+        MONAD_DEBUG_ASSERT(nonce_mismatch || balance_mismatch);
```
aa755: Elaborating more on my [previous comment](https://github.com/category-labs/monad/pull/908/files/498f7bf9b344a995ee1b6373232a245923e2ba32#r2192948435).
In the above line, it could be the case that `balance_mismatch` is false (and `nonce_mismatch` is true).
In this case, we will never check the condition `actual->balance < original_state.min_balance()` (line 636 was skipped) and we may return true from this function. Therefore, I think the class invariant of `OriginalAccountState` should assert that `this->account.value().balance >= this->min_balance_`.

magedm: Thanks! Will add assertion.

libs/execution/src/monad/state2/block_state.cpp
aa755: It seems that `local` is indeed guaranteed to not be `std::nullopt` here. Because:
- The first thing this function (fix_account_mismatch) does is:
```
    if (is_dead(original)) {
        return false;
```
This means, the speculative execution of this transaction is based on top of a state where this account exists (is not null):
```
inline constexpr bool is_dead(std::optional<Account> const &account)
{
    return !account.has_value() || is_empty(account.value());
}
```
- It seems an account can only go back to `std::nullopt` if the transaction does SUICIDE. however, the `account.reset()` to handle that only happens in: `State::destruct_suicides()`, which is called only in `execute_final`, which is only called AFTER `can_merge()`, which is the only caller of this function (fix_account_mismatch).

```
~/f/monad  > grep "destruct_suicides" -r . --include='*.cpp' --exclude='test_state.cpp'
./libs/execution/src/monad/execution/execute_transaction.cpp:    state.destruct_suicides<rev>();
~/f/monad  >
```

aa755: @magedm : do you agree with the above analysis? If true, it would shorten my proof of this function as I would not need to reason about what happens in the case `local` is null. This case will be ruled out by a new precondition, which untimately comes from `execute_imp2` : I would put a new postcondition on `execute_imp2` to capture that it does not reset any account.

magedm: @aa755  Yes. I agree with you analysis.

category/execution/ethereum/state3/account_state.hpp
```diff
@@ -86,4 +86,56 @@ class AccountState final : public AccountSubstate
     }
 };
 
+class OriginalAccountState final : public AccountState
```
aa755, line 89: It seems the fields of `AccountSubState` are not really relevant to `OriginalAccountState`, in the sense that the fields of the `AccountSubState` subobject of an `OriginalAccountState` are never updated: the fields of the `AccountSubState` seem relevant to `State::current_` only, not `State::original_`. 
@magedm , can you confirm this is your understanding too?

If yes, it could make sense to do another refactoring, e.g. have a `CurrentAccountState` which has `AccountSubstate` field and change `class AccountState : public AccountSubstate{` to `class AccountState{`. `CurrentAccountState` will be derived from `AccountState`, as does `OriginalAccountState`. 

The above refactoring is largely orthogonal to this MR: the main brach had the same (very minor) issue which may not be worth fixing at this time. Also, this refactoring is not really needed for my proof. I just wanted to confirm my understanding as for my Coq proofs, I need to articulate how every field in `State` corresponds to states in an idealized sequential EVM execution.

magedm: My understanding agrees with your observation. However, I haven't checked that. This PR does not depend on it.  For this PR, my goal was to relax merge validation without changing unrelated state actions.

category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +598,79 @@ class State
...
+            }
+            if (nonce_mismatch) {
+                MONAD_ASSERT(original->nonce < actual->nonce);
+                auto const diff = actual->nonce - original->nonce;
+                local->nonce += diff;
+            }
+            if (balance_mismatch) {
+                if (original->balance > actual->balance) {
+                    auto const diff = original->balance - actual->balance;
+                    MONAD_ASSERT(local->balance >= diff);
```
aa755: @magedm : I am having difficulty proving in Coq that this assertion will never get violated. I am wondering whether you already have a reasoning in mind?

magedm: @aa755 : The intent of the assertion is that it should hold if all balance subtractions were properly associated with calls to set_min_balance on the original account state. The calculation of min_balance is a bit complicated but IIUC coq should be able to handle it.
A possible explanation for why coq cannot prove the assertion is that the assertion is based on assuming that every call to subtract_from_balance is preceded by an appropriate check of the balance in sender_has_balance. Does coq make this connection? I.e., that every call to subtract_from_balance is covered by a call to sender_has_balance.

aa755: Just FYI:
It seems the proof of this assertion really depends on the following (precondition): `orignalBal - localBal <= minBal`. As we discussed, this suffices:
```
    Lemma balassertion (minBal localBal orignalBal actualBal: N):
    orignalBal - localBal <= minBal (* crucial precondition *)
   ->  minBal <= actualBal (* otherwise the function returns false and doesnt reach the assertion *)
   ->  actualBal < orignalBal (* the condition for the branch enclosing this assertion *)
   -> (orignalBal - actualBal) <= localBal (* the assertion *).
    Proof.
      lia. (* success *)
```
The precondition indeed always holds because of the way `sender_has_balance` works, as your intuition suggested:

let `totD_t` be the total debits to the account so far in the tx after step/time t of executing it. `totC_t` is similarly the total credits so far. let `f` be the final step and 0 be the initial state (`totD_0 = totC_0 = 0`).
```
   minBal_t = max_i(totD_i - totC_i) for i in 0..t
   minBal_f = max_i(totD_i - totC_i) for i in 0..f
   max_i(totD_i - totC_i) for i in 0..f >= (totD_f - totC_f) // by definition of max_i
   totD_f - totC_f = originalBal - localBal
   minBal_f  >=  originalBal - localBal // combining the above 3 lines
```
I will add a class invariant to `State` asserting that for all account `ac`, `original_[ac]->min_balance  >=  original_[ac]->balance - recent[ac]->balance`

category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +598,79 @@ class State
...
+                if (original->balance > actual->balance) {
+                    auto const diff = original->balance - actual->balance;
+                    MONAD_ASSERT(recent->balance >= diff);
+                    recent->balance -= diff;
+                }
+                else {
+                    auto const diff = actual->balance - original->balance;
+                    recent->balance += diff;
+                }
+            }
```
aa755: Because of how modular arithmetic works, those ~8 lines can be shortened to the following (assuming + - ops on `::intx::uint256` are done upto mod 2^256):
```suggestion
           recent->balance= (recent->balance + (actual->balance - original->balance))
```
But I am not sure whether that is easier or harder to understand. So I don't have any strong opinion for/against this change: feel free to ignore

aa755: I proved the following in Coq to justify the equivalence of the old snippet and the proposed new line:

```coq
Require Import ZArith.
Require Import stdpp.decidable.
Require Import bluerock.auto.cpp.specs.
Require Import bluerock.auto.cpp.tactics4.
Definition mod256 : Z := 2 ^ 256.

Definition mod256z (x : Z) : Z := Z.modulo x mod256.

Open Scope Z_scope.

(* model of the current code for updating recent->balance. the output of this fn is stored there *)
Definition update_balance
           (original actual recent : Z) : Z :=
  if bool_decide (original > actual) then
    let diff   := mod256z (original - actual) in
    mod256z (recent - diff)
  else
    let diff   := mod256z (actual - original) in
    mod256z (recent + diff).


(* model of the proposed change to the logic for  updating recent->balance. the output of this fn is stored there *)
Definition update_balance_mod
           (original actual recent : Z) : Z := 
  mod256z (recent + (actual - original)).

Require Import bluerock.auto.cpp.tactics4.

Lemma changeIsEquivalentToOrig: forall original actual recent,
    update_balance original actual recent
    = update_balance_mod original actual recent.
Proof using.
  intros.
  unfold update_balance, update_balance_mod.
  simpl. unfold mod256z.
  case_bool_decide.
  - Arith.mod_simpl. f_equiv. lia.
  - Arith.mod_simpl. reflexivity.
Qed.
  
```

magedm: Thanks @aa755 ! Updated PR.

category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +598,73 @@ class State
...
+                //   is larger than the other.
+                recent->balance += (actual->balance - original->balance);
+            }
+        }
+        if (nonce_mismatch) {
+            original->nonce = actual->nonce;
+        }
+        if (balance_mismatch) {
+            original->balance = actual->balance;
+        }
```
aa755: Just curious: what do you think about removing unnecessary branching? Feel free to ignore
```suggestion
            MONAD_ASSERT(original->nonce <= actual->nonce);
            recent->nonce += actual->nonce - original->nonce;
             // Modular arithmetic properly handles both cases:
             //   whether actual->balance or original->balance
             //   is larger than the other.
            recent->balance += (actual->balance - original->balance);
            
        }
         original->nonce = actual->nonce;
         original->balance = actual->balance;
```

magedm: It was convenient for extra logging. I updated the PR to make more concise.

PR #924, opened 2024-09-24
Support GDB user space thread debugging

ned14:
Make Monad programs which use `monad_context` tell GDB to load our custom `libthread_db.so.1` which reports to a suitable fork of GDB all of our running contexts as GDB userspace threads. They can then be switched into, state examined etc as if kernel threads.

I have placed the suitable fork of GDB as a prebuilt binary at:

https://github.com/monad-crypto/binutils-gdb/releases/download/20240930/gdb.gz

As this support completely replaced the GDB pretty printer support, that has been removed.

I also fixed a very minor thing not worth its own PR: the default `ReadOnlyOnDiskDbConfig` default read buffers and i/o uring entries were far too low at eight each. This would quite substantially impact performance negatively. So I raised them to 128 each.


libs/runloop/CMakeLists.txt
```diff
@@ -16,12 +19,39 @@ add_library(monad_context_c STATIC
...
+set_target_properties(monad_context_c_libthread_db PROPERTIES
+  POSITION_INDEPENDENT_CODE ON
+  OUTPUT_NAME "thread_db"
+  SUFFIX ".so.1"
+)
+target_link_libraries(monad_context_c_libthread_db PRIVATE monad_context_c-custom-binutils-gdb)
+# We need to inhibit all dependencies on anything else to prevent clashing with GDB's runtime
+# when loaded in as a plugin. TODO FIXME this is failing to link. If it does ever cause
+# problems with GDB, diagnosing and fixing this should be revisited.
+#target_link_options(monad_context_c_libthread_db PRIVATE -static)
```
kjcamann, line 51: Is this not needed anymore?

libs/runloop/src/monad/context/context_switcher.h
```diff
@@ -295,7 +272,7 @@ namespace monad
         //! smart pointer
         inline context_ptr make_context(
             monad_context_switcher impl, monad_context_task task,
-            struct monad_context_task_attr &attr)
+            const struct monad_context_task_attr &attr)
```
kjcamann, line 275: Given that we use the trailing `const` usually, I am guessing you did not run `clang-format`?

libs/runloop/src/monad/context/context_switcher_none.c
```diff
@@ -2,11 +2,47 @@
...
 
+LINUX_THREAD_DB_USER_THREADS_ATOMIC(thread_db_userspace_threads_t *)
+_thread_db_userspace_threads[4];
+
+static void __attribute__((constructor)) _thread_db_userspace_threads_init()
+{
+    assert(_thread_db_userspace_threads[0] == nullptr);
+    void *mem = malloc(4096);
+    if (mem != nullptr) {
+        size_t bytes = 4096;
```
kjcamann: Would be slightly nicer as:

```
size_t bytes = 4096;
void *mem = malloc(size);
if (men != nullptr) {
  ...
```

Also, why `4096`? I get it's a default and can rescaled but what is it based on?

Also, given that we're running before `main` I am not sure what you _should_ do if somehow `mem` _is_ equal to `nullptr`. Should you die here? Does `FILE *stderr` even work at this point, or do you need to `dprintf` to `STDERR_FILENO`?

libs/runloop/src/monad/context/context_switcher_none.c
```diff
@@ -212,6 +203,37 @@ extern void monad_context_reparent_switcher(
...
+                if (mem != nullptr) {
+                    free(mem);
+                }
+                thread_db_slot = allocate_thread_db_userspace_thread_index();
+            }
+            thread_db_slot = ~thread_db_slot;
+            memcpy(
+                (size_t *)&context->thread_db_slot,
+                &thread_db_slot,
+                sizeof(size_t));
```
kjcamann: It seems a little messy/confusing to do the reallocation logic in this spot. It breaks up the flow of what we're doing (trying to make a new context). Could it be moved to a helper function?

libs/runloop/CMakeLists.txt
ned14: I'd prefer to keep it as the lack of it will surely bite us in the ass and that comment is to say basically TODO FIXME

libs/runloop/src/monad/context/context_switcher_none.c
ned14: Improved. And yes, fprintf won't necessarily work at that point.

libs/runloop/src/monad/context/context_switcher_none.c
ned14: Fixed

libs/runloop/src/monad/context/context_switcher.h
ned14: Actually I did, but for some reason clang format didn't fix it. Manually fixed.

PR #936, opened 2024-10-01
EIP-1153 (`TLOAD` / `TSTORE`) support

Baltoli:
This is part of the queue of PRs we need to support Cancun; it relies on #911 being merged first.

This PR implements host support for transient storage using a version stack of storage states, rather than the explicit journaling scheme suggested in EIP-1153.


libs/execution/src/monad/state3/state.hpp
```diff
@@ -27,6 +27,7 @@
 #include <bit>
 #include <cstddef>
 #include <cstdint>
+#include <iostream>
```
kkuehlz: leftover from debugging?

Baltoli: correct, removed

PR #939, opened 2024-10-02
Fix small issues with new i/o executor preventing migration of fibers to Ken's executor

ned14:
When creating a running fiber bridge between executors with Ken's
executor at https://github.com/monad-crypto/monad/pull/928, it
was realised that there were a few small issues with the current
new i/o executor which prevented the migration of running fibers
out of the i/o executor. This PR fixes those small issues, and
comes with a new test which ensures there won't be regression
again here in the future.

### DIFF
--- a/libs/runloop/src/monad/async/test/foreign_executor.cpp
+++ b/libs/runloop/src/monad/async/test/foreign_executor.cpp
@@ -1,0 +1,177 @@
+#include <atomic>
+#include <gtest/gtest.h>
+
+#include "../../test_common.hpp"
+
+#include <monad/context/config.h>
+
+#include "../executor.h"
+#include "../task.h"
+
+TEST(foreign_executor, works)
+{
+    monad_async_executor_attr ex_attr{};
+    ex_attr.io_uring_ring.entries = 4;
+    auto ex = make_executor(ex_attr);
+    auto switcher = make_context_switcher(monad_context_switcher_sjlj);
+    static auto print_stats = [](monad_async_task task, char const *desc) {
+        std::cout << desc << ":\n   ticks_when_submitted = "
+                  << task->ticks_when_submitted
+                  << "\n   ticks_when_attached = " << task->ticks_when_attached
+                  << "\n   ticks_when_detached = " << task->ticks_when_detached
+                  << "\n   ticks_when_suspended_awaiting = "
+                  << task->ticks_when_suspended_awaiting
+                  << "\n   ticks_when_suspended_completed = "
+                  << task->ticks_when_suspended_completed
+                  << "\n   ticks_when_resumed = " << task->ticks_when_resumed
+                  << "\n   total_ticks_executed = "
+                  << task->total_ticks_executed << std::endl;
+    };
+    monad_async_task_attr t_attr{};
+    auto task = make_task(switcher.get(), t_attr);
+
+    struct shared_t
+    {
+        monad_context context;
+        struct monad_async_task_head saved_async_task;
+        int invokable_called{0}, resumed{0};
+    } shared;
+
+    shared.context = task->derived.context;
+    task->derived.user_ptr = (void *)&shared;
+    task->derived.user_code =
+        +[](monad_context_task context_task) -> monad_c_result {
+        monad_async_task task = (monad_async_task)context_task;
+        auto *shared = (shared_t *)context_task->user_ptr;
+        print_stats(task, "Just after first attach before 10 ms suspend");
+        CHECK_RESULT(monad_async_task_suspend_for_duration(
+            nullptr,
+            ((monad_async_task)task),
+            10000000ULL)); // 10 milliseconds
+        print_stats(
+            task, "After 10 ms suspend before suspend_save_detach_and_invoke");
+
+        auto r = monad_async_task_suspend_save_detach_and_invoke(
+            task,
+            &shared->saved_async_task,
+            +[](monad_context_task context_task) -> monad_c_result {
+                auto *shared = (shared_t *)context_task->user_ptr;
+                shared->invokable_called++;
+                return monad_c_make_success(0);
+            });
+        CHECK_RESULT(r);
+        if (r.value != 5) {
+            abort();
+        }
+        print_stats(
+            &shared->saved_async_task,
+            "After suspend_save_detach_and_invoke in 'naked' context before "
+            "raw suspend 1");
+        context_task->context->switcher.load(std::memory_order_acquire)
+            ->suspend_and_call_resume(context_task->context, nullptr);
+        print_stats(task, "After raw suspend now back within the executor 1");
+
+        r = monad_async_task_suspend_save_detach_and_invoke(
+            task,
+            nullptr,
+            +[](monad_context_task context_task) -> monad_c_result {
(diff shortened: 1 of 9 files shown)


libs/runloop/src/monad/context/context_switcher.h
```diff
@@ -57,7 +64,7 @@ typedef struct monad_context_task_head
 #endif
 } *monad_context_task;
 #if __STDC_VERSION__ >= 202300L || defined(__cplusplus)
-static_assert(sizeof(struct monad_context_task_head) == 56);
+static_assert(sizeof(struct monad_context_task_head) == 64);
```
kjcamann, line 67: While you're already changing this `static_assert` stuff, we don't need these conditional compilation guards if they're only for checking for earlier support than C23 because `_Static_assert` can now be spelled `static_assert`. If it breaks bindgen, you can add `-std=c23` to the CMake command, like I do in my PR that moves `monad_c_result` to `libs/core`.

libs/runloop/src/monad/async/test/foreign_executor.cpp
```diff
@@ -0,0 +1,177 @@
...
+        monad_context context;
+        struct monad_async_task_head saved_async_task;
+        int invokable_called{0}, resumed{0};
+    } shared;
+
+    shared.context = task->derived.context;
+    task->derived.user_ptr = (void *)&shared;
+    task->derived.user_code =
+        +[](monad_context_task context_task) -> monad_c_result {
+        monad_async_task task = (monad_async_task)context_task;
```
kjcamann, line 44: IMO these kinds of tests are easier to follow if the you make the user code that runs on a different context be physically in a different function (e.g., a free function at file scope) rather than a lambda in the "main" function.

If it is in inside a lambda, it is (1) indented too much but more importantly (2) harder to "see" / understand that you are in a different but "co-equal" thread of execution, since you nominally appear to physically be in the same definition context. Even when you know what's going on, you have to constantly remind yourself. Or at least, I do!

libs/runloop/src/monad/async/test/foreign_executor.cpp
```diff
@@ -0,0 +1,177 @@
...
+        auto *shared = (shared_t *)context_task->user_ptr;
+        print_stats(task, "Just after first attach before 10 ms suspend");
+        CHECK_RESULT(monad_async_task_suspend_for_duration(
+            nullptr,
+            ((monad_async_task)task),
+            10000000ULL)); // 10 milliseconds
```
kjcamann, line 50: Nit: since you're in C++ here, this might look nicer as:

`to_nanos(std::chrono::milliseconds{10})`. Then you'll need `to_nanos` is in `cpp_helpers.hpp` as some constexpr template powered by duration_cast, e.g.

```
template <typename T>
constexpr uint64_t to_nanos(std::chrono::duration<T> d) {
    return std::chrono::duration_cast<std::chrono::nanoseconds>(d).count();
}
```

libs/runloop/src/monad/async/test/foreign_executor.cpp
```diff
@@ -0,0 +1,177 @@
...
+        << "\nBack in main after executor has said there is no more work 1."
+        << std::endl;
+    // Manually resume the context to pretend we are a foreign executor
+    memset(
+        ((char *)task.get()) + sizeof(struct monad_context_task_head),
+        0xff,
+        sizeof(struct monad_async_task_head) -
+            sizeof(struct monad_context_task_head));
+    auto res = monad_c_make_success(5);
+    memcpy((void *)&task->derived.result, &res, sizeof(monad_c_result));
```
kjcamann, line 111: Maybe it's too late in the day for me but why is this not `task->derived.result = monad_c_make_success(0)`

libs/runloop/src/monad/async/test/foreign_executor.cpp
```diff
@@ -0,0 +1,177 @@
...
+        struct monad_async_task_head saved_async_task;
+        int invokable_called{0}, resumed{0};
+    } shared;
+
+    shared.context = task->derived.context;
+    task->derived.user_ptr = (void *)&shared;
+    task->derived.user_code =
+        +[](monad_context_task context_task) -> monad_c_result {
+        monad_async_task task = (monad_async_task)context_task;
+        auto *shared = (shared_t *)context_task->user_ptr;
```
kjcamann, line 45: I know it might be too late to change this now (it would be a huge refactoring at this point) but I never liked the `typedef struct tag {} *tag` pattern, where the pointer nature gets baked into the typename _if_ the thing is not just an opaque handle from the perspective of the user.

Here, you are _required_ to know that the thing is actually going to be a pointer whose layout is explicitly known to you for the purpose of the API (not just for inlining sake). Concretely, the user knows that he can and should use `context_task->user_ptr` thus they know it's not just a handle but a structure type with known members. The fact that they can write the `->` here is invisible in the code, unless you go look at the header.

This is sort of an unrelated thing to this PR so probably I shouldn't bother saying it, but it confused me when reading it. Also, you and I are writing the first C libraries and I don't want the aggregate "C style" of the project to look inconsistent by the time this is open source.

Currently no one is happy since I'm using the `_t` trailing typedef suffixes (which you don't like) and you're doing the `*` inclusion (which I don't like). So this comment doesn't relate to the PR, but this line of code reminded me that we should agree on something soon.

libs/runloop/src/monad/context/context_switcher.h
ned14: Once more of your PRs go in, I'll be making a PR that does a pass refactoring my stuff onto your stuff e.g. use monad abort instead of C abort etc. I'll also be able to hard assume C23.

libs/runloop/src/monad/async/test/foreign_executor.cpp
ned14: I guess I'm used to it from ASIO programming. You want me to hoist out the functions for this PR?

libs/runloop/src/monad/async/test/foreign_executor.cpp
ned14: I'd find to_nanos a bit anodyne personally. There is an argument that since 64 bit all durations should be in nanoseconds always everywhere period.

libs/runloop/src/monad/async/test/foreign_executor.cpp
ned14: result is const because it's only supposed to be set by executors etc.

This test is pretending to be a foreign executor, but it's not being compiled within an executor implementation, so it is const and if we're going to fake setting it, we have to use memcpy to work around us being not a real executor.

libs/runloop/src/monad/async/test/foreign_executor.cpp
ned14: This is indeed something we previously discussed and we did not reach consensus. I won't reiterate those points again, but I will say there are idioms of practice and it's okay if two libraries use differing idioms if they are designed differently. Their idiom says "I am designed like this", and that's okay. It reminds you how to speak for each API set.

I also think the upper layers of code won't expose these lower layers in any case. The lower layers will be wrapped into higher layers of abstraction.

PR #952, opened 2024-10-09
Fix a performance regression in context switching caused by the GDB IPC. Thanks to Ken for spotting and reporting it!

ned14:


libs/runloop/src/monad/context/context_switcher_fcontext.c
```diff
@@ -1,4 +1,5 @@
 // #define MONAD_CONTEXT_PRINTING 1
+// #define MONAD_CONTEXT_DISABLE_GDB_IPC
```
kjcamann, line 2: Are these manually commented/uncommented? Usually I make things like this an `option` in the CMake build, and then if the option is set, it sets the compiler definition. It doesn't seem like others are doing this, but also they don't have as much conditional compilation as we do. If there's no general feeling against it, I think it would be good to do. Then you can leave the code the same, but have different cmake build dirs with different options (I have around 10 of them with all different kinds of settings, different compilers, TSAN, etc.). @jhunsaker any thoughts on this?

ned14: It's only there for debugging purposes.

PR #963, opened 2024-10-14
New i/o executor bug fixes for cancellation

ned14:
To the new i/o executor add a test that cancellation of a suspend for duration works. This found lots of bugs all of which are now fixed.

More tests shall follow for other operations such as i/o in future
PRs.

### DIFF
--- a/libs/runloop/src/monad/async/executor.c
+++ b/libs/runloop/src/monad/async/executor.c
@@ -599,10 +599,14 @@ static inline monad_c_result monad_async_executor_run_impl(
                     cqe->flags);
                 fflush(stdout);
 #endif
-                if (cqe->user_data == 0 && cqe->res == 0 && cqe->flags == 0) {
-                    // Spurious empty CQE. Appears to be generated when you
-                    // cancel an op which would have produced a CQE.
-                    continue;
+                if (cqe->user_data == 0 && cqe->res <= 0 && cqe->flags == 0) {
+                    // Empty CQE. Theroretically no longer possible since recent
+                    // other changes.
+                    fprintf(
+                        stderr,
+                        "FATAL: Empty CQE received. This should supposedly "
+                        "never happen.\n");
+                    abort();
                 }
                 struct monad_async_task_impl *task;
                 monad_async_io_status *iostatus;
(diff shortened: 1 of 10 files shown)


libs/runloop/src/monad/async/test/cancellation.cpp
```diff
@@ -0,0 +1,236 @@
...
+            }
+            return monad_c_make_success(0);
+        };
+
+        std::vector<task_ptr> tasks;
+#if 1
+        tasks.reserve(1);
+#else
+        tasks.reserve(ex_attr.io_uring_ring.entries + 8);
+#endif
```
kjcamann: Is this `#if 1` something you meant to cleanup later? Also there's an `#if 0` about a hundred lines down from here, same question there.

libs/runloop/src/monad/async/test/cancellation.cpp
```diff
@@ -0,0 +1,236 @@
...
+            to_result(
+                monad_async_task_attach(ex.get(), tasks.back().get(), nullptr))
+                .value();
+        }
+
+        monad::small_prng rand;
+
+        struct timespec nowait
+        {
+        };
```
kjcamann: Our formatting rules make this tagged structure declaration with value initialization look very odd (i.e., like it is a definition of a new structure type). If you write it as `= {}` does it look any better? And since we don't need the `struct` tag in C++ it might be clearer without it. Also, `const`.

libs/runloop/src/monad/async/test/cancellation.cpp
```diff
@@ -0,0 +1,236 @@
...
+        auto const begin = std::chrono::steady_clock::now();
+        do {
+            auto const v = rand();
+            task_ptr &i = tasks[v % tasks.size()];
+            assert(i);
+            if ((v >> 29) == 0) {
+                // Implicit cancellation
+                i.reset();
+                implicit_cancels++;
+            }
```
kjcamann, line 73: You lost me here. What is this testing / what is an implicit cancellation?

libs/runloop/src/monad/async/executor.c
```diff
@@ -599,10 +599,14 @@ static inline monad_c_result monad_async_executor_run_impl(
                     cqe->flags);
                 fflush(stdout);
 #endif
-                if (cqe->user_data == 0 && cqe->res == 0 && cqe->flags == 0) {
-                    // Spurious empty CQE. Appears to be generated when you
-                    // cancel an op which would have produced a CQE.
-                    continue;
+                if (cqe->user_data == 0 && cqe->res <= 0 && cqe->flags == 0) {
```
kjcamann, line 602: Should you `MONAD_ASSERT` this so you get a stack trace?

libs/runloop/src/monad/async/test/cancellation.cpp
ned14: Fixed

libs/runloop/src/monad/async/test/cancellation.cpp
ned14: As the code would suggest, implicit cancellation is when you destroy a state in the middle of doing i/o.

libs/runloop/src/monad/async/test/cancellation.cpp
ned14: Fixed

libs/runloop/src/monad/async/executor.c
ned14: I've avoided any dependency on monad_core until now because it prevented the building of dependency free binaries.

https://github.com/monad-crypto/monad/pull/981 fixes this exact problem by splitting monad_core into the nice bit and the naughty bit, and once that is merged then I can safely add a dependency to the new monad_core_min library.

PR #979, opened 2024-10-25
Add db data auto expiration logic, can be enabled through `StateMachine::auto_expire()` api

Chen-Yifan:
Adds the data auto expiration logic. 
When the part of trie has `StateMachine::auto_expire()` returns true, it will expire any nodes that are out of history window during upsert. This handles all corner cases like only single child left after expiration, expire is also tested with compaction.

**The algo:**
At a trie section where auto expiration is enabled, we execute the data expiration logic for all non-update paths, specifically we expire all branches whose `node->subtrie_min_version(i) < current_version - history_length + 1`.
For branches need both data expiration and compaction, we prioritize expiration over compaction.
At a specific node, when done applying updates, expiration or compaction over all branches, we are ready to recreate the parent node, before which, we check how many valid branches are left, the same as how we handle erases.

Special cases:
- If 0 valid child left, meaning that whole branch is erased, we propagate the erasure upward to its ancestor, so on so forth.
- If 1 node left, the recreated parent will be the same as the single child except that its path is coalesced:`concat(orig_parent->path, single_branch, single_child->path)`.

### DIFF
--- a/libs/db/src/monad/mpt/node.cpp
+++ b/libs/db/src/monad/mpt/node.cpp
@@ -432,7 +432,7 @@ void ChildData::copy_old_child(Node *const old, unsigned const i)
     min_offset_fast = old->min_offset_fast(index);
     min_offset_slow = old->min_offset_slow(index);
     subtrie_min_version = old->subtrie_min_version(index);
-    cache_node = true;
+    cache_node = ptr != nullptr;
 
     MONAD_DEBUG_ASSERT(is_valid());
 }
--- a/libs/db/src/monad/mpt/trie.cpp
+++ b/libs/db/src/monad/mpt/trie.cpp
@@ -41,16 +41,16 @@ using namespace MONAD_ASYNC_NAMESPACE;
  `*_prefix_index_start` is the starting nibble index in current function frame
 */
 void dispatch_updates_flat_list_(
-    UpdateAuxImpl &, StateMachine &, UpwardTreeNode &parent, ChildData &,
+    UpdateAuxImpl &, StateMachine &, UpdateTNode &parent, ChildData &,
     Node::UniquePtr old, Requests &, NibblesView path, unsigned prefix_index);
 
 void dispatch_updates_impl_(
-    UpdateAuxImpl &, StateMachine &, UpwardTreeNode &parent, ChildData &,
+    UpdateAuxImpl &, StateMachine &, UpdateTNode &parent, ChildData &,
     Node::UniquePtr old, Requests &, unsigned prefix_index, NibblesView path,
     std::optional<byte_string_view> opt_leaf_data, int64_t version);
 
 void mismatch_handler_(
-    UpdateAuxImpl &, StateMachine &, UpwardTreeNode &parent, ChildData &,
+    UpdateAuxImpl &, StateMachine &, UpdateTNode &parent, ChildData &,
     Node::UniquePtr old, Requests &, NibblesView path,
     unsigned old_prefix_index, unsigned prefix_index);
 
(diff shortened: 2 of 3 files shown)


libs/db/src/monad/mpt/test/db_test.cpp
```diff
@@ -1192,18 +1191,157 @@ TEST_F(OnDiskDbFixture, rw_query_old_version)
     EXPECT_EQ(bad_read.error(), DbError::key_not_found);
 }
 
+TEST(DbTest, auto_expire_large_set)
+{
+    std::filesystem::path const dbname{
+        MONAD_ASYNC_NAMESPACE::working_temporary_directory() /
+        "monad_db_test_auto_expire_large_set"};
```
ned14: We probably ought to fix all the tests in here to unlink the temporary file at test end. Something like:

`auto undb = monad::make_scope_exit([&]() noexcept { std::filesystem::remove(dbname); };`

libs/db/src/monad/mpt/test/test_fixtures_base.hpp
```diff
@@ -165,7 +165,8 @@ namespace monad::test
     static_assert(alignof(StateMachineVarLenTrieWithPrefix<>) == 8);
 
     template <
-        class Compute, bool enable_compaction = false, size_t cache_depth = 6>
+        class Compute, bool enable_compaction = false,
```
ned14: Kinda getting to the cusp of where a config structure would be better. Sequences of bool template parameters are easy to mistype.

libs/db/src/monad/mpt/trie.cpp
```diff
@@ -474,6 +628,99 @@ struct compaction_receiver
...
+                single_child->version)};
+    }
+    uint16_t total_child_data_size = 0;
+    // no need to update version (max of children or itself)
+    std::vector<unsigned> orig_indexes;
+    std::vector<uint16_t> child_data_offsets;
+    orig_indexes.reserve(number_of_children);
+    child_data_offsets.reserve(number_of_children);
+    for (unsigned orig_j = 0, bit = 1;
+         orig_j < (unsigned)std::popcount(orig->mask);
```
ned14: popcount can be a very expensive operation.

libs/db/src/monad/mpt/test/db_test.cpp
Chen-Yifan: Illia had some work that did it, will revisit this later https://github.com/monad-crypto/monad/pull/808/files#diff-ce89ce5e6b1d881d0ffdc687fa2710abd8e46e49e3adb4098c0e640d47c4b6c0R94 
nvm, let me add that now

Chen-Yifan: updated

libs/db/src/monad/mpt/node.cpp
```diff
@@ -432,7 +432,7 @@ void ChildData::copy_old_child(Node *const old, unsigned const i)
     min_offset_fast = old->min_offset_fast(index);
     min_offset_slow = old->min_offset_slow(index);
     subtrie_min_version = old->subtrie_min_version(index);
-    cache_node = true;
+    cache_node = bool(ptr);
```
ned14: (ptr != nullptr)

libs/db/src/monad/mpt/node.cpp
```diff
@@ -432,7 +432,7 @@ void ChildData::copy_old_child(Node *const old, unsigned const i)
     min_offset_fast = old->min_offset_fast(index);
     min_offset_slow = old->min_offset_slow(index);
     subtrie_min_version = old->subtrie_min_version(index);
-    cache_node = true;
+    cache_node = ptr != nullptr;
```
ned14, line 435: I would tend to bracket such things as it's easy to misread what is being done here later, but I won't hold up this PR for that.

PR #987, opened 2024-10-30
mpt2: Add proof validation

kkuehlz:
* Add in memory RLP types
* add recursive rlp encode and decode
* Add proof validation
* Unit tests


libs/mpt2/src/monad/mpt2/proof.cpp
```diff
@@ -0,0 +1,134 @@
...
+        return ProofError::InvalidKey;
+    }
+
+    BOOST_OUTCOME_TRY(
+        auto const decoded_proof, rlp::decode_item(encoded_proof));
+    BOOST_OUTCOME_TRY(
+        auto const proof, UnwrapItemOrError<RawList>(decoded_proof));
+
+    std::vector<bytes32_t> hashes;
+    hashes.reserve(proof.size());
```
maxkozlovsky, line 72: Should be resize instead of reserve ( or emplace_back instead of []).

PR #988, opened 2024-10-30
As a prelude to extending DB version history length, when creating a storage pool now allocate three read-write chunks instead of one

ned14:
(the second and third cnv chunks will be used to store DB version history in a later PR).

### DIFF
--- a/libs/async/src/monad/async/storage_pool.cpp
+++ b/libs/async/src/monad/async/storage_pool.cpp
@@ -19,6 +19,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <filesystem>
+#include <format>
 #include <iostream>
 #include <limits>
 #include <memory>
@@ -491,13 +492,17 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
     for (auto const &device : devices_) {
         if (device.is_file() || device.is_block_device()) {
             auto const devicechunks = device.chunks();
-            MONAD_ASSERT(devicechunks > 0);
+            if (devicechunks < 4) {
+                throw std::runtime_error(std::format(
+                    "Device {} has {} chunks the minimum allowed is four.",
+                    device.current_path().c_str(),
+                    devicechunks));
+            }
             MONAD_DEBUG_ASSERT(
                 devicechunks <= std::numeric_limits<uint32_t>::max());
-            if (devicechunks > 1) {
-                chunks.push_back(devicechunks - 1);
-                total += devicechunks - 1;
-            }
+            // Take off three for the cnv chunks
+            chunks.push_back(devicechunks - 3);
+            total += devicechunks - 3;
             fnv1a_hash<uint32_t>::add(
                 hashshouldbe, static_cast<uint32_t>(devicechunks));
             fnv1a_hash<uint32_t>::add(
@@ -537,14 +542,20 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
             }
         }
     }
-    // First block of each device goes to conventional, remainder go to
+    // First three blocks of each device goes to conventional, remainder go to
     // sequential
-    chunks_[cnv].reserve(devices_.size());
-    for (auto &device : devices_) {
-        chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 0);
-    }
+    chunks_[cnv].reserve(devices_.size() * 3);
     chunks_[seq].reserve(total);
     if (flags.interleave_chunks_evenly) {
+        for (auto &device : devices_) {
+            chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 0);
+        }
+        for (auto &device : devices_) {
+            chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 1);
+        }
+        for (auto &device : devices_) {
+            chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 2);
+        }
         // We now need to evenly spread the sequential chunks such that if
         // device A has 20, device B has 10 and device C has 5, the interleaving
         // would be ABACABA i.e. a ratio of 4:2:1
@@ -553,7 +564,7 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
         for (size_t n = 0; n < chunks.size(); n++) {
             chunkratios[n] = double(total) / static_cast<double>(chunks[n]);
             chunkcounts[n] = chunkratios[n];
-            chunks[n] = 1;
+            chunks[n] = 3;
         }
         while (chunks_[seq].size() < chunks_[seq].capacity()) {
             for (size_t n = 0; n < chunks.size(); n++) {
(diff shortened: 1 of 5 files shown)


libs/db/src/monad/mpt/test/cli_tool_test.cpp
```diff
@@ -137,15 +137,15 @@ struct cli_tool_fixture
...
+                          (3 + Config.chunks_max / 2) *
                                   MONAD_ASYNC_NAMESPACE::AsyncIO::
                                       MONAD_IO_BUFFERS_WRITE_SIZE +
                               24576)) {
                 abort();
             }
             if (-1 == truncate(
                           dbpath2b,
-                          (1 + Config.chunks_max / 2) *
+                          (3 + Config.chunks_max / 2) *
```
Chen-Yifan: why are we adding 2 chunks when truncate file?

ned14: It turns out all the maths around `Config.chunks_max` was wrong originally, we should have always been adding one and with this PR it is now two. Previously we added nothing, and we should have been.

In that line there, we wish to round up the result of dividing by two, so we add one beforehand. So adding three is the result of the fix of the missing add of two, plus one to round up.

libs/async/src/monad/async/storage_pool.cpp
```diff
@@ -537,12 +537,15 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
...
-    // First block of each device goes to conventional, remainder go to
+    // First two blocks of each device goes to conventional, remainder go to
     // sequential
-    chunks_[cnv].reserve(devices_.size());
+    chunks_[cnv].reserve(devices_.size() * 2);
     for (auto &device : devices_) {
         chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 0);
     }
+    for (auto &device : devices_) {
+        chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 1);
```
Chen-Yifan: can we merge this to one loop? or the order in `chunks_` matters?

libs/async/src/monad/async/storage_pool.cpp
```diff
@@ -495,8 +495,8 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
             MONAD_DEBUG_ASSERT(
                 devicechunks <= std::numeric_limits<uint32_t>::max());
             if (devicechunks > 1) {
```
Chen-Yifan: `> 2`?

libs/async/src/monad/async/storage_pool.cpp
```diff
@@ -577,9 +580,9 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
     }
     else {
         for (size_t deviceidx = 0; deviceidx < chunks.size(); deviceidx++) {
-            for (size_t n = 1; n <= chunks[deviceidx]; n++) {
+            for (size_t n = 0; n < chunks[deviceidx]; n++) {
                 chunks_[seq].emplace_back(
-                    std::weak_ptr<class chunk>{}, devices_[deviceidx], n);
+                    std::weak_ptr<class chunk>{}, devices_[deviceidx], 2 + n);
```
Chen-Yifan: Is here a bug to use consecutive `deviceidx`? what if some devices have less than two chunks available, you have ignored them when doing `chunks.push_back(devicechunks - 2)` on line 498.

libs/async/src/monad/async/storage_pool.cpp
ned14: Ah, you found a bug! Thank you. It is now fixed.

libs/async/src/monad/async/storage_pool.cpp
ned14: Good catch!

libs/async/src/monad/async/storage_pool.cpp
ned14: In the non-interleaving branch, yes they are supposed to be consecutive. I fixed the test above so now this should be fine.

libs/async/src/monad/async/storage_pool.cpp
```diff
@@ -537,14 +537,17 @@ void storage_pool::fill_chunks_(creation_flags const &flags)
...
-        chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 0);
-    }
+    chunks_[cnv].reserve(devices_.size() * 2);
     chunks_[seq].reserve(total);
     if (flags.interleave_chunks_evenly) {
+        for (auto &device : devices_) {
+            chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 0);
+        }
+        for (auto &device : devices_) {
+            chunks_[cnv].emplace_back(std::weak_ptr<class chunk>{}, device, 1);
```
Chen-Yifan, line 554: can this be merged to one loop as what you did below? or just pull that out of the if else

libs/async/src/monad/async/storage_pool.cpp
Chen-Yifan: saw you added `chunks_[cnv]` lines above which is good. But I was more saying all elements in `chunks` are devices that have greater than 2 chunks, but those devices are not necessarily consecutive ones that you could index by `devices_[deviceidx]`. A bug could occur if you have `device0, device1, device2` where device 1 has only 1 chunk available. I think the logic here is definitely complex, it might be easier we just throw if any of the devices user input doesn't have 2 chunks available.

libs/async/src/monad/async/storage_pool.cpp
ned14: If interleaving chunks, do interleave the chunks.

If not interleaving chunks, do not interleave the chunks (see loop below)

libs/async/src/monad/async/storage_pool.cpp
ned14: Ah I understand you now. Give me a few mins.

libs/async/src/monad/async/storage_pool.cpp
Chen-Yifan: what's the point of interleaving cnv chunks?

libs/async/src/monad/async/storage_pool.cpp
ned14: Now we refuse devices with less than three chunks.

libs/async/src/monad/async/storage_pool.cpp
ned14: Originally, it was to spread i/o load across multiple physical devices so we could get an idea of performance with PCIe 5.0 devices using 2x PCIe 4.0 devices. Our current node writer design would need upgrading to not suck with that configuration, but otherwise it works very well, you get 180% of performance for a random read workload. It wouldn't be much work to get multi device support working at high speeds I think.

Of course as of early next year we could just buy a PCIe 5.0 device and call it a day, but then we might want to simulate PCIe 6.0 device. Dunno, it seemed to me at the time a worthwhile thing to have and it cost little to implement.

ned14: Oh I just noticed you were asking specifically about interleaving cnv chunks.

I guess it's mainly to honour what the API asked for. If they ask for interleaved, they get interleaved. It would be more surprising if the request were ignored.

Chen-Yifan: Ok, was thinking about the way we uses those cnv chunks in triedb, we mmap `db_metadata` to it, so I think we want the two chunks on each device to be consecutive?

ned14: I'm currently erring on a single cnv chunk for both copies of the version history, so 128Mb of storage for each ring. I'll make it easy to be able to swap that for 256Mb for each ring without breaking DB disc format down the line. Just seems to me 16M entries for version history is surely enough? 

16M entries is 932 hours at five versions per second. 932 hours is 38 days, more than a month.

Chen-Yifan: not my call here, we discussed it to be 2^25, we should raise in the thread if any concern

ned14: PR updated to allocate three read-write chunks, one whole chunk per DB history copy.

PR #997, opened 2024-11-01
Use pool chunks to store the DB version history

ned14:
The existing 64k DB version history on disk format is still supported.
If a new DB is created, and there are enough spare read-write blocks
in the storage pool, two of those blocks are used to store the DB
version history instead. The 64k DB version history storage in this
case is used to store the list of chunks which make up the DB version
history, thus offering the option of a future extension point if it
became necessary.

Two copies of the DB history are still stored, and the dirty flag
is still held for each during update. Each chunk contains the two
copies. Storage could be extended, if desired, to 100% of the storage
pool.

It was discovered during testing that performance was being unduly
impacted by a repeated linear scan of DB version history. This has
been replaced with a state based linear scan which does not repeat
work performed.

### DIFF
--- a/libs/db/src/monad/mpt/update_aux.cpp
+++ b/libs/db/src/monad/mpt/update_aux.cpp
@@ -1,4 +1,5 @@
 #include <monad/async/config.hpp>
+#include <monad/async/detail/scope_polyfill.hpp>
 #include <monad/async/detail/start_lifetime_as_polyfill.hpp>
 #include <monad/async/storage_pool.hpp>
 #include <monad/core/assert.h>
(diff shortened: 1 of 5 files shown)


libs/db/src/monad/mpt/update_aux.cpp
```diff
@@ -482,8 +548,51 @@ void UpdateAuxImpl::set_io(
...
+                              tofill,
+                              chunk->capacity(),
+                              (off_t)fdw.second));
+                storage.cnv_chunks[storage.cnv_chunks_len++].cnv_chunk_id = n;
+                db_metadata_[0].main->history_length +=
+                    chunk->capacity() / sizeof(chunk_offset_t);
+            }
+        }
+        else {
+            db_metadata_[0].main->history_length =
```
Chen-Yifan, line 583: so the max history length can be determined given the number of cnv chunks a storage pool has. 
can you please add a helper member func `uint64_t history_ring_buffer_capacity()`?

libs/db/src/monad/mpt/update_aux.cpp
```diff
@@ -482,8 +548,51 @@ void UpdateAuxImpl::set_io(
...
+                auto fdw = chunk->write_fd(chunk->capacity());
+                MONAD_ASSERT(
+                    -1 != ::pwrite(
+                              fdw.first,
+                              tofill,
+                              chunk->capacity(),
+                              (off_t)fdw.second));
+                storage.cnv_chunks[storage.cnv_chunks_len++].cnv_chunk_id = n;
+                db_metadata_[0].main->history_length +=
+                    chunk->capacity() / sizeof(chunk_offset_t);
```
Chen-Yifan: why is it not divided by 2?

libs/db/src/monad/mpt/update_aux.cpp
ned14: I called the function `version_history_max_possible()`.

libs/db/src/monad/mpt/update_aux.cpp
ned14: Nice catch thank you!

PR #1011, opened 2024-11-08
dynamic history length adjustment based on disk usage, kick off compaction based on disk usage too

Chen-Yifan:
replaying ethereum history
https://jenkins.devcore4.com/job/execution_replay_ethereum/job/manual_replay_entire_history/13/console


libs/db/src/monad/mpt/trie.hpp
```diff
@@ -738,6 +740,11 @@ class UpdateAuxImpl
         return io != nullptr;
     }
 
+    double disk_usage() const
+    {
+        return 1 - num_chunks(chunk_list::free) / (double)io->chunk_count();
```
ned14: Safer to do the entire calculation in double

PR #1024, opened 2024-11-14
Allow node finds to load the same node twice

ned14:
https://github.com/monad-crypto/monad-internal/issues/810 reports this assertion failure:

```
Assertion 'sender->root_.node->next(branch_index) == nullptr' failed.
```

This assertion checks that when the i/o completion arrives to say that the raw data for a node has been read, that that node isn't already loaded.

It should never occur that a node is loaded twice. We have a hash table of i/o offsets issued to prevent this. I checked the logic around the hash table modifications, it is correct. I checked that the hash table hash function and comparison function ignores the chunk offset bits apart from chunk id and offset (it does). There should be no way possible for a node to **ever** be loaded twice. I spend some hours tracing through every possible combination of state within this code -- the assertion should never, ever fail.

Yet obviously it did. My current best guess is one of three things could be at cause:

1. Memory corruption.
2. The unordered dense hash map has some very rare and subtle bug.
3. A duplicate completion event is issued for an i/o read in very rare circumstances (possibly a bug or quirk in io_uring, or in our deferred completions framework).

I'm going to assume the cause is (3).

To prevent the above issue happening again, I made the code which handles node loading be tolerant to being called multiple times.

This is not a PR which can provide a test proving that the bug is fixed, so I have not done so. The bug should never occur. The fact that it did suggests one of the three causes above. None of those are testable either.

### DIFF
--- a/libs/db/src/monad/mpt/find_notify_fiber.cpp
+++ b/libs/db/src/monad/mpt/find_notify_fiber.cpp
@@ -76,20 +76,24 @@ namespace
             ResultType buffer_)
         {
             MONAD_ASSERT(buffer_);
-            MONAD_ASSERT(parent->next(branch_index) == nullptr);
-            parent->set_next(
-                branch_index,
-                detail::deserialize_node_from_receiver_result(
-                    std::move(buffer_), buffer_off, io_state));
-            auto *const node = parent->next(branch_index);
             auto const offset = parent->fnext(branch_index);
-            auto it = inflights.find(offset);
-            auto pendings = std::move(it->second);
-            inflights.erase(it);
-            for (auto &cont : pendings) {
-                MONAD_ASSERT(cont(NodeCursor{*node}));
+            auto *node = parent->next(branch_index);
+            MONAD_DEBUG_ASSERT(node == nullptr);
+            if (node == nullptr) {
+                parent->set_next(
+                    branch_index,
+                    detail::deserialize_node_from_receiver_result(
+                        std::move(buffer_), buffer_off, io_state));
+                node = parent->next(branch_index);
             }
-            return;
+            auto it = inflights.find(offset);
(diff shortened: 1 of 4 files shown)


libs/db/src/monad/mpt/find_request_sender.cpp
```diff
@@ -61,37 +61,37 @@ struct find_request_sender::find_receiver
...
+                sender->subtrie_with_sender_lifetime_ = std::move(node_ptr);
+            }
+            else {
+                sender->root_.node->set_next(branch_index, std::move(node_ptr));
+            }
         }
-        if (sender->inflights_ != nullptr) {
-            auto it = sender->inflights_->find(next_offset);
+        auto it = sender->inflights_.find(next_offset);
+        if (it != sender->inflights_.end()) {
```
Chen-Yifan, line 84: thought it would always hit in the `inflight_` hash table?

libs/db/src/monad/mpt/find_notify_fiber.cpp
```diff
@@ -76,20 +76,24 @@ namespace
...
+                    detail::deserialize_node_from_receiver_result(
+                        std::move(buffer_), buffer_off, io_state));
+                node = parent->next(branch_index);
+            }
             auto it = inflights.find(offset);
-            auto pendings = std::move(it->second);
-            inflights.erase(it);
-            for (auto &cont : pendings) {
-                MONAD_ASSERT(cont(NodeCursor{*node}));
+            if (it != inflights.end()) {
```
Chen-Yifan, line 90: so as this one

libs/db/src/monad/mpt/find_request_sender.cpp
ned14: Note the:

> I'm going to assume the cause is (3).

above.

PR #1076, opened 2025-01-06
Fix breakage of Rust bindings if using latest Rust toolchain.

ned14:


libs/runloop/CMakeLists.txt
```diff
@@ -82,31 +81,37 @@ target_include_directories(monad_async_c PUBLIC "src")
...
     --no-copy "monad_c_.*"
     --no-copy "monad_context.*"
+    --no-copy "outcome_.*"
+    --no-debug "cxx_status_code_system"
     --with-derive-default
     --must-use-type cxx_result_status_code_system_monad_async
     --
     -I ${CMAKE_CURRENT_LIST_DIR}/src
+    -I ${CMAKE_CURRENT_LIST_DIR}/../core/src
     -D MONAD_CONTEXT_DISABLE_INLINE_CUSTOM_GDB_THREAD_DB_LOAD
```
kjcamann, line 111: Would you mind adding `-std=c23` to this command line? I guess you aren't using anything from it in these headers, but I do explicitly require elsewhere in the project.

libs/async/src/monad/async/test/sender_receiver.cpp
```diff
@@ -16,6 +15,7 @@
 #include <monad/async/util.hpp>
 #include <monad/core/array.hpp>
 #include <monad/core/assert.h>
+#include <monad/core/scope_polyfill.hpp>
```
kjcamann, line 18: I vaguely remember having to do something with this file to compile with clang-19. So some of these changes might be because of the libclang19 used by `bindgen`, in which case we really do need to merge this.

libs/runloop/CMakeLists.txt
ned14: There is good reason to not turn that on - Rust build breaks with one of the compilers if you do. As a result, I've used C 23 in the source files, but kept the C header files C11 compatible.

I _think_ I've fixed the Rust build incompatibility with C 23 issue in my build system reforms of https://github.com/category-labs/monad-bft/pull/1314. If you can categorically confirm trunk BFT does not break if I flip on C 23 here for both GCC and clang, I can go ahead and add your request.

libs/async/src/monad/async/test/sender_receiver.cpp
ned14: One of my other unmerged PRs fixes that exact line above. But I'm always getting in trouble for mixing separate concerns into PRs, so I'll have to say no to that one (but absolutely let's merge my other outstanding PRs ASAP and fix it that way).

PR #1086, opened 2025-01-09
Run CI on every commit

tongzhi1998:
Running CI on every commit upon git push.


0cjs (approved):

0xSqualo:
@tongzhi1998 wouldn’t `publish-status` post `CI Docker Workflow: success` for every pushed commit as soon as `list-commits` finishes, even if the Docker build matrix later fails or never runs?

PR #1115, opened 2025-01-27
Bump evmone to v0.13.0, intx to v0.12.0

goodlyrottenapple:


libs/execution/src/monad/execution/code_analysis.hpp
```diff
@@ -12,7 +12,7 @@ using CodeAnalysis = ::evmone::baseline::CodeAnalysis;
 
 inline CodeAnalysis analyze(byte_string_view const &code)
 {
-    return evmone::baseline::analyze(EVMC_FRONTIER, code); // TODO rev
+    return evmone::baseline::analyze(code, false); // TODO rev
```
Baltoli: Seems like they've removed the revision parameter:
```suggestion
    return evmone::baseline::analyze(code, false);
```

libs/execution/src/monad/execution/baseline_execute.cpp
```diff
@@ -33,13 +33,21 @@
 
 MONAD_NAMESPACE_BEGIN
 
+struct VMDeleter
```
goodlyrottenapple: @Baltoli I presume this codepath will eventually be folded into the (almost) identical looking interface in the compiler repo?

Baltoli: Yep, when we merge the two together properly this will go away.

PR #1130, opened 2025-01-30
runloop_monad: Consume from consensus WAL

kkuehlz:
* chain: Update devnet and testnet to cancun
* runloop_monad: Consume from consensus WAL
* includes unit tests of various states the wal could be in
* rlp: Add encode_consensus_block_body

### DIFF
--- a/libs/execution/src/monad/chain/monad_devnet.cpp
+++ b/libs/execution/src/monad/chain/monad_devnet.cpp
@@ -15,7 +15,7 @@ uint256_t MonadDevnet::get_chain_id() const
 
 evmc_revision MonadDevnet::get_revision(uint64_t const, uint64_t const) const
 {
-    return EVMC_SHANGHAI;
+    return EVMC_CANCUN;
 }
 
 MONAD_NAMESPACE_END
--- a/libs/execution/src/monad/chain/monad_testnet.cpp
+++ b/libs/execution/src/monad/chain/monad_testnet.cpp
@@ -15,7 +15,7 @@ uint256_t MonadTestnet::get_chain_id() const
 
 evmc_revision MonadTestnet::get_revision(uint64_t const, uint64_t const) const
 {
-    return EVMC_SHANGHAI;
+    return EVMC_CANCUN;
 }
 
 MONAD_NAMESPACE_END


libs/execution/src/monad/execution/wal_emitter.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+std::optional<WalEmitter::Result> WalEmitter::next()
+{
+    monad_wal_action action;
+    auto const pos = cursor_.tellg();
+    if (MONAD_LIKELY(cursor_.read(
+            reinterpret_cast<char *>(&action), sizeof(monad_wal_action)))) {
+        auto const header_filename =
+            evmc::hex(byte_string_view{std::bit_cast<bytes32_t>(action.id)}) +
+            ".header";
```
kjcamann: IMO this looks clearer as:

```
auto const header_filename = fmt::format("{}.header", action.id);
```

You may need to include `monad/core/fmt/bytes_formatter.hpp` and you may still need a bit cast if there's no implicit conversion to whatever type is actually formattable.

libs/execution/src/monad/execution/wal_emitter.hpp
```diff
@@ -0,0 +1,52 @@
...
+};
+
+static_assert(sizeof(enum monad_wal_action_type) == 1);
+static_assert(alignof(enum monad_wal_action_type) == 1);
+
+struct monad_wal_action
+{
+    enum monad_wal_action_type action;
+    struct evmc_bytes32 id;
+};
```
kjcamann: A minor stylistic thing: this is named and treated like it's a C type (`monad_`-prefixed, `struct` and `enum` elaborated-type specifiers) but it's in a C++ header file. I do like that its C declarative form emphasizes the binary-standard nature of it.

Later on, I will propose moving my `monad_event_` binary standard types into a separate library and renaming the prefix to `monad_c_`. If that ends up happening, I will move this type there, so I do suggest leaving it as C-like for now.

libs/execution/src/monad/execution/wal_emitter.cpp
```diff
@@ -0,0 +1,127 @@
...
+        auto const header_data = slurp_file(ledger_dir_ / header_filename);
+        byte_string_view header_view{header_data};
+        auto const header_res = rlp::decode_consensus_block_header(header_view);
+        MONAD_ASSERT_PRINTF(
+            !header_res.has_error(),
+            "Could not rlp decode file %s",
+            header_filename.c_str());
+
+        auto const body_filename =
+            evmc::hex(header_res.value().block_body_id) + ".body";
```
kjcamann: Same thing about `fmt::format`

libs/execution/src/monad/execution/wal_emitter.cpp
```diff
@@ -0,0 +1,127 @@
...
+    {
+        MONAD_ASSERT(
+            std::filesystem::exists(path) &&
+            std::filesystem::is_regular_file(path));
+        std::ifstream is(path);
+        MONAD_ASSERT(is);
+        return {
+            std::istreambuf_iterator<char>(is),
+            std::istreambuf_iterator<char>()};
+    }
```
kjcamann: There is a tricky unsafeness to this:

- std::istreambuf_iterator will propagate exceptions
- `std::ifstream` (IIRC) won't throw exceptions by default. It _can_, but you need to configure it to do that
- The stream is OK after `MONAD_ASSERT(is)`, but `read(2)` could fail later, and if it does, you return a zero or partially-read `byte_string`

This is a bit academic because the only reason I think it would fail would be if you get `EINTR` (someone hits Ctrl^C during the read), so your slurped file is wrong but there's no indication of that. Even though you're going to exit anyway in that case, if it causes the program to behave strangely, then it might die abnormally (by signal) and not clean up right. I think you should check the stream state after the read is finished to have confidence in it.

cmd/monad/main.cpp
```diff
@@ -327,9 +328,17 @@ int main(int const argc, char const *argv[])
                 stop);
         case CHAIN_CONFIG_MONAD_DEVNET:
         case CHAIN_CONFIG_MONAD_TESTNET:
+            WalEmitter emitter(block_db_path);
+            auto const bft_block_id =
+                bft_id_for_finalized_block(db, init_block_num);
+            monad_wal_action const rewind_action{
+                .action = MONAD_WAL_PROPOSE, .id = bft_block_id};
+            if (emitter.rewind_to(rewind_action)) {
+                emitter.next(); // skip this proposal
```
kjcamann: This seems a bit confusing: the way this reads, it seems like you rewind to the proposal step of the last finalized block, then skip the proposal step (that's fine). But shouldn't you keep skipping stuff until the the finalization of said block, which you know happens (because it's in the db).

I am not so much looking for the answer to this, but for something about the code (a comment perhaps) to make it more evident what is happening here.

kkuehlz: I'll add a comment. The problem is the `BlockHashChain` type. It has in memory state of proposals, and we replay them to rebuild it. We could traverse the DB to rebuild all these in memory, but I have a new design for the BlockHashChain that relies entirely on the DB, so I will change this once I fix that.

Good eye, though.

PR #1155, opened 2025-02-07
Improve `MONAD_ASSERT` in `libs/async` with more information.

ned14:


libs/async/src/monad/async/io.cpp
```diff
@@ -382,7 +395,12 @@ void AsyncIO::submit_request_(
     /* Do sanity check to ensure initiator is definitely appending where
     they are supposed to be appending.
     */
-    MONAD_ASSERT((chunk_and_offset.offset & 0xffff) == (offset & 0xffff));
+    MONAD_ASSERT_PRINTF(
+        (chunk_and_offset.offset & 0xffff) == (offset & 0xffff),
+        "where we are appending %u is not where we are supposed to be "
+        "appending %llu",
+        (chunk_and_offset.offset & 0xffff),
```
Chen-Yifan, line 402: should also print chunk id

libs/async/src/monad/async/io.cpp
```diff
@@ -810,7 +828,7 @@ unsigned char *AsyncIO::poll_uring_while_no_io_buffers_(bool is_write)
                 << " within_completions_count = "
                 << detail::AsyncIO_per_thread_state().within_completions_count
                 << std::endl;
-            MONAD_ASSERT("no i/o buffers remaining" == nullptr);
+            MONAD_ABORT();
```
Chen-Yifan: MONAD_ABORT("no i/o buffers remaining")

libs/async/src/monad/async/io.hpp
```diff
@@ -143,7 +143,8 @@ class AsyncIO final
 
     file_offset_t chunk_capacity(size_t id) const noexcept
     {
-        MONAD_ASSERT(id < seq_chunks_.size());
+        MONAD_ASSERT_PRINTF(
+            id < seq_chunks_.size(), "id %zu size %zu", id, seq_chunks_.size());
```
Chen-Yifan: size is too brief, seq chunks size %zu

PR #1254, opened 2025-03-25
Check version history for correctness during version history length changes.

ned14:
Rename a bunch of functions to better say what they actually do.

### DIFF
--- a/libs/db/src/monad/mpt/update_aux.cpp
+++ b/libs/db/src/monad/mpt/update_aux.cpp
@@ -217,15 +217,36 @@ void UpdateAuxImpl::fast_forward_next_version(
 }
 
 void UpdateAuxImpl::update_history_length_metadata(
-    uint64_t const history_len) noexcept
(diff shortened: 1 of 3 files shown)


libs/db/src/monad/mpt/update_aux.cpp
```diff
@@ -217,15 +217,36 @@ void UpdateAuxImpl::fast_forward_next_version(
...
+        MONAD_ASSERT(
+            new_history_length > 0 && new_history_length <= ro.capacity());
+        auto const max_version = ro.max_version();
+        if (max_version != INVALID_BLOCK_ID) {
+            auto const current_history_length =
+                reinterpret_cast<std::atomic_uint64_t *>(&m->history_length)
+                    ->load(std::memory_order_acquire);
+            if (new_history_length > current_history_length) {
+                for (auto version = max_version - new_history_length;
+                     version < max_version - current_history_length;
```
Chen-Yifan: `version <= max_version - current_history_length;` ?

PR #1262, opened 2025-04-03
monad_mpt --reset-history-length option to open a rwdb with fixed history

Chen-Yifan:
length


libs/db/src/monad/mpt/cli_tool_impl.cpp
```diff
@@ -382,6 +382,7 @@ struct impl_t
     bool truncate_database = false;
     bool create_empty_database = false;
     std::optional<uint64_t> rewind_database_to;
+    std::optional<uint64_t> history_length;
```
ned14: I think if the command line option is named `--reset-history-length`, then this variable should be named `reset_history_length`

PR #1300, opened 2025-05-02
Fix refund and status issue for selfdestruct CallFrame

tongzhi1998:


Baltoli:
Cancun spec tests pass; presumably no longer an issue

PR #1304, opened 2025-05-05
Add Prestate Tracer

tongzhi1998:
This PR adds prestate tracer, which has 2 modes
1. PreState mode: Return the prestate prior to a tx, a "stateless witness"
2. StateDiff mode: Return the diff in state caused by a tx

Instead of `rlp` encoding the trace, this PR serialize the state into `json` and then binary encode it to store in `triedb`, making decoding on the `bft` side much easier. 

Peformance:
3M (in-memory): tps=18829  gps=643 M
12M: tps=6944  gps=430 M
15M: tps=5230  gps=442 M
18M: tps=2940  gps=316 M
19M: tps=3706  gps=331 M


Baltoli:
Superseded

PR #1322, opened 2025-05-15
Bugfix rodb NodeCache, add stress test

Chen-Yifan:
### read_only_stress_test for nonblocking rodb
Add a stress test for nonblocking rodb, which helps replicate the assertions in rpc crash https://github.com/category-labs/category-internal/issues/1596. The stress test runs fine for a few hours after the fix.

The command to run the read_only_stress_test for nonblocking RODb in the newly introduced mode:
```
./build/libs/db/src/monad/mpt/test/read_only_db_stress_test \
                   --db test.db \
                   --num-nodes-per-version 50 \
                   --overwrite-keys-mode \
                   --timeout 86400 
```
### Nonblocking RODb bug fix

Fix nonblocking RODb race due to physical offset being recycled by RWDb while the offset is still accessible in NodeCache

The solution is to index by virtual chunk offset, which is always unique. 
main changes are in `find_owning_notify_fiber()`, other diff are due to `UpdateAuxImpl::physical_to_virtual()` return type change. 
We had to change db_metadata::chunk_info_t to make sure load and store of insertion count are atomic.

---

### Future Work:
- locking system around root offset ring buffer
- extend virtual offset address space or assert upon wrapping around.

### DIFF
--- a/libs/db/src/monad/mpt/detail/db_metadata.hpp
+++ b/libs/db/src/monad/mpt/detail/db_metadata.hpp
@@ -264,6 +264,14 @@ namespace detail
             return &chunk_info[idx];
         }
 
+        chunk_info_t atomic_load_chunk_info(
+            uint32_t const idx, std::memory_order load_ord =
+                                    std::memory_order_seq_cst) const noexcept
+        {
+            return reinterpret_cast<std::atomic<chunk_info_t> const *>(at(idx))
+                ->load(load_ord);
+        }
+
         chunk_info_t const &operator[](uint32_t idx) const noexcept
         {
             MONAD_DEBUG_ASSERT(idx < chunk_info_count);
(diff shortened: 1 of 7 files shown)


libs/db/src/monad/mpt/detail/db_metadata.hpp
```diff
@@ -333,50 +341,62 @@ namespace detail
...
+                auto *tail = at_(list.end);
+                auto const insertion_count = tail->insertion_count() + 1;
+                MONAD_DEBUG_ASSERT(
+                    tail->next_chunk_id == chunk_info_t::INVALID_CHUNK_ID);
+                info.insertion_count0_ = uint32_t(insertion_count) & 0x3ff;
+                info.insertion_count1_ =
+                    uint32_t(insertion_count >> 10) & 0x3ff;
+                list.end = tail->next_chunk_id = i->index(this) & 0xfffffU;
+            }
+            reinterpret_cast<std::atomic<chunk_info_t> *>(i)->store(
```
maxkozlovsky, line 368: shouldn't the chunk info be filled before the entry is added to the list? can somebody access the pointer between list.end assignment and chunk_info assignment?

Chen-Yifan: I think it doesn't matter either we update list.end first or update chunk_info first. Only writer uses on list.begin and list.end value and relies on those to know the insertion_count for the next append, readers currently only cares about the id to count mapping in chunk_info_t. And our db only has one writer at a time.

PR #1338, opened 2025-06-03
Integrate monad vm

andreaslyn:


libs/execution/src/monad/db/db_cache.hpp
```diff
@@ -174,18 +160,15 @@ class DbCache final : public Db
...
             consensus_header,
             receipts,
             call_frames,
             senders,
             transactions,
             ommers,
             withdrawals);
-        proposals_.commit(
-            std::move(state_deltas), std::move(code), consensus_header.round);
-        MONAD_ASSERT(!state_deltas);
```
Baltoli, line 187: Did you mean to delete this line?
```diff
-        MONAD_ASSERT(!state_deltas);
```

libs/execution/src/monad/db/trie_rodb.hpp
```diff
@@ -81,21 +82,21 @@ class TrieRODb final : public ::monad::Db
...
+        // TODO read intercode object
         auto code_leaf_res = db_.find(
             prefix_cursor_,
             mpt::concat(
                 CODE_NIBBLE,
                 mpt::NibblesView{to_byte_string_view(code_hash.bytes)}),
             block_number_);
         if (!code_leaf_res.has_value()) {
-            return std::make_shared<CodeAnalysis>(analyze({}));
+            MONAD_ASSERT(code_hash == NULL_HASH);
```
0xSqualo: Wouldn't this new assert abort the process if someone triggered a `read_code(...)` with a code hash that doesn't exist?

libs/execution/src/monad/db/db_cache.hpp
andreaslyn: I was considering keeping that line, but it is dead code, right?

libs/execution/src/monad/db/trie_rodb.hpp
andreaslyn: Good spot. I was planing on removing this assertion, because we are relying on this behavior in some unit tests using `trie_db`. It is safe to remove the assertion, because have another assertion in `block_state.cpp` which checks this in production
```
code_hash == NULL_HASH || result->code_size() != 0
```

libs/execution/src/monad/db/db_cache.hpp
Baltoli: Yeah on a second look I think it's just maximally defensive programming; the standard mandates that `state_deltas` has `nullptr` stored after being moved from.

PR #1362, opened 2025-06-13
EIP-7702

Baltoli:
Needs:
- [x] ~https://github.com/category-labs/monad/pull/968~
  - ~Direct dependency; must be merged first.~
- [x] ~https://github.com/category-labs/monad/pull/1387~
  - ~Here there are several `TODO(BSC)` comments that need to be addressed once the dependent PR is merged; there's a bit of difficulty keeping this PR rebased while the compiler merge is in progress.~
  - ~**No relevant tests will run until this is merged**.~
- [x] https://github.com/category-labs/monad-compiler/pull/470
  - ~Needs to reach here via the compiler in whatever form that takes (submodule / inlined repo).~
  - ~This PR is now folded into the current state of the `vm-develop` branch~

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -31,6 +31,7 @@
         "string_view": "cpp",
         "unordered_map": "cpp",
         "unordered_set": "cpp",
-        "vector": "cpp"
+        "vector": "cpp",
+        "optional": "cpp"
     }
 }
--- a/category/execution/ethereum/execute_block.cpp
+++ b/category/execution/ethereum/execute_block.cpp
@@ -203,7 +203,6 @@ Result<std::vector<Receipt>> execute_block(
     fiber::PriorityPool &priority_pool, BlockMetrics &block_metrics,
     std::vector<std::unique_ptr<CallTracerBase>> &call_tracers)
 {
-    (void)authorities;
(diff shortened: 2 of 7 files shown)


libs/execution/src/monad/execution/precompiles.cpp
```diff
@@ -102,6 +102,16 @@ std::optional<evmc::Result> check_call_precompile(evmc_message const &msg)
         return std::nullopt;
     }
 
+    if constexpr (rev >= EVMC_PRAGUE) {
+        // EIP-7702 specifies that precompiles don't actually get called when
+        // they're the target of a delegation.
+        auto const delegated = false; // TODO(BSC)
+        // auto const delegated = (msg.flags & EVMC_DELEGATED) != 0;
```
dhil: I think this condition is right, because at this point we know that the `address` is a precompile, so if it is an EIP-7702 delegated call then we have nothing to execute. The compiler PR 470 sets `EVMC_DELEGATED` now.

Baltoli: Yeah I think the condition is correct also; the TODO is just because before updating the compiler and doing the repo merge, we didn't have access to `EVMC_DELEGATED` in evmc

PR #1381, opened 2025-06-28
Add block metrics

magedm:
Sample block metrics:
```
// bl   block number
// ro   round number
// ts   block start timestamp
// tx   number of transactions
// rt   number of retries (transactions that failed merge validation)
// rtp  percentage of retries
// sr   sender recovery time (microseconds)
// txe  transaction parallel execution time (includes optimistic execution and re-execution after validation failure)
// cmt  block commit time
// tot  total block execution time
// tpse tps based on transaction execution time only
// tps  tps based on total block time
// gas  gas
// gpse gps based on transaction execution time only
// gps  gps based on total block time
// ae   accounts cache misses where the account was empty
// ane  accounts cache misses where the account was nonempty
// sz   storage cache miss where the value was zero
// snz  storage cache miss where the value was nonzero
// ac   accounts cache size (max 10M)
// sc   storage cache size (max 10M)

2025-07-02 21:26:17.177409348 [255384] runloop_ethereum.cpp:186 LOG_INFO        __exec_block,bl=19000099,ts=1751491577114,tx=  164,rt= 163,rtp=99.39%,sr= 2304µs,txe= 42773µs,cmt= 13254µs,tot= 62880µs,tpse= 3834,tps= 2608,gas= 14182880,gpse= 331,gps=225,ae=  23,ane=  95,sz= 158,snz= 285,ac=   14892,sc=   49640

```

### DIFF
--- a/cmd/monad/runloop_ethereum.cpp
+++ b/cmd/monad/runloop_ethereum.cpp
@@ -15,6 +15,7 @@
 #include <monad/execution/validate_block.hpp>
 #include <monad/execution/validate_transaction.hpp>
 #include <monad/fiber/priority_pool.hpp>
+#include <monad/metrics/block_metrics.hpp>
 #include <monad/procfs/statm.h>
 #include <monad/state2/block_state.hpp>
 
@@ -80,6 +81,9 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
     uint64_t const start_block_num = block_num;
     BlockDb block_db(ledger_dir);
     while (block_num <= end_block_num && stop == 0) {
+        [[maybe_unused]] auto const block_start =
+            std::chrono::system_clock::now();
+        auto const block_begin = std::chrono::steady_clock::now();
         Block block;
         MONAD_ASSERT_PRINTF(
             block_db.get(block_num, block),
@@ -93,8 +97,12 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
 
         BOOST_OUTCOME_TRY(static_validate_block(rev, block));
 
+        auto const sender_recovery_begin = std::chrono::steady_clock::now();
         auto const recovered_senders =
             recover_senders(block.transactions, priority_pool);
+        [[maybe_unused]] auto const sender_recovery_time =
+            std::chrono::duration_cast<std::chrono::microseconds>(
+                std::chrono::steady_clock::now() - sender_recovery_begin);
         std::vector<Address> senders(block.transactions.size());
         for (unsigned i = 0; i < recovered_senders.size(); ++i) {
             if (recovered_senders[i].has_value()) {
@@ -112,6 +120,7 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
                 ? std::nullopt
                 : std::make_optional(block.header.number - 1));
         BlockState block_state(db, vm);
+        BlockMetrics block_metrics;
         BOOST_OUTCOME_TRY(
             auto results,
             execute_block(
@@ -121,7 +130,8 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
                 senders,
                 block_state,
                 block_hash_buffer,
-                priority_pool));
+                priority_pool,
+                block_metrics));
 
         std::vector<Receipt> receipts(results.size());
         std::vector<std::vector<CallFrame>> call_frames(results.size());
@@ -132,6 +142,7 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
         }
 
         block_state.log_debug();
+        auto const commit_begin = std::chrono::steady_clock::now();
         block_state.commit(
             MonadConsensusBlockHeader::from_eth_header(block.header),
             receipts,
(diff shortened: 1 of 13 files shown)


cmd/monad/runloop_ethereum.cpp
```diff
@@ -80,6 +80,8 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
     uint64_t const start_block_num = block_num;
     BlockDb block_db(ledger_dir);
     while (block_num <= end_block_num && stop == 0) {
+        auto const tstart = std::chrono::system_clock::now();
```
maxkozlovsky: Could use a single variable here, calling now() two times in a row seems redundant.

cmd/monad/runloop_ethereum.cpp
```diff
@@ -170,6 +183,34 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
...
+                           std::chrono::steady_clock::now() - tbegin)
+                           .count());
+        LOG_INFO(
+            "__exec_block,bl={:8},ro={:8},ts={}"
+            ",tx={:5},rt={:4},rtp={:5.2f}%"
+            ",sr={:5},txe={:6},cmt={:6},tot={:6},tpse={:5},tps={:5}"
+            ",gas={:9},gpse={:4},gps={:3}{}",
+            block.header.number,
+            block.header.number,
+            std::chrono::duration_cast<std::chrono::milliseconds>(
```
maxkozlovsky, line 177: printing start is redundant since log prints time and start is (current time - block_time)

cmd/monad/runloop_ethereum.cpp
```diff
@@ -93,8 +95,14 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
 
         BOOST_OUTCOME_TRY(static_validate_block(rev, block));
 
+        auto const tbegin_sender_recovery = std::chrono::steady_clock::now();
         auto const recovered_senders =
             recover_senders(block.transactions, priority_pool);
+        uint64_t const sender_recovery_time =
+            (uint64_t)(std::chrono::duration_cast<std::chrono::microseconds>(
```
maxkozlovsky: maybe make a function since similar code is repeated several times
```
auto elapsed_ms(std::chrono::steady_clock::time_point start_time) {
    return std::chrono::duration_cast<std::chrono::microseconds>(
        std::chrono::steady_clock::now() - start_time
    ).count();
}
```

cmd/monad/runloop_ethereum.cpp
magedm: I don't have an immediate use for the timestamp.

@msetrin Do you want an explicit timestamp or as suggested it can be infered from the block execution time and the log timestamp?

cmd/monad/runloop_ethereum.cpp
magedm: The calls to now() are for two different clocks. Can the two calls be combined?

cmd/monad/runloop_ethereum.cpp
magedm: Will do. Thanks!

cmd/monad/runloop_ethereum.cpp
maxkozlovsky: Not really, missed that it is two different clocks

cmd/monad/runloop_ethereum.cpp
msetrin: > log prints time and start is (current time - block_time)

Is this actually true?  Does log time actually map exactly to the the end of block precessing time?

cmd/monad/runloop_ethereum.cpp
```diff
@@ -170,6 +182,34 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
...
+            (uint64_t)std::chrono::duration_cast<std::chrono::microseconds>(
+                std::chrono::steady_clock::now() - block_begin)
+                .count();
+        LOG_INFO(
+            "__exec_block,bl={:8},ro={:8},ts={}"
+            ",tx={:5},rt={:4},rtp={:5.2f}%"
+            ",sr={:5},txe={:6},cmt={:6},tot={:6},tpse={:5},tps={:5}"
+            ",gas={:9},gpse={:4},gps={:3}{}",
+            block.header.number,
+            block.header.number,
```
msetrin: Is this correct?  You seem to be using block.header.number for both block and round.

cmd/monad/runloop_ethereum.cpp
magedm: Quill takes the timestamp immediately on the caller thread.
https://quillcpp.readthedocs.io/en/latest/timestamp_types.html
The time measurement is negligible anyway compared to the latencies that we're measuring so it might be cleaner to keep the explicit timestamp.

cmd/monad/runloop_ethereum.cpp
magedm: In Ethereum replay, the round number is the same as the block number because proposals are always finalized.

PR #1422, opened 2025-07-23
Ensure usage of `asmjit` dual mapping

Baltoli:
Raised by @0xSqualo initially; this PR sets a flag to ensure that we always force asmjit to use dual-mapped JIT pages (their documentation and implementation details are out of sync on whether this is the case by default, so we set the flag ourselves).

### DIFF
--- a/category/vm/libs/vm/src/monad/vm/compiler.hpp
+++ b/category/vm/libs/vm/src/monad/vm/compiler.hpp
@@ -85,6 +85,11 @@ namespace monad::vm
         void compile_loop();
         void dispense_compile_jobs();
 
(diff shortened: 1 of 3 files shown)


category/vm/libs/vm/src/monad/vm/compiler.hpp
```diff
@@ -85,6 +85,7 @@ namespace monad::vm
         void compile_loop();
         void dispense_compile_jobs();
 
+        asmjit::JitAllocator::CreateParams asmjit_create_params_;
```
andreaslyn: This can be static and probably constexpr as well.

PR #1424, opened 2025-07-23
Byte optimisations

goodlyrottenapple:
Addresses https://github.com/category-labs/monad/issues/1463

### DIFF
--- a/category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
@@ -591,15 +591,6 @@ namespace monad::vm::compiler::native
             as_.jmp(back);
         }
 
-        for (auto const &[lbl, rpq, back] : byte_out_of_bounds_handlers_) {
-            as_.bind(lbl);
-            as_.xor_(rpq[0].r32(), rpq[0].r32());
-            as_.xor_(rpq[1].r32(), rpq[1].r32());
-            as_.xor_(rpq[2].r32(), rpq[2].r32());
-            as_.xor_(rpq[3].r32(), rpq[3].r32());
-            as_.jmp(back);
-        }
-
         error_block(error_label_, runtime::StatusCode::Error);
 
         // By putting jump table in the text section, we can use the
@@ -951,7 +942,6 @@ namespace monad::vm::compiler::native
         // size of jump table
         return code_holder_.textSection()->realSize() +
                (load_bounded_le_handlers_.size() << 5) +
-               (byte_out_of_bounds_handlers_.size() << 5) +
                (rodata_.data().size() << 5) + (bytecode_size_ << 2);
     }
 
@@ -1133,6 +1123,21 @@ namespace monad::vm::compiler::native
         }
         auto r = stack_.release_general_reg(std::move(elem));
         return {r, GeneralRegReserv{r}};
+    }
+
+    template <typename... LiveSet>
+    std::pair<StackElemRef, AvxRegReserv> Emitter::alloc_or_release_avx_reg(
+        StackElemRef elem, std::tuple<LiveSet...> const &live)
+    {
+        if (is_live(elem, live)) {
+            if (stack_.has_free_avx_reg() ||
+                (!elem->stack_offset() && !elem->general_reg() &&
+                 !elem->literal())) {
+                return alloc_avx_reg();
+            }
+        }
+        auto r = stack_.release_avx_reg(std::move(elem));
+        return {r, AvxRegReserv{r}};
     }
 
     template <typename... LiveSet, size_t... Is>
@@ -2313,11 +2318,17 @@ namespace monad::vm::compiler::native
         auto ix = stack_.pop();
         auto src = stack_.pop();
 
-        if (ix->literal() && src->literal()) {
+        if (ix->literal()) {
             auto const &i = ix->literal()->value;
-            auto const &x = src->literal()->value;
-            push(runtime::byte(i, x));
-            return;
+            if (i >= 32) {
+                push(0);
+                return;
+            }
+            if (src->literal()) {
+                auto const &x = src->literal()->value;
+                push(runtime::byte(i, x));
+                return;
+            }
         }
 
         RegReserv const ix_reserv{ix};
@@ -2325,23 +2336,33 @@ namespace monad::vm::compiler::native
 
         discharge_deferred_comparison();
 
-        if (!src->stack_offset()) {
-            mov_stack_elem_to_stack_offset(src);
+        if (src->general_reg()) {
+            if (ix->literal()) {
+                byte_literal_ix_general_reg_src(std::move(ix), std::move(src));
+            }
+            else {
+                byte_non_literal_ix_general_reg_src(
+                    std::move(ix), std::move(src));
+            }
         }
-        if (ix->literal()) {
-            byte_literal_ix(ix->literal()->value, *src->stack_offset());
-            return;
+        else if (src->avx_reg()) {
+            if (ix->literal()) {
+                byte_literal_ix_avx_reg_src(std::move(ix), std::move(src));
+            }
+            else {
+                byte_non_literal_ix_avx_reg_src(std::move(ix), std::move(src));
+            }
         }
-        if (ix->general_reg()) {
-            byte_general_reg_or_stack_offset_ix(
-                std::move(ix), *src->stack_offset());
-            return;
+        else {
+            if (ix->literal()) {
+                MONAD_VM_DEBUG_ASSERT(src->stack_offset().has_value());
+                byte_literal_ix_stack_offset_src(std::move(ix), std::move(src));
+            }
+            else {
+                byte_non_literal_ix_literal_or_stack_offset_src(
+                    std::move(ix), std::move(src));
+            }
         }
-        if (!ix->stack_offset()) {
-            mov_avx_reg_to_stack_offset(ix);
-        }
-        byte_general_reg_or_stack_offset_ix(
-            std::move(ix), *src->stack_offset());
     }
 
     // Discharge
@@ -3656,68 +3677,6 @@ namespace monad::vm::compiler::native
         }
     }
 
-    void Emitter::byte_literal_ix(uint256_t const &ix, StackOffset src)
-    {
-        if (ix >= 32) {
-            return push(0);
-        }
-        int64_t const i = 31 - static_cast<int64_t>(ix[0]);
-
-        auto [dst, dst_reserv] = alloc_general_reg();
-        Gpq256 const &gpq = general_reg_to_gpq256(*dst->general_reg());
-
-        as_.xor_(gpq[0].r32(), gpq[0].r32());
-        as_.xor_(gpq[1].r32(), gpq[1].r32());
-        as_.xor_(gpq[2].r32(), gpq[2].r32());
-        as_.xor_(gpq[3].r32(), gpq[3].r32());
-        auto m = stack_offset_to_mem(src);
-        m.addOffset(i);
-        as_.mov(gpq[0].r8Lo(), m);
-
-        stack_.push(std::move(dst));
-    }
-
-    void Emitter::byte_general_reg_or_stack_offset_ix(
-        StackElemRef ix, StackOffset src)
-    {
-        auto [dst, dst_reserv] = alloc_general_reg();
-
-        Gpq256 const &dst_gpq = general_reg_to_gpq256(*dst->general_reg());
-
-        as_.mov(dst_gpq[0], 31);
-        as_.xor_(dst_gpq[1].r32(), dst_gpq[1].r32());
-        as_.xor_(dst_gpq[2].r32(), dst_gpq[2].r32());
-        as_.xor_(dst_gpq[3].r32(), dst_gpq[3].r32());
-        if (ix->general_reg()) {
-            Gpq256 const &ix_gpq = general_reg_to_gpq256(*ix->general_reg());
-            as_.sub(dst_gpq[0], ix_gpq[0]);
-            as_.sbb(dst_gpq[1], ix_gpq[1]);
-            as_.sbb(dst_gpq[2], ix_gpq[2]);
-            as_.sbb(dst_gpq[3], ix_gpq[3]);
-        }
-        else {
-            MONAD_VM_DEBUG_ASSERT(ix->stack_offset().has_value());
-            x86::Mem m = stack_offset_to_mem(*ix->stack_offset());
-            as_.sub(dst_gpq[0], m);
-            for (size_t i = 1; i < 4; ++i) {
-                m.addOffset(8);
-                as_.sbb(dst_gpq[i], m);
-            }
-        }
-        auto byte_out_of_bounds_lbl = as_.newLabel();
-        auto byte_after_lbl = as_.newLabel();
-        as_.jb(byte_out_of_bounds_lbl);
-        auto m = stack_offset_to_mem(src);
-        m.setIndex(dst_gpq[0]);
-        as_.mov(dst_gpq[0].r8Lo(), m);
-        as_.bind(byte_after_lbl);
-
-        byte_out_of_bounds_handlers_.emplace_back(
-            byte_out_of_bounds_lbl, dst_gpq, byte_after_lbl);
-
-        stack_.push(std::move(dst));
-    }
-
     // Sets zero flag according to whether `e` is below `i`.
     // Does not update the low 64 bits, even when `e` is not live.
     template <typename... LiveSet>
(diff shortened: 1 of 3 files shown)


category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3585,111 +3595,268 @@ namespace monad::vm::compiler::native
...
-    void Emitter::byte_literal_ix(uint256_t const &ix, StackOffset src)
+    // Sets zero flag according to whether `e` is below `i` when
+    // `compute_i_minus_e_minus_one` = false. Sets sign flag according to
+    // whether `e` is greater than or equal `i` when
+    // `compute_i_minus_e_minus_one` = true. Additionally, if `e` is below `i`,
+    // and `compute_i_minus_e_minus_one` = true, rax will conain the value
+    // `i - 1 - e - 1` otherwise rax can have any value.
+    template <bool compute_i_minus_e_minus_one, typename... LiveSet>
+    void Emitter::cmp_stack_elem_to_uint16(
+        StackElemRef e, uint16_t i, std::tuple<LiveSet...> const &live)
```
goodlyrottenapple: I originally modified this function by adding a template flag, not realising there was a recent addition of the avx case. At this point, perhaps I should just make a separate version which does the `i - 1 - e` computation?

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2241,35 +2241,45 @@ namespace monad::vm::compiler::native
...
-        RegReserv const ix_reserv{ix};
+        GeneralRegReserv const ix_reserv{ix};
         RegReserv const src_reserv{src};
 
         discharge_deferred_comparison();
 
-        if (!src->stack_offset()) {
-            mov_stack_elem_to_stack_offset(src);
+        if (!ix->literal() && !ix->general_reg() && !ix->stack_offset()) {
+            mov_avx_reg_to_stack_offset(ix);
```
andreaslyn: This is not necessary. See the new implementation of `cmp_stack_elem_to_uint16`

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
andreaslyn: I think that this functionality is not needed

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3585,111 +3595,268 @@ namespace monad::vm::compiler::native
...
+        auto zero_reg = dst_gpq[1];
+
+        if (ix->literal()) {
+            auto const &i = ix->literal()->value;
+            MONAD_VM_DEBUG_ASSERT(i < 32);
+            src_mem.addOffset(31 - static_cast<int64_t>(i[0]));
+            as_.movzx(dst_gpq[0].r32(), src_mem);
+        }
+        else {
+            cmp_stack_elem_to_uint16<true>(ix, 32, std::make_tuple(src));
```
andreaslyn: Except from the stack offset case, the algorithms work without restricting to ix <= 32. All that is needed is to check that the upper 192 bits are clear. You can add a separate function for this check. For stack offset you can use the existing `cmp_stack_elem_to_uint16`

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3585,111 +3595,268 @@ namespace monad::vm::compiler::native
...
+                auto ix0 = ix_gpq[0];
+                ix_less_than = [this, ix0](uint32_t c) {
+                    this->as_.cmp(ix0, c);
+                };
+            }
+            else {
+                MONAD_VM_DEBUG_ASSERT(ix->stack_offset().has_value());
+                auto ix0 = stack_offset_to_mem(*ix->stack_offset());
+                ix_less_than = [this, ix0](uint32_t c) {
+                    this->as_.cmp(ix0, c);
```
andreaslyn: Put ix0 in a register to avoid loading memory for each comparison.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3585,111 +3595,268 @@ namespace monad::vm::compiler::native
...
+            ix_less_than(24);
+            as_.cmovb(dst_gpq[0], dst_gpq[1]);
+            ix_less_than(16);
+            as_.cmovb(dst_gpq[0], dst_gpq[2]);
+            ix_less_than(8);
+            as_.cmovb(dst_gpq[0], dst_gpq[3]);
+
+            as_.xor_(zero_reg.r32(), zero_reg.r32());
+
+            cmp_stack_elem_to_uint16<true>(ix, 32, std::make_tuple(dst));
```
andreaslyn: Just need to check whether the high 192 bits are clear

category/vm/test/unit/emitter_tests.cpp
```diff
@@ -2397,10 +2397,31 @@ TEST(Emitter, byte)
     asmjit::JitRuntime rt;
     pure_bin_instr_test(rt, BYTE, &Emitter::byte, 31, 1, 1);
     pure_bin_instr_test(
-        rt, BYTE, &Emitter::byte, 0, {0, 0, 0, 0x8877665544332211}, 0x88);
+        rt,
```
andreaslyn, line 2400: See the test in `byte_bench/main.cpp`:
```
    for (uint64_t idx = 0; idx <= 64; ++idx) {
        result.emplace_back(idx, x, f(idx, x));
    }
    for (uint64_t idx = std::numeric_limits<uint64_t>::max();
            idx > 0;
            idx = idx >> 1) {
        result.emplace_back(idx, x, f(idx, x));
    }
    uint64_t idx = uint64_t{1} << 63;
    result.emplace_back(idx, x, f(idx, x));
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2241,35 +2241,45 @@ namespace monad::vm::compiler::native
...
+            }
+            if (src->literal()) {
+                auto const &x = src->literal()->value;
+                push(runtime::byte(i, x));
+                return;
+            }
         }
 
-        RegReserv const ix_reserv{ix};
+        GeneralRegReserv const ix_reserv{ix};
```
andreaslyn: We will probably want to reserve avx register as well

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3585,111 +3595,268 @@ namespace monad::vm::compiler::native
...
+        as_.xor_(dst_gpq[3].r32(), dst_gpq[3].r32());
+
+        auto ix_temp_reg = x86::rax;
+        auto zero_reg = dst_gpq[1];
+
+        if (ix->literal()) {
+            auto const &i = ix->literal()->value;
+            MONAD_VM_DEBUG_ASSERT(i < 32);
+            src_mem.addOffset(31 - static_cast<int64_t>(i[0]));
+            as_.movzx(dst_gpq[0].r32(), src_mem);
```
andreaslyn: We do not have benchmarks for the cases where ix is literal. This is not the only viable implementation

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3585,111 +3595,268 @@ namespace monad::vm::compiler::native
...
-        auto byte_after_lbl = as_.newLabel();
-        as_.jb(byte_out_of_bounds_lbl);
-        auto m = stack_offset_to_mem(src);
-        m.setIndex(dst_gpq[0]);
-        as_.mov(dst_gpq[0].r8Lo(), m);
-        as_.bind(byte_after_lbl);
 
-        byte_out_of_bounds_handlers_.emplace_back(
-            byte_out_of_bounds_lbl, dst_gpq, byte_after_lbl);
+        as_.shrx(dst_gpq[0], dst_gpq[0], ix_temp_reg);
```
andreaslyn: This should for sure not be shared between the implementations, but perhaps the avx register desination is good when is is literal. We need benchmarks.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: do we want a jump if the upper bits aren't clear or do the same as now and put in a value such that executing the rest of the code will yield 0 in the `dst[0]` register?

andreaslyn: The branchless algorithm

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: As suggested, i have pulled out the versions when ix is a literal into separate methods which can be optimised further separately.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        as_.shl(x86::rax, 3);
+
+        // select double word from src based on dword_index
+        as_.vpermps(dst_ymm, scratch_ymm, src_ymm);
+        as_.vmovq(scratch_ymm.xmm(), x86::rax);
+        // shift right by sub_byte_index bytes
+        as_.vpsrlq(dst_ymm.xmm(), dst_ymm.xmm(), scratch_ymm.xmm());
+        // clear everything but the lowest byte
+        auto const mask = rodata_.add16(0xff, 0);
+        as_.vpand(dst_ymm.xmm(), dst_ymm.xmm(), mask);
```
goodlyrottenapple: will be swapping this version out for:

```
[[gnu::noinline]]
uint256 byte_avx3(uint64_t byte_index0, uint256 const &x0)
{
    static uint32_t const bound = 1;

    register uint64_t byte_index asm("rsi") = byte_index0;

    asm (
        "sub $31, %%rsi\n\t"
        "cmoval %[bound], %%esi\n\t"
        "neg %%esi\n\t"
        : "+r" (byte_index)
        : [bound] "m" (bound)
        : "cc", "memory");

    register __m256i sub_byte_index asm("ymm1");
    register __m256i dword_index asm("ymm0");
    asm (
        "vmovd %%esi, %%xmm1\n\t"
        "vpsrld $2, %%xmm1, %%xmm0\n\t"
        : "=x" (dword_index), "=x" (sub_byte_index)
        : "r" (byte_index)
    );

    __m256i x;
    std::memcpy(&x, &x0, 32);

    register __m256i pre_result asm("ymm0") =
        _mm256_permutevar8x32_epi32(x, dword_index);

    auto shift = static_cast<uint32_t>(byte_index) & ~uint8_t{0x1c};

    asm (
        "vpcmpeqd %%xmm1, %%xmm1, %%xmm1\n\t"
        "vpinsrb  $0, %[shift], %%xmm1, %%xmm1\n\t"
        "vpshufb %%xmm1, %%xmm0, %%xmm0\n\t"
        : "+x" (pre_result)
        : [shift] "r" (shift)
        : "ymm1");
    return std::bit_cast<uint256>(pre_result);
}
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2275,35 +2265,55 @@ namespace monad::vm::compiler::native
...
+            if (!src->stack_offset()) {
+                mov_stack_elem_to_stack_offset(src);
+            }
+            if (ix->literal()) {
+                byte_literal_ix_stack_offset_src(std::move(ix), std::move(src));
+            }
+            else {
+                byte_non_literal_ix_stack_offset_src(
+                    std::move(ix), std::move(src));
+            }
```
andreaslyn: I think we can have a `byte_literal_ix_mem_src` function, working for any `x86::Mem`. Then we can read the byte from the literal from ro-data, see `rodata_.add32(...)`. The stack offset src case can also be implemented by using `byte_literal_ix_mem_src`.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        stack_.push(std::move(dst));
+    }
+
+    void
+    Emitter::byte_literal_ix_general_reg_src(StackElemRef ix, StackElemRef src)
+    {
+        MONAD_VM_DEBUG_ASSERT(ix->literal().has_value());
+        MONAD_VM_DEBUG_ASSERT(src->general_reg().has_value());
+
+        auto dst = release_general_reg(std::move(src), std::make_tuple(ix));
```
andreaslyn: I think that we do not generally want to release the general reg from src. I think `alloc_or_release_general_reg` is better here. It implemented in this PR https://github.com/category-labs/monad/pull/1430/files#diff-334b54aac2e8dbf4d22c629d7bdbbb257fa45dbcc93a92e2a1ce06031117821dR1124

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+
+        auto const &i = ix->literal()->value;
+        MONAD_VM_DEBUG_ASSERT(i < 32);
+        uint64_t const byte_index = 31 - i[0];
+        uint64_t const word_index = byte_index >> 3;
+        if (word_index != 0) {
+            as_.mov(dst_gpq[0], dst_gpq[word_index]);
+        }
+        as_.mov(x86::rax.r32(), (byte_index & 7) << 3);
+        as_.shrx(dst_gpq[0], dst_gpq[0], x86::rax);
```
andreaslyn: Shift is only necessary if `byte_index & 7` is non-zero. Use `shr` instead of `shlx` when shifting by a constant

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        auto dst = release_general_reg(std::move(src), std::make_tuple(ix));
+        Gpq256 const &dst_gpq = general_reg_to_gpq256(*dst->general_reg());
+
+        auto const &i = ix->literal()->value;
+        MONAD_VM_DEBUG_ASSERT(i < 32);
+        uint64_t const byte_index = 31 - i[0];
+        uint64_t const word_index = byte_index >> 3;
+        if (word_index != 0) {
+            as_.mov(dst_gpq[0], dst_gpq[word_index]);
+        }
```
andreaslyn:
```suggestion
        Gpq256 &dst_gpq = general_reg_to_gpq256(*dst->general_reg());

        auto const &i = ix->literal()->value;
        MONAD_VM_DEBUG_ASSERT(i < 32);
        uint64_t const byte_index = 31 - i[0];
        uint64_t const word_index = byte_index >> 3;
        std::swap(dst_gpq[0], dst_gpq[word_index])
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        as_.mov(x86::rax.r32(), (byte_index & 7) << 3);
+        as_.shrx(dst_gpq[0], dst_gpq[0], x86::rax);
+        as_.movzx(dst_gpq[0].r32(), dst_gpq[0].r8Lo());
```
andreaslyn: Use `mov` instead of `movzx` if `byte_index & 7 == 7`

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        as_.movzx(dst_gpq[0].r32(), src_mem);
+        // if ix > 31 (i.e. byte_check_bounds_compute_idx set ZF=0), we copy
+        // 0 into the dst[0] register
+        as_.cmovnz(dst_gpq[0], zero_reg);
+
+        stack_.push(std::move(dst));
+    }
+
+    void
+    Emitter::byte_literal_ix_general_reg_src(StackElemRef ix, StackElemRef src)
```
andreaslyn, line 3818: You can add the comments to `byte_literal_ix_general_reg_src` as TODOs, if you prefer to implement these optimizations in a follow up PR.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        stack_.push(std::move(dst));
+    }
+
+    void Emitter::byte_non_literal_ix_general_reg_src(
+        StackElemRef ix, StackElemRef src)
+    {
+        MONAD_VM_DEBUG_ASSERT(!ix->literal().has_value());
+        MONAD_VM_DEBUG_ASSERT(src->general_reg().has_value());
+
+        auto dst = release_general_reg(std::move(src), std::make_tuple(ix));
```
andreaslyn: I think that `alloc_or_release_general_reg` from the PR is more suitable https://github.com/category-labs/monad/pull/1430

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3674,312 @@ namespace monad::vm::compiler::native
...
+        Gpq256 const &dst_gpq = general_reg_to_gpq256(*dst->general_reg());
+
+        x86::Gpq ix0;
+        if (ix->general_reg()) {
+            Gpq256 const &ix_gpq = general_reg_to_gpq256(*ix->general_reg());
+            ix0 = ix_gpq[0];
+        }
+        else if (ix->stack_offset()) {
+            //  as_.int3();
```
andreaslyn:
```suggestion
```

category/vm/test/unit/emitter_tests.cpp
```diff
@@ -2397,10 +2397,46 @@ TEST(Emitter, byte)
...
+         idx = idx >> 1) {
+        pure_bin_instr_test(
+            rt, BYTE, &Emitter::byte, {idx, 0, 0, 0}, {-1, -1, -1, -1}, 0);
+        pure_bin_instr_test(
+            rt, BYTE, &Emitter::byte, {0, idx, 0, 0}, {-1, -1, -1, -1}, 0);
+        pure_bin_instr_test(
+            rt, BYTE, &Emitter::byte, {0, 0, idx, 0}, {-1, -1, -1, -1}, 0);
+        pure_bin_instr_test(
+            rt, BYTE, &Emitter::byte, {0, 0, 0, idx}, {-1, -1, -1, -1}, 0);
+    }
```
andreaslyn: Copy paste the new shl test from this PR https://github.com/category-labs/monad/pull/1430 instead

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: see `byte_non_literal_ix_literal_or_stack_offset_src`

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,342 @@ namespace monad::vm::compiler::native
...
+            src_mem.setIndex(dst_gpq[0]);
+            as_.movzx(dst_gpq[0], src_mem);
+        }
+        else {
+            // asmjit cannot emit a [base_label + index_register + offset]
+            // memory operand, so in the case of src being in rodata, we move
+            // the address of the literal to rax and then emmit a [base_register
+            // + index_register] version of movzx
+            as_.lea(x86::rax, rodata_.add32(src->literal()->value));
+            as_.movzx(dst_gpq[0], x86::byte_ptr(x86::rax, dst_gpq[0]));
```
goodlyrottenapple: not sure if this is optimal?

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,342 @@ namespace monad::vm::compiler::native
...
+            src_mem.setSize(1);
+
+            src_mem.setIndex(dst_gpq[0]);
+            as_.movzx(dst_gpq[0], src_mem);
+        }
+        else {
+            // asmjit cannot emit a [base_label + index_register + offset]
+            // memory operand, so in the case of src being in rodata, we move
+            // the address of the literal to rax and then emmit a [base_register
+            // + index_register] version of movzx
```
andreaslyn: This an x86 restriction

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,342 @@ namespace monad::vm::compiler::native
...
+            alloc_or_release_general_reg(std::move(src), std::make_tuple(ix));
+        Gpq256 const &dst_gpq = general_reg_to_gpq256(*dst->general_reg());
+
+        auto const &i = ix->literal()->value;
+        MONAD_VM_DEBUG_ASSERT(i < 32);
+        uint64_t const byte_index = 31 - i[0];
+        uint64_t const word_index = byte_index >> 3;
+        if (src_gpq[0] != dst_gpq[0] || word_index != 0) {
+            as_.mov(dst_gpq[0], src_gpq[word_index]);
+        };
```
andreaslyn: I think you forgot this
```suggestion
        std::swap(dst_gpq[0], src_gpq[word_index]);
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,342 @@ namespace monad::vm::compiler::native
...
+        StackElemRef ix, StackElemRef src_orig)
+    {
+        MONAD_VM_DEBUG_ASSERT(!ix->literal().has_value());
+        MONAD_VM_DEBUG_ASSERT(src_orig->avx_reg().has_value());
+        auto [scratch, scratch_reserv] = alloc_avx_reg();
+        auto scratch_ymm = avx_reg_to_ymm(*scratch->avx_reg());
+
+        auto [dst, src] = [&] {
+            // ix must be on the stack or in a general register at this point
+            // so we don't have to include it in the liveness check here
```
andreaslyn: I think that `ix` can be in avx reg location here

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: Is that sound now that we use `alloc_or_release_general_reg`? What if `src` is still live?

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: ah good spot, this invariant is no longer true

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: thnx will update the comment. wasn't sure if there was a different trick to it.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
andreaslyn: Right, we can instead do
```suggestion
        if (&src_gpq == &dst_gpq) {
            std::swap(dst_gpq[0], dst_gpq[word_index]);
        }
        else if (word_index != 0) {
            as_.mov(dst_gpq[0], src_gpq[word_index]);
        }
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,340 @@ namespace monad::vm::compiler::native
...
+        // set rax to 0xffffffff if any of the upper 192 bits of ix are set or
+        // ix[0] > 31
+        as_.cmovnz(x86::rax, rodata_.add8(32));
+        as_.sub(x86::rax, 31);
+        as_.cmova(x86::rax, rodata_.add8(1));
+        as_.neg(x86::rax);
+
+        // copy the double word from src containing the index byte
+        // into dst_ymm
+        as_.vmovd(scratch_ymm.xmm(), x86::rax);
```
andreaslyn:
```suggestion
        as_.cmova(x86::eax, rodata_.add4(1));
        as_.neg(x86::eax);

        as_.vmovd(scratch_ymm.xmm(), x86::eax);
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,340 @@ namespace monad::vm::compiler::native
...
+        stack_.push(std::move(dst));
+    }
+
+    void Emitter::byte_non_literal_ix_avx_reg_src(
+        StackElemRef ix, StackElemRef src_orig)
+    {
+        MONAD_VM_DEBUG_ASSERT(!ix->literal().has_value());
+        MONAD_VM_DEBUG_ASSERT(src_orig->avx_reg().has_value());
+        auto [scratch, scratch_reserv] = alloc_avx_reg();
+        auto scratch_ymm = avx_reg_to_ymm(*scratch->avx_reg());
```
andreaslyn: Allocate destination before scratch, so destination is likely to have the lower avx reg

category/vm/test/unit/emitter_tests.cpp
```diff
@@ -2397,10 +2397,50 @@ TEST(Emitter, byte)
...
+            i,
+            i | (uint256_t{1} << 65),
+            i | (uint256_t{1} << 128),
+            i | (uint256_t{1} << 224),
+            i | (uint256_t{1} << 255)};
+        for (auto const &i : indices) {
+            pure_bin_instr_test(
+                rt, BYTE, &Emitter::byte, i, value, runtime::byte(i, value));
+            pure_bin_instr_test(
+                rt, SAR, &Emitter::byte, i, ~value, runtime::byte(i, ~value));
```
andreaslyn:
```suggestion
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3686,340 @@ namespace monad::vm::compiler::native
...
+
+        // setting scratch_ymm to all 1s will set all dst_ymm bytes to 0
+        // when used as vpshufb control mask
+        as_.vpcmpeqd(scratch_ymm.xmm(), scratch_ymm.xmm(), scratch_ymm.xmm());
+        // if ix <= 31, rax & 0xffffffe3 is equivalent to rax % 4
+        // which will be used to move the correct sub-byte of dst_ymm
+        // to 0th position
+        // if ix > 31, 0xe3 will be inserted instead and since the MSB of 0xe3
+        // is 1, vpshufb will zero out the lowest byte of dst_ymm
+        as_.and_(x86::rax, 0xffffffe3);
```
andreaslyn:
```suggestion
        // if ix <= 31, eax & 0xe3 is equivalent to eax % 4
        // which will be used to move the correct sub-byte of dst_ymm
        // to 0th position
        // if ix > 31, eax & 0xe3 == 0x3e will be inserted instead and since
        // the MSB of 0xe3 is 1, vpshufb will zero the lowest byte of dst_ymm
        as_.and_(x86::eax, 0xe3);
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3726,6 +3701,322 @@ namespace monad::vm::compiler::native
         }
     }
 
+    // Sets zero flag according to whether the upper 192 bits of `e` are set.
+    template <typename... LiveSet>
+    void Emitter::check_stack_elem_top_192_bits_empty(
```
andreaslyn: This function is now merged into vm-develop. It is called `test_high_bits192`.

PR #1433, opened 2025-07-28
Optimize shifts

andreaslyn:


category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -1119,6 +1119,22 @@ namespace monad::vm::compiler::native
...
+        StackElemRef elem, std::tuple<LiveSet...> const &live)
+    {
+        if (is_live(elem, live)) {
+            if (stack_.has_free_general_reg() ||
+                (!elem->stack_offset() && !elem->avx_reg() &&
+                 !elem->literal())) {
+                return alloc_general_reg();
+            }
+        }
+        auto r = stack_.release_general_reg(std::move(elem));
```
dhil, line 1134: I believe the `release_general_reg` operation is only defined when the `elem` is held in a general register. So, I suggest we debug assert it.
```suggestion
        MONAD_VM_DEBUG_ASSERT(elem.general_reg().has_value());
        auto r = stack_.release_general_reg(std::move(elem));
```

dhil: Or maybe this ought to be asserted in `move_general_reg` in `virtual_stack.cpp` instead.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -4110,135 +4130,342 @@ namespace monad::vm::compiler::native
...
-            std::move(value), 32, std::tuple_cat(std::make_tuple(shift), live));
-
-        auto [dst, dst_reserv] = alloc_general_reg();
-        Gpq256 &dst_gpq = general_reg_to_gpq256(*dst->general_reg());
+        if (value->avx_reg()) {
+            return shift_avx_reg_by_non_literal<shift_type>(
+                std::move(shift), std::move(value), live);
+        }
+        else if (value->literal()) {
+            mov_literal_to_avx_reg(value);
```
dhil, line 4140: I suppose it is possible that we may need to spill an avx register here, whilst there may be a general purpose register available. Perhaps as a future todo; does it make sense to check here whether we can avoid spilling by checking whether we have enough free general register to hold the literal? Or perhaps this becomes a non-issue when the shift by literal algorithms are implemented?

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -4727,6 +4954,44 @@ namespace monad::vm::compiler::native
         as_.adc(gpq[3], 0);
     }
 
+    // Will not mutate the lower 64 bits, even when `elem` is not live.
+    template <typename... LiveSet>
+    void Emitter::test_high_bits192(
+        StackElemRef elem, std::tuple<LiveSet...> const &live)
+    {
+        MONAD_VM_DEBUG_ASSERT(!elem->literal());
```
dhil: Shouldn't we discharge deferred comparisons here?

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -4110,135 +4130,342 @@ namespace monad::vm::compiler::native
...
+            shift, 257, std::tuple_cat(std::make_tuple(value), live));
+        mov_stack_elem_low64_to_gpq(std::move(shift), x86::rax);
+        as_.cmovnz(x86::eax, rodata_.add4(256));
+
+        // Allocate result before temporary avx registers, so that result
+        // is likely to have lower avx reg, which better avoids spill.
+        auto [result, result_reserv] = alloc_avx_reg();
+        auto const out = avx_reg_to_ymm(*result->avx_reg());
+
+        MONAD_VM_ASSERT(in != out);
```
dhil: Minor: I am wondering why this isn't a debug assert like the assertion at the beginning of this function. I suppose if either assertion fails it is fatal. 

Anyways, if this assertion were ever to trigger, then there would be something really wrong with the register allocation and spilling procedure. Looking at that code of that procedure, it seems to me that this assertion can never trigger because of reserve on line 4313 above.

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
andreaslyn: There is a stronger assertion in `move_general_reg`
```
        MONAD_VM_ASSERT(general_reg_stack_elems_[reg.reg] == &src);
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
andreaslyn: That is not supposed to be there!

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
andreaslyn: I have added this instead
```
MONAD_VM_DEBUG_ASSERT(!stack_.has_deferred_comparison());
```

category/vm/libs/compiler/src/monad/vm/compiler/ir/x86/emitter.cpp
andreaslyn: Unlikely that we will run out of avx registers. I am not worried about that.

dhil: OK fine. Nonetheless, I'd be interested in an example that triggers this pathological case. I may look into it when I have some spare cycles.

andreaslyn: See the fuzzer generator.hpp
```c++
        // With 75% probability, use 14 of the 16 available avx
        // registers immediately, to increase probability of running
        // out of avx registers.
        with_probability(eng, 0.75, [&](auto &) {
            program.push_back(NonTerminator{CALLVALUE}); // uses 1 avx register
            program.push_back(NonTerminator{GASPRICE}); // uses 1 avx register
            // Use 12 more avx registers:
            for (int i = 0; i < 12; ++i) {
                // [PREV, CALLVALUE, ...]
                program.push_back(NonTerminator{DUP2});
                // [CALLVALUE, PREV, CALLVALUE, ...]
                program.push_back(NonTerminator{DUP2});
                // [PREV, CALLVALUE, PREV, CALLVALUE, ...]
                program.push_back(NonTerminator{AND});
                // [PREV & CALLVALUE, PREV, CALLVALUE, ...]
                program.push_back(NonTerminator{SWAP1});
                // [PREV, PREV & CALLVALUE, CALLVALUE, ...]
                program.push_back(NonTerminator{SWAP2});
                // [CALLVALUE, PREV & CALLVALUE, PREV, ...]
                program.push_back(NonTerminator{SWAP1});
                // [PREV & CALLVALUE, CALLVALUE, PREV, ...]
            }
        });
```

PR #1440, opened 2025-07-30
Fix `CREATE` initcode size bug

Baltoli:


category/vm/libs/runtime/src/monad/vm/runtime/create.hpp
```diff
@@ -39,7 +39,8 @@ namespace monad::vm::runtime
         }
 
         if constexpr (Rev >= EVMC_SHANGHAI) {
-            if (MONAD_VM_UNLIKELY(*size > 0xC000)) {
+            if (MONAD_VM_UNLIKELY(
+                    *size > ctx->chain_params.max_initcode_size())) {
```
andreaslyn: Let us avoid the clutter and keep init code size hardcoded, and instead update evmone. Remember runtime unit tests for the init code size bounds `MAX_INITCODE_SIZE` and `MAX_INITCODE_SIZE+1`, because that is not covered by the fuzzer.

andreaslyn: Note that we already have `MAX_INITCODE_SIZE` in evmone constants.hpp, so perhaps we do not even need to modify evmone.

Baltoli: The underlying problem is that we can't hard-code the max code size inside the VM, because it's a dynamic property of the active chain (in Monad, but not in Ethereum):

```cpp
size_t MonadChain::get_max_code_size(
    uint64_t const block_number, uint64_t const timestamp) const
{
    auto const monad_rev = get_monad_revision(block_number, timestamp);
    if (MONAD_LIKELY(monad_rev >= MONAD_TWO)) {
        return 128 * 1024;
    }
    else if (monad_rev >= MONAD_ZERO) {
        return MAX_CODE_SIZE_EIP170;
    }
    else {
        MONAD_ABORT("invalid revision");
    }
}
```

andreaslyn:
```c++
#ifdef MONAD_COMPILER_TESTING

inline const auto DEBUG_TRACE_ENV =
    std::getenv("EVMONE_DEBUG_TRACE");
inline const bool DEBUG_TRACE_ENABLED =
    DEBUG_TRACE_ENV &&
    std::strcmp(DEBUG_TRACE_ENV, "1") == 0;

inline const auto MONAD_COMPILER_FUZZING_ENV =
    std::getenv("MONAD_COMPILER_FUZZING");
inline const bool MONAD_COMPILER_FUZZING =
    MONAD_COMPILER_FUZZING_ENV &&
    std::strcmp(MONAD_COMPILER_FUZZING_ENV, "1") == 0;

inline const bool SAVE_EVM_STACK =
    MONAD_COMPILER_FUZZING;

inline const auto MAX_CODE_SIZE =
    MONAD_COMPILER_FUZZING
    ? MONAD_MAX_CODE_SIZE
    : ETHEREUM_MAX_CODE_SIZE;

/// The limit of the size of init codes for contract creation
/// defined by [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860)
inline const auto MAX_INITCODE_SIZE = 2 * MAX_CODE_SIZE;

#else

constexpr bool DEBUG_TRACE_ENABLED = false;
constexpr bool SAVE_EVM_STACK = false;
constexpr auto MAX_CODE_SIZE = ETHEREUM_MAX_CODE_SIZE;

/// The limit of the size of init codes for contract creation
/// defined by [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860)
constexpr auto MAX_INITCODE_SIZE = 2 * MAX_CODE_SIZE;

#endif
```

andreaslyn: The problem is missing parameterization in monad revision

andreaslyn:
```
MAX_CODE_SIZE<MONAD_ZERO> = ETHEREUM_MAX_CODE_SIZE;
MAX_CODE_SIZE<MONAD_ONE> = MAX_CODE_SIZE<MONAD_ZERO>;
MAX_CODE_SIZE<MONAD_TWO> = 128 * 1024;

MAX_INIT_CODE_SIZE<mrev> = 2 * MAX_CODE_SIZE<mrev>;

Comment (andreaslyn) [reply category/vm/libs/runtime/src/monad/vm/runtime/create.hpp]: Since we do not have the monad revision available, I think passing chain config in context is fine. Note that max code size in test_vm.cpp should depend on whether fuzzer is enabled. See constants.hpp in evmone. Will need to change evmone to use `MAX_INITCODE_SIZE` instead of hardcoded 0xc000.

File: category/vm/libs/runtime/src/monad/vm/runtime/types.hpp [hunk 2]
```diff
@@ -151,15 +151,28 @@ namespace monad::vm::runtime
         }
     };
 
+    struct ChainParams
+    {
+        std::size_t max_code_size;
+
+        constexpr std::size_t max_initcode_size() const noexcept
```
andreaslyn: The max initcode size should be a separate parameter, because it is currently 0xc000, which is not equal to 2*max_code_size.

category/vm/test/unit/runtime/create_tests.cpp
```diff
@@ -117,3 +117,31 @@ TEST_F(RuntimeTest, Create2Constantinople)
     ASSERT_EQ(ctx_.gas_remaining, 915624);
     ASSERT_EQ(ctx_.gas_refund, 10);
 }
+
+TEST_F(RuntimeTest, CreateMaxCodeSize)
```
andreaslyn: I think we are missing a test for the out of bound case 128*1024+1

Baltoli: These tests can't test errors: they will try to call `ctx_.exit(...)` which crashes as there isn't a landing pad set up.

Baltoli: We could probably fix the tests to allow that but it's a bigger change than this PR

category/vm/libs/runtime/src/monad/vm/runtime/types.hpp
Baltoli: When is the max initcode size *not* double the max code size?

Baltoli: When max initcode size is 0xC000 in EIP-3860; max code size is 0x6000

andreaslyn: Right now, max code size is `128*1024` and max init code size is 0xc000, which is not equal to `2*128*1024`

andreaslyn: On the monad blockchain

Baltoli: 🤦 right, we broke the intended semantics. Point taken, thanks

category/vm/test/unit/runtime/create_tests.cpp
andreaslyn: If the fuzzer is reaching this case then we do not need to unit test it.

andreaslyn: Looks like the fuzzer will actually hit this error case. Can happen when generating trivial create instructions, I think

category/execution/ethereum/chain/ethereum_mainnet.cpp
```diff
@@ -159,6 +159,14 @@ size_t EthereumMainnet::get_max_code_size(
                : std::numeric_limits<size_t>::max();
 }
 
+size_t EthereumMainnet::get_max_initcode_size(
+    uint64_t const block_number, uint64_t const timestamp) const
+{
+    return get_revision(block_number, timestamp) >= EVMC_SHANGHAI
+               ? MAX_INITCODE_SIZE_EIP3860
+               : std::numeric_limits<size_t>::max();
```
andreaslyn, line 167: Looks like this is not the right init code size for revisions before shanghai

andreaslyn: But perhaps will work in practice for some reason?

jhunsaker: it was simply limited by gas before?

> Currently, on London, with 30M gas limit, it would be possible to trigger jumpdest-analysis of a total ~1.3GB of initcode.

jhunsaker: Previous code had no limit for prior to shanghai

```
        if constexpr (Rev >= EVMC_SHANGHAI) {
            if (MONAD_VM_UNLIKELY(*size > 0xC000)) {
                ctx->exit(StatusCode::OutOfGas);
            }
        }
```

andreaslyn: Right, was not entirely sure about adding a limit. But I guess it will run out of gas when init code size is larger than size_t max, as pointed out earlier.

PR #1442, opened 2025-07-30
[compiler] Log asm produced in the emitter tests

laurenthuberdeau:
## Context

I often want to see the code that caused some emitter test to fail. This requires setting the `asm_log_path` parameter on the `CompilerConfig` which can be annoying, especially when you want to see the code of multiple tests.

This PR sets the `asm_log_path` parameter by default such that every test case get its own log file. On failure, the path of the log file is printed to let the user know about the existence of this debug file like so:

```
/home/lhuberdeau/monad/category/vm/test/unit/emitter_tests.cpp:244: Failure
Expected equality of these values:
  uint256_t::load_le(ret.size)
    Which is: 32-byte object <AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA FF-FF FF-FF FF-FF>
  result
    Which is: 32-byte object <AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA AA-AA FF-FF FF-FF>

See disassembly at:
/tmp/monad_vm_test_logs/emitter_test_0.s
```

A few things I wasn't sure about:
- ~The path can be pretty long, I thought about hashing it to make it shorter but then it would lose any meaning.~
- ~The log files are written to `/tmp/` but maybe writing them to the `build` directory would make finding those files easier.~
- ~On my machine, it makes the execution time jump of `vm-unit-tests` from 9s to 25s. That's probably too long, but maybe it's worth if we find it useful enough. Maybe we want to condition the logging on a flag?~

### DIFF
--- a/test/vm/unit/emitter_tests.cpp
+++ b/test/vm/unit/emitter_tests.cpp
@@ -25,6 +25,8 @@
 #include <category/vm/runtime/types.hpp>
 #include <category/vm/runtime/uint256.hpp>
 
+#include "test_params.hpp"
+
 #include <asmjit/core/globals.h>
 #include <asmjit/core/jitruntime.h>
 
@@ -51,6 +53,58 @@ using namespace monad::vm::runtime;
 
 namespace
 {
+    static int test_emitter_ix = 0;
+
+    std::string new_emitter_asm_log_path()
+    {
+        test_emitter_ix++;
+        return std::string(std::format(
+            "/tmp/monad_vm_test_logs/emitter_test_{}.s", test_emitter_ix));
+    }
+
+    struct TestEmitter : Emitter
+    {
+
+        std::string log_path_storage_;
+
+        CompilerConfig add_asm_log_path(CompilerConfig c, std::string log_path)
+        {
+            if (!c.asm_log_path &&
+                monad::vm::compiler::test::params.dump_asm_on_failure) {
+                c.asm_log_path = log_path.c_str();
+            }
+            return c;
+        }
+
+        // Default log path, used if c.asm_log_path is not set. The only reason
+        // to have this constructor is so that the lifetime of the
+        // new_emitter_asm_log_path extends to after the Emitter constructor.
+        TestEmitter(
+            asmjit::JitRuntime const &rt, code_size_t bytecode_size,
+            CompilerConfig const &c = {},
+            std::string const &log_path = new_emitter_asm_log_path())
+            : Emitter(rt, bytecode_size, add_asm_log_path(c, log_path))
+            , log_path_storage_(log_path)
+        {
+        }
+
+        // Override finish_contract to flush debug_logger_'s file handle
+        entrypoint_t finish_contract(asmjit::JitRuntime &rt)
+        {
+            auto entrypoint = Emitter::finish_contract(rt);
+
+            // Flush the debug logger in case the code segfaults before the
+            // Emitter destructor is called.
+            flush_debug_logger();
+            if (monad::vm::compiler::test::params.dump_asm_on_failure) {
+                std::cout << "See disassembly at:\n  " << log_path_storage_
+                          << std::endl;
+            }
+
+            return entrypoint;
+        }
+    };
+
     evmc::address max_address()
     {
         evmc::address ret;
@@ -214,7 +268,7 @@ namespace
                     right, Emitter::location_type_to_string(right_loc)) << std::endl;
 #endif
 
-        Emitter emit{rt, ir.codesize};
+        TestEmitter emit{rt, ir.codesize};
         (void)emit.begin_new_block(ir.blocks()[0]);
         emit.push(right);
         if (dup) {
@@ -276,7 +330,7 @@ namespace
             << std::endl;
 #endif
 
-        Emitter emit{rt, ir.codesize};
+        TestEmitter emit{rt, ir.codesize};
         (void)emit.begin_new_block(ir.blocks()[0]);
         emit.push(input);
         if (dup) {
@@ -427,7 +481,7 @@ namespace
                        {PUSH0, PUSH0, PUSH0, JUMP, JUMPDEST, RETURN});
 
         asmjit::JitRuntime rt;
-        Emitter emit{rt, ir.codesize};
+        TestEmitter emit{rt, ir.codesize};
 
         for (auto const &[k, _] : ir.jump_dests()) {
             emit.add_jump_dest(k);
@@ -537,7 +591,7 @@ namespace
         auto ir =
             get_jumpi_ir(deferred_comparison, swap, dup, jumpdest_fallthrough);
 
-        Emitter emit{rt, ir.codesize};
+        TestEmitter emit{rt, ir.codesize};
 
         for (auto const &[k, _] : ir.jump_dests()) {
             emit.add_jump_dest(k);
@@ -654,7 +708,7 @@ namespace
              RETURN});
 
         asmjit::JitRuntime rt;
-        Emitter emit{rt, ir.codesize};
+        TestEmitter emit{rt, ir.codesize};
 
         (void)emit.begin_new_block(ir.blocks()[0]);
         emit.push(1);
@@ -724,7 +778,7 @@ namespace
 TEST(Emitter, empty)
 {
     asmjit::JitRuntime rt;
-    Emitter emit{rt, code_size_t{}};
+    TestEmitter emit{rt, code_size_t{}};
 
     entrypoint_t entry = emit.finish_contract(rt);
     auto ctx = test_context();
@@ -740,7 +794,7 @@ TEST(Emitter, empty)
 TEST(Emitter, stop)
 {
     asmjit::JitRuntime rt;
-    Emitter emit{rt, bin<1>};
+    TestEmitter emit{rt, bin<1>};
     emit.stop();
 
     entrypoint_t entry = emit.finish_contract(rt);
@@ -755,7 +809,7 @@ TEST(Emitter, stop)
 TEST(Emitter, fail_with_error)
 {
     asmjit::JitRuntime const rt;
-    Emitter emit{rt, bin<1>};
+    TestEmitter emit{rt, bin<1>};
     // Test that asmjit error handler is in place:
     EXPECT_THROW(emit.fail_with_error(asmjit::kErrorOk), Emitter::Error);
 }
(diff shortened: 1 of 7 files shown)


category/vm/test/unit/emitter_tests.cpp
```diff
@@ -180,11 +180,13 @@ namespace
...
     void pure_bin_instr_test_instance(
-        asmjit::JitRuntime &rt, PureEmitterInstr instr, uint256_t const &left,
-        Emitter::LocationType left_loc, uint256_t const &right,
-        Emitter::LocationType right_loc, uint256_t const &result,
-        basic_blocks::BasicBlocksIR const &ir, bool dup)
+        bool &test_success, asmjit::JitRuntime &rt, PureEmitterInstr instr,
+        uint256_t const &left, Emitter::LocationType left_loc,
+        uint256_t const &right, Emitter::LocationType right_loc,
+        uint256_t const &result, basic_blocks::BasicBlocksIR const &ir,
+        bool dup, char const *asm_log_path)
```
andreaslyn: Consider changing this to
```suggestion
        bool dup, bool dump_asm_log)
```

category/vm/test/unit/emitter_tests.cpp
```diff
@@ -180,11 +180,13 @@ namespace
...
     {
+        test_success = false;
```
andreaslyn: I think that this is not necessary

category/vm/test/unit/emitter_tests.cpp
```diff
@@ -197,7 +199,7 @@ namespace
                     right, Emitter::location_type_to_string(right_loc)) << std::endl;
 #endif
 
-        Emitter emit{rt, ir.codesize};
+        Emitter emit{rt, ir.codesize, {.asm_log_path = asm_log_path}};
```
andreaslyn: Have `asm_log_path` be non-empty string iff `dump_asm_log` is true.
Make sure the `emit` object goes out of scope immediately after calling `emit.finish_contract`.
Print the asm log path when the `emit` object is out of scope iff `dump_asm_log` is true.

category/vm/test/unit/emitter_tests.cpp
```diff
@@ -296,15 +308,37 @@ namespace
...
+                .name,
+            Emitter::location_type_to_string(left_loc),
+            Emitter::location_type_to_string(right_loc),
+            left,
+            right);
+    }
+
     void pure_bin_instr_test(
         asmjit::JitRuntime &rt, EvmOpCode opcode, PureEmitterInstr instr,
         uint256_t const &left, uint256_t const &right, uint256_t const &result)
```
andreaslyn: A new argument `bool dump_asm_log = false` would allow enabling this selectively during development.

category/vm/test/unit/emitter_tests.cpp
laurenthuberdeau: We still need to compute the path when `!success` so I prefer to compute keep that logic out in `pure_bin_instr_test`.

category/vm/test/unit/emitter_tests.cpp
laurenthuberdeau: Added a CLI argument instead

category/vm/test/unit/emitter_tests.cpp
andreaslyn: Just print the path regardless of success

category/vm/test/unit/emitter_tests.cpp
andreaslyn: A cli argument requires this solution be complete. All tests should dump the assembly in this case. Currently most tests do not dump.

category/vm/test/unit/emitter_tests.cpp
laurenthuberdeau: I worry this will make the output of the tests very noisy, since `pure_bin_instr_test` tries with all possible operands combinations (32 for each `pure_bin_instr_test` calls). I think keeping the existing behavior of no output on success is preferable.

category/vm/test/unit/emitter_tests.cpp
laurenthuberdeau: Good point, I'll review the other tests to make sure they write their assembly.

category/vm/test/unit/emitter_tests.cpp
andreaslyn: That is fair. The reason I am trying to change this is because it will not work in general. It fails to print the path at least when the native execution segfaults.

laurenthuberdeau: Ah I hadn't considered this case

category/vm/test/unit/emitter_tests.cpp
laurenthuberdeau: All emitter tests should now log their assembly.

test/vm/unit/emitter_tests.cpp
```diff
@@ -32,8 +34,69 @@ using namespace monad::vm::compiler;
...
+// Helper macros to make the tests more readable.
+// GoogleTests macros return a stream that can be used to append
+// additional information to the assertion failure message.
+#define ASM_LOG_PATH_MSG                                                       \
+    (monad::vm::compiler::test::params.dump_asm_on_failure                     \
+         ? std::format(                                                        \
+               "See disassembly at:\n  {}", current_emitter_asm_log_path())    \
+         : std::string())
+#define ASSERT(cond, ...)                                                      \
+    ASSERT_##cond(__VA_ARGS__) << ASM_LOG_PATH_MSG << std::endl
```
andreaslyn: Not convinced that this is worth it, just to be able to print the path.

laurenthuberdeau: Is the concern that we'll forget to use this macro and have a mix of `ASSERT` and `ASSERT_{pred}` macros in the test file? If so, I suppose we could redefine GTests's macros to include the path by default, but that's a much bigger chance.

andreaslyn: My concern is rooted in the introduction of unnecessary complexity to critical testing infrastructure. A new assert functionality which introduces a new state dependency is not trivial. Compare this to printing the path as part of `finish_contract`, which would entirely prevent this complexity.

laurenthuberdeau: Fair, I can see that this change has a much larger footprint than we initially expected. Printing the path in `finish_contract` is just too noisy to be useful in my opinion as many tests end up generating hundreds if not thousands of assembly log files like in the samples below.

I don't feel strongly for this feature, so I don't mind closing the PR if we think it's not worth the extra complexity.

<details>

<summary>A few test outputs</summary>

```
lhuberdeau@devnet22:~/monad$ ./build/test/vm/unit/vm-unit-tests --gtest_filter=Emitter.signextend --dump-asm
Note: Google Test filter = Emitter.signextend
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Emitter
[ RUN      ] Emitter.signextend
See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_1.s
See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_2.s

[...] <- 4668 more lines!

See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_2336.s
[       OK ] Emitter.signextend (906 ms)
[----------] 1 test from Emitter (906 ms total)

lhuberdeau@devnet22:~/monad$ ./build/test/vm/unit/vm-unit-tests --gtest_filter=Emitter.byte --dump-asm
Note: Google Test filter = Emitter.byte
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Emitter
[ RUN      ] Emitter.byte
See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_1.s
See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_2.s

[...] <- 11068 more lines!

See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_5536.s
[       OK ] Emitter.byte (1991 ms)
[----------] 1 test from Emitter (1991 ms total)

lhuberdeau@devnet22:~/monad$ ./build/test/vm/unit/vm-unit-tests --gtest_filter=Emitter.eq --dump-asm
Note: Google Test filter = Emitter.eq
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Emitter
[ RUN      ] Emitter.eq
See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_1.s
See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_2.s

[...] <- 892 more lines!

See disassembly at:
  /tmp/monad_vm_test_logs/emitter_test_448.s
[       OK ] Emitter.eq (164 ms)
[----------] 1 test from Emitter (164 ms total)
```

</details>

andreaslyn: This is presumably only enabled in case there is a failing test. In which case one can filter out the failing test. Then right over rhe gtest error message we have the path.

andreaslyn: --gtest_filter=failingtest

andreaslyn:
```
$ ./build/test/vm/unit/vm-unit-tests --gtest_filter=failingtest --dump-asm | grep ": Failure" -B2
```

laurenthuberdeau: It now writes the path to stdout in `finish_contract`.

test/vm/unit/main.cpp
```diff
@@ -0,0 +1,27 @@
...
+    CLI::App app{"Monad VM unit tests", "vm-unit-tests"};
+    app.add_flag(
+        "--dump-asm",
+        monad::vm::compiler::test::params.dump_asm_on_failure,
+        "Save assembly on failure");
+    CLI11_PARSE(app, argc, argv);
+
+    // Create test log directory
+    std::filesystem::path test_log_dir = "/tmp/monad_vm_test_logs";
+    if (!std::filesystem::exists(test_log_dir)) {
```
andreaslyn:
```suggestion
    bool const needs_test_logs =
        monad::vm::compiler::test::params.dump_asm_on_failure;
    if (needs_test_logs && !std::filesystem::exists(test_log_dir)) {
```

laurenthuberdeau: Good catch!

PR #1443, opened 2025-07-30
Prestate tracing

dhil:
Rebase of #1304 + compilation error fixes.

### DIFF
--- a/category/execution/ethereum/db/test/test_db.cpp
+++ b/category/execution/ethereum/db/test/test_db.cpp
@@ -77,8 +77,7 @@ namespace
             .value();
     // clang-format on
     auto const CA1_CODE_HASH = to_bytes(keccak256(CA1_CODE));
-    auto const CA1_CODE_ICODE =
-        vm::make_shared_intercode(CA1_CODE);
+    auto const CA1_CODE_ICODE = vm::make_shared_intercode(CA1_CODE);
 
     // clang-format off
     auto const CA2_CODE = 
@@ -86,8 +85,7 @@ namespace
             .value();
     // clang-format on
     auto const CA2_CODE_HASH = to_bytes(keccak256(CA2_CODE));
-    auto const CA2_CODE_ICODE =
-        vm::make_shared_intercode(CA2_CODE);
+    auto const CA2_CODE_ICODE = vm::make_shared_intercode(CA2_CODE);
 
     constexpr auto key1 =
         0x00000000000000000000000000000000000000000000000000000000cafebabe_bytes32;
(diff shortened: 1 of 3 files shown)


category/execution/ethereum/trace/prestate_tracer.cpp
```diff
@@ -0,0 +1,293 @@
...
+        res[key_json] = value_json;
+    }
+    return res;
+}
+
+json account_to_json(std::optional<Account> const &account)
+{
+    json res = json::object();
+    if (MONAD_UNLIKELY(!account.has_value())) {
+        // If account is created, then only show 'balance = "0x0"'
```
andreaslyn, line 72:
```suggestion
        // If account is not created, then only show 'balance = "0x0"'
```

category/execution/ethereum/db/trie_db.cpp
```diff
@@ -280,27 +287,51 @@ void TrieDb::commit(
...
-                .value = chunk,
-                .incarnation = false,
-                .next = UpdateList{},
-                .version = static_cast<int64_t>(block_number_)}));
-            ++chunk_index;
-        }
+        chunk_and_upsert(call_frame_view, call_frame_updates);
+
+        // prestate traces
+        auto const &prestate_trace_json = state_to_json(pre_state_traces[i]);
```
andreaslyn, line 316: Not sure what the requirements are here, but looks like it would be an essentially free reduction of database size to switch to RLP encoding instead of JSON.

category/execution/ethereum/db/test/test_db.cpp
```diff
@@ -864,6 +931,118 @@ TYPED_TEST(DBTest, commit_call_frames)
...
+
+    for (uint64_t txn = 0; txn < NUM_TXNS; ++txn) {
+        auto const &res_json =
+            read_pre_state_traces(this->db, tdb.get_block_number(), txn);
+        ASSERT_TRUE(!res_json.empty());
+        EXPECT_EQ(res_json, expected_json);
+    }
+}
+
+TYPED_TEST(DBTest, commit_state_deltas_traces)
```
andreaslyn, line 989: I could be wrong, but looks like we are missing a test for the case where state deltas JSON string exceeds `mpt::MAX_VALUE_LEN_OF_LEAF` in size, and hence is split into multiple chunks in the DB.

category/execution/ethereum/db/trie_db.cpp
dhil: I think we will abandon this approach. In #1471 there is no DB interaction.

PR #1447, opened 2025-07-31
Bletner/llvm_optimize_pass

brett-monad:
enable LLVM O3 optimize pass
assorted cleanups

### DIFF
--- a/category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
+++ b/category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
@@ -6,6 +6,7 @@
 #include <monad/vm/runtime/call.hpp>
 #include <monad/vm/runtime/create.hpp>
 #include <monad/vm/runtime/data.hpp>
+#include <monad/vm/runtime/detail.hpp>
 #include <monad/vm/runtime/environment.hpp>
 #include <monad/vm/runtime/keccak.hpp>
 #include <monad/vm/runtime/log.hpp>
@@ -30,35 +31,6 @@ namespace monad::vm::llvm
     inline std::string term_name(Terminator term)
     {
         return std::format("{}", term);
-    }
-
-    inline bool has_ctx_param(Instruction const &instr)
-    {
-        auto oc = instr.opcode();
-        return (
-            oc == Gas || oc == Number || oc == MSize || oc == CodeSize ||
-            oc == Timestamp || oc == ReturnDataSize || oc == ChainId ||
-            oc == Difficulty || oc == BlobBaseFee || oc == BaseFee ||
-            oc == GasPrice || oc == Coinbase || oc == Address || oc == Caller ||
-            oc == Origin || oc == GasLimit || oc == SStore || oc == BlobHash ||
-            oc == BlockHash || oc == TLoad || oc == MLoad ||
-            oc == CallDataLoad || oc == SelfBalance || oc == ExtCodeHash ||
-            oc == ExtCodeSize || oc == SLoad || oc == Balance || oc == Sha3 ||
-            oc == Exp || oc == Log || oc == MStore || oc == MStore8 ||
-            oc == TStore || oc == CallDataCopy || oc == CodeCopy ||
-            oc == MCopy || oc == ReturnDataCopy || oc == Create ||
-            oc == Create2 || oc == ExtCodeCopy || oc == DelegateCall ||
-            oc == StaticCall || oc == Call || oc == CallCode ||
-            oc == CallValue || oc == CallDataSize);
-    }
-
-    inline bool has_gas_param(Instruction const &instr)
-    {
-        auto oc = instr.opcode();
-        return (
-            oc == Gas || oc == SStore || oc == Create || oc == Create2 ||
-            oc == DelegateCall || oc == StaticCall || oc == Call ||
-            oc == CallCode);
     }
 
     struct Emitter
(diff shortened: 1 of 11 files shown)


category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
```diff
@@ -335,21 +310,17 @@ namespace monad::vm::llvm
 
                 std::vector<Value *> args;
 
-                if (has_ctx_param(instr)) {
-                    args.push_back(ctx_ref);
-                }
+                args.push_back(ctx_ref);
+
+                auto *g =
+                    llvm.lit(64, static_cast<uint64_t>(base_gas_remaining));
```
Baltoli, line 316: Maybe a debug assert here that `base_gas_remaining` is non-negative?

category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
```diff
@@ -860,10 +948,99 @@ namespace monad::vm::llvm
...
+        {
+            llvm.save_insert();
+            auto [f, arg] = llvm.internal_function_definition(
+                instr_name(instr),
+                llvm.word_ty,
+                {llvm.ptr_ty(context_ty),
+                 llvm.int_ty(64),
+                 llvm.word_ty,
+                 llvm.word_ty,
+                 llvm.word_ty});
```
Baltoli, line 976: The signature for these calls seems like it's worth an abstraction:
```suggestion
                runtime_arg_types(3));
```

category/vm/libs/llvm/src/monad/vm/llvm/execute.cpp
```diff
@@ -41,7 +33,8 @@ namespace monad::vm::llvm
     };
 
     template <evmc_revision Rev>
-    std::shared_ptr<LLVMState> compile_impl(std::span<uint8_t const> code)
+    std::shared_ptr<LLVMState>
+    compile_impl(std::span<uint8_t const> code, std::string &dbg_nm)
```
Baltoli, line 37: I'd either add an overload or a defaulted optional parameter here; it's a bit neater if we don't have to think about the debug name unless we explicitly have to. Also not sure it needs to be a mutable reference?

category/vm/libs/llvm/src/monad/vm/llvm/execute.cpp
```diff
@@ -50,32 +43,23 @@ namespace monad::vm::llvm
...
-        long dbg_pid = 0;
-
-        if (dbg_dump) {
-            auto now = std::chrono::system_clock::now();
-            auto duration = now.time_since_epoch();
-            auto milliseconds =
-                std::chrono::duration_cast<std::chrono::milliseconds>(duration)
-                    .count();
-            dbg_pid = milliseconds;
+        if (dbg_nm != "") {
```
Baltoli, line 46: See above re. optional

category/vm/libs/llvm/src/monad/vm/llvm/llvm_state.hpp
```diff
@@ -187,6 +216,11 @@ namespace monad::vm::llvm
             return ir.CreateIntCast(a, int_ty(256), false, "cast_word");
         };
 
+        Value *cast_bool(Value *a)
+        {
+            return ir.CreateIntCast(a, int_ty(1), false, "cast_word");
```
Baltoli, line 221:
```suggestion
            return ir.CreateIntCast(a, int_ty(1), false, "cast_bool");
```

category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
```diff
@@ -512,13 +494,27 @@ namespace monad::vm::llvm
             return f;
         }
 
+        Function *init_stop()
+        {
+            llvm.save_insert();
```
Baltoli, line 499: Given that you're doing save / restore at function scope boundaries, a neat trick you could do here is to make a little RAII object that takes care of remembering to restore for you. Pseudo-C++:
```cpp
struct Saver {
   Saver(LLVMState& llvm) { llvm.save_insert();    }
  ~Saver()                { llvm.restore_insert(); }
};
```

category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
```diff
@@ -375,12 +346,22 @@ namespace monad::vm::llvm
...
+            llvm.call_void(check_gas_f, {ctx_ref});
+        };
+
         void emit_jump(Value *v)
         {
             if (ir.jump_dests().size() == 0) {
                 exit_(ctx_ref, StatusCode::Error);
             }
             else {
+                emit_gas_check();
```
andreaslyn, line 364: I think that this can be exploited with one large basic block, due to only performing gas check in the end.

category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
```diff
@@ -584,54 +574,120 @@ namespace monad::vm::llvm
...
+            auto *else_lbl = llvm.basic_block("else_lbl", fun);
+            llvm.insert_at(entry);
 
-            auto *pred1 = llvm.or_(pred, high_pred);
+            auto *gas_ref = context_gep(ctx_ref, context_offset_gas_remaining);
+            auto *gas = llvm.load(llvm.int_ty(64), gas_ref);
+            auto *gas_pred = llvm.slt(gas, llvm.lit(64, 0));
 
-            llvm.condbr(pred1, then_lbl, else_lbl);
+            llvm.condbr(gas_pred, then_lbl, else_lbl);
```
andreaslyn, line 685: Perhaps the performance issues with many JUMPDESTs is due to this branch having inaccurate branch weights https://llvm.org/docs/LangRef.html#branch-weights

category/vm/libs/llvm/src/monad/vm/llvm/emitter.hpp
andreaslyn: Perhaps something similar to `STATIC_GAS_THREASHOLD` in x86.cpp can be used to resolve this problem. Something like this
```c++
        int32_t const acc = *accumulated_base_gas + block_base_gas;
        if (acc >= STATIC_GAS_CHECK_THRESHOLD) {
            *accumulated_base_gas = 0;
            emit_gas_check();
        }

```

andreaslyn: Put that right after decrementing gas remaining in basic block prologue.

PR #1451, opened 2025-08-01
Relaxed merge validation of account balance

magedm:
Extracted from [PR#908](https://github.com/category-labs/monad/pull/908) focusing on relaxing balance validation.

This PR allows merge validation to succeed for transactions executed in parallel with conflicts on account balances as long as all operations on such balances were relative and not dependent on the exact balance values.

### DIFF
--- a/category/execution/ethereum/evm.cpp
+++ b/category/execution/ethereum/evm.cpp
@@ -48,17 +48,26 @@ bool sender_has_balance(State &state, evmc_message const &msg) noexcept
     auto const &account = state.recent_account(msg.sender);
     uint256_t const balance = account.has_value() ? account->balance : 0;
     auto &original_state = state.original_account_state(msg.sender);
+    // RELAXED MERGE
+    // if current balance has at least message value, then:
+    // 1. compute the amount that current balance exceeds message value
+    // 2. require that the original balance at merge time is at least the
+    // original balance used during this execution less said excess
     if (balance >= value) {
-        auto const diff = balance - value;
+        uint256_t const diff = balance - value;
         auto const &original = original_state.account_;
-        auto const original_balance =
+        uint256_t const original_balance =
             original.has_value() ? original->balance : 0;
-        if (original_balance > diff) {
-            auto const min_balance = original_balance - diff;
+        if (original_balance > diff) { // avoid underflow
+            uint256_t const min_balance =
+                original_balance -
+                diff; // original balance - current balance + value
             original_state.set_min_balance(min_balance);
         }
     }
     else {
+        // otherwise require that original balance at merge time matches
+        // original balance used during this execution exactly
         original_state.set_validate_exact_balance();
     }
     return balance >= value;
--- a/category/execution/ethereum/state2/block_state.cpp
+++ b/category/execution/ethereum/state2/block_state.cpp
@@ -141,15 +141,19 @@ bool BlockState::can_merge(State &state) const
     auto &original = state.original();
     for (auto &kv : original) {
         Address const &address = kv.first;
-        OriginalAccountState &original_state = kv.second;
-        auto const &account = original_state.account_;
-        auto const &storage = original_state.storage_;
+        OriginalAccountState &account_state = kv.second;
+        auto const &account = account_state.account_;
+        auto const &storage = account_state.storage_;
         StateDeltas::const_accessor it{};
         MONAD_ASSERT(state_->find(it, address));
-        auto const &actual = it->second.account.second;
-        if ((account != actual) &&
-            !state.try_fix_account_mismatch(address, original_state, actual)) {
-            return false;
+        if (account != it->second.account.second) {
+            // RELAXED MERGE
+            // try to fix original and current in `state` to match the block
+            // state up until this transaction
+            if (!state.try_fix_account_mismatch(
+                    address, account_state, it->second.account.second)) {
+                return false;
+            }
         }
         // TODO account.has_value()???
         for (auto const &[key, value] : storage) {
--- a/category/execution/ethereum/state3/account_state.hpp
+++ b/category/execution/ethereum/state3/account_state.hpp
@@ -101,6 +101,9 @@ public:
     }
 };
 
+// RELAXED MERGE
+// track the min original balance needed at start of transaction and if the
+// original and current balances can be adjusted
 class OriginalAccountState final : public AccountState
 {
     bool validate_exact_balance_{false};
--- a/category/execution/ethereum/state3/state.hpp
+++ b/category/execution/ethereum/state3/state.hpp
@@ -68,7 +68,7 @@ class State
 
     unsigned version_{0};
 
-    bool relaxed_validation_{false};
+    bool const relaxed_validation_{false};
 
 public:
     OriginalAccountState &original_account_state(Address const &address)
@@ -112,20 +112,13 @@ private:
     }
 
 public:
-    State(BlockState &block_state, Incarnation const incarnation)
-        : block_state_{block_state}
-        , incarnation_{incarnation}
-    {
-    }
-
     State(
         BlockState &block_state, Incarnation const incarnation,
-        Address const &sender, uint64_t const nonce)
+        bool relaxed_validation = false)
         : block_state_{block_state}
         , incarnation_{incarnation}
-        , relaxed_validation_{true}
+        , relaxed_validation_{relaxed_validation}
     {
-        set_original_nonce(sender, nonce);
     }
 
     State(State &&) = delete;
@@ -191,11 +184,6 @@ public:
         }
 
         --version_;
-    }
-
-    [[nodiscard]] bool relaxed_validation() const
-    {
-        return relaxed_validation_;
     }
 
     ////////////////////////////////////////
@@ -339,7 +327,7 @@ public:
 
         MONAD_ASSERT_THROW(
             std::numeric_limits<uint256_t>::max() - delta >=
-            account.value().balance,
+                account.value().balance,
             "balance overflow");
 
         account.value().balance += delta;
@@ -640,11 +628,15 @@ public:
         account.value().incarnation = incarnation_;
     }
 
+    // RELAXED MERGE
+    // if original and current can be adjusted to satisfy min balance, adjust
+    // both values for merge
     bool try_fix_account_mismatch(
         Address const &address, OriginalAccountState &original_state,
         std::optional<Account> const &actual)
     {
         auto &original = original_state.account_;
+        // verify original used and original found are otherwise the same
         if (is_dead(original)) {
             return false;
         }
@@ -661,15 +653,18 @@ public:
             return false;
         }
         MONAD_ASSERT(original->balance != actual->balance);
-        if (!relaxed_validation()) {
+        // is relaxed merge disabled
+        if (!relaxed_validation_) {
             return false;
         }
         if (original_state.validate_exact_balance()) {
             return false;
         }
+        // original balance does not meet min required
         if (actual->balance < original_state.min_balance()) {
             return false;
         }
+        // adjust balances
         auto it = current_.find(address);
         if (it != current_.end()) {
             MONAD_ASSERT(it->second.size() == 1);
--- a/category/execution/ethereum/execute_transaction.cpp
+++ b/category/execution/ethereum/execute_transaction.cpp
@@ -328,14 +328,18 @@ Result<evmc::Result> ExecuteTransaction<rev>::execute_impl2(State &state)
 {
     auto const sender_account = state.recent_account(sender_);
     auto const &icode = state.get_code(sender_)->intercode();
-    auto const validate_txn = [this, &state, &sender_account, &icode] {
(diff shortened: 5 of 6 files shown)


category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +581,52 @@ class State
...
+        if (original->code_hash != actual->code_hash) {
+            return false;
+        }
+        if (original->incarnation != actual->incarnation) {
+            return false;
+        }
+        if (original->nonce != actual->nonce) {
+            return false;
+        }
+        MONAD_DEBUG_ASSERT(original->balance != actual->balance);
```
0xSqualo: Is this intentional to be left in here?

category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +581,52 @@ class State
...
+        if (it != current_.end()) {
+            MONAD_ASSERT(it->second.size() == 1);
+            auto &recent_state = it->second.recent();
+            auto &recent = recent_state.account_;
+            if (!recent) {
+                return false;
+            }
+            // Modular arithmetic properly handles both cases: whether
+            // actual->balance or original->balance is larger than the other.
+            recent->balance += actual->balance - original->balance;
```
0xSqualo: Can you elaborate on this to double check? Perhaps unit tests around this area could be needed to ensure correctness in all scenarios for this.

To reduce errors in future changes in the code it could make sense to pull into the two separate branches for readability:

```c++
if (actual->balance > original->balance) {
    recent->balance += actual->balance - original->balance;
} else {
    recent->balance -= original->balance - actual->balance;
}
```

category/execution/ethereum/state2/block_state.cpp
```diff
@@ -120,15 +120,17 @@ vm::SharedVarcode BlockState::read_code(bytes32_t const &code_hash)
     }
 }
 
-bool BlockState::can_merge(State const &state)
+bool BlockState::can_merge(State &state) const
 {
     MONAD_ASSERT(state_);
-    for (auto const &[address, account_state] : state.original_) {
+    for (auto &[address, account_state] : state.original_) {
```
0xSqualo: Did `address` drop its `const`-ness? Could make sense to pull it out of the pair as a `const` reference.

category/execution/ethereum/state3/account_state.hpp
```diff
@@ -86,4 +86,45 @@ class AccountState final : public AccountSubstate
...
+
+    void set_validate_exact_balance()
+    {
+        validate_exact_balance_ = true;
+    }
+
+    void set_min_balance(uint256_t const &value)
+    {
+        MONAD_DEBUG_ASSERT(account_.has_value());
+        MONAD_DEBUG_ASSERT(account_->balance >= value);
```
0xSqualo: Should these be regular asserts?

category/execution/ethereum/evm.cpp
```diff
@@ -32,8 +32,22 @@ MONAD_NAMESPACE_BEGIN
 bool sender_has_balance(State &state, evmc_message const &msg) noexcept
 {
     auto const value = intx::be::load<uint256_t>(msg.value);
-    auto const balance =
-        intx::be::load<uint256_t>(state.get_balance(msg.sender));
+    auto const &account = state.recent_account(msg.sender);
+    auto const balance = account.has_value() ? account->balance : 0;
+    auto &original_state = state.original_account_state(msg.sender);
+    if (balance >= value) {
```
0xSqualo, line 56: Could you short circuit here if `balance == value`? Since the `diff` would equal 0, and then none of the write logic inside this first `if {}` branch would be triggered -> perhaps as an `if-else` in an additional branch.

category/execution/ethereum/state3/state.hpp
```diff
@@ -377,12 +397,14 @@ class State
...
             add_to_balance(beneficiary, account.value().balance);
             account.value().balance = 0;
+            original_account_state(address).set_validate_exact_balance();
         }
         else {
             if (address != beneficiary ||
                 account->incarnation == incarnation_) {
                 add_to_balance(beneficiary, account.value().balance);
                 account.value().balance = 0;
+                original_account_state(address).set_validate_exact_balance();
```
0xSqualo, line 430: To double check, should this be set only if the `address != beneficiary`?

I am also wondering if this should be called for `address` **and** `beneficiary`.

category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +581,52 @@ class State
...
+        }
+        if (original_state.validate_exact_balance()) {
+            return false;
+        }
+        if (actual->balance < original_state.min_balance()) {
+            return false;
+        }
+        auto it = current_.find(address);
+        if (it != current_.end()) {
+            MONAD_ASSERT(it->second.size() == 1);
```
0xSqualo, line 670: Should this assert? or return `false` (and perhaps log?) as the mismatch could not be fixed.

category/execution/ethereum/evm.cpp
magedm: What's the motivation for changing this? I'd expect balance > value in most cases.

category/execution/ethereum/state2/block_state.cpp
magedm: Do you mean the following or is there a better way?
```
for (auto &kv : state.original_) {
    Address const &address = kv.first;
    OriginalAccountState &account_state = kv.second;
```

category/execution/ethereum/state3/account_state.hpp
magedm: I'll change to regular assert.
Is there a general preference for regular asserts?

category/execution/ethereum/state3/state.hpp
magedm: > To double check, should this be set only if the address != beneficiary?
I think so. Yes.

If rev >= EVMC_CANCUN and address == beneficiary, the code doesn't access balance in this function.

> I am also wondering if this should be called for address and beneficiary.

This function only adds to beneficiary balance. There is no dependence on the exact value of beneficiary balance (in part because there is no overflow check on balance). So calling set_validate_exact_balance for beneficiary is unnecessary.

category/execution/ethereum/state3/state.hpp
magedm: This is intentional. The function try_fix_account_mismatch is called only if there is mismatch and since at this point we know that everything else matches, then it must be a balance mismatch.

category/execution/ethereum/state3/state.hpp
magedm: Transaction execution (even a speculative execution) should end up with any account found in current to have exactly one version in the version stack.
But since this code is executed only in speculative execution with account mismatch, it may be safer to change this assert to return false (which I wouldn't expect to happen).

If we return false here, the code becomes more robust but the code logic becomes less clear implying that having version stack size != 1 is  a possibility. What do you think?

category/execution/ethereum/state2/block_state.cpp
0xSqualo: Yes unfortunately I think this is only way in this case.

category/execution/ethereum/state3/account_state.hpp
0xSqualo: This was more regarding if these conditions should _never_ be broken at runtime (which I assume they shouldn't in this specific case).

If it is a state that probably should never happen, but is recoverable, then the debug asserts make sense from my side to catch those cases during testing automation or running debug builds otherwise.

category/execution/ethereum/state3/state.hpp
0xSqualo: Got it - in that case is it worth upgrading this assert to also be in the release builds? (as discussed in the other comment).

category/execution/ethereum/state3/state.hpp
magedm: Changing the if else code to the one line version was based on this comment from @aa755:
https://github.com/category-labs/monad/pull/908#discussion_r2239932744
And it passed Coq formal verification by @aa755 
https://github.com/category-labs/monad/pull/908#discussion_r2239970515

This code is went through passed several replays of 100K blocks from Ethereum block 19M.
I am open to changing it back to if else to be on the safe side.

category/execution/ethereum/evm.cpp
0xSqualo: It was more of a consideration if this was considered a hot path that could somewhat frequently occur, but in that case it could make sense to leave as is.

category/execution/ethereum/state3/state.hpp
0xSqualo: >  But since this code is executed only in speculative execution with account mismatch, it may be safer to change this assert to return false (which I wouldn't expect to happen).

Understood - perhaps in this case it makes sense to return `false` and change this to a debug assert?

category/execution/ethereum/state3/state.hpp
pleasew8t: I agree that separate branches make sense, for readability and to avoid any issues if any of the underlying assumptions change. The alternative is to add a unit test, as Magnus suggested, even just for the purpose of regression testing.

category/execution/ethereum/state3/state.hpp
aa755: if this assertion gets violated, it means something has gone horribly wrong and we dont understand what. In such a case, retrying may not really solve the problem. IMO, in this case, it would be better to halt the chain temporarily rather than letting the code do something potentially much more dangerous and go unnoticed (which a DEBUG assert will do in release builds?).

magedm: @0xSqualo, Any comment on @aa755 's comment.
A version stack with size != 1 at the end of execution should never happen. The question is if it is better to stop execution immediately or return false to let execution continue.

category/execution/ethereum/state3/state.hpp
jhunsaker: looks like this was changed per the suggestions - marking resolved

category/execution/ethereum/state3/state.hpp
magedm: Input from @jhunsaker that this should be an assert.

category/execution/ethereum/state3/state.hpp
```diff
@@ -206,6 +225,7 @@ class State
     bytes32_t get_balance(Address const &address)
     {
         auto const &account = recent_account(address);
+        original_account_state(address).set_validate_exact_balance();
```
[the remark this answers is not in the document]
magedm: The call to recent_account(address) guarantees that an entry exists in original_.
Furthermore, even if it were possible for the entry to be missing from original_, creating an entry in the map is the desired effect.

category/execution/ethereum/state3/account_state.hpp
```diff
@@ -101,4 +101,45 @@ class AccountState final : public AccountSubstate
...
+    }
+
+    void set_validate_exact_balance()
+    {
+        validate_exact_balance_ = true;
+    }
+
+    void set_min_balance(uint256_t const &value)
+    {
+        MONAD_ASSERT(account_.has_value());
```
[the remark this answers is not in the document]
magedm: This function is never called with a nullopt account. It is called only if original_balance > diff.

category/execution/ethereum/evm.cpp
```diff
@@ -45,8 +45,22 @@ MONAD_ANONYMOUS_NAMESPACE_BEGIN
...
+        auto const diff = balance - value;
+        auto const &original = original_state.account_;
+        auto const original_balance =
+            original.has_value() ? original->balance : 0;
+        if (original_balance > diff) {
```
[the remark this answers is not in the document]
magedm: What subtraction?

category/execution/ethereum/state3/state.hpp
```diff
@@ -574,6 +596,57 @@ class State
...
+            auto &recent = recent_state.account_;
+            if (!recent) {
+                return false;
+            }
+            if (actual->balance > original->balance) {
+                recent->balance += actual->balance - original->balance;
+            }
+            else {
+                MONAD_ASSERT(
+                    recent->balance >= (original->balance - actual->balance));
```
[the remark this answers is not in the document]
magedm: The assertion is reached only if actual->balance <= original->balance. There couldn't be underflow.

category/execution/ethereum/state3/state.hpp
```diff
@@ -574,6 +596,57 @@ class State
...
+        }
+        if (original_state.validate_exact_balance()) {
+            return false;
+        }
+        if (actual->balance < original_state.min_balance()) {
+            return false;
+        }
+        auto it = current_.find(address);
+        if (it != current_.end()) {
+            MONAD_ASSERT(it->second.size() == 1);
```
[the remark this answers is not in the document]
magedm: The assertion is intentional. Correct execution must guarantee exactly one entry in the version stack.

PR #1452, opened 2025-08-01
Add --trace_calls flag 

Chen-Yifan:


cmd/monad/main.cpp
```diff
@@ -205,6 +208,10 @@ int main(int const argc, char const *argv[])
         "event_trace", quill::file_handler(trace_log, handler_cfg));
 #endif
 
+#ifdef ENABLE_CALL_TRACING
+    enable_call_tracing(trace_calls);
+#endif
```
Baltoli: Do we still need the compile-time option? I might be misunderstanding but I thought the point of this PR was to reduce the number of config options and allow call tracing to be enabled at zero runtime cost without having to build it in?

PR #1468, opened 2025-08-05
Wide and truncating multiplication for arbitrary-width integers

m-alvarez:
Wide and truncating multiplication for arbitrary-width integers.

Currently only 4x4->4 and 4x4->8 versions are used, but more exotic widths are necessary for Barrett reduction.

### DIFF
--- a/category/vm/runtime/uint256.hpp
+++ b/category/vm/runtime/uint256.hpp
@@ -32,7 +32,7 @@
 #endif
 
 #ifndef __BMI2__
-#  error "Target architecture must support BMI2 (for MULX)"
+    #error "Target architecture must support BMI2 (for MULX)"
 #endif
 
 // GCC's overeager SLP vectorizer sometimes pessimizes code. For functions that
@@ -732,36 +732,40 @@ namespace monad::vm::runtime
     }
 
     [[gnu::always_inline]]
-    inline constexpr std::pair<uint64_t, uint64_t>
-    mulx_constexpr(uint64_t const x, uint64_t const y) noexcept
+    inline constexpr void mulx_constexpr(
+        uint64_t const x, uint64_t const y, uint64_t &r_hi,
+        uint64_t &r_lo) noexcept
     {
-        uint128_t const prod = static_cast<uint128_t>(x) * static_cast<uint128_t>(y);
-        uint64_t const hi = static_cast<uint64_t>(prod >> uint128_t{64});
-        uint64_t const lo = static_cast<uint64_t>(prod);
-        return {hi, lo};
+        uint128_t const prod =
+            static_cast<uint128_t>(x) * static_cast<uint128_t>(y);
+        r_hi = static_cast<uint64_t>(prod >> uint128_t{64});
+        r_lo = static_cast<uint64_t>(prod);
     }
 
     [[gnu::always_inline]]
-    inline std::pair<uint64_t, uint64_t>
-    mulx_intrinsic(uint64_t const x, uint64_t const y) noexcept
+    inline void mulx_intrinsic(
+        uint64_t const x, uint64_t const y, uint64_t &r_hi,
+        uint64_t &r_lo) noexcept
     {
+        /*
         uint64_t hi;
         uint64_t lo;
+        */
         asm("mulx %[x], %[lo], %[hi]"
-            : [hi] "=r"(hi), [lo] "=r"(lo)
+            : [hi] "=r"(r_hi), [lo] "=r"(r_lo)
             : [x] "r"(x), [y] "d"(y));
-        return {hi, lo};
     }
 
     [[gnu::always_inline]]
-    inline constexpr std::pair<uint64_t, uint64_t>
-    mulx(uint64_t const x, uint64_t const y) noexcept
+    inline constexpr void mulx(
+        uint64_t const x, uint64_t const y, uint64_t &r_hi,
+        uint64_t &r_lo) noexcept
     {
         if consteval {
-            return mulx_constexpr(x, y);
+            return mulx_constexpr(x, y, r_hi, r_lo);
         }
         else {
-            return mulx_intrinsic(x, y);
+            return mulx_intrinsic(x, y, r_hi, r_lo);
         }
     }
 
@@ -769,47 +773,48 @@ namespace monad::vm::runtime
     using words_t = std::array<uint64_t, M>;
 
     [[gnu::always_inline]]
-    inline std::tuple<uint64_t, uint64_t, uint64_t> adc_3(
-        std::tuple<uint64_t, uint64_t, uint64_t> const x,
-        std::tuple<uint64_t, uint64_t> const y) noexcept
+    inline void adc_3(
+        uint64_t x_2, uint64_t x_1, uint64_t x_0, uint64_t const y_1,
+        uint64_t const y_0, uint64_t &r_2, uint64_t &r_1,
+        uint64_t &r_0) noexcept
     {
-        auto [x_2, x_1, x_0] = x;
-        auto [y_1, y_0] = y;
         asm("addq %[y_0], %[x_0]\n"
             "adcq %[y_1], %[x_1]\n"
             "adcq $0, %[x_2]"
             : [x_0] "+r"(x_0), [x_1] "+r"(x_1), [x_2] "+r"(x_2)
             : [y_0] "r"(y_0), [y_1] "r"(y_1)
             : "cc");
-        return {x_2, x_1, x_0};
+        r_2 = x_2;
+        r_1 = x_1;
+        r_0 = x_0;
     }
 
     [[gnu::always_inline]]
-    inline std::pair<uint64_t, uint64_t>
-    adc_2(std::pair<uint64_t, uint64_t> const x, uint64_t const y_0) noexcept
+    inline void adc_2(
+        uint64_t x_1, uint64_t x_0, uint64_t const y_0, uint64_t &r_1,
+        uint64_t &r_0) noexcept
     {
-        auto [x_1, x_0] = x;
         asm("addq %[y_0], %[x_0]\n"
             "adcq $0, %[x_1]"
             : [x_0] "+r"(x_0), [x_1] "+r"(x_1)
             : [y_0] "r"(y_0)
             : "cc");
-        return {x_1, x_0};
+        r_1 = x_1;
+        r_0 = x_0;
     }
 
     [[gnu::always_inline]]
-    inline std::pair<uint64_t, uint64_t> adc_2(
-        std::pair<uint64_t, uint64_t> const x,
-        std::pair<uint64_t, uint64_t> const y) noexcept
+    inline void adc_2(
+        uint64_t x_1, uint64_t x_0, uint64_t const y_1, uint64_t const y_0,
+        uint64_t &r_1, uint64_t &r_0) noexcept
     {
-        auto [x_1, x_0] = x;
-        auto [y_1, y_0] = y;
         asm("addq %[y_0], %[x_0]\n"
             "adcq %[y_1], %[x_1]"
             : [x_0] "+r"(x_0), [x_1] "+r"(x_1)
             : [y_0] "r"(y_0), [y_1] "r"(y_1)
             : "cc");
-        return {x_1, x_0};
+        r_1 = x_1;
+        r_0 = x_0;
     }
 
     template <size_t I, size_t R, size_t M>
@@ -820,8 +825,18 @@ namespace monad::vm::runtime
     {
         if constexpr (I < std::min(R, M)) {
             if constexpr (I + 1 < R) {
-                auto const [hi, lo] = mulx(x[I], y);
-                std::tie(carry, result[I]) = adc_2({hi, lo}, carry);
+                uint64_t hi;
+                uint64_t lo;
+                mulx(x[I], y, hi, lo);
+                adc_2(
+                    // Input 1
+                    hi,
+                    lo,
+                    // Input 2
+                    carry,
+                    // Output
+                    carry,
+                    result[I]);
                 mul_line_recur<I + 1, R, M>(x, y, result, carry);
             }
             else {
@@ -841,7 +856,7 @@ namespace monad::vm::runtime
         words_t<R> &__restrict__ result) noexcept
     {
         uint64_t carry;
-        std::tie(carry, result[0]) = mulx(y, x[0]);
+        mulx(y, x[0], carry, result[0]);
 
         mul_line_recur<1, R, M>(x, y, result, carry);
     }
@@ -855,15 +870,35 @@ namespace monad::vm::runtime
         if constexpr (J + 1 < M && I + J < R) {
             if constexpr (I + J + 2 < R) {
                 // We need c_lo, c_hi
-                auto const [hi, lo] = mulx(x[J + 1], y_i);
-                std::tie(c_hi, c_lo, result[I + J]) =
-                    adc_3({hi, lo, result[I + J]}, {c_hi, c_lo});
+                uint64_t hi;
+                uint64_t lo;
+                mulx(x[J + 1], y_i, hi, lo);
+                adc_3(
+                    // Input 1
+                    hi,
+                    lo,
+                    result[I + J],
+                    // Input 2
+                    c_hi,
+                    c_lo,
+                    // Result
+                    c_hi,
+                    c_lo,
+                    result[I + J]);
             }
             else if constexpr (I + J + 1 < R) {
                 // We only need c_lo
                 uint64_t const lo = x[J + 1] * y_i;
-                std::tie(c_lo, result[I + J]) =
-                    adc_2({lo, result[I + J]}, {c_hi, c_lo});
+                adc_2(
+                    // Input 1
+                    lo,
+                    result[I + J],
+                    // Input 2
+                    c_hi,
+                    c_lo,
+                    // Output
+                    c_lo,
+                    result[I + J]);
             }
             else {
                 // We're done, we don't need subsequent results
@@ -873,9 +908,16 @@ namespace monad::vm::runtime
         }
         else {
             if constexpr (I + M < R) {
-                auto [hi, lo] = adc_2({c_hi, c_lo}, result[I + M - 1]);
-                result[I + M - 1] = lo;
-                result[I + M] = hi;
+
+                adc_2(
+                    // Input 1
+                    c_hi,
+                    c_lo,
+                    // Input 2
+                    result[I + M - 1],
+                    // Output
+                    result[I + M],
+                    result[I + M - 1]);
             }
             else if constexpr (I + M < R + 1) {
                 result[I + M - 1] += c_lo;
@@ -904,7 +946,7 @@ namespace monad::vm::runtime
         uint64_t c_lo;
 
         if constexpr (I + 1 < R) {
-            std::tie(c_hi, c_lo) = mulx(x[0], y_i);
+            mulx(x[0], y_i, c_hi, c_lo);
         }
         else {
             c_hi = 0;
@@ -947,7 +989,9 @@ namespace monad::vm::runtime
         for (size_t j = 0; j < N; j++) {
             uint64_t carry = 0;
             for (size_t i = 0; i < M && i + j < R; i++) {
-                auto const [hi, lo] = mulx(x[i], y[j]);
+                uint64_t hi;
+                uint64_t lo;
+                mulx(x[i], y[j], hi, lo);
 
                 auto const [s0, c0] = addc(lo, result[i + j], false);
                 auto const [s1, c1] = addc(s0, carry, false);


test/vm/unit/uint256_tests.cpp
```diff
@@ -18,6 +18,12 @@
 
 using namespace monad::vm::runtime;
 
+TEST(uint256, intx_binary_compatible)
+{
+    ASSERT_EQ(alignof(uint256_t), alignof(::intx::uint256));
+    ASSERT_EQ(sizeof(uint256_t), sizeof(::intx::uint256));
```
Baltoli: Can these just be static asserts?

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+    template <size_t R, size_t M>
+    [[gnu::always_inline]]
+    inline constexpr void mul_line(
+        words_t<M> const &x, uint64_t const y,
+        NO_AVX_COPY uint64_t *__restrict__ result) noexcept
+    {
+        uint64_t carry;
+        std::tie(carry, result[0]) = mulx(y, x[0]);
+#pragma GCC unroll(M)
+        for (size_t i = 1; i < std::min(R, M); i++) {
```
andreaslyn: Consider implementing this with template programming instead of loop unrolling, because loop unrolling can prevent optimizations, if the unrolling happens too late.

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+    }
+
+    template <size_t M>
+    using words_t = std::array<uint64_t, M>;
+
+#if defined(__GNUC__) && !defined(__clang__)
+    #define NO_AVX_COPY volatile
+#else
+    #define NO_AVX_COPY
+#endif
```
andreaslyn: Do you know alternatives to this?

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+        // Clang does not need this case analysis, but GCC can't figure out what
+        // portion of the multiplication is dead code
+        if (i + 1 < R) {
+            std::tie(c_hi, c_lo) = mulx(x[0], y_i);
+        }
+        else {
+            c_lo = x[0] * y_i;
+        }
+#pragma GCC unroll(M)
+        for (size_t j = 0; j + 1 < M && i + j < R; j++) {
```
andreaslyn: It is possible that template programming will allow for better code gen here as well.

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+    operator*(uint256_t const &lhs, uint256_t const &rhs) noexcept
+    {
+        if consteval {
+            return uint256_t{truncating_mul<uint256_t::num_words>(
+                lhs.as_words(), rhs.as_words())};
+        }
+        else {
+            uint256_t result;
+            monad_vm_runtime_mul(&result, &lhs, &rhs);
+            return result;
```
andreaslyn: The interpreter should actually call this directly, because it can do
```c++
auto const &lhs = pop();
monad_vm_runtime_mul(*stack_top, lhs, *stack_top);
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+        // Instead, we widen the carry and skew the loop so that every iteration
+        // computes
+        //     result[i+j] = result[i+j] + c_lo
+        //     c_lo = prod_lo + c_hi (+ carry)
+        //     c_hi = prod_hi (+ carry)
+        uint64_t c_hi;
+        uint64_t c_lo;
+
+        // Clang does not need this case analysis, but GCC can't figure out what
+        // portion of the multiplication is dead code
```
andreaslyn:
```suggestion
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+        if (i + 1 < R) {
+            std::tie(c_hi, c_lo) = mulx(x[0], y_i);
+        }
+        else {
+            c_lo = x[0] * y_i;
+        }
+#pragma GCC unroll(M)
+        for (size_t j = 0; j + 1 < M && i + j < R; j++) {
+            // Clang does not need this case analysis, but GCC can't figure out
+            // what portion of the multiplication is dead code
```
andreaslyn:
```suggestion
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+        return adc_2_intrinsic(x_1, x_0, y_1, y_0);
+#endif
+    }
+
+    // result[i .. min(i + M + 1, R)) += y_i * x[0 .. M)
+    template <size_t R, size_t M>
+    [[gnu::always_inline]]
+    inline constexpr void mul_add_line(
+        words_t<M> const &x, size_t const i, uint64_t const y_i,
+        NO_AVX_COPY uint64_t *__restrict__ result) noexcept
```
andreaslyn: The `NO_AVX_COPY` is forcing GCC to spill all result words to memory, right?

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +742,284 @@ namespace monad::vm::runtime
...
+        words_t<M> const &x, uint64_t const y,
+        NO_AVX_COPY uint64_t *__restrict__ result) noexcept
+    {
+        uint64_t carry;
+        std::tie(carry, result[0]) = mulx(y, x[0]);
+#pragma GCC unroll(M)
+        for (size_t i = 1; i < std::min(R, M); i++) {
+            // Without the explicit conditional, GCC will not realize that the
+            // higher word of the multiplication is unused in the last iteration
+            // of the loop when R = M and use wide multiplication instead
```
andreaslyn:
```suggestion
```
Comment unnecessary, because the alternative is literally telling the compiler to emit a wide multiplication, so gcc is right.

category/vm/runtime/uint256.hpp
m-alvarez: There is a number of alternatives
1. Manually insert memory barriers after every store `result[i] = ...`
  * This also results in the entire `result` array being spilled to memory, and I found it to have worse performance than the `volatile` writes.
2. Shift more of the code to assembly.
3. Suppress vectorization with pragmas
  * `target` pragmas prevent inlining of stdlib functions so the performance tanks
  * `optimize("no-tree-vectorize")` is actually enough to suppress all the vector instructions for both stand-alone `truncating_mul` and for the Barrett reduction use-case
4. All of the vector instructions in a single `truncating_mul` can be eliminated with elementwise casts to `volatile` (instead of casting the entire array), which also results in better codegen for `truncating_mul`. However this doesn't eliminate all of the vector instructions when doing Barrett reduction.

I think since the main use-case for making `truncating_mul` fast is Barrett reduction, and the Barrett functions will most often be called without inlining (as they will be runtime calls from compiled code), we can go with option 3, with the caveat that anyone that calls `truncating_mul` may have to transitively disable tree vectorization

andreaslyn: An alternative would be to put the specializations we care about in a separate .cpp file, then compile it into 
the corresponding .s file with clang. Now commit the .s file to git and use that instead of the .cpp file when compiling the VM. Include a makefile/script for generating the .s file, for future updates.

andreaslyn: Would of course prevent inline, but that should not have significant impact.

category/vm/runtime/uint256.hpp
m-alvarez: Yes. This turned out to make a big performance difference. Moving to a non-volatile version and suppressing vectorization explicitly is very beneficial for performance of sequences of multiplications (exponentiation, Barrett reduction)

category/vm/runtime/uint256.hpp
m-alvarez: Multiplication-by-reciprocal methods benefit a lot from inlining; without it, on multi-word integers I have benchmarks showing they're worse than naive division. However, the multiplication-by-reciprocal versions of `udivrem`, `addmod` and `mulmod` don't need to be inlined themselves, so we could either lower those to assembly or (what I am doing currently) make them `noinline` and turn off vectorization locally. This gives good performance so far

category/vm/fuzzing/generator/generator.hpp
```diff
@@ -867,7 +868,7 @@ namespace monad::vm::fuzzing
                         g, forward_jds_begin, forward_jds_end);
                 }));
 
-            auto const *bs = intx::as_bytes(jd);
+            auto const *bs = runtime::as_bytes(jd);
```
andreaslyn: I think we are missing a unit test for each new `as_bytes` transmute function. Alternatively, use reinterpret cast here and remove the `as_bytes` transmute functions.

category/vm/runtime/uint256.hpp
```diff
@@ -17,18 +17,26 @@
...
 
 #ifndef __AVX2__
     #error "Target architecture must support AVX2"
 #endif
 
+#if defined(__GNUC__) && !defined(__clang__)
+    #define MONAD_VM_NO_VECTORIZE __attribute__((optimize("no-tree-vectorize")))
+#else
+    #define MONAD_VM_NO_VECTORIZE
+#endif
```
andreaslyn: Presumably this macro is used when it is known that vectorization is inefficient, so why allow clang vectorization when this macro is present?

category/vm/runtime/uint256.hpp
```diff
@@ -416,30 +405,14 @@ namespace monad::vm::runtime
...
-                uint256_t result;
-                monad_vm_runtime_mul(&result, &lhs, &rhs);
-                return result;
-            }
-        }
+        operator*(uint256_t const &lhs, uint256_t const &rhs) noexcept;
 
+        MONAD_VM_NO_VECTORIZE
         [[gnu::always_inline]]
         inline constexpr uint256_t &operator*=(uint256_t const &rhs) noexcept
```
andreaslyn: I think we lose the vectorization guarantees when this function is inlined. We need to specify the external functions with `MONAD_VM_NO_VECTORIZE` as noinline, right?

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +743,308 @@ namespace monad::vm::runtime
...
+        return {r_2, r_1, r_0};
+    }
+
+    [[gnu::always_inline]]
+    inline constexpr std::tuple<uint64_t, uint64_t, uint64_t> adc_3(
+        std::tuple<uint64_t, uint64_t, uint64_t> const x,
+        std::tuple<uint64_t, uint64_t, uint64_t> const y) noexcept
+    {
+#if defined(__GNUC__) && !defined(__clang__)
+        return adc_3_asm(x, y);
```
andreaslyn: Can this `adc_3_asm` be used independently of compiler?

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +743,308 @@ namespace monad::vm::runtime
...
+    [[gnu::always_inline]]
+    inline constexpr void mul_add_line(
+        words_t<M> const &x, uint64_t const y_i,
+        words_t<R> &__restrict__ result) noexcept
+    {
+        // A naive implementation would use a single carry variable. However
+        // this means on every iteration we compute
+        //     result[i+j] = result[i+j] + prod_lo + carry
+        // which needs to propagate two carry bits. This causes a slew of
+        // setb/xor/movxz instructions to be inserted.
```
andreaslyn: I agree that the naive algorithm is explicitly telling the compiler to propagate two carry bits, but I would be surprised if any compiler guarantees this.
```suggestion
        // which needs to propagate two carry bits. This causes a slew of
        // setb/xor/movxz instructions to be inserted.
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +743,308 @@ namespace monad::vm::runtime
...
+        // Instead, we widen the carry and skew the loop so that every iteration
+        // computes
+        //     result[i+j] = result[i+j] + c_lo
+        //     c_lo = prod_lo + c_hi (+ carry)
+        //     c_hi = prod_hi (+ carry)
+        uint64_t c_hi;
+        uint64_t c_lo;
+
+        // Clang does not need this case analysis, but GCC can't figure out what
+        // portion of the multiplication is dead code
```
andreaslyn: Is this behavior of both clang and gcc documented somewhere?

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +743,308 @@ namespace monad::vm::runtime
...
+    [[gnu::always_inline]]
+    inline std::pair<uint64_t, uint64_t>
+    mulx_intrinsic(uint64_t const x, uint64_t const y) noexcept
+    {
+        uint64_t hi;
+#if defined(__GNUC__) && !defined(__clang__)
+        uint64_t lo;
+        asm("mulx %[x], %[lo], %[hi]"
+            : [hi] "=r"(hi), [lo] "=r"(lo)
+            : [x] "rm"(x), [y] "d"(y));
```
andreaslyn: I have experienced problems with inline asm memory input operands being unnecessarily copied to the stack, so I personally never use memory input operands.

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +743,308 @@ namespace monad::vm::runtime
...
+        auto const lo = static_cast<uint64_t>(prod);
+        return {hi, lo};
+    }
+
+    [[gnu::always_inline]]
+    inline std::pair<uint64_t, uint64_t>
+    mulx_intrinsic(uint64_t const x, uint64_t const y) noexcept
+    {
+        uint64_t hi;
+#if defined(__GNUC__) && !defined(__clang__)
```
andreaslyn: Seemingly unnecessary to branch on compiler to avoid inline assembly just for clang

category/vm/runtime/uint256.hpp
m-alvarez: The problem is not vectorization per se, but GCC's overeager SLP vectorizer, which is a [known issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105617). Clang tends to only generate vector instructions for things like zeroing out local variables, which seems inoffensive.

category/vm/runtime/uint256.hpp
m-alvarez: No, but the comment is outdated and now unnecessary: since we are using inline assembly to call `mulx`, it's to be expected that GCC will not be able to narrow it to `imul` if the upper part is not used.

category/vm/runtime/uint256.hpp
m-alvarez: This is a very long-standing issue with Clang, see [here](https://bugs.llvm.org/show_bug.cgi?id=20197)

category/vm/fuzzing/generator/generator.hpp
m-alvarez: Since `runtime::as_bytes` is only ever used here, I replaced it by a cast and removed it.

category/vm/runtime/uint256.hpp
m-alvarez: Yes. I've now removed the intrinsics version of `adc_2` and `adc_3` and removed all the conditional compilation directives. Since Clang cannot deal with multiple constraint alternatives efficiently, all of the "rm" operands are now "r". This causes more `mov` instructions to be inserted but no discernible impact to performance, either when benchmarking multiplication on its own or when benchmarking the entire Barrett reduction operators.

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +731,265 @@ namespace monad::vm::runtime
...
+    }
+
+    template <size_t R, size_t M, size_t N>
+    [[gnu::always_inline]]
+    inline words_t<R>
+    truncating_mul_runtime(words_t<M> const &x, words_t<N> const &y) noexcept
+        requires(0 < R && R <= M + N)
+    {
+        words_t<R> result;
+        mul_line<R>(x, y[0], result);
```
andreaslyn, line 977: Looks like this is an access out of bounds when `N == 0` is true

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +731,265 @@ namespace monad::vm::runtime
...
+        truncating_mul_runtime_recur<1, R, M, N>(x, y, result);
```
andreaslyn, line 978: I guess this will work, but not sure whether we need to support the cases `M == 0` or `N == 0`,
```suggestion
        words_t<R> result{0};
        truncating_mul_runtime_recur<0, R, M, N>(x, y, result);
```

m-alvarez: Good catch. The proposed change fixes the issues but incurs M unnecessary additions. I think requiring `M > 0 && N > 0` is reasonable.

category/vm/runtime/uint256.hpp
andreaslyn: Is there a concern about making this noinline?

m-alvarez: We can annotate all external functions as `MONAD_VM_NO_VECTORIZE` and move `[[gnu::noinline]]` to the definition of `MONAD_VM_NO_VECTORIZE`.
There is a small concern that `exp` uses multiplication but doesn't trigger the SLP vectorizer, possibly because all the operands can be kept in registers, and so annotating it as `NO_VECTORIZE` (and therefore no inline) means we miss out on inlining for `exp` without it being necessary. We could just leave `exp` as it is, but I don't like the idea of suppressing vectorization only for some functions because the optimizer happens to do the right thing.

andreaslyn: Can `exp` use an internal multiplication which is annotated with `MONAD_VM_NO_VECTORIZE` and `always_inline` instead of an external function?

test/vm/unit/uint256_tests.cpp
```diff
@@ -436,15 +453,64 @@ TEST(uint256, arithmetic)
...
+                truncating_mul_runtime<2 * uint256_t::num_words>(
+                    x.as_words(), y.as_words());
+            ASSERT_EQ(full_product_rt, intx_product);
+
+            auto const full_product_ce =
+                truncating_mul_constexpr<2 * uint256_t::num_words>(
+                    x.as_words(), y.as_words());
+            ASSERT_EQ(full_product_ce, intx_product);
+
+            check_truncating_mul<1>(
```
andreaslyn: Perhaps it would be good to have unit tests for the cases `(M = 1,N = 1, R = 1..2)`, `(M = 2, N = 1, R = 1..3)`, `(M = 1, N = 2, R = 1..3)`.

category/vm/runtime/uint256.hpp
m-alvarez: Yes, but `MONAD_VM_NO_VECTORIZE` will not have any effect in that case as inlined functions will be vectorized as part of the callee.

andreaslyn: I would expect all functions using multiplication be annotated with `MONAD_VM_NO_VECTORIZE`. If those functions are external then additionally `noinline`, and if internal then `always_inline` for the most part.

m-alvarez: I think this is sensible. I have proceeded as follows:
* Mark `operator*`, `operator*=`, `exp` as `noinline`
* Add a note on `MONAD_VM_NO_VECTORIZE` explaining this

andreaslyn (approved):

andreaslyn (dismissed):

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +727,272 @@ namespace monad::vm::runtime
         }
     }
 
+    [[gnu::always_inline]]
+    inline constexpr std::pair<uint64_t, uint64_t>
+    mulx_constexpr(uint64_t const x, uint64_t const y) noexcept
+    {
+        auto const prod = static_cast<uint128_t>(x) * y;
+        auto const hi = static_cast<uint64_t>(prod >> 64);
+        auto const lo = static_cast<uint64_t>(prod);
```
dhil: I prefer to be explicit about the integer types and their ranks here. Feel free to ignore.
```suggestion
        uint128_t const prod = static_cast<uint128_t>(x) * static_cast<uint128_t>(y);
        uint64_t const hi = static_cast<uint64_t>(prod >> uint128_t{64});
        uint64_t const lo = static_cast<uint64_t>(prod);
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +727,272 @@ namespace monad::vm::runtime
...
+     * corresponds to full precision multiplication
+     */
+    template <size_t R, size_t M, size_t N>
+    MONAD_VM_NO_VECTORIZE [[gnu::always_inline]]
+    inline constexpr words_t<R>
+    truncating_mul(words_t<M> const &x, words_t<N> const &y) noexcept
+        requires(0 < R && 0 < M && 0 < N && R <= M + N)
+    {
+        if consteval {
+            return truncating_mul_constexpr<R>(x, y);
```
dhil: In critical code, I'll always err on the side of being explicit. Again, feel free to ignore.
```suggestion
            return truncating_mul_constexpr<R, M, N>(x, y);
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +727,272 @@ namespace monad::vm::runtime
...
+        }
+        else {
+            return truncating_mul_runtime<R>(x, y);
```
dhil:
```suggestion
            return truncating_mul_runtime<R, M, N>(x, y);
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +727,272 @@ namespace monad::vm::runtime
...
+    inline void mul_add_line_recur(
+        words_t<M> const &x, uint64_t const y_i,
+        words_t<R> &__restrict__ result, uint64_t c_hi, uint64_t c_lo) noexcept
+    {
+        if constexpr (J + 1 < M && I + J < R) {
+            if constexpr (I + J + 2 < R) {
+                // We need c_lo, c_hi
+                auto const [hi, lo] = mulx(x[J + 1], y_i);
+                std::tie(c_hi, c_lo, result[I + J]) =
+                    adc_3({hi, lo, result[I + J]}, {c_hi, c_lo});
```
dhil: Do these tuple-arguments generate as tight code as we'd like? C.f. https://godbolt.org/z/3odjxbh1W -- I hope here the compiler can eliminate the overhead because it can see the implementation.

andreaslyn: I was thinking about this as well. I think the compilers provide is not such guarantees, so I would not personally be using it in runtime code.

category/vm/runtime/uint256.hpp
```diff
@@ -17,29 +17,31 @@
...
+#include <cstring>
 #include <format>
 #include <immintrin.h>
-#include <intx/intx.hpp>
 #include <limits>
 #include <stdexcept>
 
 #ifndef __AVX2__
     #error "Target architecture must support AVX2"
 #endif
```
aa755, line 32: I guess most CPUs that support AVX2 also support BMI2, but we may want to be safe:

```suggestion
#endif

#ifndef __BMI2__
#  error "Target architecture must support BMI2 (for MULX)"
#endif
```

category/vm/runtime/uint256.hpp
```diff
@@ -777,6 +727,272 @@ namespace monad::vm::runtime
...
+    template <size_t I, size_t R, size_t M>
+    [[gnu::always_inline]]
+    inline void mul_add_line(
+        words_t<M> const &x, uint64_t const y_i,
+        words_t<R> &__restrict__ result) noexcept
+    {
+        // A naive implementation would use a single carry variable. However
+        // this means on every iteration we compute
+        //     result[i+j] = result[i+j] + prod_lo + carry
+        // which needs to propagate two carry bits. This causes a slew of
```
aa755, line 938: If most of the monad validators use Broadwell or later CPUs, even if not all, we may want to provide a separate codepath (by checking for `#ifdef __ADX__`) where we exploit ADCX/ADOX to have dual carry chains.  I guess most usages will be for 4*4->8, so we can have a specialization for this case where we reuse [this kernel](https://github.com/herumi/mcl/blob/877aa82ecc55dc111922492ad600e177d78dfc29/src/asm/bint-x64-amd64.S#L5233)

aa755 (dismissed):
I am not an expert of the codegen quality guarantees of gcc/clang so for me it is hard to understand whether this conceptually elegant multiplication  implementation which uses easy-to-understand abstractions will produce the most performant machine code. For example, it is not clear whether items will be spilled to memory after inlining all the template-recursive calls to `mul_add_line_recur`. 

Even if the current compiler produces good code, the future versions may not. Is there a reason to not implement custom assembly kernels for the shapes we need, like this [kernel](https://github.com/herumi/mcl/blob/877aa82ecc55dc111922492ad600e177d78dfc29/src/asm/bint-x64-amd64.S#L5233) (this uses ADX so we would need to tweak it).  I understand that reasoning about assembly by eyeballing may be hard but we can do Coq proofs to guarantee correctness. The arithmetic used in this multiplication should be easy to reason about using Coq libraries.  IIUC, the kernel will use <10 x86 instructions so writing semantics of them should also not be hard.

EVM ops should only need the 4*4->8 kernel. Does Barret reduction require many different shapes? Even in that case, we can write a Gallina(Coq) function that produces a kernel given as inputs the sizes of inputs and output (type `N->N->N-> list X86Instruction`) and prove that function generically correct (for all sizes). Then our build system can just generate the .S files buy running that Gallina function.

category/vm/runtime/uint256.hpp
m-alvarez: This would be great to have if we could do it generically (I tried but couldn't find a way to get GCC to emit ADCX/ADOX instructions). I think specialized kernels for fixed sizes are unlikely to be very helpful here: a 4x4->8 kernel would only be used by `mulmod` and Barrett reduction uses very different multiplication sizes depending on the number of significant bits of the input.

PR #1471, opened 2025-08-05
Add prestate tracing support to`eth_call`

dhil:
Rebase of #1330. 

This patch is simpler than #1443 in that it doesn't commit the prestate traces to DB.

I think there are some improvements that can be made to this patch:
* Geth prestate tracer has some configurable options, which can disable serialization of bytecode, storage, toggle between diff mode and "regular" mode.
* ~~The prestate traces are materialized only then to be immediately serialized to Json, we could skip this intermediate step and construct the Json objects directly (ought to reduce the memory footprint).~~


category/execution/ethereum/trace/tracer_config.h
```diff
@@ -0,0 +1,18 @@
...
+{
+#endif
+
+enum tracer_config
+{
+    NOOP_TRACER = 0,
+    CALL_TRACER = 1,
+    PRESTATE_TRACER = 2,
+    STATEDIFF_TRACER = 3,
+};
```
Baltoli: I think the flag in `monad/main.cpp` should be updated to allow node operators to specify prestate or statediff tracing, right?

Baltoli: May also be worth commenting that this is intended as an exclusive flag, and isn't a bitfield.

category/execution/ethereum/state3/state.hpp
```diff
@@ -559,6 +559,22 @@ class State
...
+    Map<Address, AccountState> get_original_accounts() const
+    {
+        return original_;
+    }
+
+    Map<Address, AccountState> get_current_accounts() const
+    {
+        Map<Address, AccountState> current;
+        for (auto const &[address, stack] : current_) {
+            current.emplace(address, stack.recent());
```
Baltoli: How come this is `get_current_accounts` but references `stack.recent()` rather than `stack.current()`?

category/execution/ethereum/test/test_prestate_tracer.cpp
```diff
@@ -0,0 +1,120 @@
...
+                "balance":"0x1f4",
+                "code":"0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0160005500",
+                "nonce":1,
+                "storage":{
+                    "0x00000000000000000000000000000000000000000000000000000000cafebabe":"0x0000000000000000000000000000000000000000000000000000000000000003",
+                    "0x1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c1c":"0x0000000000000000000000000000000000000000000000000000000000000003"
+                }
+            }
+        },
+        "pre":{"0x0000000000000000000000000000000000000100":{"balance":"0x0"}}
```
Baltoli: Very minor nit: might be slightly easier to parse this visually if this was formatted like `post` (i.e. the address etc. are indented).

category/execution/ethereum/trace/tracer_config.h
dhil: Yes, it is a good question. So with this patch prestate tracing is enabled when performing the RPC call. The caller chooses between prestate or statediff. Maybe we need a flag that can enable/disable this facility at the RPC altogether?

category/execution/ethereum/state3/state.hpp
dhil: With `VersionStack::current(unsigned const version)` I need to pick some version, which may cause the account state to be copied -- otherwise I think it amounts to the same? That's my current reading of the code.

category/rpc/test/test_eth_call.cpp
```diff
@@ -981,3 +985,285 @@ TEST_F(EthCallFixture, static_precompile_OOG_with_trace)
...
+        to_vec(rlp::encode_address(std::make_optional(from)));
+    auto const rlp_block_id = to_vec(rlp_finalized_id);
+
+    auto executor = monad_eth_call_executor_create(
+        1, 1, node_lru_size, max_timeout, max_timeout, dbname.string().c_str());
+    auto state_override = monad_state_override_create();
+
+    // PreState
+    {
+        struct callback_context ctx;
```
dhil: I believe the root cause of flaky test(s) is here. 

There is a race condition here between the promise inside the `callback_context` (on the main test thread) and the `monad_eth_call_executor` (on a separate thread). What can happen here is that the destructor of `callback_context` is invoked, and thereby erasing the promise, before the `monad_eth_call_executor` is completely done with the promise. I believe a necessary criterion for correctness here is that `callback_context` must outlive `monad_eth_call_executor`.

~~If this criterion is sound, then as a corollary most of the tests in `EthCallFixture` are broken. The reason we don't observe it may just be because there is so much work in between the call to `monad_eth_call_executor` and the destruction of the `callback_context`.~~ Edit: actually, the destruction order is OK, because `monad_eth_call_executor` is manually destroyed before the implicit destruction of `callback_context` in most test cases.

I was able to reproduce this flakiness with a test that does a single call to `monad_eth_call_executor` simply by enclosing it in a local scope with the `callback_context`.

I was able to reproduce the flakiness on `main` with just a single call to `monad_eth_call_executor`.

category/execution/ethereum/trace/prestate_tracer.cpp
```diff
@@ -0,0 +1,313 @@
...
+            if (!(original_account.has_value() &&
+                  current_account.has_value())) {
+                post[address_key] = account_to_json(current_account, state);
+            }
+            else {
+                if (original_account->balance != current_account->balance) {
+                    post[address_key]["balance"] =
+                        "0x" + intx::to_string(current_account->balance, 16);
+                }
+                if (original_account->code_hash != current_account->code_hash) {
```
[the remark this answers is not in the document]
dhil: Added a safety justification.

category/execution/ethereum/trace/tracer_config.h
Baltoli: > So with this patch prestate tracing is enabled when performing the RPC call

ah, that's what I was missing. Got it!

category/execution/ethereum/state3/state.hpp
```diff
@@ -574,6 +574,22 @@ class State
...
     }
+
+    ////////////////////////////////////////
+
+    Map<Address, AccountState> get_original_accounts() const
+    {
+        return original_;
+    }
+
+    Map<Address, AccountState> get_current_accounts() const
```
kjcamann: I think these should not be forced to make copies, i.e., they should return a reference `Map<Address, AccountState> const &`.

It's true that in the place where you use them, you need to make copies because you erase things out of them. But it should be the call-site that knows whether it's just going to erase/needs a copy, and perform its local initialization accordingly.

If you do that, then you can't use `auto x = get_original_accounts()` later, since it will deduce the reference type, you'll need to explicitly say the type name so you get copy initialization. But this way leaves it possible for other parts of the code to "just observe".

dhil: I think the better option here is to use your visitor design, and then inside `prestate_tracer.cpp` simply use the visitor to project the necessary information. What do you think?

dhil: I've implemented your suggestion in https://github.com/category-labs/monad/pull/1471/commits/f17d59399384257115626d16dd99d6f7436c36b5.

category/execution/ethereum/trace/prestate_tracer.cpp
```diff
@@ -0,0 +1,317 @@
...
+            auto &current_account_state = it->second;
+            auto &current_account = current_account_state.account_;
+            auto &current_storage = current_account_state.storage_;
+            auto &original_account = original_account_state.account_;
+            auto &original_storage = original_account_state.storage_;
+            if (MONAD_UNLIKELY(!current_account.has_value())) {
+                // current account is destructed
+                if (MONAD_UNLIKELY(!original_account.has_value())) {
+                    // both original and current = NULL => no delta
+                    continue;
```
kjcamann: I am curious how this can happen, if it can.

category/execution/ethereum/trace/prestate_tracer.cpp
```diff
@@ -0,0 +1,317 @@
...
+                StateDelta state_delta{
+                    .account = {std::nullopt, current_account},
+                    .storage = generate_storage_deltas({}, current_storage),
+                };
+                state_deltas.emplace(address, std::move(state_delta));
+            }
+            else {
+                // remove unchanged storage values
+                for (auto it = original_storage.begin();
+                     it != original_storage.end();) {
```
kjcamann: Could you hoist the initializer and make this a while loop? A for loop with no update expression is quite strange-looking.

category/execution/ethereum/trace/prestate_tracer.cpp
dhil: Good question. The short answer is: I don't know. 

The slightly longer answer: I didn't write this piece of code, but I did reason type-directed about it. Clearly, the case `current_account = std::nullopt` and `original_account = std::nullopt` is a possible habitant of the product type `std::optional<Account> * std::optional<Account>`. Though, that does not mean the case manifests at runtime.

I guess a possible interpretation of the four possible habitants are:
* `(std::nullopt, Account)`: account created
* `(Account, std::nullopt)`: account destructed
* `(Account, Account)`: account modified
* `(std::nullopt, std::nullopt)`: account never touched

If this reading is correct, then I suppose a reasonable implementation would never materialise the case `(std::nullopt, std::nullopt)`.

So question: Should we change this to a `MONAD_ASSERT(false)` rather than `continue`?

jhunsaker: created and destructed in same tx? at first glance, this case doesn't stand out to me as an issue

dhil: In that case my suggested interpretation is clearly wrong, and the code should be good as-is.

PR #1467, opened 2025-08-05
Optimize shift by literal

andreaslyn:
Closes https://github.com/category-labs/monad/issues/1461

PR #1484, opened 2025-08-06
Register reservation improvements

andreaslyn:
Limit scope of reservations
Add missing reservations

Depends on https://github.com/category-labs/monad/pull/1467

### DIFF
--- a/category/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/compiler/ir/x86/emitter.cpp
@@ -1076,7 +1076,9 @@ namespace monad::vm::compiler::native
         if (is_live(elem, live) && !elem.stack_offset() && !elem.literal() &&
             !elem.avx_reg()) {
             if (stack_.has_free_general_reg()) {
+                elem.reserve_general_reg();
                 auto [new_elem, reserv] = alloc_general_reg();
+                elem.unreserve_general_reg();
                 mov_general_reg_to_gpq256(
                     *elem.general_reg(),
                     general_reg_to_gpq256(*new_elem->general_reg()));
@@ -1118,6 +1120,7 @@ namespace monad::vm::compiler::native
             if (stack_.has_free_general_reg() ||
                 (!elem->stack_offset() && !elem->avx_reg() &&
                  !elem->literal())) {
+                GeneralRegReserv const reserv{elem};
                 return alloc_general_reg();
             }
         }
@@ -1133,6 +1136,7 @@ namespace monad::vm::compiler::native
             if (stack_.has_free_avx_reg() ||
                 (!elem->stack_offset() && !elem->general_reg() &&
                  !elem->literal())) {
+                AvxRegReserv const reserv{elem};
                 return alloc_avx_reg();
             }
         }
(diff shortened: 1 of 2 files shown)


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -1076,7 +1076,9 @@ namespace monad::vm::compiler::native
         if (is_live(elem, live) && !elem.stack_offset() && !elem.literal() &&
             !elem.avx_reg()) {
             if (stack_.has_free_general_reg()) {
+                elem.reserve_general_reg();
                 auto [new_elem, reserv] = alloc_general_reg();
+                elem.unreserve_general_reg();
```
Baltoli, line 1081: If we wanted to keep `[un]reserve_general_reg()` private given they're only being used publicly here, we could do something like:
```cpp
auto [new_elem, reserv] = [&] {
  RegReserv const elem_reserv{elem};
  return alloc_general_reg();
}();
```

Baltoli: But it depends if you want to keep the granular public approach for future usage; I'd probably have a weak preference to keep them private and keep the RAII wrappers as the only way to reserve registers.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2298,7 +2306,11 @@ namespace monad::vm::compiler::native
             return pre_dst;
         }
 
-        discharge_deferred_comparison();
+        {
+            RegReserv const pre_dst_reserv{pre_dst};
+            RegReserv const pre_src_reserv{pre_src};
+            discharge_deferred_comparison();
+        }
```
Baltoli, line 2313: For this common pattern, we could add an overload to `discharge_deferred_comparison` that takes a pack of references to stack elements that need to be reserved. Not sure if that's clearer or not, but it keeps the code a bit less noisy perhaps? Roughly:
```cpp
template <typename... Ts>
void Emitter::discharge_deferred_comparison(Ts... reserved)
  requires(std::conjunction_v<std::is_same<Ts, StackElemRef>> && sizeof...(Ts) > 0)
{
  auto const reservations = std::array{reserved...};
}
```

dhil: I was thinking the same, but then I was thinking it doesn't really scale the same way as using block syntax, say, if we ever need to sequence multiple actions. I'd be fine either way here.

category/vm/compiler/ir/x86/emitter.cpp
andreaslyn: The problem is that `elem` is a `StackElem&` and `RegReserv` requires `StackElemRef`. Adding a `RegReserv` specialization for `StackElem&` would not be a safe interface either, because the `StackElem&` can get deallocated before `RegReserv` goes out of scope. I did consider an alternative CPS style approach to perform the reserve/unreserve, but I that seems to be an overkill for an internal interface.

category/vm/compiler/ir/x86/emitter.cpp
andreaslyn: The current pattern is to have the caller perform register reservations. It is possible to change it to something like the live set approach, as suggested by @Baltoli. But it is not clear whether this complexity is worthwhile.

PR #1486, opened 2025-08-07
Add Quill-based logging to the VM

### LINKED ISSUE #1434: Compiler: Logging
It will be good to have info from the async compile thread, which is not part of the fast path. Once in a while log
* Info: the average estimated native code size of compiled contracts.
* Info: the average bytecode size of compiled contracts.
* Info: average native code size was a factor of the average bytecode size.
* Info: the maximum estimated native code size of compiled contract.
* Info: the maximum bytecode size of compiled contract.
* Info: number of elements in varcode cache.
* Info: approximate total size of varcode cache elements in MB.

Use proper logging function instead of `std::cerr` in x86.cpp
* Warning: when compilation fails due to estimated native code size too large. 
* Error: when compilation fails for another reason.

When debug logging is enabled we can moreover add logging to the `VM` class. Once in a while log
* Debug: the total number of `execute_intercode` calls.
* Debug: the total number of `execute_native_entrypoint` calls.
* Debug: the total number of `execute_raw` calls.

Since `VM` is part of the fast path, we should put the debug logging related work in unlikely branches,
```
if (MONAD_VM_UNLIKELY(is_debug_logging)) {
    ...
}
```

mkolosick:
- Adds new compile flag MONAD_COMPILER_STATS

closes #1434

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,8 @@ option(MONAD_COMPILER_COVERAGE "Build with coverage" OFF)
 option(MONAD_COMPILER_TESTING "Build compiler tests" OFF)
 option(MONAD_COMPILER_BENCHMARKS "Build compiler benchmarks" OFF)
 option(MONAD_COMPILER_DUMP_ASM "Dump assembly files into build/asm" OFF)
+option(MONAD_COMPILER_STATS "Print statistics about the compiler" OFF)
+option(MONAD_COMPILER_HOT_PATH_STATS "Print statistics about the vm" OFF)
 
 if(NOT DEFINED BUILD_SHARED_LIBS)
   # If this isn't defined, evmone has the audacity to define it itself (to ON)
@@ -75,6 +77,14 @@ function(monad_compile_options target)
 
   if(MONAD_COMPILER_TESTING)
     target_compile_definitions(${target} PUBLIC "MONAD_COMPILER_TESTING=1")
+  endif()
+
+  if(MONAD_COMPILER_STATS)
+      target_compile_definitions(${target} PUBLIC "MONAD_COMPILER_STATS=1")
+  endif()
+
+  if(MONAD_COMPILER_HOT_PATH_STATS)
+      target_compile_definitions(${target} PUBLIC "MONAD_COMPILER_HOT_PATH_STATS=1")
   endif()
 
   target_compile_options(
--- a/category/vm/compiler.hpp
+++ b/category/vm/compiler.hpp
@@ -17,6 +17,8 @@
 
 #include <category/vm/code.hpp>
 #include <category/vm/compiler/ir/x86.hpp>
+#include <category/vm/utils/debug.hpp>
+#include <category/vm/utils/log_utils.hpp>
 #include <category/vm/varcode_cache.hpp>
 
 #include <evmc/evmc.h>
@@ -28,12 +30,110 @@
 #include <asmjit/x86.h>
 
 #include <atomic>
+#include <chrono>
 #include <condition_variable>
 #include <thread>
 
 namespace monad::vm
 {
     using CompilerConfig = compiler::native::CompilerConfig;
+
+    struct CompilerStats
+    {
+        EuclidMean<uint64_t> avg_native_code_size_;
+        EuclidMean<uint64_t> avg_compiled_bytecode_size_;
+        GeoMean<double> avg_native_code_ratio_;
+        EuclidMean<int64_t> avg_compile_time_;
+        std::atomic<uint64_t> max_native_code_size_{0};
+        std::atomic<uint64_t> max_compiled_bytecode_size_{0};
+        std::atomic<uint64_t> num_compiled_contracts_{0};
+        std::atomic<int64_t> max_compile_time_{0};
+        std::atomic<uint64_t> num_unexpected_compilation_errors_{0};
+        std::atomic<uint64_t> num_size_out_of_bound_compilation_errors_{0};
+
+        // must be called non-concurrently
+        void event_new_compiled_code_cached(
+            SharedIntercode const &icode, SharedNativecode const &ncode,
+            auto compile_start, auto compile_end) noexcept
+        {
+            if constexpr (utils::collect_monad_compiler_stats) {
+                switch (ncode->error_code()) {
+                case Nativecode::ErrorCode::Unexpected:
+                    num_unexpected_compilation_errors_.fetch_add(
+                        1, std::memory_order_release);
+                    break;
+
+                case Nativecode::ErrorCode::SizeOutOfBound:
+                    num_size_out_of_bound_compilation_errors_.fetch_add(
+                        1, std::memory_order_release);
+                    break;
+
+                case Nativecode::ErrorCode::NoError:
+                    auto native_code_size_estimate =
+                        ncode->code_size_estimate();
+                    auto bytecode_size = icode->code_size();
+                    avg_native_code_size_.update(native_code_size_estimate);
+                    avg_compiled_bytecode_size_.update(bytecode_size);
+                    if (bytecode_size > 0) {
+                        avg_native_code_ratio_.update(
+                            static_cast<double>(native_code_size_estimate) /
+                            static_cast<double>(bytecode_size));
+                    }
+                    num_compiled_contracts_.fetch_add(
+                        1, std::memory_order_release);
+                    max_native_code_size_ = std::max(
+                        max_native_code_size_.load(std::memory_order_acquire),
+                        native_code_size_estimate);
+                    max_compiled_bytecode_size_ = std::max(
+                        max_compiled_bytecode_size_.load(
+                            std::memory_order_acquire),
+                        bytecode_size);
+                    break;
+                }
+                auto compile_time =
+                    std::chrono::duration_cast<std::chrono::microseconds>(
+                        compile_end - compile_start)
+                        .count();
+                avg_compile_time_.update(compile_time);
+                max_compile_time_ = std::max(
+                    max_compile_time_.load(std::memory_order_acquire),
+                    compile_time);
+            }
+        }
+
+        std::string
+        print_stats(uint64_t cache_size, uint64_t cache_weight) const
+        {
+            if constexpr (utils::collect_monad_compiler_stats) {
+                return std::format(
+                    ",avg_native_code_size={}B,avg_compiled_bytecode_size={}B"
+                    ",avg_native_code_ratio={:.2f}"
+                    ",max_native_code_size={}B,max_compiled_bytecode_size={}B"
+                    ",num_compiled_contracts={}"
+                    ",avg_compile_time={}µs,max_compile_time={}µs"
+                    ",num_unexpected_compilation_errors={},num_size_out_of_"
+                    "bound_compilation_errors={}"
+                    ",varcode_cache_size={},varcode_cache_weight={}kB",
+                    avg_native_code_size_.get(),
+                    avg_compiled_bytecode_size_.get(),
+                    avg_native_code_ratio_.get(),
+                    max_native_code_size_.load(std::memory_order_acquire),
+                    max_compiled_bytecode_size_.load(std::memory_order_acquire),
+                    num_compiled_contracts_.load(std::memory_order_acquire),
+                    avg_compile_time_.get(),
+                    max_compile_time_.load(std::memory_order_acquire),
+                    num_unexpected_compilation_errors_.load(
+                        std::memory_order_acquire),
+                    num_size_out_of_bound_compilation_errors_.load(
+                        std::memory_order_acquire),
+                    cache_size,
+                    cache_weight);
+            }
+            else {
+                return "";
+            }
+        }
+    };
 
     class Compiler
     {
@@ -91,6 +191,12 @@ namespace monad::vm
             return varcode_cache_.set_warm_cache_kb(warm_kb);
         }
 
+        std::string print_stats() const
+        {
+            return stats_.print_stats(
+                varcode_cache_.size(), varcode_cache_.approx_weight());
+        }
+
         // For testing: wait for compile job queue to become empty.
         void debug_wait_for_empty_queue();
 
@@ -116,5 +222,7 @@ namespace monad::vm
         std::atomic_flag stop_flag_;
         size_t compile_job_soft_limit_;
         bool enable_async_compilation_;
+
+        CompilerStats stats_;
     };
 }
--- a/category/vm/compiler/ir/x86.cpp
+++ b/category/vm/compiler/ir/x86.cpp
@@ -25,10 +25,12 @@
 
 #include <asmjit/core/jitruntime.h>
 
+#include <quill/detail/LogMacros.h>
+// clang-tidy incorrectly views this macro as unused
+#include <quill/Quill.h> // IWYU pragma: keep
+
 #include <cstddef>
 #include <cstdint>
-#include <format>
-#include <iostream>
 #include <limits>
 #include <memory>
 #include <span>
@@ -483,19 +485,13 @@ namespace monad::vm::compiler::native
             }
         }
         catch (Emitter::Error const &e) {
-            std::cerr << std::format(
-                             "ERROR: X86 emitter: failed compile: {}", e.what())
-                      << std::endl;
+            LOG_ERROR("ERROR: X86 emitter: failed compile: {}", e.what());
             return std::make_shared<Nativecode>(rt, rev, nullptr, 0);
         }
         catch (SizeEstimateOutOfBounds const &e) {
-            if (config.verbose) {
-                std::cerr
-                    << std::format(
-                           "WARNING: X86 emitter: native code out of bound: {}",
-                           e.size_estimate)
-                    << std::endl;
-            }
+            LOG_WARNING(
+                "WARNING: X86 emitter: native code out of bound: {}",
+                e.size_estimate);
             return std::make_shared<Nativecode>(
                 rt, rev, nullptr, e.size_estimate);
         }
--- a/category/vm/utils/log_utils.hpp
+++ b/category/vm/utils/log_utils.hpp
@@ -1,0 +1,69 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <atomic>
+#include <cmath>
+
+template <class T>
+concept Numeric = std::is_arithmetic_v<T>;
+
+// not safe for concurrent updates
+// can be safely read concurrent to updating
+template <Numeric T>
+struct EuclidMean
+{
+    std::atomic<T> running_avg_{0};
+    std::atomic<T> count_{0};
+
+    void update(T new_value) noexcept
+    {
+        auto count = count_.load(std::memory_order_acquire);
+        auto running_avg = running_avg_.load(std::memory_order_acquire);
+        auto new_avg = (running_avg * count + new_value) / (count + 1);
+        running_avg_.store(new_avg, std::memory_order_release);
+        count_.fetch_add(1, std::memory_order_release);
+    }
+
+    T get() const noexcept
+    {
+        return running_avg_.load(std::memory_order_acquire);
+    }
+};
+
+// not safe for concurrent updates
+// can be safely read concurrent to updating
+template <Numeric T>
+struct GeoMean
+{
+    std::atomic<T> running_avg_{0};
+    std::atomic<T> count_{0};
+
+    void update(T new_value) noexcept
+    {
+        auto count = count_.load(std::memory_order_acquire);
+        auto running_avg = running_avg_.load(std::memory_order_acquire);
+        auto new_avg =
+            (running_avg * count + std::log2(new_value)) / (count + 1);
+        running_avg_.store(new_avg, std::memory_order_release);
+        count_.fetch_add(1, std::memory_order_release);
+    }
+
+    T get() const noexcept
+    {
+        return std::exp2(running_avg_.load(std::memory_order_acquire));
+    }
+};
--- a/category/vm/utils/lru_weight_cache.hpp
+++ b/category/vm/utils/lru_weight_cache.hpp
@@ -117,10 +117,16 @@ namespace monad::vm::utils
         }
 
         /// Get approximate total weight of the cached elements.
-        uint64_t approx_weight()
+        uint64_t approx_weight() const
         {
             return static_cast<uint64_t>(
                 weight_.load(std::memory_order_acquire));
+        }
+
+        /// Return the number of cached elements.
+        size_t size() const noexcept
+        {
+            return hmap_.size();
         }
 
         // For testing: to check internal invariants. Not safe with
--- a/category/vm/varcode_cache.hpp
+++ b/category/vm/varcode_cache.hpp
@@ -63,6 +63,18 @@ namespace monad::vm
         // Cache weight of the given code size.
         static std::uint32_t code_size_to_cache_weight(std::size_t code_size);
 
+        /// Get approximate total weight of the cached elements.
+        uint64_t approx_weight() const
+        {
+            return weight_cache_.approx_weight();
+        }
+
+        /// Return the number of cached elements.
+        size_t size() const noexcept
+        {
+            return weight_cache_.size();
+        }
+
     private:
         WeightCache weight_cache_;
         std::uint32_t warm_cache_kb_;
--- a/category/vm/vm.cpp
+++ b/category/vm/vm.cpp
@@ -98,6 +98,7 @@ namespace monad::vm
         evmc_host_context *context, evmc_message const *msg,
         std::span<uint8_t const> code)
     {
+        stats_.event_execute_raw();
         auto const stack_ptr = stack_allocator_.allocate();
         auto ctx = runtime::Context::from(
             memory_allocator_, host, context, msg, {code.data(), code.size()});
@@ -112,6 +113,7 @@ namespace monad::vm
         evmc_host_context *context, evmc_message const *msg,
         SharedIntercode const &icode)
     {
+        stats_.event_execute_intercode();
         uint8_t const *const code = icode->code();
         size_t const code_size = icode->code_size();
         auto const stack_ptr = stack_allocator_.allocate();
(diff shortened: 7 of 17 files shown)


category/vm/compiler/ir/x86.cpp
```diff
@@ -468,19 +470,13 @@ namespace monad::vm::compiler::native
...
         }
         catch (Emitter::Error const &e) {
-            std::cerr << std::format(
-                             "ERROR: X86 emitter: failed compile: {}", e.what())
-                      << std::endl;
+            LOG_ERROR("ERROR: X86 emitter: failed compile: {}", e.what());
             return std::make_shared<Nativecode>(rt, rev, nullptr, 0);
         }
         catch (SizeEstimateOutOfBounds const &e) {
-            if (config.verbose) {
```
andreaslyn, line 492: I think we can remove the verbose flag now

category/vm/varcode_cache.hpp
```diff
@@ -4,6 +4,37 @@
...
+        {
+            count_.fetch_add(1, std::memory_order_release);
+            sum_.fetch_add(new_value, std::memory_order_release);
+        }
+
+        T get()
+        {
+            return sum_.load(std::memory_order_acquire) /
+                   count_.load(std::memory_order_acquire);
+        }
```
Baltoli: This isn't critical code so won't break anything per se, but I think you can maybe get an incorrect result by interleaving a call to `get()` between the updates to `count_` and `sum_` in `update()`?

category/vm/varcode_cache.hpp
```diff
@@ -4,8 +4,44 @@
...
 {
+#ifdef MONAD_COMPILER_STATS
+    template <class T>
+    concept Numeric = std::is_arithmetic_v<T>;
+
+    // Works so long as sum_ won't overflow.
+    // This is true of `native_code_size` and `bytecode_size` which are
+    // bounded by `code_size_hard_upper_bound` and transcations
+    // respectively. It is also true of `native_code_ratio` due to the above
+    // being bounded.
```
andreaslyn: Nice

cmd/monad/runloop_ethereum.cpp
```diff
@@ -187,7 +187,11 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
             output_header.gas_used /
                 (uint64_t)std::max(1L, block_metrics.tx_exec_time().count()),
             output_header.gas_used / (uint64_t)std::max(1L, block_time.count()),
-            db.print_stats());
+            db.print_stats(),
+            vm.print_compiler_stats());
```
andreaslyn, line 207: Cool

category/vm/vm.hpp
```diff
@@ -68,5 +68,72 @@ namespace monad::vm
...
+
+    private:
+#ifndef NDEBUG
+        struct VmStats
+        {
+            std::atomic<uint64_t> execute_intercode_call_count_{0};
+            std::atomic<uint64_t> execute_native_entrypoint_call_count_{0};
+            std::atomic<uint64_t> execute_raw_call_count_{0};
+
+            // do we care about the atomicity?
```
andreaslyn: Yes, VM is called concurrently

category/vm/varcode_cache.hpp
```diff
@@ -4,8 +4,44 @@
 #include <category/vm/utils/evmc_utils.hpp>
 #include <category/vm/utils/lru_weight_cache.hpp>
 
+#ifdef MONAD_COMPILER_STATS
+namespace
+{
+}
+#endif
+
```
andreaslyn: Typo?

category/vm/varcode_cache.hpp
```diff
@@ -45,11 +81,75 @@ namespace monad::vm
...
+                max_bytecode_size_.store(
+                    std::max(bytecode_size, max_bytecode_size),
+                    std::memory_order_release);
+            }
+
+            std::string print_stats()
+            {
+                std::string str = std::format(
+                    "average_native_code_size={},average_bytecode_size={},"
+                    "average_native_code_ratio={:.2f}"
```
andreaslyn: Let us abbreviate to "avg". It is actually not the average bytecode size, but only the average among compiled contracts.
```suggestion
                    "avg_native_code_size={},avg_compiled_bytecode_size={},"
                    "avg_native_code_ratio={:.2f}"
```

category/vm/varcode_cache.hpp
```diff
@@ -45,11 +81,75 @@ namespace monad::vm
...
     private:
         WeightCache weight_cache_;
         std::uint32_t warm_cache_kb_;
+
+#ifdef MONAD_COMPILER_STATS
+        struct VarcodeCacheStats
+        {
+            Average<uint64_t> average_native_code_size_;
+            Average<uint64_t> average_bytecode_size_;
+            Average<double> average_native_code_ratio_; // average of ratios
```
andreaslyn: This is perhaps unintuitive, but the standard mean is not the correct average of ratios. The geometric mean is the average of ratios. This came as a surprise to me when I realized it. See for an explanation https://www.notion.so/2025-04-11-Average-of-benchmark-results-1d275b0ba8408066be62fb7578f7a01f

category/vm/varcode_cache.hpp
```diff
@@ -45,11 +81,75 @@ namespace monad::vm
...
+                        average_native_code_ratio_.update(
+                            static_cast<double>(native_code_size) /
+                            static_cast<double>(bytecode_size));
+                    }
+                }
+                auto max_native_code_size =
+                    max_native_code_size_.load(std::memory_order_acquire);
+                max_native_code_size_.store(
+                    std::max(native_code_size, max_native_code_size),
+                    std::memory_order_release);
```
andreaslyn: There is potential for a race condition here, when `max_native_code_size_` is updated immediately after loading. This race condition is not reachable at the moment, because `event_new_code_cached` is only called by the compiler thread.

Perhaps we can move these stats to the compiler class instead, to ensure that this event is only invoked synchronously by the compiler thread. We do not need the fields of this be atomic.

andreaslyn: Would be nice to additionally have the average compile time and maximum compile time part of the stats.

category/vm/varcode_cache.hpp
```diff
@@ -4,8 +4,44 @@
...
+    // bounded by `code_size_hard_upper_bound` and transcations
+    // respectively. It is also true of `native_code_ratio` due to the above
+    // being bounded.
+    template <Numeric T>
+    struct Average
+    {
+        std::atomic<T> sum_{0};
+        std::atomic<T> count_{0};
+
+        void update(T new_value)
```
Baltoli: It's not critical here, but it's best practice to mark
- member functions as `const` when possible (if a member function isn't `const`, you can't call it through a non-const reference or pointer, and you end up propagating more mutable references through your code than you might like to)
- functions as `noexcept` when possible. This is much less important than `const`, but can be a subtle footgun in some places (for example, it's possible to pessimise class behaviour with respect to the standard containers if its move constructor isn't `noexcept`).
```suggestion
        void update(T new_value) const noexcept
```

category/vm/varcode_cache.hpp
```diff
@@ -45,11 +81,75 @@ namespace monad::vm
             warm_cache_kb_ = warm_kb;
         }
 
+        std::string print_stats()
```
Baltoli: N.B. this can't be `noexcept` because `std::format` [can throw](https://en.cppreference.com/w/cpp/utility/format/format.html).
```suggestion
        std::string print_stats() const
```

category/vm/varcode_cache.hpp
```diff
@@ -45,11 +81,75 @@ namespace monad::vm
...
+                    std::max(native_code_size, max_native_code_size),
+                    std::memory_order_release);
+                auto max_bytecode_size =
+                    max_bytecode_size_.load(std::memory_order_acquire);
+                max_bytecode_size_.store(
+                    std::max(bytecode_size, max_bytecode_size),
+                    std::memory_order_release);
+            }
+
+            std::string print_stats()
```
Baltoli:
```suggestion
            std::string print_stats() const
```

category/vm/varcode_cache.hpp
```diff
@@ -45,11 +81,75 @@ namespace monad::vm
...
+                std::string str = std::format(
+                    "average_native_code_size={},average_bytecode_size={},"
+                    "average_native_code_ratio={:.2f}"
+                    ",max_native_code_size={},max_bytecode_size={}",
+                    average_native_code_size_.get(),
+                    average_bytecode_size_.get(),
+                    average_native_code_ratio_.get(),
+                    max_native_code_size_.load(std::memory_order_acquire),
+                    max_bytecode_size_.load(std::memory_order_acquire));
+                return str;
```
Baltoli: Trivial:
```suggestion
                return std::format(
                    "average_native_code_size={},average_bytecode_size={},"
                    "average_native_code_ratio={:.2f}"
                    ",max_native_code_size={},max_bytecode_size={}",
                    average_native_code_size_.get(),
                    average_bytecode_size_.get(),
                    average_native_code_ratio_.get(),
                    max_native_code_size_.load(std::memory_order_acquire),
                    max_bytecode_size_.load(std::memory_order_acquire));
```

category/vm/vm.hpp
```diff
@@ -68,5 +68,72 @@ namespace monad::vm
             evmc_host_interface const *, evmc_host_context *,
             evmc_message const *, SharedIntercode const &,
             compiler::native::entrypoint_t);
+
+        std::string print_vm_stats()
```
Baltoli:
```suggestion
        std::string print_vm_stats() const
```

category/vm/vm.hpp
```diff
@@ -68,5 +68,72 @@ namespace monad::vm
...
+        {
+#ifndef NDEBUG
+            return stats_.print_stats();
+#else
+            return "";
+#endif
+        }
+
+        std::string print_compiler_stats()
```
Baltoli:
```suggestion
        std::string print_compiler_stats() const
```

category/vm/vm.hpp
```diff
@@ -68,5 +68,72 @@ namespace monad::vm
...
+            void clear_stats()
+            {
+                execute_intercode_call_count_.store(
+                    0, std::memory_order_release);
+                execute_native_entrypoint_call_count_.store(
+                    0, std::memory_order_release);
+                execute_raw_call_count_.store(0, std::memory_order_release);
+            }
+
+            std::string print_stats()
```
Baltoli: I think there's the potential for this API to be a bit confusing: the other `print_*` methods can be `const` and just return a formatted string, but this one mutates the internal cache state as well.

Baltoli: If you'd rather keep it as it is, then C++ has a useful feature [`nodiscard`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard.html) that will have the compiler shout at you if you don't hang on to the return value from this function.

category/vm/utils/lru_weight_cache.hpp
```diff
@@ -108,6 +108,12 @@ namespace monad::vm::utils
                 weight_.load(std::memory_order_acquire));
         }
 
+        /// Return the number of cached elements.
+        size_t size()
```
Baltoli:
```suggestion
        size_t size() const noexcept
```

category/vm/varcode_cache.cpp
```diff
@@ -12,6 +12,16 @@
 
 namespace monad::vm
 {
+#ifdef MONAD_COMPILER_STATS
```
Baltoli: Rather than checking an `#ifdef` every time we want to dispatch on stats enabled / disabled, it might be a good idea to define a `constexpr` variable based on the flag definition (for example, as we do [here](https://github.com/category-labs/monad/blob/4fac17448ec8246391b6598c7f2940ceb0a9d553/category/vm/utils/debug.hpp) for dispatching on whether the fuzzer is enabled or disabled - though note that this code needs to also have a runtime value in the enabled case, as it's also controlled by an environment variable).

Would look something like this:
```cpp
#ifdef MONAD_COMPILER_STATS
  constexpr inline bool stats_enabled = true;
#else
  constexpr inline bool stats_enabled = false;
#endif
```

Using a `constexpr` variable with `if constexpr (...) { ... }` means that the compiler can do better semantic analysis and warnings, as the else-branch code is deleted textually in a macro-based approach. Generally scoping macro flag usage as high and tightly as possible is good, to the extent that it's possible.

category/vm/varcode_cache.hpp
```diff
@@ -4,8 +4,44 @@
...
 
+#ifdef MONAD_COMPILER_STATS
+namespace
+{
+}
+#endif
+
 namespace monad::vm
 {
+#ifdef MONAD_COMPILER_STATS
```
Baltoli: Let's not gate class definitions behind this flag; it's a small marginal increase to compile time and it's worth making sure this compiles in CI etc.

category/vm/vm.hpp
andreaslyn: Right there is a difference between `VmStats` and the other stats in that `VmStats` is reporting only numbers since last time `print_stats` was called and the others are reporting number based on the entire history. The tps and gps numbers are only based on the value since last log, so I think clearing the stats as in `VmStats::print_stats` is consistent with this.

category/vm/varcode_cache.hpp
andreaslyn: Oh and the total number of compiled contracts is interesting to know as well

category/vm/vm.hpp
mkolosick: That is where I took this pattern from. Resetting feels like a correct design as these counts feel more like a timer to me whereas the other stats are snapshots of a state in a way. 

Some possible alternatives besides `nodiscard` would be changing to the name `print_and_reset_counters` or adding a separate `reset_counters` method. `print_and_reset_counters` seems ideal to me.

Baltoli: `print_and_reset_counters` seems fine to me; happy with that design. It should still be `nodiscard` though, I think

andreaslyn: I think way in which tps/gps is reported at the moment, is to report the numbers since last log, and report the historic numbers immediately before terminating the client. Perhaps we can use the same pattern for all our stats with reasonably few changes?

category/vm/vm.cpp
```diff
@@ -17,6 +17,17 @@
 
 namespace monad::vm
 {
+#ifndef NDEBUG
```
andreaslyn: Instead of NDEBUG, I think we can add a separate MONAD_COMPILER_HOT_PATH_STATS macro to enable/disable stats from the hot path. This would allow stats in a release build. When guarded behind this macro, we can simply print the vm specific stats together with all the other stats.

Baltoli: Yes I'm generally in favour of having our own flags rather than relying on debug mode

category/vm/varcode_cache.hpp
mkolosick: Okay I had "known" that the geomean was supposedly the appropriate mean for this circumstance but when I looked I couldn't find an explanation that clicked for me. This helped and also seeing the log formulation rather than the square root formulation relieves some of my doubts.

andreaslyn: Right, it is not clear to me how the geometric "intuition" of geometric mean is justifying it being the average of ratios.

category/vm/varcode_cache.hpp
mkolosick: Ah yes this is true. I thought this was fine because `get` was only being called in the top-level runloops but the compiler runs async so that's not actually sufficient. Need to rethink that concurrency of this and the max a little more and see if there's a way to avoid a `compare_exchange` loop.

category/vm/compiler/ir/x86.cpp
mkolosick: done

category/vm/varcode_cache.hpp
mkolosick: yes

category/vm/varcode_cache.hpp
mkolosick: done

category/vm/varcode_cache.hpp
mkolosick: changed

category/vm/varcode_cache.hpp
mkolosick: Moved to `compiler.hpp`/`compiler.cpp` and added new stats.

category/vm/varcode_cache.hpp
mkolosick: done

category/vm/varcode_cache.hpp
mkolosick: done

category/vm/varcode_cache.hpp
mkolosick: done

category/vm/vm.hpp
mkolosick: done

category/vm/vm.hpp
mkolosick: done

category/vm/utils/lru_weight_cache.hpp
mkolosick: done

category/vm/varcode_cache.cpp
mkolosick: done

category/vm/varcode_cache.hpp
mkolosick: done

category/vm/vm.cpp
mkolosick: done

category/vm/varcode_cache.hpp
mkolosick: done

category/vm/compiler.hpp
```diff
@@ -13,13 +15,87 @@
...
+                        compile_end - compile_start)
+                        .count();
+                avg_compile_time_.update(compile_time);
+                max_compile_time_ = std::max(max_compile_time_, compile_time);
+            }
+        }
+
+        // TODO: this is called in a separate thread from the compiler so can
+        // race with `event_new_compiled_code_cached`. Technically this allows
+        // data races for all of the stats
```
mkolosick: I have split the vm stats into per block (printed in each `runloop_` method) and overall (printed in `main` before exit) stats. I changed the compiler stats to only log on exit and to track the stats in the compiler thread. Technically we should still need the compiler stats to be atomic as the compiler thread still exists and might be doing work when we are logging in `main`. Do we care about this possibility and if so should these stats be made into atomics?

CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ option(MONAD_COMPILER_COVERAGE "Build with coverage" OFF)
 option(MONAD_COMPILER_TESTING "Build compiler tests" OFF)
 option(MONAD_COMPILER_BENCHMARKS "Build compiler benchmarks" OFF)
 option(MONAD_COMPILER_DUMP_ASM "Dump assembly files into build/asm" OFF)
+option(MONAD_COMPILER_STATS "Print statistics about the compiler" OFF)
+option(MONAD_VM_HOT_PATH_STATS "Print statistics about the vm" OFF)
```
andreaslyn: Let us change to `MONAD_COMPILER_HOT_PATH_STATS` for consistency with the other options.

category/vm/compiler.hpp
```diff
@@ -13,13 +15,87 @@
...
 namespace monad::vm
 {
     using CompilerConfig = compiler::native::CompilerConfig;
 
+    struct CompilerStats
+    {
+        Average<uint64_t> avg_native_code_size_;
+        Average<uint64_t> avg_compiled_bytecode_size_;
+        Average<double> avg_native_code_ratio_; // geomean
+        Average<long> avg_compile_time_;
```
andreaslyn: I think that this technically should be `int64_t`, because `microseconds` requires at least signed 55 bit int type. Same applies to `max_compile_time_`.

category/vm/compiler.hpp
```diff
@@ -13,13 +15,87 @@
...
+
+        // must be called non-concurrently
+        void event_new_compiled_code_cached(
+            SharedIntercode const &icode, SharedNativecode const &ncode,
+            auto compile_start, auto compile_end) noexcept
+        {
+            if constexpr (utils::collect_monad_compiler_stats) {
+                auto native_code_size_estimate = ncode->code_size_estimate();
+                auto bytecode_size = icode->code_size();
+                // TODO: native_code_size = 0 => contract not compiled?
```
andreaslyn: Yes that means compilation has failed. Can you add two counters for that, to track number of failed compilation attempts? There are two ways compilation can fail. See `Nativecode::error_code()`

category/vm/compiler.hpp
andreaslyn: Ouch, good catch. Yes let us change all the fields back to atomic for this reason.

category/vm/utils/log_utils.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+template <class T>
+concept Numeric = std::is_arithmetic_v<T>;
+
+// does not work concurrently
+template <Numeric T>
+struct Average
+{
+    T running_avg_{0};
```
andreaslyn:
```suggestion
    T total_{0};
```

category/vm/utils/log_utils.hpp
```diff
@@ -0,0 +1,23 @@
...
+    T count_{0};
+
+    void update(T new_value) noexcept
+    {
+        running_avg_ = (running_avg_ * count_ + new_value) / (count_ + 1);
```
andreaslyn:
```suggestion
        total_ += new_value;
```

category/vm/utils/log_utils.hpp
```diff
@@ -0,0 +1,23 @@
...
+        ++count_;
+    }
+
+    T get() const noexcept
+    {
+        return running_avg_;
```
andreaslyn:
```suggestion
        return total_ / count_;
```

cmd/monad/runloop_monad.cpp
```diff
@@ -226,6 +226,9 @@ Result<std::pair<bytes32_t, uint64_t>> propose_block(
             (uint64_t)std::max(1L, block_metrics.tx_exec_time().count()),
         output_header.gas_used / (uint64_t)std::max(1L, block_time.count()),
         db.print_stats());
+    LOG_DEBUG(
+        "__exec_block_vm,{}",
+        vm.print_and_reset_block_counts());
```
andreaslyn: Now that this is behind a separate macro, we can log together with the other stats at the INFO level.

category/vm/compiler.hpp
```diff
@@ -13,13 +15,87 @@
...
 
 namespace monad::vm
 {
     using CompilerConfig = compiler::native::CompilerConfig;
 
+    struct CompilerStats
+    {
+        Average<uint64_t> avg_native_code_size_;
+        Average<uint64_t> avg_compiled_bytecode_size_;
+        Average<double> avg_native_code_ratio_; // geomean
```
andreaslyn: Consider changing these to `EuclidMean<uint64_t>` and `GeometricMean<double>` respectively, to avoid calculating the base 2 logarithm for each stats event. The `EuclidMean` class would be what is currently called `Average` the `GeometricMean` would be similar, but use multiplication instead of addition and `n`-root instead of division, eg init with `T total_{1}`, update by `total_ *= new_value` and average with `std::pow(total_, 1.0 / count_)`.

mkolosick: I think we need to stick to using `lg2`. At least on the runs I've been doing the geomean ratio is about 12 which is approximately 2^3 so it would only take about 20 multiplications to overflow a 64 bit integer.

mkolosick: *it's a double so it'd be about 300 multiplications but that's still not that many.

category/vm/utils/log_utils.hpp
```diff
@@ -0,0 +1,66 @@
...
+        // ```
+        // total_ += new_value
+        // ++count_
+        // ```
+        // then a `get` can potentially occur between the updates correct?
+        // ```
+        // total_ += new_value
+        // get() ~> { total_ / count_}
+        // ++count_
+        // ```
```
mkolosick: One last TODO decision. I believe we would have to do extra synchronization if we're not keeping the running average. @andreaslyn

andreaslyn: Right, let us keep is as is.

category/vm/compiler.hpp
```diff
@@ -13,13 +15,114 @@
...
+                avg_compile_time_.update(compile_time);
+                max_compile_time_ = std::max(
+                    max_compile_time_.load(std::memory_order_acquire),
+                    compile_time);
+            }
+        }
+
+        // TODO: this is called in a separate thread from the compiler so can
+        // race with `event_new_compiled_code_cached`. Technically this allows
+        // data races for all of the stats
```
andreaslyn: I guess we can change this comment now. Something along the lines of: The printed numbers are not necessarily reflecting exactly the same point in time, in case of concurrent update to stats.

Not worried about the issues with concurrency here, but an alternative would be to print the stats from the compiler thread instead of printing from main thread.

cmd/monad/main.cpp
```diff
@@ -408,7 +408,8 @@ int main(int const argc, char const *argv[])
             "Finish running, finish(stopped) block number = {}, "
             "number of blocks run = {}, time_elapsed = {}, num transactions = "
             "{}, "
-            "tps = {}, gps = {} M",
+            "tps = {}, gps = {} M"
+            ",{}",
```
andreaslyn: I think that this will print just a comma when printing of compiler stats is disabled

cmd/monad/main.cpp
```diff
@@ -417,7 +418,11 @@ int main(int const argc, char const *argv[])
                 std::max(1UL, static_cast<uint64_t>(elapsed.count())),
             result.assume_value().second /
                 (1'000'000 *
-                 std::max(1UL, static_cast<uint64_t>(elapsed.count()))));
+                 std::max(1UL, static_cast<uint64_t>(elapsed.count()))),
+            vm.print_compiler_stats());
+        LOG_DEBUG(
+            "vm_counts, {}",
+            vm.print_total_counts());
```
andreaslyn:
```suggestion
            vm.print_compiler_stats(),
            vm.print_total_counts());
```
(including the necessary change to the format string)

category/vm/utils/log_utils.hpp
andreaslyn:
```suggestion
```

cmd/monad/runloop_ethereum.cpp
```diff
@@ -187,7 +187,8 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
             output_header.gas_used /
                 (uint64_t)std::max(1L, block_metrics.tx_exec_time().count()),
             output_header.gas_used / (uint64_t)std::max(1L, block_time.count()),
-            db.print_stats());
+            db.print_stats(),
+            vm.print_and_reset_block_counts());
```
andreaslyn: It is interesting to know how the compiler stats are evolving, so let us print these once in a while as well. The compiler stats are not just based on the previous block, but printing them here is alright, I think.

test/vm/unit/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 add_executable(vm-unit-tests
+    "${PROJECT_SOURCE_DIR}/test/unit/common/src/test/main.cpp"
```
andreaslyn, line 17: Why is this necessary?

mkolosick: Quill needs to be initialized. I took it from `category/mpt/CmakeLists.txt'. Basically it pulls in `test/unit/common/include/monad/test/environment.hpp` which starts Quill up top before running the tests.

PR #1491, opened 2025-08-08
[cmake] build libblst.a ourselves

kjcamann:
The build system for libblst.a is a custom bash script. Previously we used ExternalProject_Add to build it, then created an imported target (blst::blst) to link the compiled build artifact.

When targets using libblst.a are built via the Rust build system (build.rs calling CMake) this intermittently fails in the CI environment and this failure is not reproducible anywhere else.

Since this is only two files, just build it ourselves for now. The compile options and definitions we use are the same as are echoed to terminal when running `build.sh` manually.

### DIFF
--- a/cmake/blst.cmake
+++ b/cmake/blst.cmake
@@ -1,22 +1,17 @@
-include(ExternalProject)
-
 set(BLST_SOURCE_DIR "${PROJECT_SOURCE_DIR}/third_party/blst")
-set(BLST_CC "${CMAKE_C_COMPILER} ${CMAKE_C_FLAGS}")
 
-ExternalProject_Add(
-    blst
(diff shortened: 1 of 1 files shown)


cmake/blst.cmake
```diff
@@ -1,22 +1,17 @@
...
-)
+enable_language(ASM)
+
+add_library(blst STATIC
+    "${BLST_SOURCE_DIR}/build/assembly.S"
+    "${BLST_SOURCE_DIR}/src/server.c")
+
+target_include_directories(blst PUBLIC ${BLST_SOURCE_DIR}/bindings)
+
+# The compiltion options and defintions match what build.sh would do; if you
```
Baltoli:
```suggestion
# The compilation options and definitions match what build.sh would do; if you
```

Issue #1460, opened 2025-05-02
Use the safe bin type for byte code size


andreaslyn:
We are currently using `size_t` for code size. It would be good to have a stronger guarantee about the size of code, so that we can safely make assumptions about code size, such as coercing to `uint32_t` and knowing that emitter will not run out of memory. We can reuse the runtime `Bin` type for this purpose. Change code size everywhere to this the safe bin tyoe, for example changes will be required in the `Intercode` and also the x86 compiler functions with span or code + code 
 size arguments.

PR #1494, opened 2025-08-11
Switch to the Bin type for code size in the vm

goodlyrottenapple:
Implements #1460

### DIFF
--- a/category/vm/compiler.cpp
+++ b/category/vm/compiler.cpp
@@ -26,6 +26,7 @@
 #include <cstddef>
 #include <memory>
 #include <thread>
+#include <variant>
 
 namespace monad::vm
 {
@@ -61,7 +62,7 @@ namespace monad::vm
         CompilerConfig const &config)
     {
         return compiler::native::compile(
-            asmjit_rt_, {icode->code(), icode->code_size()}, rev, config);
+            asmjit_rt_, icode->code(), icode->code_size(), rev, config);
     }
 
     SharedNativecode Compiler::cached_compile(
@@ -146,7 +147,7 @@ namespace monad::vm
                     code_hash,
                     icode,
                     std::make_shared<Nativecode>(
-                        asmjit_rt_, revision, nullptr, 0));
+                        asmjit_rt_, revision, nullptr, std::monostate{}));
             }
 
             bool const erase_ok = compile_job_map_.erase(acc);
--- a/category/vm/compiler.hpp
+++ b/category/vm/compiler.hpp
@@ -44,8 +44,8 @@ namespace monad::vm
         utils::EuclidMean<uint64_t> avg_compiled_bytecode_size_;
         utils::GeoMean<double> avg_native_code_ratio_;
         utils::EuclidMean<int64_t> avg_compile_time_;
-        std::atomic<uint64_t> max_native_code_size_{0};
-        std::atomic<uint64_t> max_compiled_bytecode_size_{0};
+        std::atomic<uint32_t> max_native_code_size_{0};
+        std::atomic<uint32_t> max_compiled_bytecode_size_{0};
         std::atomic<uint64_t> num_compiled_contracts_{0};
         std::atomic<int64_t> max_compile_time_{0};
         std::atomic<uint64_t> num_unexpected_compilation_errors_{0};
@@ -70,8 +70,8 @@ namespace monad::vm
 
                 case Nativecode::ErrorCode::NoError:
                     auto native_code_size_estimate =
-                        ncode->code_size_estimate();
-                    auto bytecode_size = icode->code_size();
+                        *ncode->code_size_estimate();
+                    auto bytecode_size = *icode->code_size();
                     avg_native_code_size_.update(native_code_size_estimate);
                     avg_compiled_bytecode_size_.update(bytecode_size);
                     if (bytecode_size > 0) {
--- a/category/vm/compiler/ir/basic_blocks.hpp
+++ b/category/vm/compiler/ir/basic_blocks.hpp
@@ -20,6 +20,7 @@
 #include <category/vm/core/assert.h>
 #include <category/vm/core/cases.hpp>
 #include <category/vm/evm/opcodes.hpp>
+#include <category/vm/interpreter/intercode.hpp>
 
 #include <evmc/evmc.h>
 
@@ -220,12 +221,36 @@ namespace monad::vm::compiler::basic_blocks
          * Construct basic blocks from a bytecode program.
          */
         template <evmc_revision Rev = EVMC_LATEST_STABLE_REVISION>
-        explicit BasicBlocksIR(
-            std::span<std::uint8_t const>, RevisionMarker<Rev> = {});
+        BasicBlocksIR(
+            std::uint8_t const *, interpreter::code_size_t,
+            RevisionMarker<Rev> = {});
 
         template <evmc_revision Rev = EVMC_LATEST_STABLE_REVISION>
-        BasicBlocksIR(
-            std::initializer_list<std::uint8_t>, RevisionMarker<Rev> = {});
+        [[gnu::always_inline]]
+        static constexpr BasicBlocksIR unsafe_from(
+            std::initializer_list<std::uint8_t const> bytes,
+            RevisionMarker<Rev> rm = {})
+        {
+            MONAD_VM_ASSERT(bytes.size() <= *interpreter::code_size_t::max());
+            return BasicBlocksIR(
+                std::data(bytes),
+                interpreter::code_size_t::unsafe_from(
+                    static_cast<uint32_t>(bytes.size())),
+                rm);
+        }
+
+        template <evmc_revision Rev = EVMC_LATEST_STABLE_REVISION>
+        [[gnu::always_inline]]
+        static constexpr BasicBlocksIR unsafe_from(
+            std::span<std::uint8_t const> bytes, RevisionMarker<Rev> rm = {})
+        {
+            MONAD_VM_ASSERT(bytes.size() <= *interpreter::code_size_t::max());
+            return BasicBlocksIR(
+                bytes.data(),
+                interpreter::code_size_t::unsafe_from(
+                    static_cast<uint32_t>(bytes.size())),
+                rm);
+        }
 
         /**
          * The basic blocks in the program.
@@ -244,7 +269,7 @@ namespace monad::vm::compiler::basic_blocks
         }
 
         /// Size of bytecode
-        uint64_t codesize;
+        interpreter::code_size_t codesize;
 
         /**
          * Retrieve a block by its identifier.
@@ -332,6 +357,9 @@ namespace monad::vm::compiler::basic_blocks
     template <evmc_revision Rev, typename... Args>
     BasicBlocksIR make_ir(Args &&...);
 
+    template <evmc_revision Rev, typename... Args>
+    BasicBlocksIR unsafe_make_ir(Args &&...);
+
     template <evmc_revision Rev>
     std::variant<Instruction, Terminator, JumpDest> BasicBlocksIR::scan_from(
         std::span<std::uint8_t const> bytes, std::uint32_t &current_offset)
@@ -405,8 +433,9 @@ namespace monad::vm::compiler::basic_blocks
 
     template <evmc_revision Rev>
     BasicBlocksIR::BasicBlocksIR(
-        std::span<std::uint8_t const> bytes, RevisionMarker<Rev>)
-        : codesize(bytes.size())
+        std::uint8_t const *bytes, interpreter::code_size_t byte_count,
+        RevisionMarker<Rev>)
+        : codesize(byte_count)
     {
         using monad::vm::Cases;
 
@@ -423,9 +452,8 @@ namespace monad::vm::compiler::basic_blocks
         auto current_offset = std::uint32_t{0};
         auto first = true;
 
-        auto const byte_count = bytes.size();
-        while (current_offset < byte_count) {
-            auto inst = scan_from<Rev>(bytes, current_offset);
+        while (current_offset < *byte_count) {
+            auto inst = scan_from<Rev>({bytes, *byte_count}, current_offset);
 
             if (first && std::holds_alternative<JumpDest>(inst)) {
                 add_jump_dest();
@@ -455,9 +483,10 @@ namespace monad::vm::compiler::basic_blocks
                         // case we don't want to immediately FallThrough
                         // again, but instead just advance tok and mark
                         // the block as being a jumpdest
-                        if (current_offset < byte_count) {
+                        if (current_offset < *byte_count) {
                             auto next_offset = current_offset;
-                            auto next_inst = scan_from<Rev>(bytes, next_offset);
+                            auto next_inst = scan_from<Rev>(
+                                {bytes, *byte_count}, next_offset);
                             if (std::holds_alternative<JumpDest>(next_inst)) {
                                 current_offset += 1;
                                 add_jump_dest();
@@ -490,17 +519,17 @@ namespace monad::vm::compiler::basic_blocks
         }
     }
 
-    template <evmc_revision Rev>
-    BasicBlocksIR::BasicBlocksIR(
-        std::initializer_list<std::uint8_t> bytes, RevisionMarker<Rev> rm)
-        : BasicBlocksIR(std::span(bytes), rm)
-    {
-    }
-
     template <evmc_revision Rev, typename... Args>
     BasicBlocksIR make_ir(Args &&...args)
     {
         return BasicBlocksIR(
+            std::forward<Args>(args)..., RevisionMarker<Rev>{});
+    }
+
+    template <evmc_revision Rev, typename... Args>
+    BasicBlocksIR unsafe_make_ir(Args &&...args)
+    {
+        return BasicBlocksIR::unsafe_from(
             std::forward<Args>(args)..., RevisionMarker<Rev>{});
     }
 }
--- a/category/vm/compiler/ir/local_stacks.cpp
+++ b/category/vm/compiler/ir/local_stacks.cpp
@@ -255,7 +255,7 @@ namespace monad::vm::compiler::local_stacks
         , codesize(ir.codesize)
     {
         for (auto &blk : ir.blocks()) {
-            blocks.push_back(convert_block(std::move(blk), codesize));
+            blocks.push_back(convert_block(std::move(blk), *codesize));
         }
     }
 
--- a/category/vm/compiler/ir/x86.cpp
+++ b/category/vm/compiler/ir/x86.cpp
@@ -20,6 +20,7 @@
 #include <category/vm/compiler/ir/x86/types.hpp>
 #include <category/vm/compiler/types.hpp>
 #include <category/vm/core/assert.h>
+#include <category/vm/runtime/bin.hpp>
 
 #include <evmc/evmc.h>
 
@@ -33,11 +34,12 @@
 #include <cstdint>
 #include <limits>
 #include <memory>
-#include <span>
+#include <variant>
 
 using namespace monad::vm::compiler;
 using namespace monad::vm::compiler::basic_blocks;
 using namespace monad::vm::compiler::native;
+using namespace monad::vm::interpreter;
 
 namespace
 {
@@ -293,18 +295,13 @@ namespace
         }
     }
 
-    struct SizeEstimateOutOfBounds
-    {
-        size_t size_estimate;
-    };
-
     [[gnu::always_inline]]
-    inline void
-    require_code_size_in_bound(Emitter &emit, size_t max_native_size)
+    inline void require_code_size_in_bound(
+        Emitter &emit, native_code_size_t max_native_size)
     {
         size_t const size_estimate = emit.estimate_size();
-        if (MONAD_VM_UNLIKELY(size_estimate > max_native_size)) {
-            throw SizeEstimateOutOfBounds{size_estimate};
+        if (MONAD_VM_UNLIKELY(size_estimate > *max_native_size)) {
+            throw Nativecode::SizeEstimateOutOfBounds{size_estimate};
         }
     }
 
@@ -324,7 +321,7 @@ namespace
     template <evmc_revision rev>
     void emit_instrs(
         Emitter &emit, Block const &block, int32_t instr_gas,
-        size_t max_native_size, CompilerConfig const &config)
+        native_code_size_t max_native_size, CompilerConfig const &config)
     {
         MONAD_VM_ASSERT(instr_gas <= std::numeric_limits<int32_t>::max());
         int32_t remaining_base_gas = instr_gas;
@@ -410,7 +407,7 @@ namespace
         for (auto const &[d, _] : ir.jump_dests()) {
             emit.add_jump_dest(d);
         }
-        size_t const max_native_size =
+        native_code_size_t const max_native_size =
             max_code_size(config.max_code_size_offset, ir.codesize);
         int32_t accumulated_base_gas = 0;
         for (Block const &block : ir.blocks()) {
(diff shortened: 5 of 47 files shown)


category/vm/interpreter/intercode.hpp
```diff
@@ -36,25 +37,23 @@ namespace monad::vm::interpreter
             return padded_code_;
         }
 
-        std::size_t code_size() const noexcept
+        monad::vm::code_size_t code_size() const noexcept
```
andreaslyn: We can add a `size_t size()` function, which would be a simple version for external users.

category/vm/runtime/bin.hpp
```diff
@@ -109,4 +117,18 @@ namespace monad::vm::runtime
...
+    {
+        return Bin<std::min(M, N)>::unsafe_from(std::min(*x, *y));
+    }
+
+    template <std::size_t M, std::size_t N>
+    [[gnu::always_inline]]
+    constexpr bool operator<(Bin<M> lhs, Bin<N> rhs) noexcept
+    {
+        return *lhs < *rhs;
+    }
```
andreaslyn: This should not be necessary
```suggestion
```

category/vm/runtime/bin.hpp
```diff
@@ -109,4 +117,18 @@ namespace monad::vm::runtime
     {
         return Bin<std::max(M, N)>::unsafe_from(std::max(*x, *y));
     }
+
+    template <std::size_t M, std::size_t N>
+    [[gnu::always_inline]]
+    constexpr Bin<std::min(M, N)> min(Bin<M> x, Bin<N> y) noexcept
```
andreaslyn, line 138: Need a comprehensive unit test for this function.

category/vm/core/types.hpp
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include <category/vm/runtime/bin.hpp>
```
andreaslyn: This is a cyclic dependency: runtime depends on core and core depends on runtime.

category/vm/core/types.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <span>
+
+namespace monad::vm
+{
+    using code_size_t = monad::vm::runtime::Bin<20>;
```
andreaslyn: This can be part of `intercode.hpp`, where it is needed.

category/vm/core/types.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    template <std::uint32_t x>
+    static constexpr code_size_t code_size = code_size_t::unsafe_from(x);
```
andreaslyn: This unsafe function can be removed. One can use `bin<x>` instead, which is safe.

category/vm/core/types.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    using native_code_size_t = monad::vm::runtime::Bin<27>;
```
andreaslyn: This can be moved to `x86/types.hpp`

category/vm/core/types.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    class code_span : public std::span<uint8_t const>
```
andreaslyn: Consider removing this class. Otherwise, needs to be moved.

category/vm/compiler/ir/x86/types.hpp
```diff
@@ -90,14 +94,16 @@ namespace monad::vm::compiler::native
     /// is possible to relax this hard upper bound, but native code
     /// size should be smaller than 2GB to avoid overflowing relative
     /// offsets of type `int32_t`.
-    constexpr uint64_t code_size_hard_upper_bound = 1 << 30; // 1GB
+    constexpr monad::vm::runtime::Bin<31> code_size_hard_upper_bound =
```
andreaslyn: Hmm, how can `code_size_hard_upper_bound` be `Bin<31>` and `native_code_size_t` only `Bin<27>`?

category/vm/compiler/ir/x86.hpp
```diff
@@ -39,7 +42,6 @@ namespace monad::vm::compiler::native
         // compiled, when `offset` is zero. There is a theoretical hard
         // upper bound on native code size to ensure that the emitter
         // will not overflow relative x86 memory addressing offsets.
-        return std::min(
-            offset + 32 * uint64_t{bytecode_size}, code_size_hard_upper_bound);
```
andreaslyn, line 58: Why is the hard upper bound on code size removed?

category/vm/compiler/ir/x86/types.hpp
```diff
@@ -45,9 +47,11 @@ namespace monad::vm::compiler::native
             return revision_;
         }
 
-        size_t code_size_estimate() const
+        monad::vm::native_code_size_t code_size_estimate() const
         {
-            return entrypoint_ ? code_size_estimate_ : 0;
+            return entrypoint_ ? monad::vm::native_code_size_t::unsafe_from(
+                                     static_cast<uint32_t>(code_size_estimate_))
+                               : monad::vm::native_code_size_t{};
```
andreaslyn: Should stay away from using `unsafe_from` unless we are 1000% sure what we are doing

category/vm/compiler/ir/x86/emitter.hpp
```diff
@@ -1005,7 +1006,7 @@ namespace monad::vm::compiler::native
         Stack stack_;
         bool keep_stack_in_next_block_;
         std::array<Gpq256, 3> gpq256_regs_;
-        uint64_t bytecode_size_;
+        monad::vm::code_size_t bytecode_size_;
```
andreaslyn: Generally, if it is necessary to write `monad::vm` from inside a `monad::vm` namespace, there is a problematic naming conflict which needs to be resolved.

category/vm/compiler.cpp
```diff
@@ -45,7 +46,10 @@ namespace monad::vm
         CompilerConfig const &config)
     {
         return compiler::native::compile(
-            asmjit_rt_, {icode->code(), icode->code_size()}, rev, config);
+            asmjit_rt_,
+            monad::vm::code_span{icode->code(), icode->code_size()},
```
Baltoli: This code is in `monad::vm` already:
```suggestion
           code_span{icode->code(), icode->code_size()},
```

category/vm/compiler/ir/basic_blocks.hpp
```diff
@@ -229,7 +229,7 @@ namespace monad::vm::compiler::basic_blocks
         }
 
         /// Size of bytecode
-        uint64_t codesize;
+        monad::vm::code_size_t codesize;
```
Baltoli: Again inside the `monad::vm` namespace:
```suggestion
       code_size_t codesize;
```

category/vm/compiler/ir/x86.hpp
andreaslyn: Oh nice, I see why the hard upper bound is unnecessary now. In this case, we can remove all references to `code_size_hard_upper_bound` in the code (including comments)

category/vm/compiler/ir/x86/types.hpp
andreaslyn: Having the `code_size_estimate_` field be `native_code_size_t` is perhaps better

test/vm/unit/runtime/fixture.cpp
```diff
@@ -75,7 +75,8 @@ namespace monad::vm::compiler::test
                       .code = &code_[0],
                       .return_data = {},
                       .input_data_size = sizeof(call_data_),
-                      .code_size = sizeof(code_),
+                      .code_size = code_size_t::unsafe_from(
+                          static_cast<std::uint32_t>(sizeof(code_))),
```
andreaslyn:
```suggestion
                      .code_size = bin<sizeof(code_)>,
```

category/vm/compiler/ir/local_stacks.cpp
```diff
@@ -81,7 +82,8 @@ namespace
     }
 
     void eval_instruction(
-        Instruction const &tok, std::deque<Value> &stack, uint64_t codesize)
+        Instruction const &tok, std::deque<Value> &stack,
+        monad::vm::code_size_t codesize)
```
Baltoli: One last case... will leave the rest uncommented so as not to noise up the review

category/vm/compiler/ir/x86/types.hpp
goodlyrottenapple: Yeah, I wasn't too keen on this. however, we set the value of `code_size_estimate_` past `code_size_hard_upper_bound` when the code size bound check fails. Would using an std::variant<native_code_size_t, size_t> perhaps be better? Then the implicit invariant would be that when `entrypoint_` is not null, the variant should hold a `native_code_size_t`

category/vm/compiler/ir/x86.cpp
```diff
@@ -307,7 +308,8 @@ namespace
     template <evmc_revision rev>
     void emit_instrs(
         Emitter &emit, Block const &block, int32_t instr_gas,
-        size_t max_native_size, CompilerConfig const &config)
+        monad::vm::runtime::Bin<27> max_native_size,
```
Baltoli: Several repetitions of `Bin<27>` in this file; is it worth a proper type definition?

category/vm/core/types.hpp
goodlyrottenapple: I added this to basically not have to invoke the `unsafe_from` casting of `size_t` down to a `code_size_t` in the various places we currently wrap the code pointer and size argument in std::variant.

category/vm/runtime/bin.hpp
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
Baltoli, line 16: Oops, nice catch

category/vm/core/types.hpp
andreaslyn: Looks like this literally is unsafe_from. Just use bin<x> instead

andreaslyn: Oh never mind, I am referring to code_size

andreaslyn: Just use std::span in everything before interpreter

andreaslyn: Note that just because there is a MONAD_VM_ASSERT does not mean it is safe to use unsafe_from

category/vm/compiler/ir/x86/types.hpp
andreaslyn: Need to reorg the code to fix this. My first attempt would be to make estimate_size include the check and then return std::option native_code_size_t

category/vm/core/types.hpp
goodlyrottenapple: ~~I was under the impression bin only works on constan/compile-time arguments, i.e. can′t do `bin<code.size()>`?~~ sorry github wasn't showing lates comments

andreaslyn: Yes. What I mean is that merely adding MONAD_VM_ASSERT does not make the code safe, because of the possibility of assertion failure. We are going to add this assertion as precondition to the external interface, perhaps just to intercode is were we need it.

category/vm/compiler/ir/x86.cpp
```diff
@@ -410,59 +412,77 @@ namespace
         }
         size_t const size_estimate = emit.estimate_size();
         auto entry = emit.finish_contract(rt);
-        return std::make_shared<Nativecode>(rt, rev, entry, size_estimate);
+        MONAD_VM_DEBUG_ASSERT(size_estimate <= *max_native_size);
```
goodlyrottenapple, line 427: We know this has to hold because `require_code_size_in_bound` is called just before, within the loop, otherwise `SizeEstimateOutOfBounds` would have been thrown. Hence it should be safe to cast `size_estimate` to `native_code_size_t`

category/vm/compiler/ir/basic_blocks.hpp
```diff
@@ -406,7 +407,8 @@ namespace monad::vm::compiler::basic_blocks
     template <evmc_revision Rev>
     BasicBlocksIR::BasicBlocksIR(
         std::span<std::uint8_t const> bytes, RevisionMarker<Rev>)
-        : codesize(bytes.size())
+        : codesize(interpreter::code_size_t::unsafe_from(
+              static_cast<uint32_t>(bytes.size())))
```
andreaslyn: The at the point of creating the basic block IR we already have the code size part of intercode, which we can pass. So unsafe is not necessary here.

We can have an unsafe function for constructing basic blocks for testing purposes only. For example, an `unsafe_from` static function as part of `BasicBlocksIR`,
```
static BasicBlocksIR unsafe_from(uint8_t const *code, size_t sz) {
    MONAD_VM_ASSERT(sz <= code_size_t::upper);
    return BasicBlockIR(code, code_size_t::unsafe_from(sz));
}
```

goodlyrottenapple: Will rejigg this

category/vm/compiler/ir/x86.hpp
```diff
@@ -54,7 +58,6 @@ namespace monad::vm::compiler::native
...
+        return offset + monad::vm::runtime::bin<32> * bytecode_size;
```
andreaslyn: We can restrict `native_code_size_t` to `Bin<26>`, I think:
```suggestion
        return offset + shl<5>(bytecode_size);
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -320,10 +321,10 @@ namespace monad::vm::compiler::native
     asmjit::x86::Mem Emitter::RoData::add32(uint256_t const &x)
     {
         // The total byte size of the `data_` vector is bounded via
-        // `code_size_hard_upper_bound`. We need `data_` size upper
+        // `native_code_size_t`. We need `data_` size upper
         // bounded to not overflow `int32_t` below.
-        static_assert(code_size_hard_upper_bound <= (uint64_t{1} << 31));
-        MONAD_VM_ASSERT(data_.size() < (code_size_hard_upper_bound >> 4));
+        static_assert(*native_code_size_t::max() < (uint64_t{1} << 31));
+        MONAD_VM_ASSERT(data_.size() < (*native_code_size_t::max() >> 4));
```
andreaslyn: This code was never particularly robust. I think we can improve by just doing
```c++
        // We need `data_` size upper bounded to not overflow `int32_t`
        if (data.size() >= (1 << 26)) {
            throw SizeEstimateOutOfBounds{estimate_size()};
        }
```
Just need to move `SizeEstimateOutOfBounds` to x86/types.hpp.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -364,10 +365,10 @@ namespace monad::vm::compiler::native
     asmjit::x86::Mem Emitter::RoData::add(std::array<uint8_t, N> const &x)
     {
         // The total byte size of the `data_` vector is bounded via
-        // `code_size_hard_upper_bound`. We need `data_` size upper
+        // `native_code_size_t`. We need `data_` size upper
         // bounded to not overflow `int32_t` below.
-        static_assert(code_size_hard_upper_bound <= (uint64_t{1} << 31));
-        MONAD_VM_ASSERT(data_.size() < (code_size_hard_upper_bound >> 4));
+        static_assert(*native_code_size_t::max() < (uint64_t{1} << 31));
+        MONAD_VM_ASSERT(data_.size() < (*native_code_size_t::max() >> 4));
```
andreaslyn: Same approach is better here as well.

category/vm/compiler/ir/x86.cpp
andreaslyn: Agreed

category/vm/compiler/ir/x86/types.hpp
```diff
@@ -38,7 +45,8 @@ namespace monad::vm::compiler::native
         /// If compilation failed, then `entrypoint` is `nullptr`.
         Nativecode(
             asmjit::JitRuntime &asmjit_rt, evmc_revision rev,
-            entrypoint_t entry, size_t code_size_estimate)
+            entrypoint_t entry,
+            std::variant<native_code_size_t, size_t> code_size_estimate)
```
andreaslyn: Let us have a type alias for this in `Nativecode`. Add an additional case:
```c++
// monostate -> unexpected error
// size_t -> native code size out of bound error
// native_code_size_t -> success
using CodeSizeEstimate = std::variant<std::monostate, size_t, native_code_size_t>
```

andreaslyn: Need to pass `std::monostate{}` to the `Nativecode` after catching unexpected `Error` in `x86.cpp`

andreaslyn: Debug assert in this `Nativecode` ctor to check and document the new invariant which has been introduced
```c++
MONAD_VM_DEBUG_ASSERT(
    !!entrypoint_ == std::holds_alternative<native_code_size_t>());
```

category/vm/compiler/ir/x86.cpp
```diff
@@ -425,59 +422,77 @@ namespace
...
     template <evmc_revision Rev>
     std::shared_ptr<Nativecode> compile_contract(
-        asmjit::JitRuntime &rt, std::span<uint8_t const> contract,
-        CompilerConfig const &config)
+        asmjit::JitRuntime &rt, std::uint8_t const *contract_code,
+        code_size_t contract_code_size, CompilerConfig const &config)
     {
-        auto ir = BasicBlocksIR(basic_blocks::make_ir<Rev>(contract));
+        auto ir = BasicBlocksIR(
+            basic_blocks::make_ir<Rev>(contract_code, contract_code_size));
```
andreaslyn: There was an unnecessary invocation of the move ctor here it looks like
```suggestion
        auto ir = basic_blocks::make_ir<Rev>(contract_code, contract_code_size);
```

category/vm/interpreter/intercode.cpp
```diff
@@ -26,7 +26,8 @@ namespace monad::vm::interpreter
 {
     Intercode::Intercode(std::span<std::uint8_t const> const code)
         : padded_code_(pad(code))
-        , code_size_(code.size())
+        , code_size_(
+              code_size_t::unsafe_from(static_cast<uint32_t>(code.size())))
         , jumpdest_map_(find_jumpdests(code))
     {
```
andreaslyn, line 33: Let us add (non-debug) `MONAD_VM_ASSERT(code.size() <= code_size_t::upper)` to the body of this constructor, because this is part of the external VM interface.

category/vm/llvm/execute.cpp
```diff
@@ -61,7 +61,7 @@ namespace monad::vm::llvm
         auto ptr = std::make_shared<LLVMState>();
         LLVMState &llvm = *ptr;
 
-        auto ir = BasicBlocksIR(make_ir<Rev>(code));
+        auto ir = BasicBlocksIR(unsafe_make_ir<Rev>(code));
```
andreaslyn: This is not production code, but looks like an unnecessary move ctor application.

cmd/vm/parser/parser_tool.cpp
```diff
@@ -141,7 +143,8 @@ int main(int argc, char **argv)
             auto rt = asmjit::JitRuntime{};
             monad::vm::compiler::native::compile(
                 rt,
-                opcodes,
+                opcodes.data(),
```
andreaslyn, line 146: Let us add
```
MONAD_VM_ASSERT(opcodes.size() <= code_size_t::upper);
```
before the above line `if (args.compile) {`

test/vm/compile_benchmarks/compile_benchmarks.cpp
```diff
@@ -74,7 +75,11 @@ namespace
 
         for (auto _ : state) {
             auto fn = monad::vm::compiler::native::compile(
-                rt, program, EVMC_LATEST_STABLE_REVISION);
+                rt,
+                program.data(),
+                monad::vm::interpreter::code_size_t::unsafe_from(
```
andreaslyn, line 82: Also here, `MONAD_VM_ASSERT`  before the line `auto rt = asmjit::JitRuntime{};`

test/vm/compile_benchmarks/compile_benchmarks.cpp
```diff
@@ -109,7 +114,11 @@ namespace
 
         for (auto _ : state) {
             auto ncode = monad::vm::compiler::native::compile(
-                rt, program, EVMC_LATEST_STABLE_REVISION);
+                rt,
+                program.data(),
+                monad::vm::interpreter::code_size_t::unsafe_from(
```
andreaslyn, line 123: And MONAD_VM_ASSERT, before the line `auto rt = asmjit::JitRuntime{};`

category/vm/interpreter/intercode.cpp
andreaslyn: This was always also a pre-condition before these changes, btw

andreaslyn: Perhaps adding the `MONAD_VM_ASSERT` as pre-condition to the `pad` function is better, so that `pad` will not run out of memory before hitting the assertion failure.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -319,11 +320,10 @@ namespace monad::vm::compiler::native
 
     asmjit::x86::Mem Emitter::RoData::add32(uint256_t const &x)
     {
-        // The total byte size of the `data_` vector is bounded via
-        // `code_size_hard_upper_bound`. We need `data_` size upper
-        // bounded to not overflow `int32_t` below.
-        static_assert(code_size_hard_upper_bound <= (uint64_t{1} << 31));
-        MONAD_VM_ASSERT(data_.size() < (code_size_hard_upper_bound >> 4));
+        // We need `data_` size upper bounded to not overflow `int32_t`
+        if (MONAD_VM_UNLIKELY(data_.size() > *native_code_size_t::max() >> 4)) {
```
andreaslyn: This is actually unrelated to `native_code_size_t`. I think the tightest bound is
```suggestion
        if (MONAD_VM_UNLIKELY(data_.size() >= (1 << 26))) {
```
because `data_.size() < (1 << 26)` guarantees that `data_.size() << 5` will not overflow `int32_t`.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -363,11 +363,10 @@ namespace monad::vm::compiler::native
     template <size_t N>
     asmjit::x86::Mem Emitter::RoData::add(std::array<uint8_t, N> const &x)
     {
-        // The total byte size of the `data_` vector is bounded via
-        // `code_size_hard_upper_bound`. We need `data_` size upper
-        // bounded to not overflow `int32_t` below.
-        static_assert(code_size_hard_upper_bound <= (uint64_t{1} << 31));
-        MONAD_VM_ASSERT(data_.size() < (code_size_hard_upper_bound >> 4));
+        // We need `data_` size upper bounded to not overflow `int32_t`
+        if (MONAD_VM_UNLIKELY(data_.size() > *native_code_size_t::max() >> 4)) {
```
andreaslyn: Same applies here

category/vm/compiler/ir/x86/emitter.cpp
goodlyrottenapple: Ahh thanks for the clarification. I've tweaked the comment above to make the intent clear.

category/vm/compiler.hpp
```diff
@@ -70,8 +70,8 @@ namespace monad::vm
 
                 case Nativecode::ErrorCode::NoError:
                     auto native_code_size_estimate =
-                        ncode->code_size_estimate();
-                    auto bytecode_size = icode->code_size();
+                        *ncode->code_size_estimate();
+                    auto bytecode_size = *icode->code_size();
                     avg_native_code_size_.update(native_code_size_estimate);
                     avg_compiled_bytecode_size_.update(bytecode_size);
```
goodlyrottenapple, line 76: Should this variable be called `avg_compiled_bytecode_size_`? As far as I can tell it tracks the original evm bytecode size (plus some padding)

andreaslyn: The padding is not included in the code size

goodlyrottenapple: Ah ok, still, my question is why do we call it "compiled" bytecode size?

andreaslyn: It is the average size of the bytecodes which have been compiled.

category/vm/compiler.hpp
```diff
@@ -40,12 +40,12 @@ namespace monad::vm
 
     struct CompilerStats
     {
-        EuclidMean<uint64_t> avg_native_code_size_;
-        EuclidMean<uint64_t> avg_compiled_bytecode_size_;
+        EuclidMean<uint32_t> avg_native_code_size_;
+        EuclidMean<uint32_t> avg_compiled_bytecode_size_;
```
andreaslyn: I think we can overflow `uint32_t` in `EuclidMean`. For example, if average contract size is `1 << 16` and `count` is `1 << 17`.

andreaslyn: But this may instead be an issue with `EuclidMean`

category/vm/compiler.hpp
goodlyrottenapple: Ah, just a confusing name. So just so I understand correctly, it is the average EVM bytecode size of all the contracts which have been ran through the compiler, but not the average compiled size (i.e. x86 emitted assembly)

category/vm/compiler.hpp
goodlyrottenapple: I have changed it back just in case anyway

category/vm/compiler.hpp
andreaslyn: Yes it is the average compiled bytecode size

PR #1499, opened 2025-08-12
Re-enable disassembly tests

m-alvarez:
1. Restore the disassembly tests that were silently suppressed when moving files around and updating to clang-19
2. Extend the test script to add support for running disassembly tests on both gcc and clang
3. Update the expected disassembly results
  * As seen [here](https://godbolt.org/z/G3KWEbjT9) changes to `core-spinlock.dis` are due to clang-19
  * The changes in `priority_queue.dis` are consistent with calls to `aggregator_generic::execute` no longer being inlined. This may or may not represent an actual regression


Baltoli (dismissed):

Baltoli:
Superseded

PR #1515, opened 2025-08-15
Add fuzzer for PolyTypedIR inference + type-checker

goodlyrottenapple:


test/vm/fuzzer/typechecker_fuzzer.cpp
```diff
@@ -0,0 +1,220 @@
...
+            [](auto &) { return GeneratorFocus::Generic; },
+            Choice(0.05, [](auto &) { return GeneratorFocus::Pow2; }),
+            Choice(0.8, [](auto &) { return GeneratorFocus::DynJump; }));
+
+        if (rev >= EVMC_PRAGUE && toss(engine, 0.01)) {
+            auto precompile =
+                monad::vm::fuzzing::generate_precompile_address(engine, rev);
+            known_addresses.push_back(precompile);
+        }
+
```
andreaslyn: We can remove this, because contract addresses are not affecting type inference

PR #1520, opened 2025-08-15
EIP-2935 get block hash history

dhil:


category/execution/ethereum/execute_block.cpp
```diff
@@ -169,6 +200,7 @@ Result<std::vector<ExecutionResult>> execute_block(
     MONAD_ASSERT(senders.size() == block.transactions.size());
 
     if constexpr (rev >= EVMC_PRAGUE) {
+        deploy_block_hash_history_contract(block_state, block.header);
```
Baltoli: I don't think we want to try this at the start of _every_ block, but rather only at the start of the first block at which Prague is enabled (tricky to build this given we don't yet have that forkpoint, so until we do so this draft PR is fine to try the deployment code out)

dhil: Yes, that is the thing I am unsure about, where to put the forkpoint. This is meant as a temporary placeholder.

category/execution/ethereum/execute_block.cpp
```diff
@@ -125,6 +125,37 @@ void set_block_hash_history(BlockState &block_state, BlockHeader const &header)
     }
 }
 
+// EIP-2935
+void deploy_block_hash_history_contract(BlockState &block_state, BlockHeader const &header) {
+    constexpr auto HISTORY_STORAGE_ADDRESS{
+        0x0000F90827F1C53a10cb7A02335B175320002935_address};
+    constexpr uint8_t code[] = {
```
Baltoli: Minor nitpick:
```suggestion
    static constexpr auto HISTORY_STORAGE_ADDRESS{
        0x0000F90827F1C53a10cb7A02335B175320002935_address};
    static constexpr uint8_t code[] = {
```

category/execution/ethereum/block_hash_buffer.cpp
```diff
@@ -55,6 +56,36 @@ void BlockHashBufferFinalized::set(uint64_t const n, bytes32_t const &h)
...
+    constexpr uint256_t HISTORY_SERVE_WINDOW{8191};
+    
+    MONAD_ASSERT(n < HISTORY_SERVE_WINDOW);
+    if (state_.account_exists(HISTORY_STORAGE_ADDRESS)) {
+        bytes32_t const key{
+            to_bytes(to_big_endian(n))};
+        // TODO: temporary hack to work around the signature (const &).
+        data_ = state_.get_storage(HISTORY_STORAGE_ADDRESS, key);
+        return data_;
+    }
```
Baltoli: What happens in the window before the ring buffer has been filled up?

Baltoli: The EIP says that zero values get read from slots that haven't been filled in yet; is that the same behaviour as the existing buffer?

category/execution/ethereum/block_hash_buffer.cpp
```diff
@@ -55,6 +56,37 @@ void BlockHashBufferFinalized::set(uint64_t const n, bytes32_t const &h)
     n_ = n + 1;
 }
 
+BlockHashBufferEthCall::BlockHashBufferEthCall(BlockHashBufferFinalized const &fallback, State &state)
```
omegablitz: It looks like after this PR, eth-call still only reads finalized block hashes.
Would we add support for looking up unfinalized block hashes in eth-call in a different PR?

dhil: It will read from the `HISTORY_STORAGE_ADDRESS`, so anything committed there will be retrievable. This is just a temporary fallback to make the testsuite pass until I know which timestamp the fork point will get assigned.

Issue #1524, opened 2025-08-19
Optimize exp


andreaslyn:
The exp instruction is missing optimizations in the x86 emitter.

andreaslyn:
https://github.com/category-labs/monad/pull/1526

PR #1526, opened 2025-08-19
Add fast path for exp instruction

laurenthuberdeau:
Addresses some of the optimizations mentioned in https://github.com/category-labs/monad/issues/1524.

## Optimization coverage

Here's the distribution of inputs and outputs locations for the `EXP` instruction on `main`, since it always calls the runtime function, the result is always on the stack. We use the heuristic `outputs = ["STACK"]` to know if the code generator fell back to the runtime function.

```
         inputs         |  outputs  | count  | proportion 
------------------------+-----------+--------+------------
 ["literal", "literal"] | ["STACK"] | 139329 |      94.13
 ["GPR", "literal"]     | ["STACK"] |   3072 |       2.08
 ["STACK", "literal"]   | ["STACK"] |   2919 |       1.97
 ["STACK", "STACK"]     | ["STACK"] |   1822 |       1.23
 ["AVX", "literal"]     | ["STACK"] |    558 |       0.38
 ["literal", "AVX"]     | ["STACK"] |    195 |       0.13
 ["literal", "STACK"]   | ["STACK"] |     45 |       0.03
 ["literal", "GPR"]     | ["STACK"] |     30 |       0.02
 ["GPR", "STACK"]       | ["STACK"] |     20 |       0.01
 ["STACK", "GPR"]       | ["STACK"] |     11 |       0.01
 ["GPR", "GPR"]         | ["STACK"] |      7 |       0.00
 ["STACK", "AVX"]       | ["STACK"] |      2 |       0.00
 ["AVX", "AVX"]         | ["STACK"] |      1 |       0.00
(13 rows)
```

Here's the new distribution:

```
         inputs         |   outputs   | count  | proportion 
------------------------+-------------+--------+------------
 ["literal", "literal"] | ["literal"] | 139329 |      94.13
 ["STACK", "literal"]   | ["AVX"]     |   2766 |       1.87
 ["GPR", "literal"]     | ["AVX"]     |   2726 |       1.84
 ["STACK", "STACK"]     | ["STACK"]   |   1822 |       1.23
 ["AVX", "literal"]     | ["AVX"]     |    464 |       0.31
 ["GPR", "literal"]     | ["STACK"]   |    346 |       0.23
 ["literal", "AVX"]     | ["STACK"]   |    195 |       0.13
 ["STACK", "literal"]   | ["STACK"]   |    153 |       0.10
 ["AVX", "literal"]     | ["STACK"]   |     94 |       0.06
 ["literal", "STACK"]   | ["STACK"]   |     45 |       0.03
 ["literal", "GPR"]     | ["STACK"]   |     30 |       0.02
 ["GPR", "STACK"]       | ["STACK"]   |     20 |       0.01
 ["STACK", "GPR"]       | ["STACK"]   |     11 |       0.01
 ["GPR", "GPR"]         | ["STACK"]   |      7 |       0.00
 ["STACK", "AVX"]       | ["STACK"]   |      2 |       0.00
 ["AVX", "AVX"]         | ["STACK"]   |      1 |       0.00
(16 rows)
```

All literal/literal cases are handled, meaning no contracts exceed the 500 constant folding limit. Most cases where the base is known are also handled, with less than 2% of all `EXP` instructions falling back to the runtime function.

Breaking down this 2% by literals under 65536, we can see that all the literals (except 0x2) are without obvious optimization (i.e. not powers of 2), except for `exp = 0x2` that calls the `mul` runtime function (`mul_optimized` only works when one of the side is a literal).

```
        inputs        |  outputs  | count | proportion 
----------------------+-----------+-------+------------
 ["STACK", "STACK"]   | ["STACK"] |  1822 |      66.84
 ["GPR", "0xa"]       | ["STACK"] |   344 |      12.62
 ["STACK", "0xa"]     | ["STACK"] |   148 |       5.43
 ["AVX", "0xa"]       | ["STACK"] |    87 |       3.19
 ["0x9", "AVX"]       | ["STACK"] |    48 |       1.76
 ["0x10", "AVX"]      | ["STACK"] |    48 |       1.76
 ["0x7", "AVX"]       | ["STACK"] |    48 |       1.76
 ["0x8", "AVX"]       | ["STACK"] |    48 |       1.76
 ["0x2", "STACK"]     | ["STACK"] |    43 |       1.58
 ["0x2", "GPR"]       | ["STACK"] |    21 |       0.77
 ["GPR", "STACK"]     | ["STACK"] |    20 |       0.73
 ["STACK", "GPR"]     | ["STACK"] |    11 |       0.40
 ["AVX", "literal"]   | ["STACK"] |     7 |       0.26
 ["GPR", "GPR"]       | ["STACK"] |     7 |       0.26
 ["0x2", "AVX"]       | ["STACK"] |     3 |       0.11
 ["0x3", "GPR"]       | ["STACK"] |     3 |       0.11
 ["STACK", "literal"] | ["STACK"] |     2 |       0.07
 ["0x7", "GPR"]       | ["STACK"] |     2 |       0.07
 ["0x8", "GPR"]       | ["STACK"] |     2 |       0.07
 ["0x9", "GPR"]       | ["STACK"] |     2 |       0.07
 ["GPR", "literal"]   | ["STACK"] |     2 |       0.07
 ["STACK", "AVX"]     | ["STACK"] |     2 |       0.07
 ["STACK", "0x9a"]    | ["STACK"] |     1 |       0.04
 ["STACK", "0x42"]    | ["STACK"] |     1 |       0.04
 ["0x3", "STACK"]     | ["STACK"] |     1 |       0.04
 ["AVX", "AVX"]       | ["STACK"] |     1 |       0.04
 ["0x5", "STACK"]     | ["STACK"] |     1 |       0.04
 ["STACK", "0x43"]    | ["STACK"] |     1 |       0.04
(28 rows)
```

### DIFF
--- a/category/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/compiler/ir/x86/emitter.cpp
@@ -97,6 +97,15 @@ namespace
             asmjit::Imm{static_cast<int32_t>(lit.value[3])}};
     }
 
+    Emitter::Mem256 const stack_offset_to_mem256(StackOffset const offset)
+    {
+        return {
+            x86::qword_ptr(x86::rbp, offset.offset * 32),
+            x86::qword_ptr(x86::rbp, offset.offset * 32 + 8),
+            x86::qword_ptr(x86::rbp, offset.offset * 32 + 16),
+            x86::qword_ptr(x86::rbp, offset.offset * 32 + 24)};
+    }
+
     x86::Mem stack_offset_to_mem(StackOffset offset)
     {
         return x86::qword_ptr(x86::rbp, offset.offset * 32);
@@ -587,6 +596,8 @@ namespace monad::vm::compiler::native
         , gpq256_regs_{Gpq256{x86::r12, x86::r13, x86::r14, x86::r15}, Gpq256{x86::r8, x86::r9, x86::r10, x86::r11}, Gpq256{x86::rcx, x86::rsi, x86::rdx, x86::rdi}}
         , bytecode_size_{codesize}
         , rodata_{as_.newNamedLabel("ROD")}
+        , exponential_constant_fold_counter_{0}
+        , accumulated_static_work_{0}
     {
 #ifdef MONAD_VM_TESTING
         as_.addDiagnosticOptions(kValidateAssembler);
@@ -998,17 +1009,23 @@ namespace monad::vm::compiler::native
         return block_prologue(b);
     }
 
-    void Emitter::gas_decrement_no_check(int32_t gas)
+    void Emitter::gas_decrement_static_work(int32_t gas)
     {
-        as_.sub(
-            x86::qword_ptr(reg_context, runtime::context_offset_gas_remaining),
-            gas);
+        if (gas) {
+            gas_decrement_no_check(gas);
+            if (!accumulate_static_work(gas)) {
+                as_.jl(error_label_);
+            }
+        }
     }
 
-    void Emitter::gas_decrement_check_non_negative(int32_t gas)
+    void Emitter::gas_decrement_unbounded_work(int32_t gas)
     {
-        gas_decrement_no_check(gas);
-        as_.jl(error_label_);
+        accumulated_static_work_ = 0;
+        if (gas) {
+            gas_decrement_no_check(gas);
+            as_.jl(error_label_);
+        }
     }
 
     void Emitter::spill_caller_save_regs(bool spill_avx)
@@ -1209,6 +1226,39 @@ namespace monad::vm::compiler::native
     bool Emitter::is_live(GeneralReg reg, std::tuple<LiveSet...> const &live)
     {
         return is_live(reg, live, std::index_sequence_for<LiveSet...>{});
+    }
+
+    void Emitter::gas_decrement_no_check(int32_t gas)
+    {
+        MONAD_VM_DEBUG_ASSERT(gas > 0);
+        as_.sub(
+            x86::qword_ptr(reg_context, runtime::context_offset_gas_remaining),
+            gas);
+    }
+
+    void Emitter::gas_decrement_no_check(x86::Gpq gas)
+    {
+        as_.sub(
+            x86::qword_ptr(reg_context, runtime::context_offset_gas_remaining),
+            gas);
+    }
+
+    bool Emitter::accumulate_static_work(int32_t work)
+    {
+        MONAD_VM_DEBUG_ASSERT(work >= 0);
+        MONAD_VM_DEBUG_ASSERT(
+            work <= std::numeric_limits<int32_t>::max() -
+                        STATIC_WORK_GAS_CHECK_THRESHOLD + 1);
+        MONAD_VM_DEBUG_ASSERT(
+            accumulated_static_work_ < STATIC_WORK_GAS_CHECK_THRESHOLD);
+
+        accumulated_static_work_ += work;
+
+        if (accumulated_static_work_ >= STATIC_WORK_GAS_CHECK_THRESHOLD) {
+            accumulated_static_work_ = 0;
+            return false;
+        }
+        return true;
     }
 
     bool Emitter::block_prologue(basic_blocks::Block const &b)
@@ -2890,7 +2940,9 @@ namespace monad::vm::compiler::native
         as_.mov(
             gpq[0],
             x86::qword_ptr(reg_context, runtime::context_offset_gas_remaining));
-        as_.add(gpq[0], remaining_base_gas);
+        if (remaining_base_gas != 0) {
+            as_.add(gpq[0], remaining_base_gas);
+        }
         as_.xor_(gpq[1].r32(), gpq[1].r32());
         as_.xor_(gpq[2].r32(), gpq[2].r32());
         as_.xor_(gpq[3].r32(), gpq[3].r32());
@@ -5661,7 +5713,7 @@ namespace monad::vm::compiler::native
     }
 
     template <bool Is32Bit, typename LeftOpType>
-    void Emitter::imul_by_gpq(x86::Gpq dst, LeftOpType left, x86::Gpq right)
+    void Emitter::gpr_mul_by_gpq(x86::Gpq dst, LeftOpType left, x86::Gpq right)
     {
         as_.mov(dst, right);
         if constexpr (Is32Bit) {
@@ -5677,9 +5729,12 @@ namespace monad::vm::compiler::native
         }
     }
 
+    // Sets overflow and carry flags according to imul
     template <bool Is32Bit, typename LeftOpType>
-    void Emitter::imul_by_int32(x86::Gpq dst, LeftOpType left, int32_t right)
+    void Emitter::gpr_mul_by_int32_via_imul(
+        x86::Gpq dst, LeftOpType left, int32_t right)
     {
+        MONAD_VM_DEBUG_ASSERT(right != 0 && right != 1);
         if constexpr (Is32Bit) {
             if constexpr (std::is_same_v<LeftOpType, x86::Gpq>) {
                 as_.imul(dst.r32(), left.r32(), right);
@@ -5694,20 +5749,83 @@ namespace monad::vm::compiler::native
     }
 
     template <bool Is32Bit, typename LeftOpType>
-    void Emitter::imul_by_rax_or_int32(
-        asmjit::x86::Gpq dst, LeftOpType left, std::optional<int32_t> i)
+    void Emitter::gpr_mul_by_uint64_via_shl(
+        x86::Gpq dst, LeftOpType left, uint64_t right)
     {
-        if (i) {
-            imul_by_int32<Is32Bit>(dst, left, *i);
+        MONAD_VM_DEBUG_ASSERT(std::popcount(right) == 1);
+        if constexpr (Is32Bit) {
+            MONAD_VM_DEBUG_ASSERT(
+                right <= std::numeric_limits<uint32_t>::max());
+            if constexpr (std::is_same_v<LeftOpType, x86::Gpq>) {
+                // Always mov when right == 1 to clear upper 32 bits of dst:
+                if (dst != left || right == 1) {
+                    as_.mov(dst.r32(), left.r32());
+                }
+            }
+            else {
+                as_.mov(dst.r32(), left);
+            }
+            if (right > 1) {
+                as_.shl(dst.r32(), std::bit_width(right) - 1);
+            }
         }
         else {
-            imul_by_gpq<Is32Bit>(dst, left, x86::rax);
+            if constexpr (std::is_same_v<LeftOpType, x86::Gpq>) {
+                if (dst != left) {
+                    as_.mov(dst, left);
+                }
+            }
+            else {
+                as_.mov(dst, left);
+            }
+            if (right > 1) {
+                as_.shl(dst, std::bit_width(right) - 1);
+            }
         }
+    }
+
+    template <bool Is32Bit, typename LeftOpType>
+    void Emitter::gpr_mul_by_uint64(
+        x86::Gpq dst, LeftOpType left, uint64_t pre_right)
+    {
+        uint64_t right = pre_right;
+        if constexpr (Is32Bit) {
+            right = static_cast<uint32_t>(pre_right);
+        }
+        if (right == 0) {
+            as_.xor_(dst.r32(), dst.r32());
+        }
+        else if (std::popcount(right) == 1) {
+            gpr_mul_by_uint64_via_shl<Is32Bit>(dst, left, right);
+        }
+        else {
+            MONAD_VM_DEBUG_ASSERT(Is32Bit || is_uint64_bounded(right));
+            int32_t r;
+            std::memcpy(&r, &right, sizeof(r));
+            gpr_mul_by_int32_via_imul<Is32Bit>(dst, left, r);
+        }
+    }
+
+    template <bool Is32Bit, typename LeftOpType>
+    void Emitter::gpr_mul_by_rax_or_uint64(
+        asmjit::x86::Gpq dst, LeftOpType left, std::optional<uint64_t> i)
+    {
+        if constexpr (Is32Bit) {
+            if (i) {
+                gpr_mul_by_uint64<Is32Bit>(dst, left, *i);
+                return;
+            }
+        }
+        else if (i && (is_uint64_bounded(*i) || std::popcount(*i) == 1)) {
+            gpr_mul_by_uint64<Is32Bit>(dst, left, *i);
+            return;
+        }
+        gpr_mul_by_gpq<Is32Bit>(dst, left, x86::rax);
     }
 
     void Emitter::mul_with_bit_size_by_rax(
         size_t bit_size, x86::Gpq const *dst, Operand const &left,
-        std::optional<int32_t> value_of_rax)
+        std::optional<uint64_t> value_of_rax)
     {
         if ((bit_size & 63) && (bit_size & 63) <= 32) {
             mul_with_bit_size_and_has_32_bit_by_rax<true>(
@@ -5723,7 +5841,7 @@ namespace monad::vm::compiler::native
     template <bool Has32Bit>
     void Emitter::mul_with_bit_size_and_has_32_bit_by_rax(
         size_t bit_size, x86::Gpq const *dst, Operand const &left,
-        std::optional<int32_t> value_of_rax)
+        std::optional<uint64_t> value_of_rax)
     {
         MONAD_VM_DEBUG_ASSERT(bit_size > 0 && bit_size <= 256);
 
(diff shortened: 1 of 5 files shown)


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,129 @@ namespace monad::vm::compiler::native
...
+            as_.neg(x86::rbx);
+            as_.cmp(gpq[i].r64(), 0);
+            as_.cmovnz(x86::rax, x86::rbx);
+        }
+
+        as_.pop(reg_context); // Restore reg context
+    }
+
+    template <evmc_revision rev>
+    void Emitter::exp_emit_gas_decrement(StackElemRef exponent_elem)
```
laurenthuberdeau: We emit a decrement instruction after each `Exp` call. For static exponents, we could move the decrement to the end of the basic block.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,129 @@ namespace monad::vm::compiler::native
...
     }
+
+    void Emitter::stack_elem_bit_width(StackElemRef elem)
+    {
+        mov_stack_elem_to_general_reg(elem);
+        MONAD_VM_ASSERT(elem->general_reg().has_value());
+
+        static_assert(reg_context == x86::rbx);
+
+        as_.push(reg_context); // We'll use reg context as second register
```
laurenthuberdeau: Bypassing the register allocation mechanism by manually spilling `reg_context` so it can be used like `x86::rax`.

category/vm/compiler/ir/x86/emitter.hpp
```diff
@@ -389,9 +390,20 @@ namespace monad::vm::compiler::native
...
+        template <evmc_revision rev>
+        void exp_emit_gas_decrement(StackElemRef);
+
+        template <evmc_revision rev>
+        bool exp_optimized(int32_t);
+
         template <evmc_revision rev>
         void exp(int32_t remaining_base_gas)
         {
+            if (exp_optimized<rev>(remaining_base_gas)) {
```
laurenthuberdeau: Getting linker errors for reasons I don't understand. The `exp_optimized` function should be instantiated at the same time as `exp`?
```
/home/lhuberdeau/monad/category/vm/compiler/ir/x86/emitter.hpp:404:(.text._ZN5monad2vm8compiler6native7Emitter3expIL13evmc_revision13EEEvi[_ZN5monad2vm8compiler6native7Emitter3expIL13evmc_revision13EEEvi]+0x47): 
undefined reference to: 
`bool monad::vm::compiler::native::Emitter::exp_optimized<(evmc_revision)13>(int)'
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,129 @@ namespace monad::vm::compiler::native
...
+        // For each sub register, compare to zero:
+        // If not zero, mov sub_reg_bit_width in accumulator
+        // where sub_reg_bit_width = 64 * (i + 1) - lzcnt(gpq[i])
+        // The sub registers are traversed from least significant to most
+        // significant so that the last most significant non-zero sub register
+        // is the one that determines the bit width.
+        for (size_t i = 0; i < gpq.size(); ++i) {
+            as_.lzcnt(x86::rbx, gpq[i].r64());
+            as_.sub(x86::rbx, 64 * (i + 1));
+            as_.neg(x86::rbx);
```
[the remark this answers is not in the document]
laurenthuberdeau: It's not possible for LZCNT to return more than 64 so not a problem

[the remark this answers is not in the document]
laurenthuberdeau: Only on old CPUs that interpret LZCNT as BSR. The LZCNT doc says:

> LZCNT differs from BSR. For example, LZCNT will produce the operand size when the input operand is zero. It should be noted that on processors that do not support LZCNT, the instruction byte encoding is executed as BSR.

The BSR doc says:

> Searches the source operand (second operand) for the most significant set bit (1 bit). If a most significant 1 bit is found, its bit index is stored in the destination operand (first operand). The source operand can be a register or a memory location; the destination operand is a register. The bit index is an unsigned offset from bit 0 of the source operand. If the content source operand is 0, the content of the destination operand is undefined.

category/vm/compiler/ir/x86/emitter.hpp
dhil: The linker cannot find the instantiation, because the definition is in the cpp file. In general there are two solutions to this problem: 1) move the definition into the header, or 2) instantiate the template definition at each possible parameter in the cpp file.

dhil: From a quick glance at this code, it is probably worth simply reifying the `rev` parameter and pass it in by value, and then perform a single dynamic conditional check to select the exponent cost.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,138 @@ namespace monad::vm::compiler::native
...
+        if (base_elem->literal() && exp_elem->literal()) {
+            auto const &base = base_elem->literal()->value;
+            auto const &exp = exp_elem->literal()->value;
+
+            // Evaluating exponentiation can be slow, so it's only done in
+            // cases where we can bound the work required.
+            //
+            // If the base is a power of 2, exponentiation is a simple shift.
+            // Otherwise, if exponent is not too large.
+            if (popcount(base) == 1 || exp <= 512) {
```
andreaslyn: No need to miss the optimization when `exp` is large, and the optimization is having the largest impact. See the issue for how to prevent exploits.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,138 @@ namespace monad::vm::compiler::native
...
+    // Discharge
+    template <evmc_revision rev>
+    bool Emitter::exp_optimized(int32_t remaining_base_gas)
+    {
+        auto base_elem = stack_.get(stack_.top_index());
+        auto exp_elem = stack_.get(stack_.top_index() - 1);
+
+        // 95% of the exp calls are between literals and almost everything else
+        // is with a literal exponent so we have fast paths for those.
+        // The base = 0, 1, 2^k cases are also optimized.
```
andreaslyn:
```suggestion
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,138 @@ namespace monad::vm::compiler::native
             as_.mov(res_mem, 0);
         }
     }
+
+    void Emitter::stack_elem_bit_width(StackElemRef elem)
+    {
+        mov_stack_elem_to_general_reg(elem);
```
andreaslyn: Need specialization of this to all possible locations of `elem`.

category/vm/compiler/ir/x86/emitter.cpp
andreaslyn: That is fine

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,138 @@ namespace monad::vm::compiler::native
...
+        MONAD_VM_ASSERT(elem->general_reg().has_value());
+
+        static_assert(reg_context == x86::rbx);
+
+        as_.push(reg_context); // We'll use reg context as second register
```
andreaslyn:
```suggestion
    auto const scratch_reg = [this] {
        if (stack_.has_free_general_reg()) {
            auto [e, _] = alloc_general_reg();
            return general_reg_to_gpq256(*e->general_reg());
        }
        else {
            as_.push(reg_context);
            return reg_context;
        }
    }();
```
Now only pop `reg_context` in the end if `scratch_reg == reg_context`

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,138 @@ namespace monad::vm::compiler::native
...
+        // significant so that the last most significant non-zero sub register
+        // is the one that determines the bit width.
+        as_.xor_(x86::rax, x86::rax);
+        for (size_t i = 0; i < gpq.size(); ++i) {
+            as_.lzcnt(x86::rbx, gpq[i].r64());
+            as_.sub(x86::rbx, 64 * (i + 1));
+            as_.neg(x86::rbx);
+            as_.cmp(gpq[i].r64(), 0);
+            as_.cmovnz(x86::rax, x86::rbx);
+        }
```
andreaslyn: Does this work?
```suggestion
as_.mov(x86::eax, 64)
for (size_t i = 0; i < 4; ++i) {
    as_.lzcnt(scratch_reg, gpq[i]);
    as_.lea(scratch_reg.r32(), x86::ptr(scratch_reg.r32(), -64 * i));
    as_.cmovnz(x86::eax, scratch_reg.r32());
}
as_.neg(x86::eax);
as_.add(x86::eax, 64);
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7394,4 +7401,138 @@ namespace monad::vm::compiler::native
...
+
+        as_.push(reg_context); // We'll use reg context as second register
+
+        auto const &gpq = general_reg_to_gpq256(*elem->general_reg());
+        // For each sub register, compare to zero:
+        // If not zero, mov sub_reg_bit_width in accumulator
+        // where sub_reg_bit_width = 64 * (i + 1) - lzcnt(gpq[i])
+        // The sub registers are traversed from least significant to most
+        // significant so that the last most significant non-zero sub register
+        // is the one that determines the bit width.
```
andreaslyn:
```suggestion
```

category/vm/compiler/ir/x86/emitter.hpp
laurenthuberdeau: That worked, thanks!

category/vm/compiler/ir/x86/emitter.cpp
laurenthuberdeau: Fixed

category/vm/compiler/ir/x86/emitter.cpp
laurenthuberdeau: Good idea to use `lea` to preserve the flags! I think we want to use `cmovnc` instead of `cmovnz` since `lzcnt` affects the flags like this:

> ZF flag is set to 1 in case of zero output (most significant bit of the source is set), and to 0 otherwise, \*\*CF flag is set to 1 if input was zero and cleared otherwise\*\*. OF, SF, PF, and AF flags are undefined.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7408,4 +7428,200 @@ namespace monad::vm::compiler::native
...
+        else if (elem->stack_offset()) {
+            array_byte_width(stack_offset_to_mem256(*elem->stack_offset()));
+        }
+        else if (elem->avx_reg()) {
+            auto avx_reg = avx_reg_to_ymm(*elem->avx_reg());
+            auto [avx_tmp_elem, _] = alloc_avx_reg();
+            auto avx_tmp = avx_reg_to_ymm(*avx_tmp_elem->avx_reg());
+            as_.vpxor(avx_tmp, avx_tmp, avx_tmp);
+            as_.vpcmpeqb(avx_tmp, avx_reg, avx_tmp); // tmp.b = (reg.b == 0)
+            as_.vpmovmskb(x86::eax, avx_tmp); // eax = mask of zero bytes
```
andreaslyn, line 7588: Cool

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7408,4 +7428,200 @@ namespace monad::vm::compiler::native
...
+    }
+
+    void Emitter::exp_emit_gas_decrement(evmc_revision rev, uint256_t exp)
+    {
+        auto exponent_cost = runtime::exp_dynamic_gas_cost_multiplier(rev);
+        auto exponent_byte_size = runtime::count_significant_bytes(exp);
+
+        if (exponent_byte_size != 0) {
+            gas_decrement_no_check(
+                static_cast<int32_t>(exponent_byte_size * exponent_cost));
```
andreaslyn: It is potentially exploitable without the non-negative gas check. An attacker can deploy a contract full of exp computations, then call the contract many times with close to zero gas. The out-of-gas error will not be detected before returning, and after computing a sequence of perhaps 40,000 exp instructions. I will open a PR to fix this.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7408,4 +7428,200 @@ namespace monad::vm::compiler::native
...
+    void Emitter::exp_emit_gas_decrement(
+        evmc_revision rev, StackElemRef exponent_elem)
+    {
+        auto exponent_cost = runtime::exp_dynamic_gas_cost_multiplier(rev);
+
+        if (exponent_elem->literal()) {
+            exp_emit_gas_decrement(rev, exponent_elem->literal()->value);
+        }
+        else {
+            discharge_deferred_comparison();
```
andreaslyn: Always discharge, because gas decrement changes the flags
```suggestion
        discharge_deferred_comparison();

        if (exponent_elem->literal()) {
            exp_emit_gas_decrement(rev, exponent_elem->literal()->value);
        }
        else {
            auto exponent_cost = runtime::exp_dynamic_gas_cost_multiplier(rev);
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7408,4 +7428,200 @@ namespace monad::vm::compiler::native
...
+            stack_elem_byte_width(exponent_elem);
+            as_.imul(x86::eax, exponent_cost);
+            gas_decrement_no_check(x86::rax);
+        }
+    }
+
+    // Discharge
```
andreaslyn:
```suggestion
    // Discharge via exp_emit_gas_decrement
```

test/vm/unit/emitter_tests.cpp
```diff
@@ -2247,6 +2334,62 @@ TEST(Emitter, mulmod)
     }
 }
 
+// Returns all initial segments of bytes of the uint256_t, shortest first.
+// Equivalent to Haskell's Data.List.inits function over the bytes of the
+// uint256_t.
+std::vector<uint256_t> uint256_t_inits(uint256_t const &n)
```
andreaslyn: Move to the anonymous namespace

test/vm/unit/emitter_tests.cpp
```diff
@@ -2247,6 +2334,62 @@ TEST(Emitter, mulmod)
...
+TEST(Emitter, exp)
+{
+    asmjit::JitRuntime rt;
+
+    uint256_t const large_int = {
+        0x0807060504030201,
+        0x100f0e0d0c0b0a09,
+        0x8887868584838281,
+        0x908f8e8d8c8b8a89};
+    uint256_t const bases[] = {0, 1, 2, 5, 8, 255, 256, 2048, large_int};
```
andreaslyn:
```suggestion
    uint256_t const bases[] = {0, 1, 2, 5, 8, 255, 256, 512, 513, 2048, large_int};
```

category/vm/compiler/ir/x86/emitter.cpp
laurenthuberdeau: Ah yes good point! I wonder if that's the issue the fuzzer is finding since it reported a case where one of the operand is a deferred comparison.

test/vm/unit/emitter_tests.cpp
laurenthuberdeau: Still in this module? When I wrote the function I wondered if we had a catch-all `uint256_t` utils module where this function would make sense.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7408,4 +7524,223 @@ namespace monad::vm::compiler::native
...
+            auto const base = base_elem->literal()->value;
+            auto const exp = exp_elem->literal()->value;
+            base_elem.reset(); // Locations not needed anymore
+            exp_elem.reset(); // Locations not needed anymore
+
+            // Evaluating exponentiation can be slow, so it's only done in
+            // cases where we can bound the work required.
+            // If the base is a power of 2, exponentiation is a simple shift.
+            // Otherwise, if exponent is not too large.
+            if (popcount(base) == 1 || exp <= 512) {
```
andreaslyn: The current implementation of EXP does not generally optimize for `popcount(base) == 1`. I think it is better to not rely on internal implementation details of EXP here.
```suggestion
            if (popcount(base) == 1) {
                stack_.pop();
                stack_.pop();
                exp_emit_gas_decrement_by_literal(exp, gas_factor);
                uint256_t shift{0};
                uint64_t const b{bit_width(base) - 1};
                if (MONAD_VM_LIKELY(b)) {
                    static constexpr uint64_t mask =
                        std::numeric_limits<uint32_t>::max();
                    shift = exp;
                    shift[0] = (exp[0] & ~mask) | b * (exp[0] & mask);
                }
                push(uint256_t{1} << shift);
                return true;
            }
            else if (exp <= 512) {
```

andreaslyn (approved):

PR #1529, opened 2025-08-20
Fix balance overflow assertions

Baltoli:
When an overflowing call is made through RPC with this change enabled, the response looks like:
```json
{"jsonrpc":"2.0","error":{"code":-32603,"message":"execution reverted","data":"0x"},"id":1}                                                      
```

### DIFF
--- a/category/execution/ethereum/state3/state.hpp
+++ b/category/execution/ethereum/state3/state.hpp
@@ -290,20 +290,31 @@ public:
         account.value().nonce = nonce;
     }
 
-    void add_to_balance(Address const &address, uint256_t const &delta)
+    [[nodiscard]] bool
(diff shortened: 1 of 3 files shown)


category/execution/ethereum/state3/state.hpp
```diff
@@ -290,20 +290,31 @@ class State
...
 
         account.value().balance += delta;
         account_state.touch();
+        return false;
+    }
+
+    void add_to_balance(Address const &address, uint256_t const &delta)
+    {
+        auto const overflow = add_to_balance_safe(address, delta);
+        MONAD_ASSERT(!overflow);
```
andreaslyn, line 317: I think that this can crash when called via selfdestruct

PR #1534, opened 2025-08-21
Exception propagation

andreaslyn:


category/core/assert_or_throw.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <category/core/assert_or_throw.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+void assert_or_throw_failed(
+    char const *expr, char const *function, char const *file, long line,
+    char const *msg)
+{
+    monad_print_assertion_failed_message(expr, function, file, line, msg);
+    throw monad::AssertionException{};
```
Baltoli: Printing unconditionally here seems a bit off if we're going to catch the exception somewhere, but I don't see a solution other than a top-level try-catch handler in `main`.

andreaslyn: Yeah an alternative would be to capture the stack trace in the assertion exception and print in main

andreaslyn: Should be fixed

category/core/assertion_exception.hpp
```diff
@@ -0,0 +1,76 @@
...
+/// of aborting the program.
+#define MONAD_ASSERT_OR_THROW(expr, ...)                                       \
+    if (MONAD_LIKELY(expr)) { /* likeliest */                                  \
+    }                                                                          \
+    else {                                                                     \
+        monad::AssertionException::throw_exception(                            \
+            #expr,                                                             \
+            __extension__ __PRETTY_FUNCTION__,                                 \
+            __FILE__,                                                          \
+            __LINE__);                                                         \
```
Baltoli: Is there a reason you can't make this a constructor of `AssertionException` and throw directly? Not that it makes a big difference, just curious

category/execution/ethereum/db/test/test_db.cpp
```diff
@@ -987,6 +988,90 @@ TYPED_TEST(DBTest, call_frames_stress_test)
...
+            block_hash_buffer,
+            pool,
+            metrics,
+            call_tracers);
+        ASSERT_TRUE(false);
+    }
+    catch (AssertionException const &e) {
+        std::cerr << "\n########## BEGIN EXPECTED STACK BACKTRACE ##########\n";
+        e.print();
+        std::cerr << "\n##########  END EXPECTED STACK BACKTRACE  ##########\n";
```
Baltoli: Could we do something more robust / less noisy here?

category/core/assertion_exception.hpp
andreaslyn: No constructor is better. Should be fixed.

category/execution/ethereum/db/test/test_db.cpp
andreaslyn: Not print the stack trace, for example.

cmd/monad/main.cpp
```diff
@@ -497,3 +498,8 @@ int main(int const argc, char const *argv[])
     }
     return result.has_error() ? EXIT_FAILURE : EXIT_SUCCESS;
 }
+catch (monad::AssertionException const &e) {
+    LOG_ERROR("Caught assertion exception");
+    e.print();
+    return EXIT_FAILURE;
+}
```
Baltoli: What's the advantage of catching this to print the backtrace, rather than using:
```cpp
    cxx_runtime_terminate_handler = std::get_terminate();
    std::set_terminate(backtrace_terminate_handler);
```
from the top of `main`? I only just noticed that code

andreaslyn: What is the benefit of this compared to catch?

PR #1536, opened 2025-08-21
Patched version of clang-tidy misc-const-correcntess to handle auto variables

mkolosick:
Builds a new plugin for clang-tidy that handles const-correctness for auto variables. Additionally fixes a bug in the pass related to `if constexpr` (https://github.com/llvm/llvm-project/issues/58778). Adds a pass to add const to arguments.

### DIFF
--- a/.github/workflows/license.sh
+++ b/.github/workflows/license.sh
@@ -39,6 +39,10 @@ exit_code=0
 for file in $(git ls-files -- '*.rs' '*.h' '*.hpp' '*.c' '*.cpp'); do
     contents=$(head -c $C_LICENSE_HEADER_LEN "$file")
 
+    if [ $(dirname "$file") == "utils/clang-tidy-auto-const" ]; then
+       continue
+    fi
+
     if [ "$contents" == "$C_LICENSE_HEADER" ]; then
         continue
     fi
--- a/scripts/check-clang-tidy.sh
+++ b/scripts/check-clang-tidy.sh
@@ -37,6 +37,16 @@ while [[ "$#" -gt 0 ]]; do
   esac
 done
 
+
+CUSTOM_PASS_OPTIONS=( )
+
+CONST_CORRECTNESS_PLUGIN="${BUILD_DIR}/utils/clang-tidy-auto-const/libConstCorrectnessChecks.so"
+
+if [ -f $CONST_CORRECTNESS_PLUGIN ]; then
+  CUSTOM_PASS_OPTIONS=( -load $CONST_CORRECTNESS_PLUGIN \
+                        -checks='-misc-const-correctness,misc-auto-const-correctness' )
+fi
+
 mapfile -t inputs < <(\
   find \
     category/core \
@@ -45,7 +55,8 @@ mapfile -t inputs < <(\
 
 "${RUN_CLANG_TIDY}"                               \
   "${inputs[@]}"                                  \
-  -header-filter "category/.*"                \
+  "${CUSTOM_PASS_OPTIONS[@]}"                     \
+  -header-filter "category/.*"                    \
   -j "$(nproc)"                                   \
   -p "${BUILD_DIR}" "$@"                          \
   -extra-arg='-Wno-unknown-warning-option'        \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,12 +153,15 @@ add_library(komihash INTERFACE)
 target_include_directories(komihash INTERFACE "third_party/komihash")
 
 # LLVM
-if(MONAD_COMPILER_LLVM)
-  # Because LLVM is a large dependency, we only try to find it if it's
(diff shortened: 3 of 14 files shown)


category/vm/runtime/memory.hpp
```diff
@@ -49,7 +49,7 @@ namespace monad::vm::runtime
     {
         auto size = ctx->get_memory_offset(*size_ptr);
         if (*size > 0) {
-            auto src = ctx->get_memory_offset(*src_ptr);
+            auto const src = ctx->get_memory_offset(*src_ptr);
             auto dst = ctx->get_memory_offset(*dst_ptr);
             ctx->expand_memory(max(dst, src) + size);
             auto size_in_words = shr_ceil<5>(size);
```
Baltoli: @mkolosick why doesn't it flag `dst` and `size_in_words`?

mkolosick: Not sure, I can dig into that. I figured once it was in a state where there didn't seem to be false positives it was worth pushing and I could dig into false negatives later.

mkolosick: Apparently it just didn't flag them in the run I did but it does now. I probably just ran before fully updating the tool.

mkolosick: Ah it was due to my new matcher being incorrect in its match all vs match one behavior.

scripts/vm/check-clang-tidy.sh
```diff
@@ -43,10 +43,12 @@ mapfile -t inputs < <(\
...
 
-"${RUN_CLANG_TIDY}"                               \
-  "${inputs[@]}"                                  \
-  -header-filter "category/vm/.*"                \
-  -j "$(nproc)"                                   \
-  -p "${BUILD_DIR}" "$@"                          \
-  -extra-arg='-Wno-unknown-warning-option'        \
+"${RUN_CLANG_TIDY}"                                                               \
+  "${inputs[@]}"                                                                  \
+  -load ${BUILD_DIR}/utils/clang-tidy-auto-const/libAutoConstCorrectnessCheck.so  \
```
Baltoli: Given that this is built conditionally, maybe worth having a check for the `.so` existing before trying to run `run-clang-tidy`?

scripts/check-clang-tidy.sh
```diff
@@ -37,6 +37,16 @@ while [[ "$#" -gt 0 ]]; do
...
 done
 
+
+CUSTOM_PASS_OPTIONS=( )
+
+CONST_CORRECTNESS_PLUGIN="${BUILD_DIR}/utils/clang-tidy-auto-const/libConstCorrectnessChecks.so"
+
+if [ -f $CONST_CORRECTNESS_PLUGIN ]; then
+  CUSTOM_PASS_OPTIONS=( -load $CONST_CORRECTNESS_PLUGIN \
+                        -checks='-misc-const-correctness,misc-auto-const-correctness,misc-arg-const-correctness' )
```
Baltoli: How come we turn off the regular `misc-const-correctness` check here?

.github/workflows/license.sh
```diff
@@ -39,6 +39,16 @@ exit_code=0
 for file in $(git ls-files -- '*.rs' '*.h' '*.hpp' '*.c' '*.cpp'); do
     contents=$(head -c $C_LICENSE_HEADER_LEN "$file")
 
+    if [ "$file" == "utils/clang-tidy-auto-const/AutoConstCorrectnessCheck.hpp" ] || \
+       [ "$file" == "utils/clang-tidy-auto-const/AutoConstCorrectnessCheck.cpp" ] || \
+       [ "$file" == "utils/clang-tidy-auto-const/ArgConstCorrectnessCheck.hpp" ] || \
+       [ "$file" == "utils/clang-tidy-auto-const/ArgConstCorrectnessCheck.cpp" ] || \
+       [ "$file" == "utils/clang-tidy-auto-const/ExprAutoMutationAnalyzer.hpp" ] || \
+       [ "$file" == "utils/clang-tidy-auto-const/ExprAutoMutationAnalyzer.cpp" ] || \
+       [ "$file" == "utils/clang-tidy-auto-const/MiscUpdateModule.cpp" ]; then
```
Baltoli: Is there code under `utils/clang-tidy-auto-const` that we _do_ want to run the license check on, or could this be a glob?

scripts/check-clang-tidy.sh
mkolosick: The `misc-auto-const-correctness` pass still does all of the `misc-const-correctness` checks as well. It's just `misc-const-correctness` patched to not ignore `auto` variables.

.github/workflows/license.sh
mkolosick: The `CMakeLists.txt` but I see now that's in a separate loop anyway.

PR #1537, opened 2025-08-21
reserve balance tfm 2

zander-xyz:


category/execution/monad/chain/monad_chain.cpp
```diff
@@ -19,8 +19,112 @@
...
+        }
+    }
+
+    // check current block
+    if (ctx.senders_and_authorities.contains(sender)) {
+        for (size_t j = 0; j < i; ++j) {
+            if (sender == ctx.senders[j]) {
+                return false;
+            }
+            if (std::ranges::contains(ctx.authorities[j], sender)) {
```
babelkushal: the spec says that IsEmptying (method: `can_sender_dip_into_reserve` ) is false if sender is delegated in the same transaction. This loop is returning true -- it is only iterating until the previous transaction, and not checking delegation in the current transaction.

category/execution/monad/chain/monad_chain.cpp
```diff
@@ -19,8 +19,112 @@
...
+        uint512_t const violation_threshold = [&] -> uint512_t {
+            uint256_t const orig_balance =
+                orig_account.has_value() ? orig_account.value().balance : 0;
+            uint256_t const reserve =
+                std::min(get_max_reserve(monad_rev, addr), orig_balance);
+            uint256_t const gas_fees =
+                uint256_t{tx.gas_limit} * gas_price(rev, tx, base_fee_per_gas);
+            if (addr == sender) {
+                if (gas_fees > reserve) {
+                    return UINT512_MAX;
```
babelkushal: ~note that `gas_fees > reserve` implies that `gas_fees > orig_balance`, which means this piece of code should not be even reachable. Just noting it here. In other words, `reserve-gas_fees` in the spec is guaranteed to be `uint` if one assumes `orig_bal >= gas_fees`~

category/execution/monad/chain/monad_chain.cpp
babelkushal: Having said that, maybe the spec is conservative and perhaps could be relaxed for allowing delegation in the current tx. Need to think a bit more about the correctness if we want to relax the spec on this front.

category/execution/monad/chain/monad_chain.cpp
babelkushal: Differs from spec. This check does not matter for non-emptying transactions as consensus would not let such a transaction through. For emptying transaction, it does matter, and impl needs to change.

PR #1555, opened 2025-08-25
Fix poll_uring_() to return completion count to match comment

Chen-Yifan:
zellic's finding

This bug doesn't cause issues now since poll_blocking()/poll_nonblocking() callers ignore the return value, but relies on io_in_flight(), which correctly implements the counting logic.

### DIFF
--- a/category/async/io.cpp
+++ b/category/async/io.cpp
@@ -483,7 +483,9 @@ void AsyncIO::poll_uring_while_submission_queue_full_()
     }
 }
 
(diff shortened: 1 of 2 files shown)


category/async/io.cpp
```diff
@@ -711,6 +713,7 @@ bool AsyncIO::poll_uring_(bool blocking, unsigned poll_rings_mask)
         return process_cqe();
```
maxkozlovsky: process_cqe() is boolean function, which works to correct value due to implicit conversion. May be make it explicit to show that this is expected behavior.

PR #1564, opened 2025-08-27
Refactor VM to use chain traits

Baltoli:


category/vm/evm/chain.hpp
```diff
@@ -0,0 +1,98 @@
...
+    public:
+        static constexpr evmc_revision evm_rev() noexcept
+        {
+            return Rev;
+        }
+
+        // Instead of storing a revision, caches should identify revision
+        // changes by storing the opaque pointer returned by this method. No two
+        // chain specializations will return the same value.
+        static constexpr void const *id() noexcept
```
andreaslyn: I think that this can simply be
```c++
static constexpr uint64_t revision_id() noexcept
{
    return static_cast<uint64_t>(Rev);
}
```

category/vm/evm/chain.hpp
```diff
@@ -0,0 +1,98 @@
...
+        requires sizeof(T) == 1;
+        { T::evm_rev() } -> std::same_as<evmc_revision>;
+        { T::id() } -> std::same_as<void const *>;
+    };
+
+    // This is a temporary workaround to account for the fact that the VM
+    // boundary uses EVM revisions as runtime values. When there's a continuous
+    // thread of templated arguments down from execute_block, this should be
+    // removed.
+    constexpr void const *revision_to_chain_id(evmc_revision const rev) noexcept
```
andreaslyn: Then this becomes
```c++
    constexpr uint64_t revision_to_revision_id(evmc_revision const rev) noexcept
    {
        return static_cast<uint64_t>(rev);
    }
```

category/vm/compiler/ir/basic_blocks.hpp
```diff
@@ -112,7 +113,7 @@ namespace monad::vm::compiler::basic_blocks
         case Jump:
             return 8;
         case SelfDestruct: {
-            if constexpr (Rev < EVMC_TANGERINE_WHISTLE) {
+            if constexpr (Chain::evm_rev() < EVMC_TANGERINE_WHISTLE) {
```
andreaslyn: I think the ultimate goal with the chain traits will be to eliminate the dependency to ethereum revision entirely, and instead only provide traits such as `std::uint16_t selfdestruct_static_gas_cost()`. Out of scope for this PR, of course, and requires making some design decisions.

Baltoli: Yes, that was my intent with this PR. We would ultimately like to get to a point where `evm_rev` is hidden, but I thought I'd keep this PR smaller given the size of the diff.

PR #1567, opened 2025-08-28
Chain traits in execution

Baltoli:


category/execution/ethereum/block_reward_test.cpp
```diff
@@ -34,6 +35,7 @@
 #include <optional>
 
 using namespace monad;
+using namespace monad::vm;
```
andreaslyn: I could be wrong, but looks like this is unnecessary, since `EvmChain` has been moved to the `monad` namespace. Perhaps also unnecessary in other files?

Baltoli: good catch, I'll cut these out

PR #1576, opened 2025-08-30
Fix VM runtime / compile time interface boundaries

Baltoli:
~Depends on https://github.com/category-labs/monad/pull/1583~


category/vm/evm/explicit_evm_chain.hpp
mkolosick, line 1: `EXPLICIT_EVM_CHAIN_MEMBER_LIST` makes sense as a fix to the issues I was having.

category/vm/interpreter/execute.cpp
mkolosick, line 1: This makes sense for now. Note to self to run a code size vs locality performance tradeoff of the macro version. Otherwise this is definitely cleaner for review.

Baltoli: Yep, I'd be interested to see an analysis there. My gut feeling is that it's not hugely critical given this trampoline is called only once per contract.

mkolosick, line 1: I believe it is technically possible for `monad_vm_interpreter` trampoline and `monad_vm_core_loop` to end up compiled with different calling conventions due to their different linkages but it appears that neither gcc nor clang currently do that.

category/vm/interpreter/entry.S
mkolosick, line 1: Why don't the `.cfi_startproc` work?

category/vm/interpreter/execute.cpp
```diff
@@ -13,136 +13,70 @@
...
+                "identical");
+
             auto *const stack_top = stack_ptr - 1;
             auto const *const stack_bottom = stack_top;
-            auto const *const instr_ptr = analysis.code();
-            auto const gas_remaining = ctx.gas_remaining;
+            auto const *const instr_ptr = analysis->code();
+            auto const gas_remaining = ctx->gas_remaining;
 
             instruction_table<traits>[*instr_ptr](
```
mkolosick, line 58: Okay and the extra argument for the `core_loop` function pointer isn't used within the core loop so this doesn't lose us an argument register. I'm fully content with this design then.

category/vm/compiler.cpp
```diff
@@ -134,25 +148,22 @@ namespace monad::vm
...
             MONAD_VM_ASSERT(find_ok);
-            auto const &[revision, icode, config] = acc->second;
+            auto const &[compile_fn, chain_id, icode, config] = acc->second;
 
             if (MONAD_VM_LIKELY(enable_async_compilation_)) {
                 // It is possible that a new async compile request with the same
                 // intercode arrives right after we erase from
                 // `compile_job_map_` below. Therefore we use `cached_compile`,
                 // because it first checks whether the intercode is already
                 // compiled.
```
andreaslyn, line 158: I think this comment can be moved to the compile lambda

PR #1579, opened 2025-08-31
[staking]: Change result of precompile_get_valset

kkuehlz:
The result now includes the next index to use in a followup query.

### DIFF
--- a/category/execution/monad/staking/staking_contract.cpp
+++ b/category/execution/monad/staking/staking_contract.cpp
@@ -442,6 +442,22 @@ StakingContract::precompile_dispatch(byte_string_view &input)
     return dispatch_table[signature];
 }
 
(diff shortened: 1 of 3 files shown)


category/execution/monad/staking/staking_contract.cpp
```diff
@@ -442,6 +442,26 @@ StakingContract::precompile_dispatch(byte_string_view &input)
...
+
+    uint64_t const end =
+        std::min(len, static_cast<uint64_t>(start_index) + limit);
+    std::vector<u64_be> valids;
+    uint64_t i;
+    for (i = start_index; i < end; ++i) {
+        valids.push_back(valset.get(i).load());
+    }
+    bool const done = (end == len);
+    return {done, static_cast<uint32_t>(i), std::move(valids)};
```
kkuehlz, line 458: note: this cast to uint32_t is safe because the length is checked to be less than U32_MAX earlier.

PR #1581, opened 2025-08-31
Memory bounded NodeCache

Chen-Yifan:
NodeCache max memory default (`ReadOnlyOnDiskDbConfig::node_lru_max_mem`) is 100MB

### DIFF
--- a/category/mpt/node_cache.hpp
+++ b/category/mpt/node_cache.hpp
@@ -1,0 +1,97 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <category/core/lru/static_lru_cache.hpp>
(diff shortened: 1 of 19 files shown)


category/mpt/node_cache.hpp
```diff
@@ -0,0 +1,112 @@
...
+            used_bytes_ += new_node_size;
+            list_.insert(list_.begin(), node);
+            it = map_.emplace(virt_offset, list_.iterator_to(node)).first;
+        }
+
+        // trim oldest
+        if (used_bytes_ > max_bytes_) {
+            size_t const empty_end_slots = list_.size() - map_.size();
+            auto list_it = list_.rbegin();
+            std::advance(list_it, empty_end_slots);
```
maxkozlovsky: The iteration can be very slow if list_.size() sufficiently greater than map_.size(). It would be better to cache the pointer to the last non-null element, the logic can be like this:
```
if (it != map_.end()) {
    ... update existing ...
} else if (used_bytes_ + new_node_size <= max_bytes_ && std::next(last_non_null_) != list_.end()) {
    allocate std::next(last_non_null_)
} else {
   recycle last_non_null
   last_non_null_ = std::prev(last_non_null_);
}
```

When updating lru, and node is equal to last_non_null_, last_non_null_ = std::prev(last_non_null_)

Or keep two lists, one of used elements and other of free elements.

category/mpt/node_cache.hpp
```diff
@@ -0,0 +1,133 @@
...
+        , max_bytes_(max_bytes)
+        , used_bytes_{0}
+    {
+        // push nodes to free_list_
+        while (list_.begin() != list_.end()) {
+            auto it = list_.begin();
+            auto &node = *it;
+            list_.erase(it);
+            free_list_.push_back(node);
+        }
```
magedm: How about moving all nodes in list_ to free_list_ in one step? Or make free_list_ part of static_lru_cache instead of inserting nodes into list_ and then moving them to free_list_?

Chen-Yifan: updated

PR #1582, opened 2025-09-02
Monad chain traits

Baltoli:
~Depends on #1576~


category/vm/evm/explicit_traits.hpp
```diff
@@ -0,0 +1,154 @@
...
+    template decltype(f<::monad::MonadTraits<MONAD_ZERO>>)                     \
+        f<::monad::MonadTraits<MONAD_ZERO>>;                                   \
+    template decltype(f<::monad::MonadTraits<MONAD_ONE>>)                      \
+        f<::monad::MonadTraits<MONAD_ONE>>;                                    \
+    template decltype(f<::monad::MonadTraits<MONAD_TWO>>)                      \
+        f<::monad::MonadTraits<MONAD_TWO>>;                                    \
+    template decltype(f<::monad::MonadTraits<MONAD_THREE>>)                    \
+        f<::monad::MonadTraits<MONAD_THREE>>;                                  \
+    template decltype(f<::monad::MonadTraits<MONAD_FOUR>>)                     \
+        f<::monad::MonadTraits<MONAD_FOUR>>;
```
andreaslyn: Looks like this is exactly `EXPLICIT_MONAD_TRAITS(f)`. Consider refactor
```c++
#define EXPLICIT_EVM_TRAITS(f) ...
#define EXPLICIT_MONAD_TRAITS(f) ...
#define EXPLICIT_TRAITS(f) \
    EXPLICIT_EVM_TRAITS(f) \
    EXPLICIT_MONAD_TRAITS(f)
```

category/vm/evm/explicit_traits.hpp
```diff
@@ -0,0 +1,154 @@
...
+        id<::monad::MonadTraits<MONAD_ONE>>;                                   \
+    template decltype(id<::monad::MonadTraits<MONAD_TWO>>)                     \
+        id<::monad::MonadTraits<MONAD_TWO>>;                                   \
+    template decltype(id<::monad::MonadTraits<MONAD_THREE>>)                   \
+        id<::monad::MonadTraits<MONAD_THREE>>;                                 \
+    template decltype(id<::monad::MonadTraits<MONAD_FOUR>>)                    \
+        id<::monad::MonadTraits<MONAD_FOUR>>;
+
+#define CONCAT2(a, b) a##b
+#define CONCAT(a, b) CONCAT2(a, b)
```
andreaslyn: Looks like these identifiers are likely to cause name collisions. Consider moving to category/core and prefix with `MONAD_`

category/vm/evm/opcodes.hpp
```diff
@@ -750,17 +750,53 @@ namespace monad::vm::compiler
...
+            EvmTraits<previous_evm_revision(EVMC_PRAGUE)>>();
+    }
+
+    template <>
+    consteval std::array<OpCodeInfo, 256>
+    make_opcode_table<EvmTraits<EVMC_OSAKA>>()
+    {
+        return make_opcode_table<
+            EvmTraits<previous_evm_revision(EVMC_OSAKA)>>();
+    }
```
andreaslyn, line 765: I guess the intention was to have the `EVMC_OSAKA` specialization below the monad specializations

category/vm/evm/traits.hpp
```diff
@@ -16,44 +16,83 @@
...
+    concept MonadExtraTraits = requires() {
+        { T::monad_rev() } -> std::same_as<monad_revision>;
+    };
+
+    template <typename T>
+    concept Traits = requires() {
+        requires sizeof(T) == 1;
+        { T::evm_rev() } -> std::same_as<evmc_revision>;
+
+        { T::is_monad() } -> std::same_as<bool>;
```
andreaslyn: The `monad_rev()` and `is_monad()` traits are temporary as well as `evm_rev()`, I suppose

cmd/monad/runloop_ethereum.cpp
```diff
@@ -112,133 +116,145 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
...
-        auto const recovered_authorities =
-            recover_authorities(block.transactions, priority_pool);
-        [[maybe_unused]] auto const sender_recovery_time =
-            std::chrono::duration_cast<std::chrono::microseconds>(
-                std::chrono::steady_clock::now() - sender_recovery_begin);
-        std::vector<Address> senders(block.transactions.size());
-        for (unsigned i = 0; i < recovered_senders.size(); ++i) {
-            if (recovered_senders[i].has_value()) {
-                senders[i] = recovered_senders[i].value();
+        auto const body = [&]<Traits traits>() -> Result<void> {
```
andreaslyn, line 119: Given that `body` is almost the entire function, perhaps it is simpler to factor it out into a top level function

cmd/monad/runloop_monad.cpp
```diff
@@ -162,186 +164,191 @@ Result<BlockExecOutput> propose_block(
...
-    [[maybe_unused]] auto const sender_recovery_time =
-        std::chrono::duration_cast<std::chrono::microseconds>(
-            std::chrono::steady_clock::now() - sender_recovery_begin);
-    std::vector<Address> senders(block.transactions.size());
-    for (unsigned i = 0; i < recovered_senders.size(); ++i) {
-        if (recovered_senders[i].has_value()) {
-            senders[i] = recovered_senders[i].value();
+    auto const rev = chain.get_monad_revision(block.header.timestamp);
+
+    auto body = [&]<Traits traits>
```
andreaslyn: Seemingly simpler to have this `body` as a separate function as well.

cmd/monad/runloop_ethereum.cpp
```diff
@@ -112,133 +116,145 @@ Result<std::pair<uint64_t, uint64_t>> runloop_ethereum(
         evmc_revision const rev =
             chain.get_revision(block.header.number, block.header.timestamp);
```
andreaslyn, line 117: I would expect that getting revision was unnecessary now that we have traits

cmd/monad/runloop_monad.cpp
```diff
@@ -162,186 +164,191 @@ Result<BlockExecOutput> propose_block(
...
-    auto const recovered_authorities =
-        recover_authorities(block.transactions, priority_pool);
-    [[maybe_unused]] auto const sender_recovery_time =
-        std::chrono::duration_cast<std::chrono::microseconds>(
-            std::chrono::steady_clock::now() - sender_recovery_begin);
-    std::vector<Address> senders(block.transactions.size());
-    for (unsigned i = 0; i < recovered_senders.size(); ++i) {
-        if (recovered_senders[i].has_value()) {
-            senders[i] = recovered_senders[i].value();
+    auto const rev = chain.get_monad_revision(block.header.timestamp);
```
andreaslyn, line 167: Revision also unnecessary here I suppose

cmd/monad/runloop_monad.cpp
```diff
@@ -384,24 +391,39 @@ bytes32_t for_each_header(
...
+            requires(traits::is_monad())
+        {
+            std::optional<bytes32_t> next_id;
+            if (MONAD_LIKELY(rev >= MONAD_FOUR)) {
+                next_id = handle_header<MonadConsensusBlockHeaderV2>(
+                    id, data, start_exclusive, end_inclusive, fn);
+            }
+            else if (rev >= MONAD_THREE) {
+                next_id = handle_header<MonadConsensusBlockHeaderV1>(
+                    id, data, start_exclusive, end_inclusive, fn);
```
andreaslyn: I think that traits can be used here instead of `rev`
```suggestion

        auto const body = [&]<Traits traits>
            requires(traits::is_monad())
        {
            std::optional<bytes32_t> next_id;
            if constexpr (traits::monad_rev() >= MONAD_FOUR) {
                next_id = handle_header<MonadConsensusBlockHeaderV2>(
                    id, data, start_exclusive, end_inclusive, fn);
            }
            else if constexpr (traits::monad_rev() >= MONAD_THREE) {
                next_id = handle_header<MonadConsensusBlockHeaderV1>(
                    id, data, start_exclusive, end_inclusive, fn);
```

category/vm/evm/explicit_traits.hpp
Baltoli: Yes, I realised in the follow-up PR that this refactoring is necessary as well as cleaner. Will fold it in here

category/vm/evm/opcodes.hpp
Baltoli: I thought it would be cleaner to keep Ethereum and Monad specializations contiguous rather than interspersing them, but I don't feel strongly either way

category/vm/evm/traits.hpp
Baltoli: `is_monad()` will stay (so that we can write `requires(traits::is_monad())` on Monad-specific templates that should never be instantiated with Eth revisions (e.g. `check_call_monad_precompile`. `monad_rev()` is temporary.

cmd/monad/runloop_ethereum.cpp
Baltoli: You still have to get a revision at runtime based on each block's timestamp; this code is now the sole point we go from runtime -> compile time.

cmd/monad/runloop_monad.cpp
Baltoli: See above

cmd/monad/runloop_monad.cpp
Baltoli: Good catch; I rebased this from scratch onto `main` and missed these cases

cmd/monad/runloop_ethereum.cpp
Baltoli: I avoided doing so to keep the actual runloop internals as unchanged as possible vs. the traits changes (there are a lot of captures in the body lambda). It might be easier to refactor this in a followup

cmd/monad/runloop_monad.cpp
Baltoli: As above

category/vm/evm/traits.hpp
andreaslyn: I thought that the design was to use `requires(traits::feature())` rather than having monad specific features

PR #1595, opened 2025-09-03
Adjust blockhash to read from the block hash history

dhil:
This patch adjusts the implementation of the blockhash opcode to read from the block hash history whenever it is available.

### DIFF
--- a/category/execution/ethereum/block_hash_history_test.cpp
+++ b/category/execution/ethereum/block_hash_history_test.cpp
@@ -27,6 +27,7 @@
 #include <category/execution/monad/chain/monad_devnet.hpp>
 #include <category/mpt/db.hpp>
 #include <category/vm/evm/traits.hpp>
+#include <category/vm/utils/evm-as.hpp>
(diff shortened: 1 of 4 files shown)


category/execution/ethereum/block_hash_history_test.cpp
```diff
@@ -27,93 +27,107 @@
...
+        return to_bytes(to_big_endian(uint256_t{x}));
     }
+}
+
+TEST_F(BlockHistoryFixture, read_write_block_hash_history_storage)
+{
+    static constexpr uint64_t window_size = BLOCK_HISTORY_LENGTH;
+
+    this->deploy_contract();
+    this->fill_history(1, window_size);
```
Baltoli: why do you need `this->`?

dhil: I don't, just did it for readability.

dhil: I've removed the `this->` during the latest rebase.

PR #1604, opened 2025-09-04
Add UntypedIR

goodlyrottenapple:
The purpose of this IR is to aid generating code that takes advantage of statically known JUMP/JUMPI destinations via constant propagation. 
Another optimisation which this IR enables is storing block labels/addresses on the stack. This allows jumping to another block by directly loading an x86 address from the top of the stack and jumping to it, as opposed to doign a lookup in the jump table (that is, if the `JumpDest` is an `Addr{}`).
Further, the vector of coerce stack offsets should allow us to emit code to convert any EVM words which are know to be continuations/pointers into machine addresses when it is safe to do so, i.e. when the type from PolyTypedIR guarantees that the given stack element will only ever be used as an address via JUMP/JUMPI.

### DIFF
--- a/category/vm/compiler/ir/untyped.cpp
+++ b/category/vm/compiler/ir/untyped.cpp
@@ -1,0 +1,466 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/vm/compiler/ir/basic_blocks.hpp>
+#include <category/vm/compiler/ir/local_stacks.hpp>
+#include <category/vm/compiler/ir/poly_typed.hpp>
+#include <category/vm/compiler/ir/poly_typed/block.hpp>
+#include <category/vm/compiler/ir/poly_typed/kind.hpp>
+#include <category/vm/compiler/ir/untyped.hpp>
+#include <category/vm/compiler/types.hpp>
+#include <category/vm/core/assert.h>
+#include <category/vm/core/cases.hpp>
+
+#include <cstddef>
+#include <format>
+#include <limits>
+#include <optional>
+#include <span>
+#include <unordered_map>
+#include <utility>
+#include <variant>
+#include <vector>
+
+using namespace monad::vm::compiler;
+
+using monad::vm::Cases;
+
+namespace monad::vm::compiler::untyped
+{
+    UntypedIR::UntypedIR(poly_typed::PolyTypedIR &&ir)
+        : codesize{ir.codesize}
+        , jumpdests{std::move(ir.jumpdests)}
+        , blocks{build_untyped(jumpdests, std::move(ir.blocks))}
+    {
+    }
+
+    struct Ignored
+    {
+    };
+
+    std::variant<Addr, Word, Ignored> expected_jumpdest_type(
+        std::vector<poly_typed::Kind> const &dest_block_kind,
+        std::vector<poly_typed::Kind> const &output_stack_kind, size_t i)
+    {
+        if (i >= output_stack_kind.size()) {
+            return Ignored{};
+        }
+        if (std::holds_alternative<poly_typed::Cont>(*output_stack_kind[i])) {
+            if (i < dest_block_kind.size()) {
+                MONAD_VM_ASSERT(
+                    std::holds_alternative<poly_typed::Cont>(
+                        *dest_block_kind[i]) ||
+                    std::holds_alternative<poly_typed::KindVar>(
+                        *dest_block_kind[i]));
+            }
+            return Addr{};
+        }
+        else if (std::holds_alternative<poly_typed::WordCont>(
+                     *output_stack_kind[i])) {
+            if (i < dest_block_kind.size() &&
+                std::holds_alternative<poly_typed::Cont>(*dest_block_kind[i])) {
+                return Addr{};
+            }
+            else {
+                return Word{};
+            }
+        }
+        else {
+            if (i < dest_block_kind.size()) {
+                MONAD_VM_ASSERT(!std::holds_alternative<poly_typed::Cont>(
+                    *dest_block_kind[i]));
+                MONAD_VM_ASSERT(!std::holds_alternative<poly_typed::WordCont>(
+                    *dest_block_kind[i]));
+            }
+
+            return Ignored{};
+        }
+    }
+
+    std::vector<poly_typed::Value> pad_output_stack(
+        std::vector<poly_typed::Kind> const &output_stack_kind,
+        std::span<poly_typed::Value> output_stack, size_t min_params)
+    {
+        auto stack = std::vector<poly_typed::Value>(
+            output_stack.begin(), output_stack.end());
+        // We need to pad out the output stack in case the inferred output type
+        // refers to output stack items past the min_params number actually used
+        // by the block.
+        if (output_stack.size() < output_stack_kind.size()) {
+            for (size_t i = 0, param = min_params;
+                 i < output_stack_kind.size() - output_stack.size();
+                 ++i, ++param) {
+                stack.emplace_back(poly_typed::ValueIs::PARAM_ID, param);
+            }
+        }
+        return stack;
+    }
+
+    std::variant<Addr, Word> current_type(
+        std::vector<poly_typed::Kind> const &input_stack_kind,
+        std::vector<poly_typed::Value> const &output_stack, size_t i)
+    {
+        switch (output_stack[i].is) {
+        case poly_typed::ValueIs::LITERAL:
+            return Word{};
+        case poly_typed::ValueIs::PARAM_ID: {
+            MONAD_VM_ASSERT(output_stack[i].param < input_stack_kind.size());
+            poly_typed::Kind const &kind =
+                input_stack_kind[output_stack[i].param];
+
+            if (std::holds_alternative<poly_typed::Cont>(*kind)) {
+                return Addr{};
+            }
+            else {
+                return Word{};
+            }
+        }
+        default:
+            return Word{};
+        }
+    }
+
+    std::vector<size_t> collect_coercions(
+        std::vector<poly_typed::Kind> const &input_stack_kind,
+        std::vector<poly_typed::Kind> const &dest_block_kind,
+        std::vector<poly_typed::Kind> const &output_stack_kind,
+        std::vector<poly_typed::Value> const &output_stack)
+    {
+        std::vector<size_t> coersions;
+        for (size_t i = 0; i < output_stack_kind.size(); i++) {
+            auto current = current_type(input_stack_kind, output_stack, i);
+            auto expected =
+                expected_jumpdest_type(dest_block_kind, output_stack_kind, i);
+            if (!std::holds_alternative<Ignored>(expected)) {
+                if (std::holds_alternative<Word>(current) &&
+                    std::holds_alternative<Addr>(expected)) {
+                    coersions.push_back(i);
+                }
+                else if (
+                    std::holds_alternative<Addr>(current) &&
+                    std::holds_alternative<Word>(expected)) {
+                    MONAD_VM_ASSERT(false);
+                }
+                // current == Addr, jumpdest_type_may == Addr - nothing to do
+                // current == Word, jumpdest_type_may == Word - nothing to do
+            }
+        }
+        return coersions;
+    }
+
+    std::variant<std::vector<Block>, std::vector<local_stacks::Block>>
+    build_untyped(
+        std::unordered_map<byte_offset, block_id> const &jumpdests,
+        std::vector<poly_typed::Block> &&typed_blocks)
+    {
+        auto const &pre_cont = typed_blocks.front().kind;
+        if (!(pre_cont->front.size() == 0 &&
+              std::holds_alternative<poly_typed::ContVar>(pre_cont->tail))) {
+            std::vector<local_stacks::Block> untyped_blocks;
+            untyped_blocks.reserve(typed_blocks.size());
+            for (auto &tb : typed_blocks) {
+                block_id fallthrough_dest = INVALID_BLOCK_ID;
+                basic_blocks::Terminator terminator;
+                std::visit<void>(
+                    Cases{
+                        [&](poly_typed::JumpI const &t) {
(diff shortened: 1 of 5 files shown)


category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+        : codesize{ir.codesize}
+        , jumpdests{std::move(ir.jumpdests)}
+        , blocks{build_untyped(jumpdests, std::move(ir.blocks))}
+    {
+    }
+
+    std::vector<size_t> collect_coersions(
+        std::vector<poly_typed::Kind> const &dest_block_kind,
+        std::vector<poly_typed::Kind> const &ouptut_stack_kind,
+        std::vector<local_stacks::Value> const &ouptut_stack)
```
andreaslyn:
```suggestion
        std::vector<poly_typed::Kind> const &output_stack_kind,
        std::vector<poly_typed::Value> const &output_stack)
```

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+            else if (std::holds_alternative<poly_typed::Cont>(
+                         *ouptut_stack_kind[i])) {
+                // we also need to coerce any values which are literals. if the
+                // value is a param, that measn it has been passed to the
+                // current block in which case, we trust that whichever previous
+                // block we jumped out of did the work of coercing it's output
+                // stack correctly
+                if (i < dest_block_kind.size()) {
+                    MONAD_VM_ASSERT(std::holds_alternative<poly_typed::Cont>(
+                        *dest_block_kind[i]));
```
andreaslyn: The destination can have type `a,s -> Exit` and have `Cont` as first argument, so I think that this assert is not correct.

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+        return coersions;
+    }
+
+    std::optional<std::vector<Block>> build_untyped(
+        std::unordered_map<byte_offset, block_id> const &jumpdests,
+        std::vector<poly_typed::Block> &&typed_blocks)
+    {
+        auto const &pre_cont = typed_blocks.front().kind;
+        if (!(pre_cont->front.size() == 0 &&
+              std::holds_alternative<poly_typed::ContVar>(pre_cont->tail))) {
```
andreaslyn, line 170: It is possible to proceed if `pre_cont->front.size() != 0`, we will just not be able to eliminate stack underflow checks. Anyway, let us keep it as is for now, because it is simpler.

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+                }
+                if (ouptut_stack[i].is == local_stacks::ValueIs::LITERAL) {
+                    coersions.push_back(i);
+                }
+            }
+        }
+        return coersions;
+    }
+
+    std::optional<std::vector<Block>> build_untyped(
```
andreaslyn: Should be possible to return:
```
std::variant<std::vector<local_stacks::Block>, std::vector<Block>>
```
so that we do not lose the basic blocks in case type inference has failed.

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+                t = FallThrough{coerce_to_addr, fallthrough.fallthrough_dest};
+            }
+            else {
+                auto &jump_dest_value = tb.output.front();
+                poly_typed::ContKind jump_dest_kind;
+                JumpDest jump_dest;
+
+                switch (jump_dest_value.is) {
+                case local_stacks::ValueIs::COMPUTED:
+                    MONAD_VM_ASSERT(false);
```
andreaslyn: Thinking about this again, I think this is actually reachable for dead code parts of the contract. We should handle this gracefully somehow. Just return an empty vector perhaps?

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+                        jump_dest = Addr{};
+                    }
+                    else if (std::holds_alternative<poly_typed::WordCont>(
+                                 *kind)) {
+                        jump_dest_kind =
+                            std::get<poly_typed::WordCont>(*kind).cont;
+                        jump_dest = Word{};
+                    }
+                    else {
+                        MONAD_VM_ASSERT(false);
```
andreaslyn, line 299: It is probably also possible to reach this when the basic block is dead code

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,283 @@
...
+                default: {
+                    if (auto block_offset_id = jumpdests.find(
+                            static_cast<size_t>(jump_dest_value.literal));
+                        block_offset_id != jumpdests.end()) {
+                        jump_dest = block_offset_id->second;
+                        jump_dest_kind =
+                            typed_blocks[block_offset_id->second].kind;
+                    }
+                    else {
+                        MONAD_VM_ASSERT(false);
```
andreaslyn: This case is expected in well typed contracts, and can be treated exactly like the invalid instruction case.

category/vm/compiler/ir/untyped.cpp
goodlyrottenapple: Sure, should I modify `PolyTypedIR` to retain the local_stacks Blocks? Otherwise we would have to return `std::vector<poly_typed::Block>`

category/vm/compiler/ir/untyped.cpp
goodlyrottenapple: ~~yeah we can return std::nullopt or the untyped blocks instead if this is actually reachable.~~ ah i see, we want to continue rather than fully abort?

andreaslyn: Right, we can continue because the jump is unreachable.

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,294 @@
...
+                        jump_dest_kind =
+                            std::get<poly_typed::WordCont>(*kind).cont;
+                        jump_dest = Word{};
+                    }
+                    else {
+                        // Based on the kind information, we know this is a dead
+                        // branch and can therefore be marked as invalid
+                        MONAD_VM_ASSERT(
+                            std::holds_alternative<poly_typed::JumpI>(
+                                tb.terminator));
```
andreaslyn: I think assert may well be correct, but why?

category/vm/compiler/ir/poly_typed.hpp
```diff
@@ -29,6 +29,7 @@ namespace monad::vm::compiler::poly_typed
 
         interpreter::code_size_t codesize;
         std::unordered_map<byte_offset, block_id> jumpdests;
+        std::vector<local_stacks::Block> untyped_blocks;
```
andreaslyn: This is not necessary, because it is easy to convert from `poly_typed::Block` to `local_stacks::Block`

category/vm/compiler/ir/poly_typed.cpp
```diff
@@ -459,6 +459,7 @@ namespace monad::vm::compiler::poly_typed
     PolyTypedIR::PolyTypedIR(local_stacks::LocalStacksIR &&ir)
         : codesize{ir.codesize}
         , jumpdests{std::move(ir.jumpdests)}
+        , untyped_blocks{ir.blocks}
```
andreaslyn: No need to copy all basic blocks

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,294 @@
...
+    std::variant<std::vector<Block>, std::vector<local_stacks::Block>>
+    build_untyped(
+        std::unordered_map<byte_offset, block_id> const &jumpdests,
+        std::vector<poly_typed::Block> &&typed_blocks,
+        std::vector<local_stacks::Block> &&untyped_blocks)
+    {
+        auto const &pre_cont = typed_blocks.front().kind;
+        if (!(pre_cont->front.size() == 0 &&
+              std::holds_alternative<poly_typed::ContVar>(pre_cont->tail))) {
+            return untyped_blocks;
```
andreaslyn, line 216: Instead of passing all the basic blocks twice as argument, just map `std::vector<poly_typed::Block>` to `std::vector<local_stacks::Block>` here, only if it is needed

category/vm/compiler/ir/untyped.cpp
goodlyrottenapple: I'm not sure how that branch is reachable tbh. I think either we have given up on the block, i.e. the inferred type for the block is `Word... -> Exit` or else the param must either be a Cont or WordCont... at least those are the only three cases I was able to build so far...

andreaslyn: Yeah I think it is only possible to hit this case when the block is inferred to have type `Word.. -> Exit`, because of a type error. Then it is possible that `std::holds_alternative<poly_typed::Jump>(tb.terminator)` is true I think. Perhaps we should instead assert that `kind.back()` is `Word..`

andreaslyn: So we can use `DeadCode` instead of invalid instruction in this case

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,428 @@
...
+            else {
+                return Word{};
+            }
+        }
+        else {
+            if (i < dest_block_kind.size()) {
+                MONAD_VM_ASSERT(!std::holds_alternative<poly_typed::Cont>(
+                    *dest_block_kind[i]));
+                MONAD_VM_ASSERT(!std::holds_alternative<poly_typed::WordCont>(
+                    *dest_block_kind[i]));
```
andreaslyn, line 85: I think that these are possible to hit in unreachable basic block, due to possibly containing type errors. If you prefer to keep this kind of invariants, then is is possible to check whether the basic block has type `...,Word.. -> Exit` in the beginning of the `typed_blocks` loop in `build_untyped`. Immediately insert `DeadCode` if that is the case.

test/vm/unit/untyped_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+    ASSERT_EQ(jump.coerce_to_addr.size(), 1);
+    ASSERT_EQ(jump.coerce_to_addr[0], 1);
+
+    ASSERT_TRUE(std::holds_alternative<Jump>(blocks[2].terminator));
+
+    auto const &jump2 = std::get<Jump>(blocks[2].terminator);
+    ASSERT_TRUE(std::holds_alternative<Addr>(jump2.jump_dest));
+}
+
+TEST(untyped, dead_words_dot_dot_dot)
```
andreaslyn: I think it is called `ContWords` in the c++ code,
```suggestion
TEST(untyped, dead_cont_words)
```

test/vm/unit/untyped_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+    ASSERT_TRUE(
+        std::holds_alternative<std::vector<untyped::Block>>(blocks_may));
+    auto const &blocks = std::get<std::vector<untyped::Block>>(blocks_may);
+    ASSERT_EQ(blocks.size(), 3);
+    ASSERT_TRUE(std::holds_alternative<Jump>(blocks[1].terminator));
+
+    auto const &jump = std::get<Jump>(blocks[1].terminator);
+    // because the top of the stack value passed into this block
+    // is used both as a jump destination and as
+    // an EVM word for sstore, it must be an EVM word anc cannot be
```
andreaslyn:
```suggestion
    // an EVM word for sstore, it must be an EVM word and cannot be
```

category/vm/compiler/ir/untyped.cpp
goodlyrottenapple: Will do that. Btw can you ever conceivably have a type other than just `Word.. -> Exit`, e.g. `Any, Word.. -> Exit`

andreaslyn: It is possible to have `Any,Word.. -> Exit`, `(Word,Word.. -> Exit),Word.. -> Exit`, etc. This can happen if this basic block does not contain a type error itself, but there is a code path to a basic block which contains a type error.

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,460 @@
...
+        }
+    }
+
+    std::vector<poly_typed::Value> pad_output_stack(
+        std::vector<poly_typed::Kind> const &output_stack_kind,
+        std::span<poly_typed::Value> output_stack, size_t min_params)
+    {
+        auto stack = std::vector<poly_typed::Value>(
+            output_stack.begin(), output_stack.end());
+        // We need to pad out the output stack in case the infered output type
```
Baltoli:
```suggestion
        // We need to pad out the output stack in case the inferred output type
```

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,460 @@
...
+            else {
+                return Word{};
+            }
+        }
+        default:
+            return Word{};
+        }
+    }
+
+    std::vector<size_t> collect_coersions(
```
Baltoli:
```suggestion
    std::vector<size_t> collect_coercions(
```

Baltoli: etc (sorry to nitpick typos)

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,460 @@
...
+        }
+
+        std::vector<Block> blocks;
+        blocks.reserve(typed_blocks.size());
+
+        for (auto &tb : typed_blocks) {
+            if (std::holds_alternative<poly_typed::ContWords>(tb.kind->tail)) {
+                // If the inferred kind is Word... -> Exit
+                // this block cannot be reachable from the entry-point
+                // and can thus be maked as dead code
```
Baltoli:
```suggestion
                // and can thus be marked as dead code
```

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,460 @@
...
+                JumpDest jump_dest;
+
+                switch (jump_dest_value.is) {
+                case poly_typed::ValueIs::COMPUTED: {
+                    // If the jump address is not statically known, but the
+                    // entry-point block had a valid type, we must be processing
+                    // a block that is unreachable from the contract entrypoint.
+                    // Hence we can mark the block as dead code and proceed with
+                    // another block.
+                    blocks.push_back(
```
andreaslyn: I think monad vm assert(false) will work now

andreaslyn (dismissed):

category/vm/compiler/ir/untyped.cpp
```diff
@@ -0,0 +1,469 @@
...
+                    padded_output_stack);
+
+                t = FallThrough{coerce_to_addr, fallthrough.fallthrough_dest};
+            }
+            else {
+                MONAD_VM_ASSERT(tb.output.size() > 0);
+                auto &jump_dest_value = tb.output.front();
+
+                // std::span<poly_typed::Value> output_tail(tb.output.data(),
+                // tb.output.size());
```
dhil:
```suggestion
```
Old debug code?

PR #1612, opened 2025-09-04
Store delegated code inline in `Account`

Chen-Yifan:
**Account changes:**

Introduce `CodeStorage` class in `Account`, which can hold either inline delegated code or a code hash.
Preserve backward compatibility: legacy accounts in Db always store a 32-byte code hash 

**Account encoding changes:**
Update `encode_account_db()` to support inline delegated code. `Account::code_or_hash` is still encoded as an RLP string but not always 32 bytes anymore, its length determines the representation:
- `len = 0`: account has no code
- `len = 23`: inline delegated code (new format)
- `len = 32`: code hash (old or new format, both valid)

**Reader-side delegation check**
```python
def is_delegated_eoa:
     if (!account.has_code()): return False;
     if (account.inline_delegated_code()): return True;
     code = db.read_code(account.get_code_hash());
     return is_code_delegated(code);
```

### DIFF
--- a/category/execution/ethereum/core/account.hpp
+++ b/category/execution/ethereum/core/account.hpp
@@ -15,6 +15,7 @@
 
 #pragma once
 
+#include <category/core/byte_string.hpp>
 #include <category/core/bytes.hpp>
 #include <category/core/config.hpp>
 #include <category/core/int.hpp>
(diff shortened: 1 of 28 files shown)


category/execution/ethereum/db/util.cpp
```diff
@@ -456,7 +459,21 @@ namespace
...
+            else {
+                payload = payload_copy;
+            }
+            if (!payload.empty()) { // parse the extension list
+                BOOST_OUTCOME_TRY(
+                    auto extra_payload, rlp::parse_list_metadata(payload));
+                BOOST_OUTCOME_TRY(
+                    acct.delegated,
+                    rlp::decode_unsigned<unsigned char>(extra_payload));
+            }
```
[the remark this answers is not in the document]
Chen-Yifan: delegated is initialized to false by default

category/execution/ethereum/db/util.cpp
```diff
@@ -456,7 +459,21 @@ namespace
...
             acct.balance, rlp::decode_unsigned<uint256_t>(payload));
         if (!payload.empty()) {
-            BOOST_OUTCOME_TRY(acct.code_hash, rlp::decode_bytes32(payload));
+            auto const payload_copy = payload;
+            auto code_res = rlp::decode_bytes32(payload);
+            if (code_res.has_value()) {
+                acct.code_hash = code_res.value();
+            }
+            else {
+                payload = payload_copy;
```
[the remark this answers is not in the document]
Chen-Yifan: this comment is wrong

dhil (dismissed):

category/execution/ethereum/state3/state.cpp
```diff
@@ -538,7 +551,12 @@ size_t State::copy_code(
     if (MONAD_UNLIKELY(!account.has_value())) {
         return 0;
     }
-    bytes32_t const &code_hash = account.value().code_hash;
+    if (account.value().inline_delegated_code()) {
+        auto const code = account.value().code_view();
+        std::copy_n(code.data(), code.size(), buffer);
```
cmcl, line 556: Should this be checking against `offset` and `buffer_size` in a similar fashion to later on in this function?

category/execution/ethereum/core/account.hpp
```diff
@@ -25,14 +26,104 @@
...
+
+    bytes32_t get_hash() const
+    {
+        if (data_.empty()) {
+            return NULL_HASH;
+        }
+        if (is_hash()) {
+            return to_bytes(data_);
+        }
+        MONAD_ASSERT(data_.size() == DELEGATED_CODE_SIZE);
```
cmcl, line 83: See my comments around `db/util.cpp`

category/execution/ethereum/db/util.cpp
```diff
@@ -451,7 +470,11 @@ namespace
         BOOST_OUTCOME_TRY(
             acct.balance, rlp::decode_unsigned<uint256_t>(payload));
         if (!payload.empty()) {
-            BOOST_OUTCOME_TRY(acct.code_hash, rlp::decode_bytes32(payload));
+            BOOST_OUTCOME_TRY(acct.code_or_hash, rlp::decode_string(payload));
```
cmcl, line 473: Previously, `rlp::decode_bytes32` would check the payload is exactly 32 bytes and return `DecodeError::ArrayLengthUnexpected` otherwise.  `decode_string` accepts any length but `CodeStorage` is only valid for `0`,`23` and `32`.  Payloads of other lengths could abort on calls to `CodeStorage::get_hash`

category/execution/ethereum/state3/state.cpp
```diff
@@ -566,10 +584,13 @@ void State::set_code(Address const &address, byte_string_view const code)
     if (MONAD_UNLIKELY(!account.has_value())) {
         return;
     }
-
+    if (code.empty() || vm::evm::is_delegated(code)) {
+        account.value().code_or_hash = code;
+        return;
```
cmcl, line 589: Returning early skips the call to `rb_.on_set_code(address, code)`.  Is that intentional?

category/statesync/statesync_protocol.cpp
```diff
@@ -43,9 +43,8 @@ void account_update(
     using StorageDeltas = monad_statesync_client_context::StorageDeltas;
 
     if (acct.has_value()) {
-        auto const &hash = acct.value().code_hash;
-        if (hash != NULL_HASH) {
-            ctx.seen_code.emplace(hash);
+        if (acct.value().code_or_hash.is_hash()) { // code is not null or inline
+            ctx.seen_code.emplace(to_bytes(acct.value().get_code_hash()));
```
cmcl, line 47:
```suggestion
            ctx.seen_code.emplace(acct.value().get_code_hash());
```

PR #1621, opened 2025-09-06
[execution/monad]: Move system txn to trait dependent dispatch function

kkuehlz:
[execution/monad]: Remove system transactions from Chain
    
Make system transactions inaccessible from the ethereum codepath by
introducing a dispatch_transaction<Trait> function. The correct dispatch
function is chosen at link time, and no monad code is included on from
execute block.

system transactions: Stricter validation
 * Error if system transaction present before MONAD_FOUR fork
 * Error if transaction type is not legacy

### DIFF
--- a/category/execution/monad/execute_system_transaction.cpp
+++ b/category/execution/monad/execute_system_transaction.cpp
@@ -61,8 +61,7 @@ Result<Receipt> ExecuteSystemTransaction<traits>::operator()()
 {
     TRACE_TXN_EVENT(StartTxn);
 
-    BOOST_OUTCOME_TRY(static_validate_system_transaction(tx_, sender_));
-
+    BOOST_OUTCOME_TRY(static_validate_system_transaction<traits>(tx_, sender_));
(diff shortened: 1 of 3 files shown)


category/execution/monad/execute_system_transaction.cpp
```diff
@@ -61,8 +61,7 @@ Result<Receipt> ExecuteSystemTransaction<traits>::operator()()
 {
     TRACE_TXN_EVENT(StartTxn);
 
-    BOOST_OUTCOME_TRY(static_validate_system_transaction(tx_, sender_));
-
+    BOOST_OUTCOME_TRY(static_validate_system_transaction<traits>(tx_, sender_));
```
Baltoli, line 64: This should be constexpr

kkuehlz: it could, but i don't think we'll ever have any Transactions known at compile time ,so i don't think it matters.

PR #1652, opened 2025-09-11
Add fuzzer option to save generated contracts

laurenthuberdeau:
## Context

The fuzzer finds a counter-example to the changes in https://github.com/category-labs/monad/pull/1526 and I'd like to have the full contract for further analysis. This PR adds a `--debug` option to `tmux-fuzzer.sh` that sets a flag to log contracts.

### DIFF
--- a/test/vm/fuzzer/fuzzer.cpp
+++ b/test/vm/fuzzer/fuzzer.cpp
@@ -71,8 +71,8 @@ using namespace evmone::state;
 using namespace evmc::literals;
 using namespace monad;
 using namespace monad::vm::fuzzing;
-using namespace std::chrono_literals;
-using namespace std::filesystem;
+
+namespace fs = std::filesystem;
 
 using enum monad::vm::compiler::EvmOpCode;
 
@@ -562,10 +562,10 @@ static void do_run(std::size_t const run_index, arguments const &args)
                 intx::hex(intx::be::load<intx::uint256>(code_hash.bytes));
             if (args.contract_log_dir) {
                 auto const code_hash_dir =
-                    path(*args.contract_log_dir) / "code_hash";
+                    fs::path(*args.contract_log_dir) / "code_hash";
                 auto const code_buf = contract.data();
                 auto const code_len = contract.size();
-                create_directories(code_hash_dir);
+                fs::create_directories(code_hash_dir);
 
                 auto contract_path =
                     code_hash_dir / fmt::format("{}", code_hash_str);


test/vm/fuzzer/fuzzer.cpp
```diff
@@ -67,6 +72,7 @@ using namespace evmc::literals;
 using namespace monad;
 using namespace monad::vm::fuzzing;
 using namespace std::chrono_literals;
+using namespace std::filesystem;
```
Baltoli: Minor nit: generally it's best practice to avoid dragging everything into the global namespace like this. The usual idiom for unpleasantly long namespaces like `std::filesystem` is:
```suggestion
namespace fs = std::filesystem;
```

laurenthuberdeau: Fixed both `std::filesystem` and  `std::chrono_literals`

.gitmodules
```diff
@@ -53,3 +53,6 @@
 [submodule "third_party/cthash"]
 	path = third_party/cthash
 	url = https://github.com/hanickadot/cthash.git
+[submodule "third_party/creduce"]
+	path = third_party/creduce
+	url = https://github.com/csmith-project/creduce.git
```
Baltoli: Happily, they actually package this for Ubuntu so we can avoid a submodule:
```console
$ apt search creduce
creduce/plucky 2.11.0~20240909-2 amd64
  Test-Case Reduction for C Compiler Bugs
```

laurenthuberdeau: Good to know because building the library requires dependencies such as Perl and a specific LLVM version.
But I didn't mean to include this in this PR so I'll revert :)

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -545,11 +557,37 @@ static void do_run(std::size_t const run_index, arguments const &args)
...
             auto const a1 =
                 deploy_contract(monad_state, genesis_address, contract);
             MONAD_VM_ASSERT(a == a1);
+            if (args.contract_log_dir) {
+                std::cerr << fmt::format(
+                    "Deployed contract (hash {}) at address {} (size: {} bytes)\n",
+                    code_hash_str,
+                    a,
+                    contract.size());
+            }
```
andreaslyn, line 590: Logging code hash should happen before/after transition:
```
std::cout << "START evmone: " << code_hash << std::endl;
transition(... evmone_vm, ...);
std::cout << "END evmone: " << code_hash << std::endl;
std::cout << "START monad vm: " << code_hash << std::endl;
transition(..., monad_vm, ...);
std::cout << "END monad vm: " << code_hash << std::endl;
```
Remember to call keccak only if debug output is enabled.

PR #1661, opened 2025-09-15
Fix exp gas check and fuzzer issues

andreaslyn:
Notice base branch [laurent/optimized_exp](https://github.com/category-labs/monad/tree/laurent/optimized_exp)

### DIFF
--- a/category/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/compiler/ir/x86/emitter.cpp
@@ -35,8 +35,6 @@
 #include <asmjit/core/operand.h>
 #include <asmjit/x86/x86operand.h>
 
-#include <evmc/evmc.h> // For evmc_revision
-
 #include <algorithm>
 #include <array>
 #include <bit>
@@ -598,6 +596,8 @@ namespace monad::vm::compiler::native
         , gpq256_regs_{Gpq256{x86::r12, x86::r13, x86::r14, x86::r15}, Gpq256{x86::r8, x86::r9, x86::r10, x86::r11}, Gpq256{x86::rcx, x86::rsi, x86::rdx, x86::rdi}}
         , bytecode_size_{codesize}
         , rodata_{as_.newNamedLabel("ROD")}
+        , exponential_constant_fold_counter_{0}
+        , accumulated_static_work_{0}
     {
 #ifdef MONAD_VM_TESTING
         as_.addDiagnosticOptions(kValidateAssembler);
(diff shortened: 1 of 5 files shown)


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -1009,22 +1009,17 @@ namespace monad::vm::compiler::native
...
+            as_.jl(error_label_);
+        }
     }
 
-    void Emitter::gas_decrement_check_non_negative(int32_t gas)
+    void Emitter::gas_decrement_unbounded_work(int32_t gas)
     {
+        accumulated_static_work_ = 0;
         gas_decrement_no_check(gas);
         as_.jl(error_label_);
```
laurenthuberdeau: If `gas == 0`, `gas_decrement_no_check` wont set the flags and this conditional jump will use the flags from the previous instruction.

andreaslyn: Thanks, good catch. The fuzzer should catch this as well, if it is reachable. Need to investigate a bit.

andreaslyn: The fuzzer did not catch this because it is coincidentally not reachable. The `gas_decrement_unbounded_work` is only called with `gas > 0` and corresponding conditional jump in `gas_decrement_static_work` is only emitted when `gas > 0`. I will push a fix. Thanks!

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7563,61 +7660,80 @@ namespace monad::vm::compiler::native
...
+                        std::move(exp_elem), stack_.alloc_literal({1}), {}));
+                    return true;
+                }
+                mov_stack_elem_to_general_reg(exp_elem);
+                auto mul_elem = release_general_reg(std::move(exp_elem), {});
+                auto const &gp =
+                    general_reg_to_gpq256(*mul_elem->general_reg());
+                as_.mov(x86::rax, gp[0]);
+                uint8_t const b = static_cast<uint8_t>(bit_width(base) - 1);
+                if (std::popcount(b) == 1) {
```
laurenthuberdeau, line 7700: Can we use `gpr_mul_by_uint64` here to avoid reimplementing the strength reduction?

andreaslyn: Yes, but requires support in gpr_mul_by_uint64 for setting the falgs, or come up with a different algorithm here, which does not depend on multiplication overflow.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7563,61 +7660,80 @@ namespace monad::vm::compiler::native
...
+                    MONAD_VM_DEBUG_ASSERT(b >= 2 && b <= 128);
+                    gpr_mul_by_uint64_via_shl<false>(gp[0], gp[0], b);
+                    constexpr auto mask = std::numeric_limits<int32_t>::min();
+                    as_.test(x86::rax, mask);
+                    as_.cmovnz(gp[0], x86::rax);
```
laurenthuberdeau, line 7705: I'm not sure I understand this line. Why `mul_dst[0] = mul_src[0]` if `mul_src[0] - 0xffffffff != 0`?

category/vm/compiler/ir/x86/emitter.cpp
laurenthuberdeau: Ah I see that the shift instructions don't set the flags like other arithmetic instructions, good to know!

category/vm/compiler/ir/x86/emitter.cpp
andreaslyn: It uses the test instruction, which sign extends 0x800..0 to 0xfff...f8000...0. So if any of the high 33 bits of src[0] is set, then dst[0] = src[0]. Now consider the result of shifting left by dst.

laurenthuberdeau: Ah sorry, confused `test` with `cmp`. So this ensures that overflows in the multiplication causes `1 << mul_elem` to return 0 by forcing `mul_elem` to be large enough for `1 << mul_elem == 0` (i.e. larger than 256), neat!

I'm not convinced this catches all overflows however, if we use `base = (1 << 255), b = 255, src[0] = 0xf0`, then `gpr_mul_by_uint64_via_shl<false>` makes `dst[0] = 0xf0 << 255 = 0` but `src[0] & 0xffffffff80000000 = 0`, causing the final result to be 1 instead of 0.

I think a correct overflow check uses a computed mask to catch the smallest `src[0]` such that `(src[0] << (bit_width(b) - 1)) != 0`, which corresponds to `~(base - 1)` (assuming base is a power of 2).

andreaslyn: Notice that `2 <= b <= 128`, because `std::popcount(b) == 1`.

laurenthuberdeau: Ah yes, and I had missed the `bit_width` call in `gpr_mul_by_uint64_via_shl` so it makes sense now.

laurenthuberdeau (approved):

PR #1674, opened 2025-09-16
Optimize signextend by literal

andreaslyn:


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -4109,37 +4114,66 @@ namespace monad::vm::compiler::native
...
+
+        auto const src_y = avx_reg_to_ymm(*src->avx_reg());
+        auto const dst_y = avx_reg_to_ymm(*dst->avx_reg());
+        auto const tmp_y = avx_reg_to_ymm(*tmp->avx_reg());
+
+        uint256_t shuf = std::numeric_limits<uint256_t>::max();
+        if (ix >= 16) {
+            std::memset(
+                shuf.as_bytes() + ix + 1,
+                ix - 16,
```
laurenthuberdeau: Took me some time to understand why `ix - 16` and not `ix`. Can we add a comment saying that `vpshufb` operates over 128-bit lanes and so indices are in `[0,16]`? 

General question about `vpshufb`, if I'm reading the docs properly, it only uses the sign bit and the 4 least-significant bits, meaning the `- 16` could be removed if we wanted since it only affects the 5th bit?

Aglowharmony:
> Do you have the signextend microbenchmark numbers from before and after this change?

no

category/vm/compiler/ir/x86/emitter.cpp
andreaslyn: I have expanded on the comments and simplified the code to generally use `ix`

mkolosick (approved):

PR #1680, opened 2025-09-17
UntypedIR interpreter

goodlyrottenapple:
This PR creates a copy of the monad interpeter which uses information from `UntypedIR` to convert certain stack values into pointers (see `cast_pointers` and  `cast_evm_word_to_ptr`, previously called `jump_impl`, in `test/vm/untyped_ir_interpreter/instruction_table.hpp`) at the end of each block.

Specifially, we apply coercions at the JUMP/JUMPI instruction if the current PC offset is a key in the `analysis.jump_coerce_map` or `analysis.fallthrough_coerce_map` in the "else" case of JUMPI. If a block has a fallthrough terminator (specified by a new bool flag passed through as one of the interpreter args), we apply coersions from the previous block at the start of the next one at the JUMPDEST instruction, if the current PC offset is in `analysis.fallthrough_coerce_map`.

Finally, when performing a jump, we now look up whether the stack value has already been cast into a native ptr by checking the current PC offset is in `analysis.jumpdest_addr`. If not, we call `jump_impl`/`cast_evm_word_to_ptr` like in the vanilla interpreter.

### DIFF
--- a/test/vm/fuzzer/fuzzer.cpp
+++ b/test/vm/fuzzer/fuzzer.cpp
@@ -25,6 +25,7 @@
 #include "test_state.hpp"
 #include "transaction.hpp"
 
+#include <category/vm/compiler/ir/untyped.hpp>
 #include <category/vm/compiler/ir/x86/types.hpp>
 #include <category/vm/core/assert.h>
 #include <category/vm/evm/opcodes.hpp>
@@ -253,13 +254,16 @@ static evmc::address deploy_delegated_contract(
 }
 
 static evmc::address deploy_delegated_contracts(
-    State &evmone_state, State &monad_state, evmc::address const &from,
-    evmc::address delegatee)
+    BlockchainTestVM::Implementation impl, State &evmone_state,
+    State &monad_state, evmc::address const &from, evmc::address delegatee)
 {
     auto const a = deploy_delegated_contract(evmone_state, from, delegatee);
     auto const a1 = deploy_delegated_contract(monad_state, from, delegatee);
     MONAD_VM_ASSERT(a == a1);
-    assert_equal(evmone_state, monad_state);
+    assert_equal(
+        evmone_state,
+        monad_state,
+        impl != BlockchainTestVM::Implementation::UntypedIR);
     return a;
 }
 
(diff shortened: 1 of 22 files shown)


test/vm/untyped_ir_interpreter/instruction_table.hpp
```diff
@@ -0,0 +1,1951 @@
...
+        check_requirements<JUMPDEST, traits>(
+            ctx, stack_bottom, stack_top, gas_remaining);
+
+        if (prev_ptr) {
+            if (auto it = analysis.fallthrough_coerce_map.find(
+                    static_cast<std::size_t>(prev_ptr - analysis.code()));
+                it != analysis.fallthrough_coerce_map.end()) {
+                cast_pointers(analysis, stack_top, it->second);
+            }
+        }
```
goodlyrottenapple: This is where we perform the casting of stack values to pointers based on the untyped IR, but only in the fallthrough case, indicated by prev_ptr not being null. 
If prev_ptr is null, either we are at the entry-point, which does not have any coersions as the stack is empty, or the jumpdest was reached via a jump instruction, in which case the coercions have already been performed.

test/vm/untyped_ir_interpreter/instruction_table.hpp
```diff
@@ -0,0 +1,1951 @@
...
+        *stack_top = *(stack_top - N);
+        *(stack_top - N) = runtime::uint256_t{top};
+
+        MONAD_VM_NEXT(SWAP1 + (N - 1));
+    }
+
+    // Control Flow
+    namespace
+    {
+        inline std::uint8_t const *cast_evm_word_to_ptr(
```
goodlyrottenapple, line 1492: In the vanilla interpreter, this function eagerly calls error whenever the target is not a valid jumpdest. Howeverer, this check has to be delayed here because the cast values should only throw at the call-site, which may be in a different block, e.g. in case of a fallthrough.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -545,20 +585,45 @@ static void do_run(std::size_t const run_index, arguments const &args)
...
+                auto icode = monad::vm::interpreter::IntercodeUntypedIR(
+                    std::span{contract.data(), contract.size()});
+
+                if (icode.valid_ir()) {
+                    ++untyped_ir_contracts;
+                }
+                if (icode.jump_coerce_map.empty() ||
+                    icode.fallthrough_coerce_map.empty()) {
+                    continue;
+                }
```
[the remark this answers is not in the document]
goodlyrottenapple: nope

PR #1681, opened 2025-09-17
Remove poorly understood benchmarks

andreaslyn:


test/vm/execution_benchmarks/benchmarks.cpp
```diff
@@ -261,10 +261,8 @@ namespace
             load_benchmark_json(vm_performance_dir / "performanceTester.json"),
         };
 
-        for (auto const &p :
-             fs::directory_iterator(execution_benchmarks_dir / "json")) {
-            ret.emplace_back(load_benchmark_json(p));
-        }
+        // Use load_benchmark_json to load JSON benchmarks. Currently unused:
+        (void)load_benchmark_json;
```
dhil:
```suggestion
```
Maybe my eyes betray me, but it looks like `load_benchmark_json` is used on line 261.

andreaslyn: Thanks

dhil (approved):

PR #1688, opened 2025-09-17
event_interator_inline: replace local const with macro for MAX_SYNC_SPIN

iurjscsi1101500:


kkuehlz:
this define now gets included in every file, which is probably why it's local to the function. it's an implementation detail to the  spinlock

this was written before c23, which now supports `constexpr`. We plan to make more of the C code `constexpr`.

PR #1700, opened 2025-09-19
Bump `CREATE` opcode family static cost

Baltoli:


category/vm/compiler/ir/x86.cpp
```diff
@@ -27,13 +27,10 @@
 
 #include <asmjit/core/jitruntime.h>
 
-#include <quill/detail/LogMacros.h>
-// clang-tidy incorrectly views this macro as unused
-#include <quill/Quill.h> // IWYU pragma: keep
+#include <quill/Quill.h>
```
Baltoli, line 30: This clang-tidy fix is now global in the configuration file; drive-by cleanup here

category/vm/compiler/ir/x86.cpp
```diff
@@ -27,13 +27,10 @@
 
 #include <asmjit/core/jitruntime.h>
 
-#include <quill/detail/LogMacros.h>
-// clang-tidy incorrectly views this macro as unused
```
Baltoli, line 31: This clang-tidy fix is now global in the configuration file; drive-by cleanup here

category/vm/evm/opcodes.hpp
```diff
@@ -69,7 +69,7 @@ namespace monad::vm::compiler
          * Some instructions may also consume additional dynamic gas depending
          * on run-time properties (e.g. memory expansion or storage costs).
          */
-        std::uint16_t min_gas;
+        std::int32_t min_gas;
```
Baltoli: Needs to change as 160k > max uint32. Using int32 as gas is tracked as int32 in the VM anyway

category/vm/compiler/ir/x86.cpp
```diff
@@ -326,7 +323,6 @@ namespace
         Emitter &emit, Block const &block, int32_t instr_gas,
         native_code_size_t max_native_size, CompilerConfig const &config)
     {
-        MONAD_VM_ASSERT(instr_gas <= std::numeric_limits<int32_t>::max());
```
andreaslyn: Why is this not an invariant anymore?

category/vm/evm/opcodes.hpp
```diff
@@ -796,7 +796,10 @@ namespace monad::vm::compiler
     consteval std::array<OpCodeInfo, 256>
     make_opcode_table<MonadTraits<MONAD_FOUR>>()
     {
-        return make_opcode_table<MonadTraits<MONAD_FOUR>::evm_base>();
+        auto table = make_opcode_table<MonadTraits<MONAD_FOUR>::evm_base>();
+        table[CREATE].min_gas = 160'000;
+        table[CREATE2].min_gas = 160'000;
+        return table;
```
andreaslyn, line 802: Perhaps it would be good to test the new gas costs on the compiler side at some point

category/vm/compiler/ir/x86.cpp
Baltoli: It's trivially true: `instr_gas` has type `int32_t` so will always be `<= std::numeric_limits<int32_t>::max()`.

category/vm/evm/opcodes.hpp
Baltoli: yes, we should indeed improve testing for the TFM changes generally

category/vm/evm/opcodes.hpp
andreaslyn: Can change this to `uint32_t` now that gas cost is tracked by `int64_t`

PR #1706, opened 2025-09-22
Use `int64_t` for compiler block gas

Baltoli:


category/vm/compiler/ir/x86.cpp
```diff
@@ -323,11 +323,10 @@ namespace
 
     template <Traits traits>
     void emit_instrs(
-        Emitter &emit, Block const &block, int32_t instr_gas,
+        Emitter &emit, Block const &block, int64_t instr_gas,
         native_code_size_t max_native_size, CompilerConfig const &config)
     {
```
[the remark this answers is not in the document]
Baltoli: The assertion was trivial before; it wasn't asserting anything important (all `int32_t`s are in the range it asserted).

laurenthuberdeau (approved):

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -1228,7 +1228,7 @@ namespace monad::vm::compiler::native
         return is_live(reg, live, std::index_sequence_for<LiveSet...>{});
     }
 
-    void Emitter::gas_decrement_no_check(int32_t gas)
+    void Emitter::gas_decrement_no_check(int64_t gas)
```
andreaslyn, line 1231: If `gas > std::numeric_limits<int32_t>::max()` we need to do something else. For example,
```
as_.mov(x86::rax, gas);
gas_decrement_no_check(x86::rax);
```
This only works if `x86::rax` is not live everywhere this is called (directly or indirectly), and I have not checked this.

andreaslyn: Remember to add a unit test for this, or open an issue.

andreaslyn: Well, `x86::rax` can be live, just not live at the same time as `gas > INT32_MAX`

andreaslyn: Consider just jumping to error label if `gas > INT32_MAX` and `static_assert(block_gas_limit <= INT32_MAX)`.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -1243,7 +1243,7 @@ namespace monad::vm::compiler::native
             gas);
     }
 
-    bool Emitter::accumulate_static_work(int32_t work)
+    bool Emitter::accumulate_static_work(int64_t work)
```
andreaslyn, line 1265: Need to change assert here to
```
MONAD_VM_DEBUG_ASSERT(
            work <= std::numeric_limits<int64_t>::max() -
                        STATIC_WORK_GAS_CHECK_THRESHOLD + 1);
```

category/vm/compiler/ir/x86.cpp
```diff
@@ -323,11 +322,10 @@ namespace
...
-        MONAD_VM_ASSERT(instr_gas <= std::numeric_limits<int32_t>::max());
-        int32_t remaining_base_gas = instr_gas;
+        int64_t remaining_base_gas = instr_gas;
```
dhil, line 328: It is not changed as part of this PR, so Github does not seem to allow me to highlight the line, but: on line 345 we have `constexpr int32_t remaining_base_gas = 0;`, whilst technically OK, I think we should change it to `int64_t` for consistency. In addition it would also eliminate the implicit coercion in the application of `selfdestruct`.

PR #1733, opened 2025-09-26
Fuzzer: Shrink counter-examples

laurenthuberdeau:
## Context

The VM fuzzer generates contracts and messages in an attempt to find inputs that cause the compiler and evmone interpreter to produce different side effects. The contracts it generates are 1000s of instructions long, with many basic blocks, which makes it difficult to isolate the exact sequence of instructions responsible for the difference.

This PR adds a shrinker to the fuzzer, so that counter-examples are reduced to only the sequence of instructions that revealed the issue. To changes can be grouped in 2 main groups:

1. Better control of random values in the fuzzer. 
- Previously, the fuzzer generated values (contracts and messages) as it executed, with the generated contracts being represented as EVM bytecode with a single random number generator. This makes replaying and shrinking fuzzer runs more or less impossible since any change would offset the random number generator and cause a cascade of different values to be generated. 
- The fuzzer now generates all of the random values ahead of time and represent contracts in a higher-level form that makes changing contracts much simpler (e.g. blocks and sequence of instructions can be removed all at once).
2. Using the reified representation of a run introduced in the previous, the different steps taken by the fuzzer (deploy X contract, send message) can be removed, reordered and simplified by the shrinker. 

The shrinker is relatively fast and has been helpful during development of the LLVM backend. Shrinking a counter-example takes a few seconds, with usually only a few instructions (<= 20) left.

Example use of the fuzzer:

When removing the dynamic gas cost of the optimized `EXP` instruction when the base is unknown and exponent = 2:

```
            else if (exp == 2) { // x ** 2 = x * x
                stack_.pop();
                stack_.pop();
                {
                    RegReserv const base_reserv{base_elem};
                    // exp_emit_gas_decrement_by_literal(2, gas_factor);
                }
                stack_.push(std::move(base_elem));
                dup(1);
                mul(remaining_base_gas);
                return true;
            }
```

the shrinker finds the following counter-example:

```
Counter-example found by fuzzer:
Run with 2 steps
Iteration 0: Deploy contract (hash 89f5a1351a0b4ed676083a7052197823779d075beb38b3c8ae4ed38759245450) at address 0x655cb46ac0378e6a78a07d313ad2a44988eb769d (size: 235 bytes)
  0x00:
      PUSH32 0x18d6d999347de3fff2a081d3272c0315c4699357a6fedbc66ad762aac60f1608
      PUSH32 0x80000000000000000000000
      PUSH32 0xa445ada6a8f4c09c9176a90d0a5f89433c81cc9cffeed6d266e8e9f48f74d25f
      PUSH32 0x100000000
      PUSH32 0x10000000000000000000000000000000000000000
      DUP5
      DUP6
      PUSH32 0xfc62b332f3b3c713f06c8eb6710daf4a0af4824c022aa33dc85ce8a7a763a768
      PUSH32 0x2
      DUP4
      EXP
    Stop

Iteration 1: Send message:
  kind DELEGATECALL, gas 4854427, value 9beab2ea26f31c9848385921d00be7a6cbf83815baa1fc66
  code_address 0x655cb46ac0378e6a78a07d313ad2a44988eb769d, sender 0xefca25b4db15773a98fae8889b0efc6bb25357ea, recipient 0xa508d306fae702242683882fa34d622165d367bd
  input_data [64 bytes]: 7f00000000000000000000000000000000000000010000000000000000000000007fffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000
Final failure reason:
  Assertion failed: void monad::vm::fuzzing::assert_equal(const evmc::Result &, const evmc::Result &, const bool):/home/lhuberdeau/monad/test/vm/fuzzer/assertions.cpp:116 Assertion failed: evmone_result.gas_left == compiler_result.gas_left

```

### Why not use a general purpose shrinking tool?

There exists off-the-shelf tools to reduce programs generated by fuzzers. One such example is [`CReduce`](https://github.com/csmith-project/creduce), which takes a file and a predicate, and removes parts of the input files to find the smallest file that match the predicate.

This approach works fine for counter-examples that fit in a single file and that are pure. The EVM environment is stateful, with the fuzzer generating the equivalent of many programs and calling them in a specific sequence. As a result, the one-file model of CReduce is a poor fit and serializing all of the fuzzer state to a single file would likely result in long shrinking times and poor counter-examples found.

### DIFF
--- a/test/vm/fuzzer/fuzzer.cpp
+++ b/test/vm/fuzzer/fuzzer.cpp
@@ -322,6 +322,7 @@ namespace
         seed_t seed = default_seed;
         std::size_t runs = std::numeric_limits<std::size_t>::max();
         bool print_stats = false;
+        bool print_shrinker_state = false;
         BlockchainTestVM::Implementation implementation =
             BlockchainTestVM::Implementation::Compiler;
         evmc_revision revision = EVMC_OSAKA;
@@ -386,6 +387,11 @@ static arguments parse_args(int const argc, char **const argv)
         "--print-stats",
         args.print_stats,
         "Print message result statistics when logging");
+
+    app.add_flag(
+        "--print-shrinker-state",
+        args.print_shrinker_state,
+        "Print size of run when shrinking");
 
     app.add_flag(
         "--deterministic-compilation",
@@ -724,8 +730,8 @@ static Run prepare_run(arguments const &args)
 }
 
 static void do_run(
-    std::size_t const run_index, arguments const &args, Run const &run,
-    size_t &iteration_index)
+    std::size_t const run_index, arguments const &args, bool const verbose,
+    Run const &run, size_t &iteration_index)
 {
     auto const rev = args.revision;
 
@@ -795,13 +801,15 @@ static void do_run(
         iteration_index++;
     }
 
-    log_run(
-        start_time,
-        args,
-        exit_code_stats,
-        run_index,
-        iteration_index,
-        total_messages);
+    if (verbose) {
+        log_run(
+            start_time,
+            args,
+            exit_code_stats,
+            run_index,
+            iteration_index,
+            total_messages);
+    }
 }
 
 static std::optional<FuzzerAssertFailure>
@@ -809,7 +817,7 @@ try_run(arguments const &args, Run const &run)
 {
     try {
         size_t iteration_index = 0;
-        do_run(0, args, run, iteration_index);
+        do_run(0, args, false, run, iteration_index);
         return std::nullopt;
     }
     catch (FuzzerAssertFailure const &ex) {
@@ -819,16 +827,19 @@ try_run(arguments const &args, Run const &run)
 
 static void print_run_summary(Run const &run)
 {
-    size_t basic_block_count = 0;
     size_t deploy_count = 0;
     size_t message_count = 0;
+    size_t basic_block_count = 0;
+    size_t instruction_count = 0;
     for (auto const &step : run) {
         std::visit(
             monad::vm::Cases{
-                [&](DeployContract const &) {
+                [&](DeployContract const &d) {
                     deploy_count++;
-                    basic_block_count +=
-                        std::get<DeployContract>(step).contract.size();
+                    basic_block_count += d.contract.size();
+                    for (BasicBlock const &bb : d.contract) {
+                        instruction_count += bb.instructions.size();
+                    }
                 },
                 [&](SendMessage const &) { message_count++; },
                 [&](auto const &) {},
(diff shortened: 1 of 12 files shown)


test/vm/fuzzer/fuzzer.cpp
```diff
@@ -494,105 +500,375 @@ static bool toss(Engine &engine, double p)
...
+
+            This is likely too complicated and not that effective comparing to
+            simply deleting contracts/messages from the run and seeing if the
+            failure still happens. To speed up the process, we can truncate
+            the tail of the run first (after the failing contract/message), then
+            reorder a few contracts and messages and try again.
+        */
+        return run;
+    } else {
+        return shrink_singleton_run(run_index, args, failed_contract, failed_message.code_address, failed_message);
```
laurenthuberdeau: For now, only failed runs containing a single contract and call (i.e. the fuzzer failure still happens when running the contract and message in an empty environment) can be shrunk. All of the compiler "bugs" I introduced fit this criteria so I wasn't able to test the "full-run" shrinker and left it unimplemented.

The comment above describes how I would go and shrink full runs, I'm open to other ideas.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -602,7 +878,23 @@ static void run_loop(int argc, char **argv)
...
+
+        auto const &run = prepare_run(args);
+        size_t iteration_index = 0;
+        size_t message_index = 0;
+        try {
+            do_run(i, args, run, iteration_index, message_index);
+        }
+        catch (FuzzerAssertFailure const &ex) {
+            // Disable randomness in compilation for shrinking
+            args.deterministic_compilation = true;
```
laurenthuberdeau, line 1266: We don't have a good way to debug this since the moves introduced by the compiler hook are only visible in the contract assembly listing and so the virtual stack state must be inferred from it. Also, if we wanted to control the randomness of the compiler hook, we'd need to reset the state of the random number generator at the beginning of every `emit_instr` call so that removing instructions doesn't cause the hook to produce different results for subsequent instructions.

Like for shrinking entire runs, I suppose most of the fuzzer counter-examples don't depend on the hook so I disable it when shrinking.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -494,105 +500,375 @@ static bool toss(Engine &engine, double p)
...
+        } else if (contract[removed_block_ix].instructions.size() > 0) {
+            // Try to remove instructions from the block instead
+            // First try with ranges of instructions
+            auto new_contract2 = shrink_block(engine, contract, removed_block_ix);
+            if (!try_singleton_run(run_index, args, new_contract2, contract_address, failed_message)) {
+                contract = std::move(new_contract2);
+                iteration_count = 0;
+                continue;
+            } else {
+                // Idea: Replace instructions with simpler/shorter instructions?
```
laurenthuberdeau: To produce neater counter-examples, we could replace instructions that push values from the state to the stack (like `PC`, `GAS`) with PUSH of small integers. For simplicity, I decided to leave this out.

brett-monad (dismissed):

andreaslyn (commented):
I think that we are not currently in a position where we can reactor the fuzzer. The plan is to have a database of mutation tests to verify our fuzz testers. Let us revisit this when we have the necessary tools to verify the correctness of this PR.

PR #1734, opened 2025-09-26
[statesync] fix statesync connect logic

kjcamann:
This prevents an infinite loop when the consensus daemon is stopped before the execution daemon


kjcamann:
Closed in favor of the more extensive refactoring in #1930

PR #1748, opened 2025-10-03
[DRAFT] access list tracer

Baltoli:


category/execution/ethereum/trace/prestate_tracer.cpp
```diff
@@ -107,21 +109,87 @@ namespace trace
...
+            // If an address is excluded because it's always considered warm, we
+            // still want to include it in the access list if it's had storage
+            // keys set by this transaction.
+            auto const exclude =
+                keys.empty() && should_exclude_address<traits>(address);
+
+            if (!exclude) {
+                auto entry = json::object();
+                entry["address"] = bytes_to_hex(address.bytes);
+                entry["storageKeys"] = std::move(keys);
```
dhil: Alternatively, we can have the object initializer populate the entry, thereby eliminating the need for mutation (probably needs to be run through the clang formatter first).
```suggestion
                auto const entry = json::object({
                     {"address", bytes_to_hex(address.bytes)},
                     {"storageKeys", std::move(keys)}
                 });
```

PR #1750, opened 2025-10-03
db util: get_transactions()

Chen-Yifan:
Add RODb::traverse() method

### DIFF
--- a/category/mpt/test/db_test.cpp
+++ b/category/mpt/test/db_test.cpp
@@ -873,8 +873,7 @@ TEST(DbTest, history_length_adjustment_never_under_min)
     remove(dbname);
 }
 
-TEST_F(
-    OnDiskDbWithFileFixture, read_only_db_traverse_fail_upon_version_expiration)
+TEST_F(OnDiskDbWithFileFixture, read_only_db_traverse_as_version_expire)
 {
     struct TraverseMachinePruneHistory final : public TraverseMachine
     {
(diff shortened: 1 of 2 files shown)


category/execution/ethereum/db/util.hpp
```diff
@@ -167,6 +167,12 @@ mpt::Nibbles proposal_prefix(bytes32_t const &);
 
 std::vector<bytes32_t> get_proposal_block_ids(mpt::Db &, uint64_t block_number);
 
+template <typename DBType>
+    requires std::is_same_v<mpt::Db, DBType> ||
+             std::is_same_v<mpt::RODb, DBType>
+Result<std::vector<Transaction>> get_transactions(
+    DBType &, uint64_t block_number, bytes32_t const &block_id = {});
```
dhil, line 175: In addition to this function, it would be good to have something that returns a prefix of the block transactions. This sort of thing would be useful for `debug_traceTransaction`, where we need only replay part of the block state. E.g.
```c++
Result<std::vector<Transaction>> get_transactions_prefix(
    DBType &, uint64_t end_transaction, uint64_t block_number, bytes32_t const &block_id = {});
```
The semantics would be `{ tx : block.transactions | tx.index <= end_transaction }`. I think we can just assert `block.transactions.size() >= end_transaction`.

Chen-Yifan: that only needs 1 or 2 lines of code, and can be on the caller side. I think it's good to keep this PR just for general purpose function

zander-xyz (approved):

PR #1751, opened 2025-10-06
Refactor the VM interpreter

goodlyrottenapple:
Refactor the VM interpreter to make it modular for reuse in the untypedIR interpreter. This refactor also gets rid of the `MONAD_VM_NEXT*` macros.

In a followup PR, we can define the untypedIR interpreter by defining a new wrap/instruction_table pair with modified jump/jumpi and jumpdest instructions which use coersions.


category/vm/interpreter/instruction_table.hpp
```diff
@@ -411,1073 +373,633 @@ namespace monad::vm::interpreter
...
         runtime::Context &ctx, Intercode const &analysis,
         runtime::uint256_t const *stack_bottom, runtime::uint256_t *stack_top,
         std::int64_t gas_remaining, std::uint8_t const *instr_ptr)
     {
-        push_impl<N, traits>::push(
+        push_impl<N>::push(
             ctx, analysis, stack_bottom, stack_top, gas_remaining, instr_ptr);
 
-        MONAD_VM_NEXT_PUSH(PUSH0 + N);
+        instr_ptr += N;
```
goodlyrottenapple, line 974: The `MONAD_VM_NEXT_PUSH` macro would increment the instr_ptr by `1 + N`. However, now using `wrap` does `++instr_ptr;`, so we need to additionally incement by `N`.

category/vm/interpreter/call_runtime.hpp
```diff
@@ -82,10 +82,6 @@ namespace monad::vm::interpreter
 
         static_assert(
             stack_arg_count <= std::numeric_limits<std::ptrdiff_t>::max());
-        constexpr std::ptrdiff_t stack_adjustment =
-            static_cast<std::ptrdiff_t>(stack_arg_count) - (use_result ? 1 : 0);
-
-        stack_top -= stack_adjustment;
```
goodlyrottenapple, line 88: `call_runtime` was previously being called from `checked_runtime_call` which had the `stack_top` pointer passed in by value. Therefore, no modification to `stack_top` would be visible to the the caller of `checked_runtime_call`. However, now that `call_runtime` is being called directly, this incorrectly updates the stack pointer.

category/vm/interpreter/instruction_table.hpp
```diff
@@ -1553,30 +1071,19 @@ namespace monad::vm::interpreter
...
-        fuzz_tstore_stack(
-            ctx,
-            stack_bottom,
-            stack_top,
-            static_cast<uint64_t>(instr_ptr - analysis.code()));
-        check_requirements<JUMPDEST, traits>(
-            ctx, analysis, stack_bottom, stack_top, gas_remaining);
-
-        MONAD_VM_NEXT(JUMPDEST);
     }
```
goodlyrottenapple, line 1078: Perhaps `jumpdest` and `pop` should just be replaced by a single `noop` method?

PR #1758, opened 2025-10-07
Check code size explicitly in `resolve_delegation`

Baltoli:


category/vm/evm/delegation.cpp
```diff
@@ -68,11 +69,18 @@ namespace monad::vm::evm
             indicator_size + sizeof(evmc_address);
         static_assert(expected_code_size == 23);
 
+        auto const code_size = host->get_code_size(ctx, &addr);
+        if (code_size != expected_code_size) {
+            return std::nullopt;
+        }
```
dhil: This seems reasonable to me. I suppose today this code can trigger only if something goes wrong during the copy operation. As I believe it is not possible to deploy a contract whose code uses the delegator prefix and violates the equation `|code| == 23 bytes`.

category/vm/evm/delegation.cpp
```diff
@@ -68,11 +69,18 @@ namespace monad::vm::evm
             indicator_size + sizeof(evmc_address);
         static_assert(expected_code_size == 23);
```
dhil: There is some module local duplication here, should we clean it up? I think
```c++
        constexpr uint8_t indicator[] = {0xef, 0x01, 0x00};
        constexpr size_t indicator_size = std::size(indicator);
        constexpr size_t expected_code_size =
            indicator_size + sizeof(evmc_address);
        static_assert(expected_code_size == 23);
```
can be deleted as above we have
```c++
        inline constexpr std::array<uint8_t, 3>
            delegation_indicator_prefix_bytes{0xef, 0x01, 0x00};

        inline constexpr auto delegation_indicator_size =
            delegation_indicator_prefix_bytes.size() + sizeof(evmc_address);
```
So we can replace `expected_code_size` by `delegation_indicator_size` and `indicator_size` by `delegation_indicator_prefix_bytes.size()`.

category/vm/evm/delegation.cpp
Baltoli: yes, correct - this is just defence in depth and isn't currently reachable

dhil:
Reviving this PR to address #1976.

andreaslyn (approved):

PR #1768, opened 2025-10-09
revision test macros

zander-xyz:


test/unit/common/include/monad/test/monad_revision_test.hpp
```diff
@@ -0,0 +1,130 @@
...
+    {
+        using type = ::testing::Types<Ts1..., Ts2...>;
+    };
+
+    template <typename... Ts>
+    using concat_types_t = typename concat_types<Ts...>::type;
+
+    // Union of MonadRevisionTypes and EvmRevisionTypes
+    using MonadEvmRevisionTypes =
+        concat_types_t<MonadRevisionTypes, EvmRevisionTypes>;
```
Baltoli, line 72: I think the `index_sequence` machinery here is overkill; the change would be much simpler if it explicitly enumerated all the traits into a list (as the explicit traits macros do elsewhere). Then we could get ride of all this metaprogramming

jhunsaker (approved):

PR #1770, opened 2025-10-10
Interpreter refactor

goodlyrottenapple:
A slighly less involved refactor, superseeding  #1751, simply moving the body of each instruction_table method into a separate `eval_` method, such that the UntypedIR interpreter can reuse these function bodies with less duplication

### DIFF
--- a/category/vm/interpreter/call_runtime.hpp
+++ b/category/vm/interpreter/call_runtime.hpp
@@ -25,7 +25,7 @@ namespace monad::vm::interpreter
     [[gnu::always_inline]]
     inline void call_runtime(
         void (*f)(FnArgs...), runtime::Context &ctx,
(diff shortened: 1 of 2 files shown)


category/vm/interpreter/call_runtime.hpp
```diff
@@ -82,10 +82,6 @@ namespace monad::vm::interpreter
 
         static_assert(
             stack_arg_count <= std::numeric_limits<std::ptrdiff_t>::max());
-        constexpr std::ptrdiff_t stack_adjustment =
-            static_cast<std::ptrdiff_t>(stack_arg_count) - (use_result ? 1 : 0);
-
-        stack_top -= stack_adjustment;
```
andreaslyn, line 88: Nice find

andreaslyn (approved):

PR #1771, opened 2025-10-10
Fix recursion in io request resubmission

maxkozlovsky:
Avoid recursion when resubmitting requests in poll_uring_.

For delayed reads, store required arguments on deferred list and submit request to iouring directly instead of going through initiate() again.

For retrying EAGAIN errors, defer initiate() if running from inside completion.

### DIFF
--- a/category/async/io.cpp
+++ b/category/async/io.cpp
@@ -301,18 +301,26 @@ AsyncIO::~AsyncIO()
     ::close(fds_.msgwrite);
 }
 
-void AsyncIO::submit_request_(
+void AsyncIO::account_read_()
+{
+    if (++records_.inflight_rd > records_.max_inflight_rd) {
+        records_.max_inflight_rd = records_.inflight_rd;
+    }
+    ++records_.nreads;
+}
+
+void AsyncIO::submit_request_sqe_(
     std::span<std::byte> buffer, chunk_offset_t chunk_and_offset,
     void *uring_data, enum erased_connected_operation::io_priority prio)
 {
     MONAD_DEBUG_ASSERT(uring_data != nullptr);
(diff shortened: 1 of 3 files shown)


category/async/io.hpp
```diff
@@ -333,40 +348,19 @@ class AsyncIO final
...
-                    concurrent_read_ios_pending_.last = state;
-                }
-                concurrent_read_ios_pending_.count++;
-                return size_t(-1); // we never complete immediately
-            }
-        }
-
         if (capture_io_latencies_) {
             uring_data->initiated = std::chrono::steady_clock::now();
         }
```
Chen-Yifan, line 353: should this be moved to after the next code block of adding to the deferred_read? Or is it intentional for the io latency to include the deferred time?

maxkozlovsky: Queuing delays should be included in the latency. The original code was not correct in this regard.

category/async/io.cpp
```diff
@@ -488,34 +505,21 @@ size_t AsyncIO::poll_uring_(bool blocking, unsigned poll_rings_mask)
...
-                else {
-                    concurrent_read_ios_pending_.first = next;
-                }
-                concurrent_read_ios_pending_.count--;
-                state->reinitiate();
+            while (!concurrent_read_ios_pending_.empty() &&
+                   records_.inflight_rd < concurrent_read_io_limit_ &&
+                   io_uring_sq_space_left(other_ring) != 0) {
+                auto &read = concurrent_read_ios_pending_.front();
+                submit_request_sqe_(
```
Chen-Yifan, line 520: is `submit_request_sqe_` safe here? this doesn't go through the `poll_uring_while_submission_queue_full_()` code path, is it guaranteed that submission queue won't be full?

maxkozlovsky: io_uring_sq_space_left() checks for the queue space

PR #1784, opened 2025-10-15
Staking contract fuzzer

andreaslyn:
See the help message for some configuration options
```
build/category/execution/monad_staking_contract_fuzzer --help
```
For example, execute 1,000 fuzzer simulations with
```
build/category/execution/monad_staking_contract_fuzzer --runs 1000
```

### DIFF
--- a/category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
+++ b/category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
@@ -1,0 +1,152 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/execution/monad/staking/fuzzer/staking_contract_machine.hpp>
+#include <category/execution/monad/staking/test/input_generation.hpp>
+#include <category/vm/evm/monad/revision.h>
+#include <category/vm/evm/switch_traits.hpp>
+#include <category/vm/evm/traits.hpp>
+
+#include <CLI/CLI.hpp>
+
+#include <evmc/evmc.hpp>
+
+using namespace monad;
+using namespace monad::staking;
+using namespace monad::staking::test;
+using namespace monad::staking::test::fuzzing;
+using namespace evmc::literals;
+
+template <Traits traits>
+using Transition = StakingContractMachine<traits>::Transition;
+
+template <Traits traits>
+static Transition<traits>
+gen_transition(StakingContractMachine<traits> &machine)
+{
+    auto const p = machine.gen() % 100;
+    if (p < 30) {
+        if (p < 20) {
+            return Transition<traits>::syscall_reward;
+        }
+        return Transition<traits>::precompile_external_reward;
+    }
+    return machine.gen_transition();
+}
+
+template <Traits traits>
+static seed_t run_simulation(seed_t seed, uint64_t depth)
+{
+    MONAD_ASSERT(depth > 0);
+
+    std::cout << "Simulation with seed " << seed << std::endl;
+
+    auto const start_time = std::chrono::steady_clock::now();
+
+    StakingContractMachine<traits> machine{seed};
+    double success_count = 0;
+    for (size_t i = 0; i < depth; ++i) {
+        if (machine.transition(gen_transition(machine))) {
+            ++success_count;
+        }
+    }
+
+    auto const end_time = std::chrono::steady_clock::now();
+
+    auto const success_ratio = success_count / static_cast<double>(depth);
+    auto const time =
+        static_cast<double>((end_time - start_time).count()) / 1'000'000;
+
+    std::cout << "    success ratio: " << success_ratio << '\n'
+              << "  simulation time: " << time << " ms" << std::endl;
+
+    return machine.gen();
+}
+
+namespace
+{
+    struct CommandArgs
+    {
+        seed_t seed = 0;
+        uint64_t depth = 100;
+        uint64_t runs = std::numeric_limits<uint64_t>::max();
+        monad_revision monad_rev = MONAD_NEXT;
+    };
+
+    CommandArgs parse_args(int const argc, char **const argv)
+    {
+        auto app = CLI::App("Staking Contract Fuzzer");
+        auto args = CommandArgs{};
+
+        app.add_option(
+            "--seed",
+            args.seed,
+            "Seed for reproducible fuzzing (0 by default)");
+
+        app.add_option(
+            "--depth",
+            args.depth,
+            "Staking contract transitions per simulation (default 100)");
+
+        app.add_option(
+            "--runs",
+            args.runs,
+            "Number of simulations to execute (default max uint64)");
+
+        std::map<std::string, monad_revision> rev_map;
+        for (int i = static_cast<int>(MONAD_ZERO);
+             i <= static_cast<int>(MONAD_NEXT);
+             ++i) {
+            rev_map.emplace(
+                monad_revision_to_string(static_cast<monad_revision>(i)),
+                static_cast<monad_revision>(i));
+        }
+
+        app.add_option(
+               "--monad-rev",
+               args.monad_rev,
+               "Monad revision (default MONAD_NEXT)")
+            ->transform(CLI::CheckedTransformer(rev_map, CLI::ignore_case));
+
+        try {
+            app.parse(argc, argv);
+        }
+        catch (CLI::ParseError const &e) {
+            std::exit(app.exit(e));
+        }
+
+        return args;
+    }
+}
+
+int main(int argc, char **argv)
+{
+    auto const args = parse_args(argc, argv);
+
+    std::cout << "Running " << args.runs << " simulations\n"
+              << "  depth: " << args.depth << '\n'
+              << "   seed: " << args.seed << '\n'
+              << "    rev: " << monad_revision_to_string(args.monad_rev)
+              << std::endl;
+
+    auto const rev = args.monad_rev;
+    size_t seed = args.seed;
+    for (size_t i = 0; i < args.runs; ++i) {
+        seed = [rev, &seed, &args] -> seed_t {
+            SWITCH_MONAD_TRAITS(run_simulation, seed, args.depth);
+            MONAD_ABORT();
+        }();
+    }
+}
--- a/category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
+++ b/category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
@@ -1,0 +1,2079 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/execution/monad/staking/fuzzer/staking_contract_machine.hpp>
+#include <category/execution/monad/staking/test/input_generation.hpp>
+#include <category/execution/monad/staking/util/bls.hpp>
+#include <category/execution/monad/staking/util/secp256k1.hpp>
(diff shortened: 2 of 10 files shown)


category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+int main(int argc, char **argv)
+{
+    auto args = parse_args(argc, argv);
+
+    std::cout << "Running " << args.runs << " simulations\n"
+              << "  depth: " << args.depth << '\n'
+              << "   seed: " << args.seed << std::endl;
+
+    size_t seed = args.seed;
```
mkolosick, line 145: This always runs with an initial seed value of `0`. I think it would be better to follow the vm fuzzer and pick a random seed if none is provided.

category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+using namespace monad;
+using namespace monad::staking;
+using namespace monad::staking::test;
+using namespace evmc::literals;
+
+static StakingContractMachine::Transition
+gen_transition(StakingContractMachine &machine)
+{
+    auto const p = machine.gen() % 100;
```
mkolosick, line 39: It would be better to move this behavior into a `gen_int` in `StakingContractMachine` that uses `std::uniform_int_distribution` to make the range/behavior more explicit/make the distribution properly uniform.

category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    if (p < 30) {
```
mkolosick, line 40: This behavior should get moved into `machine.gen_transition`.

category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        app.add_option(
+            "--depth",
+            args.depth,
+            "Staking contract transitions per simulation (default 100)");
+
+        app.add_option(
+            "--runs",
+            args.runs,
+            "Number of simulations to execute (default max uint64)");
```
mkolosick, line 106: It would be good to add an arg for `config.enable_trace` so that it doesn't require recompilation.

category/execution/monad/staking/fuzzer/staking_contract_model.cpp
```diff
@@ -0,0 +1,670 @@
...
+    std::vector<u64_be>
+    StakingContractModel::get_validators_for_delegator(Address const &addr)
+    {
+        auto const [done, _, vs] = contract_.get_validators_for_delegator(
+            addr, 0, std::numeric_limits<uint32_t>::max());
+        MONAD_ASSERT(done);
+        return vs;
+    }
+
+    uint256_t StakingContractModel::calculate_rewards(
```
mkolosick, line 224: This could use a comment explaining what is going on. Especially as every use of the function uses inscrutable variable names.

category/execution/monad/staking/fuzzer/staking_contract_model.cpp
```diff
@@ -0,0 +1,670 @@
...
+        MONAD_ASSERT(output[31] == 1);
+        return outcome::success();
+    }
+}
+
+namespace monad::staking::test
+{
+    StakingContractModel::StakingContractModel()
+    {
+        trie_db_.commit(
```
mkolosick, line 47: Why not initialize in the exact same way as `SetUp` in `test_staking_contract.cpp`?

category/execution/monad/staking/fuzzer/staking_contract_model.hpp
```diff
@@ -0,0 +1,216 @@
...
+        // The commission rate of the active consensus validators:
+        ActiveConsensusCommissionMap active_consensus_commission_;
+
+        using DelegatorStakeMap = std::unordered_map<
+            std::tuple<uint64_t, Address>,
+            std::map<uint64_t, uint256_t, std::greater<uint64_t>>,
+            boost::hash<std::tuple<uint64_t, Address>>>;
+
+        // delegator_stake[{v, a, e}] is the stake of delegator(v, a)
+        // in epoch e:
```
mkolosick, line 64: This is a good design.

category/execution/monad/staking/fuzzer/staking_contract_fuzzer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    auto const end_time = std::chrono::steady_clock::now();
+
+    auto const success_ratio = success_count / static_cast<double>(depth);
+    auto const time =
+        static_cast<double>((end_time - start_time).count()) / 1'000'000;
+
+    std::cout << "    success ratio: " << success_ratio << '\n'
+              << "  simulation time: " << time << " ms" << std::endl;
+
+    return machine.gen();
```
mkolosick, line 75: This should get pulled out into the for loop in `main`.

category/execution/monad/staking/fuzzer/staking_contract_model.cpp
```diff
@@ -0,0 +1,670 @@
...
+            sum += withdrawal_reward(v, a, i);
+        }
+        return sum + unaccumulated_rewards(v, a);
+    }
+
+    Result<void>
+    StakingContractModel::syscall_on_epoch_change(u64_be next_epoch)
+    {
+        auto const input = abi_encode_uint(next_epoch);
+        pre_call(evmc_uint256be{});
```
mkolosick, line 312: Would it be better to have an overloaded `pre_call()` with no arguments instead?

category/execution/monad/staking/fuzzer/staking_contract_machine.hpp
```diff
@@ -0,0 +1,244 @@
...
+            syscall_reward,
+            precompile_add_validator,
+            precompile_delegate,
+            precompile_undelegate,
+            precompile_compound,
+            precompile_withdraw,
+            precompile_claim_rewards,
+            precompile_change_commission,
+            precompile_external_reward,
+            precompile_get_delegator,
```
mkolosick, line 109: This one is important to include 👍

category/execution/monad/staking/fuzzer/staking_contract_machine.hpp
```diff
@@ -0,0 +1,244 @@
...
+            boost::hash<std::tuple<uint64_t, Address>>>;
+
+        StakingContractModel model_;
+        std::mt19937_64 engine_;
+        // Whether the relatively slow public key assertions are enabled:
+        bool enable_pubkey_assertions_;
+        // Whether debug tracing is enabled:
+        bool enable_trace_;
+
+        std::vector<Address> all_addresses_;
```
mkolosick, line 55: I think these should be tracked in `StakingContractModel` instead as they are properties of the staking contract. Then `StakingContractMachine::model_precompile_*` methods would move into the respective `StakingContractModel` methods.

category/execution/monad/staking/fuzzer/staking_contract_model.hpp
```diff
@@ -0,0 +1,216 @@
...
+#include <category/execution/ethereum/db/trie_db.hpp>
+#include <category/execution/ethereum/state2/block_state.hpp>
+#include <category/execution/ethereum/state3/state.hpp>
+#include <category/execution/monad/staking/staking_contract.hpp>
+
+#include <boost/functional/hash.hpp>
+
+namespace monad::staking::test
+{
+    class StakingContractModel
```
mkolosick, line 26: There is significant repetition between `StakingContractModel` and `Stake` from
`test_staking_contract.cpp`.  This should be de-duplicated and they should both
wrap/inherit from some base class or `StakingContractModel` should just
wrap/inherit from `Stake`.

category/execution/monad/staking/fuzzer/staking_contract_model.hpp
```diff
@@ -0,0 +1,216 @@
...
+        OnDiskMachine mpt_machine_;
+        mpt::Db mpt_db_{mpt_machine_};
+        TrieDb trie_db_{mpt_db_};
+        BlockState block_state_{trie_db_, vm_};
+        State state_{block_state_, Incarnation{0, 0}};
+        NoopCallTracer call_tracer_{};
+        StakingContract contract_{state_, call_tracer_};
+
+        // An upper bound on reward rounding errors:
+        uint256_t error_bound_{};
```
mkolosick, line 38: A global error bound doesn't make sense to me. Why isn't this scoped to each validator?

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
```diff
@@ -0,0 +1,2065 @@
...
+        for (auto const &a : all_addresses_) {
+            f(a);
+        }
+    }
+
+    void StakingContractMachine::for_all_val_ids_and_addresses(
+        std::function<void(u64_be, Address const &)> f)
+    {
+        for_all_val_ids([&](u64_be v) {
+            for_all_addresses([&](Address const &a) { f(v, a); });
```
mkolosick, line 636: This loops through delegator addresses unrelated to the validator at hand? It should just iterate over the addresses relevant to the validator.

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
```diff
@@ -0,0 +1,2065 @@
...
+        });
+    }
+
+    bool StakingContractMachine::transition(Transition t)
+    {
+        TRACELN(enable_trace_, transition_string(t));
```
mkolosick: The trace should provide enough information to actually reconstruct the trace that led to the failure. The`TRACELN`s should be moved to the `syscall_`/`precompile_` methods in `StakingModelContract` (or similar) then they can include the relevant arguments to reproduce the trace. This would also handle the issue that the `model_precompile_add_validator` calls are not traced.

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
```diff
@@ -0,0 +1,2065 @@
...
+            Choice(0.80, [this](auto &) { return gen_old_address(); }));
+    }
+
+    std::pair<u64_be, Address>
+    StakingContractMachine::gen_validator_auth_address()
+    {
+        auto const &v = validator_auth_addresses_;
+        if (v.empty()) {
+            auto const [signer, msg, secp, bls, sender, value] =
+                gen_precompile_add_validator_input(
```
mkolosick, line 801: I think it would be better to avoid trying to generate an event that needs prior work to make it valid, but at minimum these checks and generating of validators should be pulled up into the relevant generator (`gen_precompile_change_commission_input` in this case).

This would also help to pull these internal calls out (similarly the `model_.syscall_snapshot` from `syscall_on_epoch_change`) as they bypass checking of post conditions.

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
```diff
@@ -0,0 +1,2065 @@
...
+                0.05,
+                [&](auto &) {
+                    return lower_bound == upper_bound ? upper_bound
+                                                      : upper_bound - 1;
+                }),
+            Choice(0.05, [&](auto &) { return lower_bound; }),
+            Choice(0.05, [&](auto &) { return upper_bound; }));
+    }
+
+    uint256_t StakingContractMachine::gen_stake(
```
mkolosick: We should document what is going on in this function and `gen_bound_biased_uint256`.

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
```diff
@@ -0,0 +1,2065 @@
...
+    {
+        uint64_t next_epoch{};
+        if (model_.epoch() == 0) {
+            next_epoch = 1 + gen() % 1'000'000;
+            auto const res = model_.syscall_on_epoch_change(next_epoch);
+            MONAD_ASSERT(res.has_value());
+        }
+        else {
+            next_epoch = model_.epoch() + 1;
+            if (!model_.in_epoch_delay_period()) {
```
mkolosick, line 1015: I think it would be better to restructure the generation of transitions to be phased (as the actual contract usage is). That is generating sequences of `action* snapshot action* epoch_change`. This also makes sure we are fuzzing cross epoch behaviors in a more robust way.

The idea to pick a random first number of epochs to jump to is good and should be kept. IIRC based on the discussion the first `syscall_on_epoch_change` occurs without a `syscall_snapshot` correct? If so, the above sequences should be changed to `(action* epoch_change(random_first_epoch)) (action* snapshot action* epoch_change)*`. This will make it more obvious that we are also testing actions occurring before the first time the epoch is set. (Unless that is not possible in which case we just do a random epoch_change first. But I think it is possible).

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
```diff
@@ -0,0 +1,2065 @@
...
+        MONAD_ASSERT(model_.epoch() == next_epoch);
+
+        for_all_val_ids([&](u64_be v) {
+            MONAD_ASSERT(
+                model_.active_consensus_stake(v) ==
+                model_.consensus_view(v).stake().load().native());
+        });
+    }
+
+    void StakingContractMachine::syscall_snapshot()
```
mkolosick: These functions interweave generating a random call and checking the post conditions. It would be better to separate these concerns. Would it make sense to move the post conditions and invariants into the `StakingContractModel` and have the `StakingContractMachine` be responsible just for generation and calling the model?

category/execution/monad/staking/fuzzer/staking_contract_machine.cpp
andreaslyn: Done

PR #1790, opened 2025-10-16
[execution] Trace state replay

dhil:
~Depends on #1762 and #1763.~

This PR is the execution part of #1753.


dhil:
Folded back into https://github.com/category-labs/monad/pull/1753.

PR #1802, opened 2025-10-20
[execution]: Use immutable data structures inside VersionStack<T>

kkuehlz:
[execution]: Add immer submodule

[execution]: Use immutable DS for types inside of VersionStack


jhunsaker:
no performance regression on replay at 12M

Issue #1602, opened 2025-09-04
Fix interpreter stalls related to bit instructions, AND,OR,XOR,NOT


andreaslyn:
Somehow `ADD; XOR` and `XOR; ADD` have lower latency than `XOR; XOR`, which is likely to be a performance bug. Also the big difference in performance between `ADD; NOT` and `XOR; NOT` is indicating a stall somewhere.

Using `force` for each 64 bit of the result seems to solve the problems.

```
$ ./build/test/vm/micro_benchmarks/vm-micro-benchmarks --title-filter "latency" --impl-filter "interp" --seq-filter "^(NOT|XOR)\n(ADD|XOR)\n$|^(ADD|XOR)\n(NOT|XOR)\n$"
interpreter
	BASIC_BIN_MATH; BASIC_BIN_MATH, constant input, latency

Baseline sequence
XOR
XOR

Results
ADD
XOR
	baseline:  0.915893 ms
	best:      0.877484 ms
	seq delta: -3.89149 ns
	total:     194 ms
XOR
ADD
	baseline:  0.926553 ms
	best:      0.858154 ms
	seq delta: -6.92999 ns
	total:     193 ms
XOR
XOR
	baseline:  0.925603 ms
	best:      0.926233 ms
	seq delta: 0.0638298 ns
	total:     200 ms

interpreter
	BASIC_UNA_MATH; BASIC_BIN_MATH, constant input, latency

Baseline sequence
XOR

Results
NOT
ADD
	baseline:  0.593755 ms
	best:      0.625125 ms
	seq delta: 3.17832 ns
	total:     131 ms
NOT
XOR
	baseline:  0.589115 ms
	best:      0.606965 ms
	seq delta: 1.80851 ns
	total:     130 ms

interpreter
	BASIC_BIN_MATH; BASIC_UNA_MATH, constant input, latency

Baseline sequence
XOR

Results
ADD
NOT
	baseline:  0.587016 ms
	best:      0.646825 ms
	seq delta: 6.05968 ns
	total:     133 ms
XOR
NOT
	baseline:  0.587555 ms
	best:      0.614035 ms
	seq delta: 2.68288 ns
	total:     129 ms
```

m-alvarez:
I noticed this while working on Barrett reduction as well, it has been fixed on that branch for a while but I've opened a standalone PR

PR #1814, opened 2025-10-22
[uint256] Force intermediate results of bitwise operations to avoid AVX stalls

m-alvarez:
As noted in #1602, bitwise operations on `uint256_t` are vectorized by GCC/Clang and incur store forwarding stalls when mixed with non-vectorized operations. This can be fixed by forcing the intermediate results.


dhil (commented):
What is the generated code diff before/after these changes?

andreaslyn (commented):
Did you forget `NOT`?

andreaslyn (approved):

PR #1823, opened 2025-10-23
[event] record state access events

kjcamann:
There are a few changes here:

- The ExecuteTransaction and ExecuteSystemTransaction functors are given a State object rather than creating one, so it can be owned by the outer transaction fiber. After the functor is done executing the transaction, this outer fiber gives the State to the recording code

- `class State` gains the `State::clear` method, so that it can be reset when re-execution is needed

- record_txn_events.cpp gains functions for recording State objects as `ACCOUNT_ACCESS_LIST_HEADER`, `ACCOUNT_ACCESS`, and `STORAGE_ACCESS` events

- All state-affecting changes to the block that occur before transaction and after the transaction are also recorded

### DIFF
--- a/category/execution/ethereum/state3/account_state.hpp
+++ b/category/execution/ethereum/state3/account_state.hpp
@@ -57,11 +57,11 @@ private:
     friend class State;
     friend class BlockState;
 
-    // the classes below can access the account_ field just for logging but
-    // CANNOT use it to make decisions affecting the final state (state of
-    // accounts) of execution.
-    friend struct trace::PrestateTracer;
-    friend struct trace::StateDiffTracer;
+    friend std::optional<Account> const &
+    get_account_for_trace(AccountState const &as)
+    {
+        return as.account_;
+    }
 
 public:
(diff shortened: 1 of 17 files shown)


category/execution/ethereum/state3/account_state.hpp
```diff
@@ -57,11 +57,11 @@ class AccountState : public AccountSubstate
     friend class State;
     friend class BlockState;
 
-    // the classes below can access the account_ field just for logging but
-    // CANNOT use it to make decisions affecting the final state (state of
-    // accounts) of execution.
-    friend struct trace::PrestateTracer;
-    friend struct trace::StateDiffTracer;
+    friend std::optional<Account> const &
```
aa755, line 60: feel free to ignore
```suggestion
    /** Do NOT use this to read the account's balance if you use the read value to make decisions affecting the final state execution (receipts or state of accounts). For that, use the account balance methods  in State:  `State::record_constraints_for_debit`, `State::get_current_balance_pessimistic`, `State::get_original_balance_pessimistic` */
    friend std::optional<Account> const &
```

PR #1825, opened 2025-10-23
Merge CacheNode and Node back to single class

maxkozlovsky:
Remove using next pointers in rodb cache, always use hash, since using a shared ptr can keep nodes that should be recycled.


category/mpt/test/db_test.cpp
```diff
@@ -1005,7 +1005,7 @@ TEST_F(OnDiskDbWithFileAsyncFixture, async_get_node_then_async_traverse)
             }
             else {
                 traverse_sender.traverse_root =
-                    copy_node<Node>(res.assume_value().get());
+                    copy_node(res.assume_value().get());
```
Chen-Yifan: no longer need copy_node() here, since the find result type now matches what traverse_root expects. There are some in other test files that we can remove too

PR #1851, opened 2025-10-30
switch vm/unit/call_tests to TraitsTest

goodlyrottenapple:


test/unit/common/include/monad/test/traits_test_constants.hpp
```diff
@@ -0,0 +1,32 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <category/vm/evm/monad/revision.h>
+
+#include <evmc/evmc.h>
+
+std::vector<std::variant<evmc_revision, monad_revision>> monad_evm_revisions()
```
goodlyrottenapple: I've had to move this function into a separate header because I
a) was getting linker errors due to multiple declarations when including `test/unit/common/include/monad/test/traits_test.hpp` in `fixture.hpp`, which is itself included in several unit test files.
b) wanted to keep `monad_evm_revisions()` in `test/unit/common`
c) didn't know how to do the CMAKE magic of moving `monad_evm_revisions()` into `traits_test.cpp` that would then be added/linked with all the unit tests

alternatively, `monad_evm_revisions()` could be inlined into `evm_tests.cpp`

category/vm/runtime/call.hpp
```diff
@@ -240,6 +240,7 @@ namespace monad::vm::runtime
         uint256_t const *args_size_ptr, uint256_t const *ret_offset_ptr,
         uint256_t const *ret_size_ptr, std::int64_t remaining_block_base_gas)
     {
+        MONAD_VM_DEBUG_ASSERT(traits::evm_rev() >= EVMC_BYZANTIUM);
```
goodlyrottenapple: Made consistent with `staticcall` which has a similar assertion. Alternatively, we should remove both.

test/vm/unit/runtime/fixture.hpp
```diff
@@ -26,15 +27,17 @@
...
 
+extern "C" void tests_trampoline(void *, void (*)(void *), void *);
+
 namespace monad::vm::compiler::test
 {
     using namespace runtime;
     using namespace evmc::literals;
 
-    class RuntimeTest : public testing::Test
+    class RuntimeTestBase
```
goodlyrottenapple, line 37: Split `RuntimeTest` into `RuntimeTestBase` + `RuntimeTest`/`RuntimeTraitsTest` to share the non-traits specific methods. Some tests do not need traits, so `RuntimeTest` will remain for those

andreaslyn (dismissed):

PR #1865, opened 2025-10-31
Added register reserv

andreaslyn:


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2270,6 +2270,7 @@ namespace monad::vm::compiler::native
             }
         }
         else {
+            AvxRegReserv const e_reserv{e};
```
laurenthuberdeau, line 2296: I think `bswap_to_ymm` supports the case where `src == dst` so we may want to allow `alloc_avx_reg` to reuse the AVX register assigned to `e`?

Baltoli (approved):

goodlyrottenapple (approved):

dhil (approved):

PR #1868, opened 2025-11-03
expand clang-tidy checks to category/async

mkolosick:
`async` directory now passes `clang-tidy` checks

### DIFF
--- a/category/async/sender_errc.hpp
+++ b/category/async/sender_errc.hpp
@@ -97,9 +97,9 @@ public:
             std::reference_wrapper<filled_write_buffer>>
             payload;
 
-        constexpr value_type() noexcept {}
+        constexpr value_type() noexcept = default;
 
-        constexpr value_type(sender_errc code_) noexcept
+        constexpr explicit value_type(sender_errc code_) noexcept
             : code(code_)
         {
         }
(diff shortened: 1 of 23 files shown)


category/async/sender_errc.hpp
```diff
@@ -276,7 +276,7 @@ namespace detail
         thread_local_sender_errc_with_payload_code_allocator() = default;
 
         template <class U>
-        constexpr thread_local_sender_errc_with_payload_code_allocator(
+        constexpr explicit thread_local_sender_errc_with_payload_code_allocator(
```
maxkozlovsky: for allocators rebind constructor should not be marked explicit I think, mark it as excepted from checks instead

maxkozlovsky (dismissed):

category/mpt/test/fiber_future_wrapped_read.cpp
```diff
@@ -40,13 +40,13 @@
...
-    : public monad::test::AsyncTestFixture<::testing::Test>
+namespace
 {
-};
+    struct FiberFutureWrappedFind
+        : public monad::test::AsyncTestFixture<::testing::Test>
+    {
+    };
 
-TEST_F(FiberFutureWrappedFind, single_thread_fibers_read)
```
Chen-Yifan, line 48: gtests don't need to be in anonymous namespace. Diff would be less noisy if moving them out in this and `mixed_async_sync_loads_test.cpp`.

Chen-Yifan (approved):

PR #1870, opened 2025-11-03
remove unused traits template argument in emitter

goodlyrottenapple:


test/vm/unit/emitter_tests.cpp
```diff
@@ -2407,7 +2395,7 @@ TEST(Emitter, exp)
         for (int i = 0; i < rep_count; ++i) {
             emit.push(513); // some exponent over 512
             emit.push(13); // base (with popcount != 1)
-            emit.exp<EvmTraits<EVMC_SPURIOUS_DRAGON>>(
+            emit.exp<EvmTraits<EVMC_LATEST_STABLE_REVISION>>(
```
goodlyrottenapple, line 2398: This test is very slow (>10s on my machine) so not sure it makes sense to turn it into a typed test. Hence just switched to `EVMC_LATEST_STABLE_REVISION` to at least detect any potential breakage when the evm revision gets updated

andreaslyn (approved):

PR #1912, opened 2025-11-13
Check reserve balance in `eth_call`

Baltoli:


msetrin (dismissed):

category/rpc/monad_executor_test.cpp
```diff
@@ -3481,3 +3482,219 @@ TEST_F(EthCallFixture, prestate_override_state)
...
+        rlp_block_id.data(),
+        rlp_block_id.size(),
+        state_override,
+        complete_callback,
+        (void *)&ctx,
+        NOOP_TRACER,
+        true);
+    f.get();
+
+    EXPECT_TRUE(ctx.result->status_code == EVMC_SUCCESS);
```
dhil: Nit, the following produces an ever-so-slightly better diagnostics message in the event of failure:
```suggestion
    EXPECT_EQ(ctx.result->status_code, EVMC_SUCCESS);
```

category/rpc/monad_executor_test.cpp
```diff
@@ -3481,3 +3482,219 @@ TEST_F(EthCallFixture, prestate_override_state)
...
+        rlp_block_id.data(),
+        rlp_block_id.size(),
+        state_override,
+        complete_callback,
+        (void *)&ctx,
+        NOOP_TRACER,
+        true);
+    f.get();
+
+    EXPECT_TRUE(ctx.result->status_code == EVMC_REVERT);
```
dhil: Nit, the following produces an ever-so-slightly better diagnostics message in the event of failure:
```suggestion
    EXPECT_EQ(ctx.result->status_code, EVMC_REVERT);
```

dhil (dismissed):

PR #1926, opened 2025-11-19
Native transfer tracer for `eth_simulateV1`

Baltoli:
Implements the execution client plumbing for the [`traceTransfers` flag in `eth_simulatev1`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth). This flag causes ERC-20 style log events to get emitted when native token transfers are made (on calls, contract creation, and selfdestruction).

The flag is not yet wired up to anything except a unit test; when the rest of `eth_simulatev1` is implemented, enabling the feature is a straightforward integration.

### DIFF
--- a/category/execution/ethereum/evm.cpp
+++ b/category/execution/ethereum/evm.cpp
@@ -54,13 +54,17 @@ namespace
         return state.record_balance_constraint_for_debit(msg.sender, value);
     }
 
-    void
-    transfer_balances(State &state, evmc_message const &msg, Address const &to)
+    template <Traits traits>
+    void transfer_balances(
+        State &state, EvmcHost<traits> &host, evmc_message const &msg,
+        Address const &to)
     {
         uint256_t const value = intx::be::load<uint256_t>(msg.value);
         state.subtract_from_balance(msg.sender, value);
         state.add_to_balance(to, value);
+        host.emit_native_transfer_event(msg.sender, to, value);
     }
+
 } // anonymous namespace
 
 template <Traits traits>
@@ -112,7 +116,8 @@ evmc::Result deploy_contract_code(
 EXPLICIT_TRAITS(deploy_contract_code);
 
 template <Traits traits>
-std::optional<evmc::Result> pre_call(evmc_message const &msg, State &state)
+std::optional<evmc::Result>
+pre_call(EvmcHost<traits> &host, evmc_message const &msg, State &state)
 {
     state.push();
 
@@ -124,7 +129,7 @@ std::optional<evmc::Result> pre_call(evmc_message const &msg, State &state)
             return evmc::Result{EVMC_INSUFFICIENT_BALANCE, msg.gas};
         }
         else if (!static_call) {
-            transfer_balances(state, msg, msg.recipient);
+            transfer_balances<traits>(state, host, msg, msg.recipient);
         }
     }
 
@@ -235,7 +240,7 @@ evmc::Result execute_create_message(
     constexpr auto starting_nonce =
         traits::evm_rev() >= EVMC_SPURIOUS_DRAGON ? 1 : 0;
     state.set_nonce(contract_address, starting_nonce);
-    transfer_balances(state, msg, contract_address);
+    transfer_balances<traits>(state, *host, msg, contract_address);
 
     evmc_message const m_call{
         .kind = EVMC_CALL,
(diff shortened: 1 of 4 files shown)


category/execution/ethereum/evmc_host.hpp
```diff
@@ -142,8 +147,12 @@ struct EvmcHost final : public EvmcHostBase
         try {
             auto const [result, transferred_balance] =
                 state_.selfdestruct<traits>(address, beneficiary);
+
             call_tracer_.on_self_destruct(
                 address, beneficiary, transferred_balance);
+            emit_native_transfer_event(
+                address, beneficiary, transferred_balance);
+
```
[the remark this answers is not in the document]
Baltoli: Test added

category/execution/ethereum/evmc_host.hpp
```diff
@@ -193,16 +202,52 @@ struct EvmcHost final : public EvmcHostBase
...
     }
+
+private:
+    void emit_native_transfer_event(
+        Address const &from, Address const &to, uint256_t const &value)
+    {
+        if (log_native_transfers_ && value > 0) {
+            constexpr auto sender =
+                0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address;
+            constexpr auto signature =
```
dhil: I suppose the storage duration doesn't matter here, though, I'd have added `static` here just so they are optimized out  at `-O0` like at `-O{1,2}`.

```suggestion
            static constexpr auto sender =
                0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address;
            static constexpr auto signature =
```

dhil (approved):

category/execution/ethereum/evmc_host.hpp
```diff
@@ -142,8 +147,12 @@ struct EvmcHost final : public EvmcHostBase
         try {
             auto const [result, transferred_balance] =
                 state_.selfdestruct<traits>(address, beneficiary);
+
             call_tracer_.on_self_destruct(
                 address, beneficiary, transferred_balance);
+            emit_native_transfer_event(
```
0xSqualo, line 154: @Baltoli won't this `emit` a transfer event even if none took place?

For [example if a user called `SELFDESTRUCT` on a later `incarnation`](https://github.com/category-labs/monad/blob/0beaae90a6bd5221d99b643bda52536d52a8bef1/category/execution/ethereum/state3/state.cpp#L418-L422), then no `value` would be transferred, but the event would still be emitted.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -142,8 +147,18 @@ struct EvmcHost final : public EvmcHostBase
...
             auto const [result, transferred_balance] =
                 state_.selfdestruct<traits>(address, beneficiary);
+
             call_tracer_.on_self_destruct(
                 address, beneficiary, transferred_balance);
+
+            // Skip emitting native transfer events when the beneficiary and
+            // destructed contract are the same account (i.e. no net transfer of
+            // funds).
+            if (result && address != beneficiary) {
```
es3n1n: Shouldn't `result` be ignored here in this check? `result` here indicates whether this is the first selfdestruct of this account, not whether a transfer occurred. 

Isn't it theoretically possible to:
1. selfdestruct
2. fund the contract again (e.g. via a CALL with value?)
3. selfdestruct again

within the same transaction? That would cause the `result` on the second selfdestruct to be `false`, and the event would not be emitted.

Worth mentioning, that from my understanding geth doesn't care whether the transfer actually took place or not. The only check they have is [value > 0](https://github.com/ethereum/go-ethereum/blob/30656d714edaf4756729bad0672cde5790128a8f/internal/ethapi/logtracer.go#L83-L85) and they call this tracer [unconditionally](https://github.com/ethereum/go-ethereum/blob/30656d714edaf4756729bad0672cde5790128a8f/core/vm/instructions.go#L938-L945). So if we want to follow geth's semantics, we might want to remove this check altogether; in my opinion, emitting only when the transfer actually took place makes sense though.

category/execution/ethereum/evm.cpp
```diff
@@ -54,13 +54,17 @@ namespace
...
-    transfer_balances(State &state, evmc_message const &msg, Address const &to)
+    template <Traits traits>
+    void transfer_balances(
+        State &state, EvmcHost<traits> &host, evmc_message const &msg,
+        Address const &to)
     {
         uint256_t const value = intx::be::load<uint256_t>(msg.value);
         state.subtract_from_balance(msg.sender, value);
         state.add_to_balance(to, value);
+        host.emit_native_transfer_event(msg.sender, to, value);
```
es3n1n, line 65: should we check for `msg.sender != to` here as well?

i think for `CALLCODE` the `msg.sender` will be equal to `to`

(or maybe just add this condition to the `emit_native_transfer_event` itself)

dhil: Yes, that is right. Thanks!

PR #1928, opened 2025-11-19
expand clang-tidy checks to category/mpt

mkolosick:


category/mpt/test/mixed_async_sync_loads_test.cpp
```diff
@@ -69,7 +73,8 @@ TEST_F(MixedAsyncSyncLoadsTest, works)
     monad::mpt::AsyncInflightNodes inflights;
     monad::mpt::NodeCache node_cache{
         1000 * monad::mpt::NodeCache::AVERAGE_NODE_SIZE};
-    std::shared_ptr<Node> cache_root = root;
+    // NOLINTNEXTLINE(performance-unnecessary-copy-initialization)
+    std::shared_ptr<Node> const cache_root = root;
```
Chen-Yifan: this line seems unnecessary, can just remove cache_root and use root in the following

category/mpt/upward_tnode.hpp
```diff
@@ -185,6 +185,9 @@ struct CompactTNode : public UpwardTreeNodeBase<CompactTNode>
     bool const cache_node{false};
     Node::SharedPtr node{nullptr};
 
+    // NOLINTBEGIN(clang-analyzer-optin.cplusplus.UninitializedObject)
```
Chen-Yifan: can we remove these UninitializedObject NOLINT since it's fixed in version 18?

PR #1940, opened 2025-11-27
Support CLZ opcode

laurenthuberdeau:
## Context

Follow-up to https://github.com/category-labs/monad/pull/1939

To be compatible with the Ethereum Fusaka network upgrade, we must support the following features:

- [EIP-7939](https://eips.ethereum.org/EIPS/eip-7939): Count leading zeros (CLZ) opcode
- [EIP-7823](https://eips.ethereum.org/EIPS/eip-7823): Set upper bounds for MODEXP
- [EIP-7883](https://eips.ethereum.org/EIPS/eip-7883): ModExp Gas Cost Increase
- [EIP-7918](https://eips.ethereum.org/EIPS/eip-7918): Blob base fee bounded by execution cost

This PR adds support for the CLZ opcode, gated by the `EVMC_OSAKA` revision.

### DIFF
--- a/category/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/compiler/ir/x86/emitter.cpp
@@ -2588,6 +2588,47 @@ namespace monad::vm::compiler::native
             std::move(shift), std::move(value), live);
     }
 
+    // Discharge
+    void Emitter::clz()
+    {
+        auto const elem = stack_.pop();
+
+        if (elem->literal()) {
+            auto const &x = elem->literal()->value;
+            push(runtime::countl_zero(x));
+            return;
+        }
+
+        discharge_deferred_comparison();
+
+        if (elem->general_reg()) {
+            auto const &gpq = general_reg_to_gpq256(*elem->general_reg());
+            std::array<asmjit::x86::Gpq, 4> const &gpq_r64 = {
+                gpq[0].r64(), gpq[1].r64(), gpq[2].r64(), gpq[3].r64()};
+            array_leading_zeros(gpq_r64);
+        }
+        else if (elem->stack_offset()) {
+            array_leading_zeros(stack_offset_to_mem256(*elem->stack_offset()));
+        }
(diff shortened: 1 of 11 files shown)


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -7928,8 +7969,36 @@ namespace monad::vm::compiler::native
...
+            // If arr[i - 1] != 0, jump to end_lbl
+            as_.jnc(end_lbl);
+        }
+        as_.bind(end_lbl);
+    }
+
     // Performs byte_width operation on array of operands. Assumes that the
     // operands are ordered from least significant to most significant.
+    // This implementation assumes that the input value is biased towards
+    // smaller values.
```
andreaslyn: I think this comment was supposed to belong to `array_leading_zeros`

laurenthuberdeau: You're right that this comment is a bit unclear. What it's trying to say is that `array_leading_zeros` and `array_byte_width` use 2 different approaches (one branchless, one branchful) for calculating more-or-less the same thing, with each approach optimizing for different inputs:
- branchful: variable latency, returning faster for wider inputs (executing only `lzcnt, jnc` if 256-bit wide), such as is the case when they are uniformly distributed.
- branchless: always the same latency, no early exit for large operands so better in the case where the input won't occupy the full 256 bits such as the exponent of the `EXP` operation.

I'll rewrite the comment to make that clearer.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2586,6 +2586,47 @@ namespace monad::vm::compiler::native
             std::move(shift), std::move(value), live);
     }
 
+    // Discharge
+    void Emitter::clz()
```
laurenthuberdeau, line 2592: Ran the fuzzer locally (with the osaka revision) for some time and it didn't find any counter-example.

PR #1944, opened 2025-11-28
Unify single buffer and scatter read handling in AsyncIO

maxkozlovsky:
Previously, single buffer reads and scatter reads had inconsistent behavior for limit checking, queueing, and statistics tracking:
- Separate counters (inflight_rd vs inflight_rd_scatter)
- Only single buffer reads were queued when over limit
- Only single buffer reads tracked bytes_read

This commit unifies the handling so both read types share:
- Single inflight_rd counter for all reads
- Same concurrent_read_io_limit_ and queueing mechanism
- Consistent bytes_read tracking
- Same dequeue behavior on completion

Implementation changes:
- deferred read queue now stores sqe entries directly
- Removed inflight_rd_scatter and max_inflight_rd_scatter counters
- Scatter reads now check limit and queue when over limit
- Scatter reads now track bytes_read by summing iovec lengths

This code was generated using Claude Sonnet 4.5

### DIFF
--- a/category/async/io.cpp
+++ b/category/async/io.cpp
@@ -282,10 +282,12 @@ void AsyncIO::account_read_(size_t size)
     records_.bytes_read += size;
 }
 
-void AsyncIO::submit_request_sqe_(
-    std::span<std::byte> buffer, chunk_offset_t chunk_and_offset,
-    void *uring_data, enum erased_connected_operation::io_priority prio)
+void AsyncIO::prepare_read_sqe_(
+    struct io_uring_sqe *sqe, std::span<std::byte> buffer,
+    chunk_offset_t chunk_and_offset, void *uring_data,
+    enum erased_connected_operation::io_priority prio)
 {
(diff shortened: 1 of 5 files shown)


category/async/io.cpp
```diff
@@ -489,10 +507,34 @@ size_t AsyncIO::poll_uring_(bool blocking, unsigned poll_rings_mask)
...
+                    read_size = sqe->len;
+                }
+                else if (sqe->opcode == IORING_OP_READV) {
+                    struct iovec const *iovecs =
+                        reinterpret_cast<struct iovec const *>(sqe->addr);
+                    read_size = iov_length(
+                        std::span<const struct iovec>(iovecs, sqe->len));
+                }
+                else {
+                    MONAD_ASSERT(false && "unexpected opcode in deferred read");
```
Chen-Yifan: can change to MONAD_ABORT();

maxkozlovsky: fixed

Issue #1904, opened 2025-11-11
Store `tx_context` as a pointer in runtime `Environment`


andreaslyn:
We can pass `tx_context` as a pointer to the VM class and into `runtime::Context::from`. Will then need to adapt compiler and interpreter accordingly. Existing tests are expected to cover this change.

PR #1954, opened 2025-12-02
Return a const pointer to `evmc_tx_context` from evmc host

goodlyrottenapple:
Implements  #1904 by changing the `evmc_host_interface` to return a const pointer to the `evmc_tx_context` struct. This pointer can then be stored in the `Context` passed to the interpreter/compiler VM.

### DIFF
--- a/category/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/compiler/ir/x86/emitter.cpp
@@ -35,6 +35,8 @@
 #include <asmjit/core/operand.h>
 #include <asmjit/x86/x86operand.h>
 
+#include <evmc/evmc.h>
+
 #include <algorithm>
 #include <array>
 #include <bit>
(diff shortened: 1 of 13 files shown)


category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -3755,9 +3776,16 @@ namespace monad::vm::compiler::native
         stack_.push(std::move(dst));
     }
 
-    void Emitter::read_context_word(int32_t offset)
+    void Emitter::read_context_word(
```
andreaslyn: I think this and `read_context_address` are better as two separate functions, to avoid some unnecessary branching. For example:

* `void read_context_word(int32_t offset)` - keep same as before this PR
* `void read_indirect_context_word(int32_t ptr_offset, int32_t offset)` - the `else` case in this PR

goodlyrottenapple: sure

brett-monad (dismissed):

Baltoli (approved):

andreaslyn (approved):

PR #1959, opened 2025-12-04
moving reserve balance API to monad_rev traits

mkolosick:


cmd/monad/runloop_monad.cpp
```diff
@@ -298,9 +298,7 @@ Result<BlockExecOutput> propose_block(
                 Transaction const &tx,
                 uint64_t const i,
                 State &state) {
-                return chain.revert_transaction(
-                    block.header.number,
-                    block.header.timestamp,
+                return chain.revert_transaction<traits>(
```
Baltoli: Can we now get rid of `Chain::revert_transaction` entirely?

Baltoli: This could just be `revert_transaction<traits>`, I think

category/execution/monad/reserve_balance.cpp
```diff
@@ -121,24 +131,26 @@ MONAD_ANONYMOUS_NAMESPACE_END
...
+        return dipped_into_reserve<traits>(
+            sender, tx, base_fee_per_gas, i, ctx, state);
     }
-    else if (monad_rev >= MONAD_ZERO) {
+    else if constexpr (traits::monad_rev() >= MONAD_ZERO) {
         return false;
     }
     else {
         MONAD_ABORT("invalid revision for revert");
     }
```
Baltoli: Now unreachable code because it's constexpr

mkolosick: Isn't this still reachable if you do say `revert_monad_transaction<EVMC_CANCUN>`?

mkolosick: Unless we add a trait bound?

Baltoli: We should make this `EXPLICIT_MONAD_TRAITS` to make that a compile / link error

Baltoli: Oh, it already is

category/execution/monad/chain/monad_chain.cpp
```diff
@@ -31,7 +31,8 @@
 #include <category/execution/monad/reserve_balance.hpp>
 #include <category/execution/monad/system_sender.hpp>
 #include <category/execution/monad/validate_monad_transaction.hpp>
-#include <category/vm/evm/switch_traits.hpp>
+#include <category/vm/evm/explicit_traits.hpp>
+#include <category/vm/evm/traits.hpp>
```
Baltoli: No longer need these includes now that the header doesn't reference traits

category/execution/monad/chain/monad_chain.hpp
```diff
@@ -20,6 +20,7 @@
 #include <category/execution/ethereum/chain/chain.hpp>
 #include <category/execution/ethereum/core/address.hpp>
 #include <category/vm/evm/monad/revision.h>
+#include <category/vm/evm/traits.hpp>
```
Baltoli: Similarly here

PR #1961, opened 2025-12-05
Tag x86 emitter stack offset locations with AVX or GPR

### LINKED ISSUE #1864: Tag x86 emitter stack offset locations with AVX or GPR
The x86 emitter does not distinguish between stack offsets which have been stored via GPR or AVX. This is problematic, becausethe emitter often prefers AVX instructions when the operands are stack offsets, which causes a store forwarding stall when stack offset was stored via GPR.

This basic block is for example problematic
```
MUL
XOR
```

The MUL instruction translates to a runtime call, which stores the mul result in a stack offset using GPRs. The XOR instruction will load the stack offset using AVX. 

The proposed solution to this is to tag stack offsets when they are known to be stored by GPRs, so that we can choose the GPR versions of instructions when one of the operands is a stack offset tagged with GPR.

goodlyrottenapple:
Track where stack offset was moved from, to help prevent store forwarding stalls. Fixes #1864.
The new microbenchmark shows the stall disappearing:

main:
```
compiler
	store forwarding stall, constant input, throughput

	baseline:  0.104079 ms
	best:      0.114349 ms
	seq delta: 1.04053 ns
	total:     24 ms

compiler
	store forwarding stall, constant input, latency

	baseline:  0.118239 ms
	best:      0.153188 ms
	seq delta: 3.54093 ns
	total:     30 ms
```

feature:
```
compiler
	store forwarding stall, constant input, throughput

	baseline:  0.104739 ms
	best:      0.105049 ms
	seq delta: 0.0314083 ns
	total:     23 ms

compiler
	store forwarding stall, constant input, latency

	baseline:  0.118699 ms
	best:      0.118269 ms
	seq delta: -0.0435664 ns
	total:     27 ms
```

These are the results for other microbenchmarks, namely pairing of BASIC_BIN_MATH; BASIC_BIN_MATH, only showing pairs where the difference in the sequence deltas between main and feature branch was >0.5ns:

throughput:
throughput

latency:
latency

Note: It appears that the results with DIV are quite noisy based on multiple measurements. This is probably due to the relative complexity of the div runtime function comparted to the other bin math operations

### DIFF
--- a/category/vm/compiler/ir/x86/emitter.cpp
+++ b/category/vm/compiler/ir/x86/emitter.cpp
@@ -479,8 +479,10 @@ namespace monad::vm::compiler::native
             mov_arg(*remaining_gas_arg_, remaining_base_gas_);
         }
         if (result_arg_.has_value()) {
-            auto result =
-                em_->stack_.alloc_stack_offset(em_->stack_.top_index() + 1);
+            // Assumes that result will be written to memory from general
+            // registers
+            auto result = em_->stack_.alloc_stack_offset(
+                em_->stack_.top_index() + 1, PrevLoc::GprReg);
             mov_arg(*result_arg_, stack_offset_to_mem(*result->stack_offset()));
             em_->stack_.push(std::move(result));
         }
@@ -1488,7 +1490,8 @@ namespace monad::vm::compiler::native
                 }
                 else {
                     MONAD_VM_DEBUG_ASSERT(d->general_reg().has_value());
-                    auto const m = stack_offset_to_mem(StackOffset{*it});
+                    auto const m =
+                        stack_offset_to_mem(StackOffset{*it, PrevLoc::GprReg});
                     // Move to final stack offset:
                     mov_general_reg_to_mem(*d->general_reg(), m);
                     // Point to next stack offset:
@@ -1503,7 +1506,9 @@ namespace monad::vm::compiler::native
             // Move to remaining final stack offsets:
             for (; it != is.end(); ++it) {
                 if (!d->stack_offset() || d->stack_offset()->offset != *it) {
-                    as_.vmovaps(stack_offset_to_mem(StackOffset{*it}), yx1);
+                    as_.vmovaps(
+                        stack_offset_to_mem(StackOffset{*it, PrevLoc::AvxReg}),
+                        yx1);
                     MONAD_COMPILER_X86_INC_FINAL_WRITE_COUNT();
                 }
             }
@@ -1582,13 +1587,16 @@ namespace monad::vm::compiler::native
                 as_.vmovaps(
                     yx2, stack_offset_to_mem(*cycle[k - 1]->stack_offset()));
                 for (int32_t const i : cycle[k]->stack_indices()) {
-                    as_.vmovaps(stack_offset_to_mem(StackOffset{i}), yx1);
+                    as_.vmovaps(
+                        stack_offset_to_mem(StackOffset{i, PrevLoc::AvxReg}),
+                        yx1);
                     MONAD_COMPILER_X86_INC_FINAL_WRITE_COUNT();
                 }
                 std::swap(yx1, yx2);
             }
             for (int32_t const i : e->stack_indices()) {
-                as_.vmovaps(stack_offset_to_mem(StackOffset{i}), yx1);
+                as_.vmovaps(
+                    stack_offset_to_mem(StackOffset{i, PrevLoc::AvxReg}), yx1);
                 MONAD_COMPILER_X86_INC_FINAL_WRITE_COUNT();
             }
         }
@@ -2099,7 +2107,7 @@ namespace monad::vm::compiler::native
     Emitter::mov_avx_reg_to_stack_offset(StackElemRef elem, int32_t preferred)
     {
         MONAD_VM_DEBUG_ASSERT(elem->avx_reg().has_value());
-        stack_.insert_stack_offset(elem, preferred);
+        stack_.insert_stack_offset(elem, preferred, PrevLoc::AvxReg);
         auto const y = avx_reg_to_ymm(*elem->avx_reg());
         as_.vmovaps(stack_offset_to_mem(*elem->stack_offset()), y);
     }
@@ -2120,7 +2128,7 @@ namespace monad::vm::compiler::native
     Emitter::mov_general_reg_to_stack_offset(StackElem &elem, int32_t preferred)
     {
         MONAD_VM_DEBUG_ASSERT(elem.general_reg().has_value());
-        stack_.insert_stack_offset(elem, preferred);
+        stack_.insert_stack_offset(elem, preferred, PrevLoc::GprReg);
         mov_general_reg_to_mem(
             *elem.general_reg(), stack_offset_to_mem(*elem.stack_offset()));
     }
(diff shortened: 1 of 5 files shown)


test/vm/micro_benchmarks/main.cpp
```diff
@@ -932,4 +937,87 @@ int main(int argc, char **argv)
...
+                              .add()
+                              .dup7()
+                              .dup7()
+                              .add()
+                              .dup8()
+                              .dup8()
+                              .add()
+                              .dup7()
+                              .dup5()
+                              .xor_() // add() in the baseline
```
andreaslyn, line 1015: Nice

test/vm/micro_benchmarks/main.cpp
```diff
@@ -932,4 +937,87 @@ int main(int argc, char **argv)
...
+                             .dup6()
+                             .add()
+                             .dup7()
+                             .dup7()
+                             .add()
+                             .dup8()
+                             .dup8()
+                             .add()
+                             .dup7()
+                             .dup5()
```
andreaslyn, line 980: Is there a particular reason to use `dup5` here instead of say `dup7`? My initial impression is that `dup7` would be more natural, or perhaps even remove this dup instruction entirely.

test/vm/micro_benchmarks/main.cpp
```diff
@@ -932,4 +937,87 @@ int main(int argc, char **argv)
...
+                             .swap9()
+                             .pop()
+                             .pop()
+                             .pop()
+                             .pop()
+                             .pop()
+                             .pop()
+                             .pop()
+                             .pop()
+                             .pop(),
```
andreaslyn: The `pop; pop; ...` sequence is problematic because of compiler optimizations. It is not a major problem, becaue the compiler does not perform this optimizations, but LLVM does eliminate the dead instructions. Consider using `add; add; ...` instead of `pop; pop; ...`.

test/vm/micro_benchmarks/main.cpp
goodlyrottenapple: I just tried a couple until I hit two that were sufficiently far back to both be spilled to memory...

category/vm/compiler/ir/x86/virtual_stack.hpp
```diff
@@ -41,9 +41,17 @@ namespace monad::vm::compiler::native
 
     bool operator==(Literal const &a, Literal const &b);
 
+    enum class PrevLoc
+    {
+        Unknown,
+        Avx,
+        Gpr
+    };
```
laurenthuberdeau, line 49: Nit: should the AVX and GPR cases match `LocationType` for consistency?

category/vm/compiler/ir/x86/virtual_stack.hpp
```diff
@@ -41,9 +41,17 @@ namespace monad::vm::compiler::native
...
+
     struct StackOffset
     {
         std::int32_t offset;
+        PrevLoc moved_from;
```
laurenthuberdeau, line 54: Have you considered placing `moved_from` on the `StackElem` structure and/or default-initializing it with `Unknown`? 

Since `get_una_arguments` and `prepare_avx_or_general_arguments_commutative` have access to the full `StackElem` structure (not just the `StackOffset`), I think it could make inserting `moved_from` information less repetitive (and make the diff smaller :)).

laurenthuberdeau (commented):
Nice!

category/vm/compiler/ir/x86/virtual_stack.hpp
goodlyrottenapple: do you mean use `PrevLoc::AvxReg` instead of `PrevLoc::Avx`?

laurenthuberdeau: Yes

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2141,7 +2148,7 @@ namespace monad::vm::compiler::native
     Emitter::mov_literal_to_stack_offset(StackElemRef elem, int32_t preferred)
     {
         MONAD_VM_DEBUG_ASSERT(elem->literal().has_value());
-        stack_.insert_stack_offset(elem, preferred);
+        stack_.insert_stack_offset(elem, preferred, PrevLoc::Unknown);
         mov_literal_to_mem<true, true>(
             elem, stack_offset_to_mem(*elem->stack_offset()));
```
andreaslyn: I think `mov_literal_to_mem` can return a `PrevLoc` instead of using unknown

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -5676,6 +5683,17 @@ namespace monad::vm::compiler::native
                 return {new_dst, new_dst, LocationType::GeneralReg};
             }
             MONAD_VM_DEBUG_ASSERT(dst->stack_offset().has_value());
+            // to prevent a store forwarding stall, load the stack offset back
+            // into gpr if we know it was previously stored there from another
+            // gpr
+            if (dst->stack_offset().value().moved_from == PrevLoc::GprReg) {
```
andreaslyn: The sanity check of `.value()` is not necessary
```suggestion
            if (dst->stack_offset()->moved_from == PrevLoc::GprReg) {
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -5724,6 +5742,14 @@ namespace monad::vm::compiler::native
                     LocationType::AvxReg};
             }
             MONAD_VM_DEBUG_ASSERT(dst->stack_offset().has_value());
+            if (dst->stack_offset().value().moved_from == PrevLoc::GprReg) {
```
andreaslyn:
```suggestion
            if (dst->stack_offset()->moved_from == PrevLoc::GprReg) {
```

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -5797,6 +5823,14 @@ namespace monad::vm::compiler::native
         }
         // Case 8: (stack, stack)
         if (dst->stack_offset() && src->stack_offset()) {
+            if (dst->stack_offset().value().moved_from == PrevLoc::GprReg) {
```
andreaslyn: Also prefer `GeneralReg` when `src` was previously in GPR
```suggestion
            if (dst->stack_offset()->moved_from == PrevLoc::GprReg ||
                 src->stack_offset()->moved_from == PrevLoc::GprReg) {
```

andreaslyn (commented):
There are some follow up optimizations, but this is a good starting point.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2141,7 +2149,9 @@ namespace monad::vm::compiler::native
     Emitter::mov_literal_to_stack_offset(StackElemRef elem, int32_t preferred)
     {
         MONAD_VM_DEBUG_ASSERT(elem->literal().has_value());
-        stack_.insert_stack_offset(elem, preferred);
+        // mov_literal_to_mem stores literals by first loading them into an avx
+        // register, hence we set it as the previous location
+        stack_.insert_stack_offset(elem, preferred, PrevLoc::AvxReg);
         mov_literal_to_mem<true, true>(
             elem, stack_offset_to_mem(*elem->stack_offset()));
```
andreaslyn, line 2156: I think we can actually assert this invariant
```suggestion
        stack_.insert_stack_offset(elem, preferred, PrevLoc::AvxReg);
        mov_literal_to_mem<true, true>(
            elem, stack_offset_to_mem(*elem->stack_offset()));
        MONAD_VM_DEBUG_ASSERT(elem->avx_reg().has_value());
```

andreaslyn (dismissed):

andreaslyn (approved):

dhil (approved):

PR #1962, opened 2025-12-05
Modexp gas changes and upper bound for Osaka (EIP-7823/EIP-7883)

laurenthuberdeau:
## Context

The library we use for EVM precompiles, Silkpre, is no longer maintained and lacks support for new precompiles and recent EIPs. While we work on our own implementation of the precompiles, this PR reimplements the `expmod_gas_cost` to support the Osaka upgrade ([EIP-7823](https://eips.ethereum.org/EIPS/eip-7823) and [EIP-7883](https://eips.ethereum.org/EIPS/eip-7883)).


category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+
+    uint64_t base_len64{static_cast<uint64_t>(base_len256.as_words()[0])};
+    uint64_t exp_len64{static_cast<uint64_t>(exp_len256.as_words()[0])};
+
+    input.erase(0, 3 * 32);
+
+    runtime::uint256_t exp_head{0}; // first 32 bytes of the exponent
+    if (input.length() > base_len64) {
+        input.erase(0, base_len64);
+        right_pad(input, 3 * 32);
```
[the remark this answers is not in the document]
laurenthuberdeau: Keeping the existing silkpre behavior for now.

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    }
+    else {
+        return 500; // From Osaka onwards, minimum gas is 500.
+    }
+}
+
+template <Traits traits>
+uint64_t expmod_gas_cost(byte_string_view const input_view)
+{
+    constexpr uint64_t min_gas{expmod_min_gas<traits>()};
```
dhil: I suppose the cost is fixed per revision, so:
```suggestion
    static constexpr uint64_t min_gas{expmod_min_gas<traits>()};
```

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+
+    if (base_len256 == 0 && mod_len256 == 0) {
+        return min_gas;
+    }
+
+    if constexpr (traits::evm_rev() >= EVMC_OSAKA) {
+        // EIP-7823: each of the length inputs (base, exponent and modulus) MUST
+        // be less than or equal to 8192 bits (1024 bytes).
+        if (base_len256 > 1024 || exp_len256 > 1024 || mod_len256 > 1024) {
+            return UINT64_MAX;
```
dhil:
```suggestion
            return std::numeric_limits<uint64_t>::max();
```

category/execution/ethereum/precompiles_test.cpp
```diff
@@ -407,10 +407,14 @@ TYPED_TEST(TraitsTest, modular_exponentiation)
...
     }
-    else {
+    else if constexpr (TestFixture::Trait::evm_rev() < EVMC_OSAKA) {
         // EIP-2565 repricing since Berlin
         do_geth_tests<typename TestFixture::Trait>(
             "Modular Exponentiation", "modexp_eip2565.json", 0x05_address);
+    } else {
+        // EIP-7883 repricing since Osaka
+        do_geth_tests<typename TestFixture::Trait>(
+            "Modular Exponentiation", "modexp_eip7883.json", 0x05_address);
```
laurenthuberdeau, line 418: Note that this test case does not cover [EIP-7823](https://eips.ethereum.org/EIPS/eip-7823).

third_party/go-ethereum/core/vm/testdata/precompiles/modexp_eip7883.json
```diff
@@ -0,0 +1,317 @@
+[
```
laurenthuberdeau, line 1: Not sure how we usually pull the geth test data in the repo. I copied it from https://github.com/ethereum/go-ethereum/blob/master/core/vm/testdata/precompiles/modexp_eip7883.json.

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    }
+}
+
+template <Traits traits>
+uint64_t expmod_gas_cost(byte_string_view const input_view)
+{
+    constexpr uint64_t min_gas{expmod_min_gas<traits>()};
+
+    std::basic_string<uint8_t> input(input_view.data(), input_view.size());
+    right_pad(input, 3 * 32);
```
Baltoli: Rather than copying the entire input then only looking at the first 96 bytes, let's instead allocate a zero-initialised buffer of 96 bytes and copy from the input into that. As written you could pass a huge input buffer and force it to be copied just to calculate the gas here.

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    runtime::uint256_t mod_len256 =
+        runtime::uint256_t::load_be_unsafe(&input.data()[64]);
+
+    if (base_len256 == 0 && mod_len256 == 0) {
+        return min_gas;
+    }
+
+    if constexpr (traits::evm_rev() >= EVMC_OSAKA) {
+        // EIP-7823: each of the length inputs (base, exponent and modulus) MUST
+        // be less than or equal to 8192 bits (1024 bytes).
```
Baltoli, line 293: Rather than writing new code that does `>= EVMC_OSAKA`, let's add new trait methods like `is_eip_7823_active` (see `traits.hpp`).

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    else if (x <= 1024) {
+        return (x_squared >> 2) + 96 * x - 3072;
+    }
+    else {
+        return (x_squared >> 4) + 480 * x - 199680;
+    }
+}
+
+static runtime::uint256_t
+mult_complexity_eip2565(runtime::uint256_t const &max_length) noexcept
```
Baltoli: Let's make this a traited function rather than stamping out copies via naming

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
-uint64_t expmod_gas_cost(byte_string_view const input)
+constexpr uint64_t expmod_min_gas()
 {
-    return silkpre_expmod_gas(
-        input.data(), input.size(), static_cast<int>(traits::evm_rev()));
+    if (traits::evm_rev() < EVMC_BERLIN) {
+        return 0; // Prior to Berlin, minimum gas is 0.
+    }
+    else if (traits::evm_rev() < EVMC_OSAKA) {
+        return 200; // From Berlin to pre-Osaka (Prague), minimum gas is 200.
```
Baltoli: These comments could mention EIPs rather than just the revision, to save a round trip to the meta EIP

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    }
+    else if (
+        runtime::count_significant_bytes(base_len256) > 8 ||
+        runtime::count_significant_bytes(exp_len256) > 8 ||
+        runtime::count_significant_bytes(mod_len256) > 8) {
+        return UINT64_MAX;
+    }
+
+    uint64_t base_len64{static_cast<uint64_t>(base_len256.as_words()[0])};
+    uint64_t exp_len64{static_cast<uint64_t>(exp_len256.as_words()[0])};
```
Baltoli: Nitpick - there's a lot of noise here, but you should just be able to `static_cast` a `uint256_t` directly to `uint64_t` because `operator uint64_t` is defined to return the LSW directly.

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    runtime::uint256_t gas;
+    if constexpr (traits::evm_rev() < EVMC_BERLIN) {
+        gas = mult_complexity_eip198(max_length) * iteration_count / 20;
+    }
+    else if constexpr (traits::evm_rev() < EVMC_OSAKA) {
+        gas = mult_complexity_eip2565(max_length) * iteration_count / 3;
+    }
+    else {
+        gas = mult_complexity_eip7883(max_length) * iteration_count;
+    }
```
Baltoli: Again this could be factored out into a separate traited function

third_party/go-ethereum/core/vm/testdata/precompiles/modexp_eip7883.json
Baltoli: That's the correct source

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +184,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+        return 500; // From Osaka onwards, minimum gas is 500.
+    }
+}
+
+template <Traits traits>
+uint64_t expmod_gas_cost(byte_string_view const input_view)
+{
+    constexpr uint64_t min_gas{expmod_min_gas<traits>()};
+
+    std::basic_string<uint8_t> input(input_view.data(), input_view.size());
```
Baltoli: Stylistic nit: isn't this type just `byte_string` more generally?

category/execution/ethereum/precompiles_impl.cpp
laurenthuberdeau: I tried to keep the original silkpre implementation as much as possible but `right_pad` and `erase` also irks me. I think we can remove all copies with `if (input_view.size() < 96) return min_gas;` and some pointer arithmetic.

category/core/runtime/uint256.hpp
```diff
@@ -652,6 +668,16 @@ namespace monad::vm::runtime
             return result;
         }
 
+        [[gnu::always_inline]] static inline constexpr uint256_t
+        load_le_unsafe(uint8_t const *bytes, size_t len) noexcept
```
laurenthuberdeau: Included for symmetry

category/core/runtime/uint256.hpp
```diff
@@ -643,6 +643,22 @@ namespace monad::vm::runtime
             return load_le_unsafe(bytes).to_be();
         }
 
+        [[gnu::always_inline]]
```
Baltoli: We can remove these now given that you're working with intx?

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +180,153 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
 
 EXPLICIT_EVM_TRAITS(blake2bf_gas_cost);
 
+template <Traits traits>
+static uint256_t mult_complexity(uint256_t const &max_len) noexcept
+{
+    if constexpr (traits::evm_rev() < EVMC_BERLIN) {
```
Baltoli: These should be the feature flag traits

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +180,153 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+        return 200;
+    }
+    else {
+        return 0; // Pre EIP-2565 (EIP-198)
+    }
+}
+
+// TODO(LH): Replace calls to this function with uint256_t::load_be_unsafe when
+// migrating off intx.
+static uint256_t uint256_load_be_unsafe(uint8_t const *bytes, size_t len)
```
Baltoli: Perhaps misnamed: the "unsafe" functions in intx just read bytes without checking if you're off the end, whereas this pads with zeros to maintain safety?

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +180,153 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    std::memcpy(intx::as_bytes(result) + (32 - len), bytes, len);
+    return intx::to_big_endian(result);
+}
+
 template <Traits traits>
 uint64_t expmod_gas_cost(byte_string_view const input)
 {
-    return silkpre_expmod_gas(
-        input.data(), input.size(), static_cast<int>(traits::evm_rev()));
+    constexpr uint64_t min_gas{expmod_min_gas<traits>()};
```
Baltoli: Nitpick: usual style is `static constexpr`; it's not a big deal but you get a stronger guarantee in places where it would _matter_ whether you're doing guaranteed CTE or not

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +180,153 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+
+    uint256_t base_len256{uint256_load_be_unsafe(
```
Baltoli: clearer to write `auto base_len256 = ...` I think? The return type is `uint256` and it avoids brace inits

category/execution/ethereum/precompiles_impl.cpp
laurenthuberdeau: Changes to `uint256_partial_load_be`.

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +182,154 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
 
 EXPLICIT_EVM_TRAITS(blake2bf_gas_cost);
 
+template <Traits traits>
+static uint256_t mult_complexity(uint256_t const &max_len) noexcept
+{
+    if constexpr (traits::eip_7883_active() || traits::eip_2565_active()) {
```
Baltoli: 7883 implies 2565 right?

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +182,154 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+        uint256_t const words{(max_len + 7) >> 3}; // ceil(max_len/8)
+        if constexpr (traits::evm_rev() < EVMC_OSAKA) {
```
Baltoli: Still want to avoid this explicit mention of a revision ideally

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -182,11 +182,154 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    }
+    else if (
+        count_significant_bytes(base_len256) > 8 ||
+        count_significant_bytes(exp_len256) > 8 ||
+        count_significant_bytes(mod_len256) > 8) {
+        return UINT64_MAX;
+    }
+
+    uint64_t base_len64{static_cast<uint64_t>(base_len256)};
+    uint64_t exp_len64{static_cast<uint64_t>(exp_len256)};
```
Baltoli: No need to name the type twice:
```suggestion
    auto const base_len64 = static_cast<uint64_t>(base_len256);
    auto const exp_len64 = static_cast<uint64_t>(exp_len256);
```

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -183,10 +183,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+
+    uint256_t const iteration_count{
+        expmod_iteration_count<traits>(exp_len256, bit_len)};
+
+    uint256_t const max_length{std::max(mod_len256, base_len256)};
+    uint256_t gas = mult_complexity<traits>(max_length) * iteration_count /
+                    expmod_gas_denominator<traits>();
+
+    if (intx::count_significant_words(gas) > 1) {
+        return UINT64_MAX;
```
dhil: Nit:
```suggestion
        return std::numeric_limits<uint64_t>::max();
```
I think we do this elsewhere in the codebase.

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -183,10 +183,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+        // be less than or equal to 8192 bits (1024 bytes).
+        if (base_len256 > 1024 || exp_len256 > 1024 || mod_len256 > 1024) {
+            return std::nullopt; // invalid input
+        }
+    }
+    else if (
+        count_significant_bytes(base_len256) > 8 ||
+        count_significant_bytes(exp_len256) > 8 ||
+        count_significant_bytes(mod_len256) > 8) {
+        return UINT64_MAX;
```
dhil: Nit:
```suggestion
        return std::numeric_limits<uint64_t>::max();
```

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -183,10 +183,152 @@ uint64_t blake2bf_gas_cost(byte_string_view const input)
...
+    auto exp_len256 =
+        input.length() >= 32
+            ? uint256_partial_load_be(
+                  &input.data()[32], std::min(32ul, input.length() - 32))
+            : uint256_t{0};
+    auto mod_len256 =
+        input.length() >= 64
+            ? uint256_partial_load_be(
+                  &input.data()[64], std::min(32ul, input.length() - 64))
+            : uint256_t{0};
```
dhil: `base_len256`, `exp_len256`, and `mod_len256` should be declared `const`, right? I cannot see that they are modified anywhere else.

PR #1966, opened 2025-12-09
Apply relaxed merge to reserve balance and validate transaction. Enable relaxed merge.

magedm:
This PR:
- Adds to State variants of check min balance functions (to support relaxing reserve balance and validate transaction)
- Relaxes reserve balance
- Relaxed validate transaction
- Enables relaxed merge validation

[Updated 2025-12-23]
- Relaxes staking [2026-03-24]

### DIFF
--- a/category/execution/ethereum/state3/state.cpp
+++ b/category/execution/ethereum/state3/state.cpp
@@ -358,7 +358,7 @@ void State::subtract_from_balance(
         account = Account{.incarnation = incarnation_};
     }
 
-    MONAD_ASSERT_THROW(delta <= account.value().balance, "balance underflow");
+    MONAD_ASSERT_THROW(check_min_balance(address, delta), "balance underflow");
 
     account.value().balance -= delta;
     account_state.touch();
@@ -683,12 +683,17 @@ bool State::try_fix_account_mismatch(
             return false;
         }
         if (actual->balance > original->balance) {
-            recent->balance += actual->balance - original->balance;
+            uint256_t const delta = actual->balance - original->balance;
+            if (std::numeric_limits<uint256_t>::max() - recent->balance <
+                delta) {
+                return false;
+            }
+            recent->balance += delta;
         }
         else {
-            MONAD_ASSERT(
-                recent->balance >= (original->balance - actual->balance));
-            recent->balance -= original->balance - actual->balance;
+            uint256_t const delta = original->balance - actual->balance;
+            MONAD_ASSERT(recent->balance >= delta);
+            recent->balance -= delta;
         }
     }
     original->balance = actual->balance;
@@ -700,13 +705,26 @@ bool State::try_fix_account_mismatch(
     return true;
 }
 
-bool State::record_balance_constraint_for_debit(
-    Address const &address, uint256_t const &debit)
+bool State::check_min_original_balance(
+    Address const &address, uint256_t const &value)
+{
+    auto &orig_state = original_account_state(address);
+    auto const &account = orig_state.account_;
+    return check_account_min_balance(orig_state, account, value);
+}
+
+bool State::check_min_balance(Address const &address, uint256_t const &value)
 {
     auto const &account = recent_account(address);
-    uint256_t const balance = account.has_value() ? account->balance : 0;
+    OriginalAccountState &orig_state = original_account_state(address);
+    return check_account_min_balance(orig_state, account, value);
+}
 
-    auto &original_state = original_account_state(address);
+bool State::check_account_min_balance(
+    OriginalAccountState &original_state, std::optional<Account> const &account,
+    uint256_t const &debit)
+{
+    uint256_t const balance = account ? account->balance : 0;
     // RELAXED MERGE
     // if current balance  >= `debit`, then:
     // 1. compute the amount that current balance exceeds `debit`
--- a/category/execution/ethereum/state3/state.cpp
+++ b/category/execution/ethereum/state3/state.cpp
@@ -369,7 +369,7 @@ void State::subtract_from_balance(
         account = Account{.incarnation = incarnation_};
     }
 
-    MONAD_ASSERT_THROW(delta <= account.value().balance, "balance underflow");
+    MONAD_ASSERT_THROW(check_min_balance(address, delta), "balance underflow");
 
     account.value().balance -= delta;
     account_state.touch();
@@ -694,12 +694,17 @@ bool State::try_fix_account_mismatch(
             return false;
         }
         if (actual->balance > original->balance) {
-            recent->balance += actual->balance - original->balance;
+            uint256_t const delta = actual->balance - original->balance;
+            if (std::numeric_limits<uint256_t>::max() - recent->balance <
+                delta) {
+                return false;
+            }
+            recent->balance += delta;
         }
         else {
-            MONAD_ASSERT(
-                recent->balance >= (original->balance - actual->balance));
-            recent->balance -= original->balance - actual->balance;
+            uint256_t const delta = original->balance - actual->balance;
+            MONAD_ASSERT(recent->balance >= delta);
+            recent->balance -= delta;
         }
     }
     original->balance = actual->balance;
@@ -711,13 +716,26 @@ bool State::try_fix_account_mismatch(
     return true;
 }
 
-bool State::record_balance_constraint_for_debit(
-    Address const &address, uint256_t const &debit)
+bool State::check_min_original_balance(
+    Address const &address, uint256_t const &value)
+{
+    auto &orig_state = original_account_state(address);
+    auto const &account = orig_state.account_;
+    return check_account_min_balance(orig_state, account, value);
+}
+
+bool State::check_min_balance(Address const &address, uint256_t const &value)
 {
     auto const &account = recent_account(address);
-    uint256_t const balance = account.has_value() ? account->balance : 0;
+    OriginalAccountState &orig_state = original_account_state(address);
+    return check_account_min_balance(orig_state, account, value);
+}
 
-    auto &original_state = original_account_state(address);
+bool State::check_account_min_balance(
+    OriginalAccountState &original_state, std::optional<Account> const &account,
+    uint256_t const &debit)
+{
+    uint256_t const balance = account ? account->balance : 0;
     // RELAXED MERGE
     // if current balance  >= `debit`, then:
     // 1. compute the amount that current balance exceeds `debit`
--- a/category/execution/ethereum/state3/state.hpp
+++ b/category/execution/ethereum/state3/state.hpp
@@ -233,17 +233,16 @@ public:
     bool try_fix_account_mismatch(
         Address const &, std::optional<Account> const &actual);
 
-    /**
-     * Checks whether the account currently has enough balance to cover `debit`
-     * and records the relaxed-merge constraints needed for that debit.
-     *
-     * NOTE: This method mutates the account's OriginalAccountState by either
-     * tightening the recorded `min_balance` or demanding exact balance
-     * validation when the balance is insufficient. Callers should treat it as
-     * a stateful helper rather than a pure predicate.
-     */
-    bool record_balance_constraint_for_debit(
-        Address const &, uint256_t const &debit);
+    // These functions check whether the account associated with the input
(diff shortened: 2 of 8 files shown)


category/execution/monad/validate_monad_transaction.cpp
```diff
@@ -36,18 +36,21 @@ Result<void> validate_monad_transaction(
...
             res.error() != TransactionError::InsufficientBalance) {
             return res;
         }
 
-        uint256_t const balance = acct.has_value() ? acct.value().balance : 0;
         uint256_t const gas_fee =
             uint256_t{tx.gas_limit} * gas_price(rev, tx, base_fee_per_gas);
-        if (MONAD_UNLIKELY(balance < gas_fee)) {
+        ConstBalanceAccessor const bal_accessor(
+            acct, state.original_account_state(sender));
```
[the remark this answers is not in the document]
magedm: Thanks. Fixed.

category/execution/ethereum/validate_transaction.cpp
```diff
@@ -266,9 +263,10 @@ EXPLICIT_TRAITS(validate_transaction);
 Result<void> validate_transaction(
     evmc_revision const rev, Transaction const &tx,
     std::optional<Account> const &sender_account,
```
aa755: It seems the intent of `validate_transaction` is to run before any changes are made to any account.  Is there any scenario where `sender_account` and `origi_state.account_` would be different?

magedm: I wasn't sure if validate_transation always operates on the original account value.
In category/rpc/monad_executor.cpp in `eth_call_impl` the `eoa` account is passed to `validate_transaction` after being modified from the original account value.

aa755 (dismissed):
I will try to prove that the `dipped_into_reserve` function in reserve_balance.cpp matches what we have in the model in Coq.

category/execution/ethereum/validate_transaction.cpp
magedm: May be true by convention but here is a counterexample in existing code of a modified account passed to validate_transaction:
https://github.com/category-labs/monad/blob/main/category/rpc/monad_executor.cpp#L341-L348
```
        std::optional<Account> eoa = acct;
        if (eoa.has_value()) {
            eoa.value().code_hash = NULL_HASH;
        }

        // Safe to pass empty code to validation here because the above override
        // will always mark this transaction as coming from an EOA.
        BOOST_OUTCOME_TRY(validate_transaction<traits>(enriched_txn, eoa, {}));
```

magedm:
> I will try to prove that the `dipped_into_reserve` function in reserve_balance.cpp matches what we have in the model in Coq.

Thank you @aa755 ! That would be great.

aa755 (approved):

category/execution/monad/reserve_balance.cpp
aa755: This `ASSERT` was there before. The proof that it will never fire in live mainnet (but can fire in the RPC path) is  much more tricky than I thought initially before working on the Coq proof of this C++ code. Also, the reason may not hold if we further optimize the reserve balance design in ways that I have always wanted to. 

The issue is that even though consensus has ensured this condition, it estimates that by considering transactions sequentially. But at this point in code, we may be doing speculative execution so the balances may be wrong/different from what will happen in the actual sequential execution, so the consensus guarantee of reserve balance being sufficient to pay at least the fee do not apply directly. Fortunately, there is no problem currently because the reserve-balance design is not very aggressive and disregards credits, even to non-delegated accounts. Because of that, consensus' balance lower bound estimates of accounts can only decrease with increasing transaction index. So, during speculative execution, when we read the account balance of the sender:
-  either the previous transaction, say `txprev`, has already finished in which case, we will read the correct balance, 
- or we read the balance from the commit (`BlockState.merge`) of a much earlier transaction, say `txfoo`. but the balance lowerbound after `txfoo` can only be higher than that after `txprev`, so even that speculated balance will be sufficient to cover the fees. Note that this reasoning fails if we make the reserve balance design smarter and also account for credits for non-delegated accounts: I think we should do this in future to make our reserve balance checks more permissive to users while insuring safety.

So, I think we should not rely on the assumption that reserve balance checks disregards credits to non-delegated accounts. I will try to sketch a way to not rely on this assumption, but that can be done in a future PR.

Baltoli:
> I will try to prove that the dipped_into_reserve function in reserve_balance.cpp matches what we have in the model in Coq.

Did this proof go through @aa755?

category/execution/monad/reserve_balance.cpp
```diff
@@ -86,9 +86,11 @@ bool dipped_into_reserve(
         // Check if dipped into reserve
         std::optional<uint256_t> const violation_threshold =
             [&] -> std::optional<uint256_t> {
-            uint256_t const orig_balance = state.get_original_balance(addr);
+            uint256_t const max_reserve = get_max_reserve<traits>(addr);
             uint256_t const reserve =
-                std::min(get_max_reserve<traits>(addr), orig_balance);
+                state.check_min_original_balance(addr, max_reserve)
+                    ? max_reserve
+                    : state.get_original_balance(addr);
```
0xSqualo, line 110: In this case `check_min_original_balance()` and [`get_original_balance()`](https://github.com/category-labs/monad/blob/cbbd825b4a5202143bbd06c471ade93015cb55dc/category/execution/ethereum/state3/account_state.hpp#L196C9-L196C35) wil overwrite `validate_exact_balance_` twice. No impact but redundant in this path.

aa755 (dismissed):

category/execution/monad/reserve_balance.cpp
magedm: Thanks for reviewing the code. This is intentional. We want a failed check min balance to set exact balance validation even if the failure path doesn't lead to a get balance call and we want get balance calls to set exect balance validation on their own. The setting of execy balance validation is intended to be idempotent and sticky (doesn't get cleared).

category/execution/ethereum/state3/state.cpp
```diff
@@ -333,7 +333,7 @@ void State::subtract_from_balance(
         account = Account{.incarnation = incarnation_};
     }
 
-    MONAD_ASSERT_THROW(delta <= account.value().balance, "balance underflow");
+    MONAD_ASSERT_THROW(check_min_balance(address, delta), "balance underflow");
```
0xSqualo, line 370: Isn't this now ordering dependent on the call above to `current_account_state(address)` above? i.e. to ensure a value exists in `current_` (if it previously didn't)?

Current implementation shouldn't have any issues, just raising for awareness / informationally for the case of future refactors. @magedm

category/execution/ethereum/state3/state.hpp
```diff
@@ -209,17 +209,17 @@ class State
...
-     * a stateful helper rather than a pure predicate.
-     */
-    bool record_balance_constraint_for_debit(
-        Address const &, uint256_t const &debit);
+    // These functions check whether the account associated with the input
+    // address has enough balance to cover the input amount and record the
+    // corresponding relaxed-merge validation requirements.
+    bool check_min_original_balance(Address const &, uint256_t const &);
+    bool check_min_balance(Address const &, uint256_t const &);
+    bool check_min_balance(Address const &, uint512_t const &);
```
0xSqualo: Is it worth defining the expected behavior for callers using `uint512_t` inputs here? (when inputs are larger than `uint256_t::max`).

magedm: The expected behavior is the same for all three functions. Since balance is only 256 bits, then checks of min balance with 512-bit values > uint256_t::max fail as expected.

category/execution/monad/reserve_balance.cpp
aa755: ~Nevermind, the condition checked by transaction validation is sufficient to imply this assert~, even when `original_bal` is speculated (thus possibly different from what consensus used in its checks for the tx):
```
Lemma foo (original_bal value gas_fee: N) :
  value + gas_fee <= original_bal (* condition checked by transaction validation in execution. original_bal may be a speculated value *)
  -> gas_fee <= 10 (* condition checked by consensus. gas_fee is fixed for a tx and not affected by speculation *)
  -> gas_fee <= original_bal `min` 10 (* this assert *).
Proof. lia. Qed.
```

aa755: While doing the Coq proof, I realized that consensus does actually not enforce `max_gas_fee <= 10` for emptying transactions. I opened a [PR](https://github.com/category-labs/monad/pull/2096) to remove the assert.

category/execution/monad/reserve_balance.cpp
```diff
@@ -212,7 +213,9 @@ uint256_t ReserveBalance::pretx_reserve(Address const &address)
 {
     MONAD_ASSERT(get_max_reserve_);
     uint256_t const max_reserve = get_max_reserve_(address);
-    return std::min(max_reserve, state_->get_original_balance(address));
+    return state_->check_min_original_balance(address, max_reserve)
+               ? max_reserve
+               : state_->get_original_balance(address);
```
aa755, line 216: perhaps a better name for `check_min_original_balance` could be `balance_is_at_least` or `balance_ge`. feel free to igore.

aa755 (dismissed):

category/execution/ethereum/state3/state.cpp
```diff
@@ -696,13 +696,33 @@ bool State::try_fix_account_mismatch(
...
-    uint256_t const balance = account.has_value() ? account->balance : 0;
+    OriginalAccountState &orig_state = original_account_state(address);
+    return check_account_min_balance(orig_state, account, value);
+}
 
-    auto &original_state = original_account_state(address);
+bool State::check_min_balance(Address const &address, uint512_t const &value)
+{
+    return value > std::numeric_limits<uint256_t>::max()
+               ? false
```
0xSqualo: @magedm Should you also call `set_validate_exact_balance()` before returning `false` here? Right now the caller in `execute_impl2` handles it, but if this function is ever called from a path that doesn't do that cleanup, the merge could proceed without an exact-balance constraint recorded. Perhaps:

```cpp
if (value > std::numeric_limits<uint256_t>::max()) {
    original_account_state(address).set_validate_exact_balance();
    return false;
}
```

magedm: @0xSqualo The logic of returning false without requiring exact validation when value > uint256_t max is that regardless of balance this function would return false. We didn't rely on the balance value in any way.

category/execution/ethereum/state3/state.cpp
magedm: Thanks but I think even now reordering is always benign. recent_account() always returns the latest value, either from current_ if the account has been touched or from original_ if it hasn't.

magedm:
@aa755 @0xSqualo I added this change to category/execution/monad/staking/staking_contract.cpp. Please review:
```
-    uint256_t const contract_balance = state_.get_balance(STAKING_CA);
     MONAD_ASSERT_THROW(
-        contract_balance >= withdrawal_amount, "withdrawal insolvent");
+        state_.check_min_balance(STAKING_CA, withdrawal_amount),
+        "withdrawal insolvent");
```

aa755 (dismissed):

category/execution/ethereum/state3/state.cpp
marcograss: from an auditing point of view and invariant point of view I find the explicit check that we had before more self-explanatory and evident

magedm:
Made changes to adapt to the change in type of 'v0' in 'validate_ethereum_transaction' and the removal of 'uin512_t'.

category/execution/ethereum/state3/state.cpp
magedm: My reason for using 'check_min_balance(address, delta)' instead of 'delta <= account.value().balance' is that the former adds a min balance predicate for checking at merge validation whereas the latter does not.  I wanted to follow the rule that whenever computation depends on balance, the code has to update/add predicates for checking at merge validation. Does that rationale change your view of this code change?

marcograss: sure this is a non issue anyway, as soon as the code is clear we can leave it like this. 

Thank you

resolving the conversation

PR #1974, opened 2025-12-10
hardened State::recent_account, relaxed validate_transaction and reserve_balance checks

aa755:
This is an alternative to #1966 with a much smaller diff and perhaps a simpler API as users of State class wont have to create balance accessors by and and worry about setting its 2 constructor arguments correctly.

The first commit can be viewed as a continuation of https://github.com/category-labs/monad/pull/1782 where I forgot to notice that `recent_account` method can also be used to get access to balance in a potentially unsafe way (unless the user is careful, they may break min_balance invaraints). So in this commit, following the strategy in https://github.com/category-labs/monad/pull/1782,  I rename it to `recent_account_pessmistic` and add `set_exact_validation` there. The next commits optimizes by avoiding the usages of this method except in call sites in test cases or RPC path.

The second and third commits optimize the code by enabling more relaxed validation (avoid using pessimistic methods): much of the code here is copied/adapted from #1966

### DIFF
--- a/category/execution/ethereum/state3/account_state.hpp
+++ b/category/execution/ethereum/state3/account_state.hpp
@@ -205,6 +205,9 @@ private:
 
     void set_min_balance(uint256_t const &value)
     {
+        if (value == 0) {
+            return;
+        }
         MONAD_ASSERT(account_.has_value());
         MONAD_ASSERT(account_->balance >= value);
         if (value > min_balance_) {


category/execution/monad/execute_system_transaction.cpp
```diff
@@ -173,7 +173,7 @@ evmc_message ExecuteSystemTransaction<traits>::to_message() const
 template <Traits traits>
 Result<void> ExecuteSystemTransaction<traits>::execute(State &state)
 {
-    auto const sender_account = state.recent_account(sender_);
+    auto const sender_account = state.recent_account_pessimistic(sender_);
```
aa755: optimize this?

PR #1979, opened 2025-12-12
Refactor MonadChainContext & remove the revert_transaction closure

goodlyrottenapple:
Removes `revert_transaction` closure passing through execution in favor of passing down the context data directly.

### DIFF
--- a/category/execution/ethereum/evm.cpp
+++ b/category/execution/ethereum/evm.cpp
@@ -25,6 +25,7 @@
 #include <category/execution/ethereum/evm.hpp>
 #include <category/execution/ethereum/evmc_host.hpp>
 #include <category/execution/ethereum/precompiles.hpp>
+#include <category/execution/ethereum/reserve_balance.hpp>
 #include <category/execution/ethereum/state3/state.hpp>
 #include <category/vm/evm/explicit_traits.hpp>
 #include <category/vm/evm/monad/revision.h>
@@ -163,9 +164,8 @@ void post_call(State &state, evmc::Result const &result)
 }
 
 template <Traits traits>
-evmc::Result create(
-    EvmcHost<traits> *const host, State &state, evmc_message const &msg,
-    std::function<bool()> const &revert_transaction)
+evmc::Result
+create(EvmcHost<traits> *const host, State &state, evmc_message const &msg)
 {
     MONAD_ASSERT(msg.kind == EVMC_CREATE || msg.kind == EVMC_CREATE2);
 
@@ -261,8 +261,16 @@ evmc::Result create(
             state, contract_address, std::move(result));
     }
 
-    if (msg.depth == 0 && revert_transaction()) {
-        result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
+    if (msg.depth == 0) {
+        if (revert_transaction<traits>(
+                msg.sender,
+                host->tx_,
+                host->base_fee_per_gas_.value_or(0),
+                host->i_,
+                state,
+                host->chain_ctx_)) {
+            result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
+        }
     }
 
     if (result.status_code == EVMC_SUCCESS) {
@@ -289,9 +297,8 @@ evmc::Result create(
 EXPLICIT_TRAITS(create);
 
 template <Traits traits>
-evmc::Result call(
-    EvmcHost<traits> *const host, State &state, evmc_message const &msg,
-    std::function<bool()> const &revert_transaction)
+evmc::Result
+call(EvmcHost<traits> *const host, State &state, evmc_message const &msg)
 {
     MONAD_ASSERT(
         msg.kind == EVMC_DELEGATECALL || msg.kind == EVMC_CALLCODE ||
@@ -317,9 +324,17 @@ evmc::Result call(
         result = state.vm().execute<traits>(*host, &msg, hash, code);
     }
 
-    if (msg.depth == 0 && revert_transaction()) {
-        result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
-        result.gas_refund = 0;
+    if (msg.depth == 0) {
+        if (revert_transaction<traits>(
+                msg.sender,
+                host->tx_,
+                host->base_fee_per_gas_.value_or(0),
+                host->i_,
+                state,
+                host->chain_ctx_)) {
+            result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
+            result.gas_refund = 0;
+        }
     }
 
     post_call(state, result);
--- a/category/execution/ethereum/evmc_host.hpp
+++ b/category/execution/ethereum/evmc_host.hpp
@@ -21,6 +21,7 @@
 #include <category/execution/ethereum/core/address.hpp>
 #include <category/execution/ethereum/evm.hpp>
 #include <category/execution/ethereum/precompiles.hpp>
+#include <category/execution/ethereum/reserve_balance.hpp>
 #include <category/execution/ethereum/state3/state.hpp>
 #include <category/execution/ethereum/trace/call_tracer.hpp>
 #include <category/execution/ethereum/transaction_gas.hpp>
@@ -52,14 +53,11 @@ protected:
     evmc_tx_context const &tx_context_;
     State &state_;
     CallTracerBase &call_tracer_;
-    std::function<bool()> revert_transaction_;
 
 public:
     EvmcHostBase(
         CallTracerBase &, evmc_tx_context const &, BlockHashBuffer const &,
-        State &, std::function<bool()> const &revert_transaction = [] {
-            return false;
-        }) noexcept;
+        State &) noexcept;
 
     virtual ~EvmcHostBase() noexcept = default;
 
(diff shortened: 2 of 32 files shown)


category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -54,6 +55,34 @@ namespace
...
+
+    static ankerl::unordered_dense::segmented_set<Address> const
+        empty_grandparent_senders_and_authorities;
+    static ankerl::unordered_dense::segmented_set<Address> const
+        empty_parent_senders_and_authorities;
+    static ankerl::unordered_dense::segmented_set<Address> const
+        empty_senders_and_authorities{};
+    static std::vector<Address> const empty_senders{Address{0}};
+    static std::vector<std::vector<std::optional<Address>>> const
+        empty_authorities{{}};
```
Baltoli, line 63: I presume having `senders_and_authorities ≠ senders ∪ authorities` doesn't break anything when you use these empty contexts?

category/execution/ethereum/test/test_monad_chain.cpp
```diff
@@ -341,42 +342,54 @@ TYPED_TEST(MonadTraitsTest, revert_transaction_dip_false)
...
+                empty_grandparent_senders_and_authorities,
+            .parent_senders_and_authorities =
+                empty_parent_senders_and_authorities,
             .senders_and_authorities = senders_and_authorities,
             .senders = senders,
             .authorities = authorities,
         };
-        EXPECT_FALSE(
-            can_sender_dip_into_reserve(Address{1}, 1, false, context));
+        EXPECT_FALSE(can_sender_dip_into_reserve<typename TestFixture::Trait>(
```
Baltoli: I wonder if you can get type deduction on `context` to work here to avoid duplication? Or happy to leave it if you think it's better to be explicit

category/execution/ethereum/evm.cpp
```diff
@@ -260,8 +259,16 @@ evmc::Result create(
             state, contract_address, std::move(result));
     }
 
-    if (msg.depth == 0 && revert_transaction()) {
-        result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
+    if constexpr (is_monad_trait_v<traits>) {
```
Baltoli: We generally want to avoid this kind of Ethereum / Monad mixing - the idiom we've settled on for this is to have two separate implementations of `revert_transaction`, one in `ethereum/` and one in `monad/`. That way the Ethereum code has no mention of Monad-specific logic.

category/execution/ethereum/evm.cpp
```diff
@@ -316,9 +322,17 @@ evmc::Result call(
         result = state.vm().execute<traits>(*host, &msg, hash, code);
     }
 
-    if (msg.depth == 0 && revert_transaction()) {
-        result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
-        result.gas_refund = 0;
+    if constexpr (is_monad_trait_v<traits>) {
```
Baltoli: Similarly here

category/execution/ethereum/evmc_host.hpp
```diff
@@ -180,9 +196,34 @@ struct EvmcHost final : public EvmcHostBase
     {
         return call_tracer_;
     }
+
+    bool revert_transaction()
+    {
+        try {
+            if constexpr (is_monad_trait_v<traits>) {
+                return ::monad::revert_monad_transaction<traits>(
```
Baltoli: Same point here re. `if constexpr` on `is_monad_trait_v`

category/execution/ethereum/execute_transaction_test.cpp
```diff
@@ -47,6 +47,37 @@ using namespace monad;
 
 using db_t = TrieDb;
 
+namespace
```
Baltoli, line 50: This whole stanza gets repeated a lot - I wonder if there's a convenient way to pull it out into a shared location?

category/rpc/monad_executor.cpp
```diff
@@ -1419,10 +1423,15 @@ struct monad_executor
                                 dynamic_cast<MonadChain *>(chain.get())
                                     ->get_monad_revision(
                                         block_header.timestamp);
+                            // NOLINTBEGIN(clang-analyzer-core.CallAndMessage)
```
Baltoli, line 1433: What's going on here that we need a linter suppression?

mkolosick: I've run into this several times where there appears to be a bug in this particular checker mistakenly labeling an argument as `undefined`. In each case I've traced through the control flow and unless I'm missing something incredibly subtle in different parts of the code base each time there's just a bug in the checker.

category/execution/ethereum/execute_transaction_test.cpp
goodlyrottenapple: Maybe best to address this in a separate PR? Took a punt on this, but afaik the function can't just live in a hpp file as it needs to refer to empty static `segmented_set`s. 

One option would be to add `empty_chain_ctx` to the `monad_execution_ethereum` cmake library, tho no other test code lives in that library, so perhaps not the best option. 

Otherwise, we would have to modify the `monad_add_test2` cmake function to compile and link the `empty_chain_ctx` cpp file with the relevant test files, which is certainly doable but seems janky.

aa755 (dismissed):
This looks good. A minor non-correctness issue is that it seems to add to the existing redundancies in arguments/fields: Many functions have multiple copies of the same data passed separately.
For example, it seems that we dont really need to add `sender_` to `EvmcHost`. It was already possible to obtain that from `msg.sender`. Also, many fields, e.g. `i_`, `context_` in `ExecuteTransactionNoValidation` are no longer needed as they are now available via the `EvmcHost` argument passed to the methods.  Also, the revert_transaction method in EvmcHost seems to be unused. I fixed these redundancies in [this commit](https://github.com/category-labs/monad/commit/18eecd63bdff6c74a573fe0ab5984b2de25536c8) , which builds and passes all tests. (The commit was written with the help of codex-cli)

cmd/monad/runloop_monad.cpp
```diff
@@ -245,30 +246,35 @@ Result<BlockExecOutput> propose_block(
...
+        parent_senders_and_authorities = parent_entry.senders_and_authorities;
         if (block.header.number > 2) {
             bytes32_t const &grandparent_id = parent_entry.parent_id;
             MONAD_ASSERT(block_cache.contains(grandparent_id));
             BlockCacheEntry const &grandparent_entry =
                 block_cache.at(grandparent_id);
-            chain_context.grandparent_senders_and_authorities =
-                &grandparent_entry.senders_and_authorities;
+            grandparent_senders_and_authorities =
+                grandparent_entry.senders_and_authorities;
```
0xSqualo: Should these be copied in this way? Surely this will introduce more overhead than before.

goodlyrottenapple: Thanks, should be fixed now. I also added a similar optimisation to `category/rpc/monad_executor.cpp`

category/execution/ethereum/execute_transaction_test.cpp
Baltoli: Sure, happy to follow this up in a subsequent PR.

> certainly doable but seems janky.

I think this is probably the best thing to do if we're repeating it that much

category/execution/ethereum/evm.cpp
```diff
@@ -261,8 +261,17 @@ evmc::Result create(
             state, contract_address, std::move(result));
     }
 
-    if (msg.depth == 0 && revert_transaction()) {
-        result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
+    if (msg.depth == 0) {
+        Address const sender{msg.sender};
+        if (revert_transaction<traits>(
+                sender,
```
Baltoli: What's the intermediate locally-scoped variable for?
```suggestion
        if (revert_transaction<traits>(
                msg.sender,
```

category/execution/ethereum/evm.cpp
```diff
@@ -317,9 +325,18 @@ evmc::Result call(
         result = state.vm().execute<traits>(*host, &msg, hash, code);
     }
 
-    if (msg.depth == 0 && revert_transaction()) {
-        result.status_code = EVMC_MONAD_RESERVE_BALANCE_VIOLATION;
-        result.gas_refund = 0;
+    if (msg.depth == 0) {
+        Address const sender{msg.sender};
+        if (revert_transaction<traits>(
+                sender,
```
Baltoli: Same here:
```suggestion
        if (revert_transaction<traits>(
                msg.sender,
```

cmd/monad/runloop_monad.cpp
```diff
@@ -245,30 +249,40 @@ Result<BlockExecOutput> propose_block(
...
     }
 
-    MonadChainContext chain_context{
-        .grandparent_senders_and_authorities = nullptr,
-        .parent_senders_and_authorities = nullptr,
+    auto const
+        &[grandparent_senders_and_authorities, parent_senders_and_authorities] =
+            [&]() -> std::tuple<
+                      ankerl::unordered_dense::segmented_set<Address> const &,
+                      ankerl::unordered_dense::segmented_set<Address> const &> {
```
Baltoli: Tiny stylistic preference: this looks a bit neater if you drop the trailing return (so it's just `[&] -> { ... }`), and then have each return statement use CTAD to construct a `std::tuple` (i.e. `return std::tuple{grandparent_entry...`).

category/execution/monad/reserve_balance.cpp
```diff
@@ -144,23 +145,24 @@ bool revert_monad_transaction(
...
 {
     if (sender_is_delegated) { // delegated accounts cannot dip
         return false;
     }
 
     // check pending blocks
     for (ankerl::unordered_dense::segmented_set<Address> const
-             *const senders_and_authorities :
+             &senders_and_authorities :
          {ctx.grandparent_senders_and_authorities,
```
aa755: will this copy the the set into the initializer list array? https://eel.is/c%2B%2Bdraft/dcl.init.list#5.sentence-2
(if this is an issue, it is a pre-existing one, so we can defer it to a new PR)

the following may be safer:

```
     // check pending blocks
-    for (ankerl::unordered_dense::segmented_set<Address> const
-             &senders_and_authorities :
-         {ctx.grandparent_senders_and_authorities,
-          ctx.parent_senders_and_authorities}) {
-        if (senders_and_authorities.contains(sender)) {
-            return false;
-        }
+    if (ctx.grandparent_senders_and_authorities.contains(sender) ||
+        ctx.parent_senders_and_authorities.contains(sender)) {
+        return false;
     }
```

or we can put pointers in the initialization list

goodlyrottenapple: Ah good catch, this wasn't an issue when the two args were pointers. I got rid of the for-loop...

PR #1986, opened 2025-12-17
Combine EVM memory regions into one transaction wide memory region

andreaslyn:


category/vm/runtime/context.S
```diff
@@ -45,13 +45,13 @@ monad_vm_runtime_increase_memory_raw:
...
-    // ctx->memory.cost = rsi;
-    mov [rbx + memory_cost_offset], rsi
-
     // deduct_gas(rcx);
     sub [rbx + gas_remaining_offset], rcx
     js .error_exit
 
+    // ctx->memory.cost = rsi;
+    mov [rbx + memory_cost_offset], rsi
+
```
andreaslyn, line 54: Unrelated, but noticed that the side effect on memory cost was not the same between compiler and interpreter in case of out-of-gas error. This is only relevant for testing purposes.

category/vm/runtime/types.hpp
```diff
@@ -104,65 +108,107 @@ namespace monad::vm::runtime
...
+                data_handle = han;
+                data = dat;
+                parent_handle = han;
+            }
+            else {
+                capacity = initial_capacity;
+                // Allocate the initial memory handle:
+                parent_handle = data_handle = data =
+                    allocator_.aligned_alloc_cached();
+            }
```
andreaslyn, line 141: This constructor will change in the follow up PR, so that `data_handle`, `data` and `capacity` is always passed as argument to the constructor. In particular, on the first call, `data_handle` and `data` will be the initial 8MB memory block, and `capacity` will be `8MB`.

category/vm/vm.cpp
```diff
@@ -97,44 +98,6 @@ namespace monad::vm
...
     template <Traits traits>
     evmc::Result VM::execute_raw(
-        evmc_host_interface const *host, evmc_host_context *host_ctx,
-        evmc_message const *msg, evmc::bytes32 const &code_hash,
-        SharedVarcode const &vcode)
-    {
-        auto const &icode = vcode->intercode();
-        auto rt_ctx = runtime::Context::from(
-            memory_allocator_, host, host_ctx, msg, icode->code_span());
-        return execute_impl<traits>(rt_ctx, code_hash, vcode);
```
andreaslyn, line 107: This functionality has essentially just been inlined at the call site, in order to allow for calling `rt_ctx.return_to<traits>(prev_rt_ctx)`

category/vm/runtime/types.hpp
```diff
@@ -104,65 +108,107 @@ namespace monad::vm::runtime
...
+            // memory index larger than Bin<30>::upper. Hence a lower bound on
+            // current gas consumption is
+            //   max_call_depth * memory_cost_from_word_count(2^20 - 1)
+            //   = 1024 * ((c * c) / 512 + (3 * c))
+            //   > 2 billion
+            //   for c = floor((2^20 - 1) / 32).
+            // This means that more than 2 billion gas must have been consumed
+            // already by the current transaction for the following assertion
+            // to fail:
+            MONAD_VM_ASSERT(x <= Bin<30>::upper);
```
jhb10c, line 192: This is enforcing an absolute memory size limit. Rn this computation is done in quadratic gas cost. If the memory cost model is changed to linear, the bound still holds as allocating enough memory to exceed Bin<30>::upper would require gas
beyond realistic transaction limits. But it drops from 2 billion to roughly 100 million.

category/core/runtime/non_temporal_memory.hpp
```diff
@@ -0,0 +1,55 @@
...
+namespace monad::vm::runtime
+{
+    inline void non_temporal_bzero(void *dest, std::size_t n)
+    {
+        MONAD_VM_DEBUG_ASSERT((reinterpret_cast<uintptr_t>(dest) & 31) == 0);
+        MONAD_VM_DEBUG_ASSERT((n & 31) == 0);
+        auto *s = reinterpret_cast<__m256i *>(dest);
+        auto *const t = reinterpret_cast<std::uint8_t *>(dest);
+        auto *const e = reinterpret_cast<__m256i *>(t + n);
+        __m256i const zero = _mm256_set_epi64x(0, 0, 0, 0);
```
dhil: Any concerns about using `setzero_si256`? I think it is semantically clearer
```suggestion
        __m256i const zero = _mm256_setzero_si256();
```

category/core/runtime/non_temporal_memory.hpp
```diff
@@ -0,0 +1,55 @@
...
+        while (s < e) {
+            _mm256_stream_si256(s, zero);
+            ++s;
+        }
+    }
```
dhil, line 37: I was thinking a bit more about this function. I have come up with an ever-so-slightly simpler alternative that generates the same code (https://godbolt.org/z/653Tj6W18):
```suggestion
    inline void non_temporal_bzero(void *dest, std::size_t n)
    {
        MONAD_VM_DEBUG_ASSERT((reinterpret_cast<uintptr_t>(dest) & 31) == 0);
        MONAD_VM_DEBUG_ASSERT((n & 31) == 0);
        auto *s = static_cast<std::uint8_t *>(dest);
        auto *const e = s + n;
        __m256i const zero = _mm256_setzero_si256();

        while (s < e) {
            _mm256_stream_si256(reinterpret_cast<__m256i *>(s), zero);
            s += 32;
        }
    }
```
The main benefit I see of this code is to eliminate the potential strict aliasing violation caused by `auto *s = reinterpret_cast<__m256i *>(dest);`.

andreaslyn: Yeah seems simpler

category/core/runtime/non_temporal_memory.hpp
```diff
@@ -0,0 +1,55 @@
...
+        MONAD_VM_DEBUG_ASSERT((n & 31) == 0);
+        auto *d = reinterpret_cast<__m256i *>(dest);
+        auto *s = reinterpret_cast<__m256i *>(src);
+        auto *const t = reinterpret_cast<std::uint8_t *>(dest);
+        auto *const e = reinterpret_cast<__m256i *>(t + n);
+        while (d < e) {
+            _mm256_stream_si256(d, *s);
+            ++d;
+            ++s;
+        }
```
dhil: Similar to the transformation in my previous comment (same generated code, c.f. https://godbolt.org/z/enennKKW5)
```suggestion
        auto *d = static_cast<std::uint8_t *>(dest);
        auto *s = static_cast<std::uint8_t *>(src);
        auto *const e = d + n;
        while (d < e) {
            _mm256_stream_si256(reinterpret_cast<__m256i *>(d), *reinterpret_cast<__m256i *>(s));
            d += 32;
            s += 32;
        }
```

category/vm/runtime/types.hpp
```diff
@@ -246,6 +294,29 @@ namespace monad::vm::runtime
...
+                MONAD_VM_DEBUG_ASSERT(memory.parent_handle == p.data_handle);
+                if (memory.data_handle != memory.parent_handle) {
+                    p.release();
+                    p.data = memory.data - p.size;
+                    p.capacity = memory.capacity + p.size;
+                    p.data_handle = memory.data_handle;
+                }
+            }
+            else {
+                memory.dealloc();
```
dhil, line 323: One thing to be vary of here is the stack thrashing phenomenon, which is a pathological case that repeatedly triggers the create/destroy operations of additional stack elements, causing the said operations to dominate. 

I am not aware of any live contracts that would exercise this phenomenon currently (we'd have to do some trial runs to figure it out). Nonetheless, the standard way to avoid this phenomenon is to cache/pool the intermediate stack segments.

I am not convinced this is a problem worth solving now, but rather, I think it is something to keep in mind for the future if we ever observe the phenomenon live on chain.

andreaslyn: I think that we will not experience stack thrashing, because the stack is kept contiguous, copying the old stack/memory into the new region.

category/core/runtime/non_temporal_memory.hpp
andreaslyn: Sure, will fix

category/vm/runtime/types.hpp
```diff
@@ -246,6 +294,29 @@ namespace monad::vm::runtime
             return Memory::Offset::unsafe_from(static_cast<uint32_t>(offset));
         }
 
+        template <Traits traits>
+        [[gnu::always_inline]]
+        void return_to(Context *parent)
+        {
+            non_temporal_bzero(memory.data, memory.size);
```
goodlyrottenapple: Potentially doing unnecessary work here if we are in the base case (i.e. no parent) but `memory.data` has been expanded, since that memory will just be `std:free`d in `memory.dealloc();` as far as I can tell?

andreaslyn: That is a good observation. Need to think a bit. Do you have a suggestion?

goodlyrottenapple: I noticed by in-lining dealloc and moving stuff around a bit so need to go back to the original code... let me see if i can come up with a suggestion...

category/vm/runtime/types.hpp
goodlyrottenapple:
```suggestion
            if (parent != nullptr) {
                non_temporal_bzero(memory.data, memory.size);
                auto &p = parent->memory;
                MONAD_VM_DEBUG_ASSERT(memory.parent_handle == p.data_handle);
                if (memory.data_handle != memory.parent_handle) {
                    p.release();
                    p.data = memory.data - p.size;
                    p.capacity = memory.capacity + p.size;
                    p.data_handle = memory.data_handle;
                }
            }
            else {
                if (memory.data_handle == memory.parent_handle) {
                    non_temporal_bzero(memory.data, memory.size);
                }
                memory.dealloc();
```

andreaslyn: Makes sense. I think we can even inline `dealloc` to just perform the `data_handle == parent_handle` check once.

category/vm/runtime/types.hpp
```diff
@@ -91,10 +93,24 @@ namespace monad::vm::runtime
...
         std::int64_t cost;
+        // Pointer to the beginning of the transaction wide memory:
+        std::uint8_t *data_handle;
+        // The `parent_handle` is a pointer to the original transaction wide
+        // memory (the transaction wide memory at the beginning of the
+        // current call frame).
+        // The `parent_handle` is kept around to allow for lazily freeing the
+        // memory pointed to by `parent_handle`. This memory is lazily freed,
+        // because the call data for the current call frame is potentially a
+        // pointer into the `parent_handle` memory.
```
goodlyrottenapple, line 112: This is great, thanks for adding!

category/vm/runtime/cached_allocator.hpp
```diff
@@ -60,11 +62,16 @@ namespace monad::vm::runtime
...
 
+        template <bool zeroed>
         CachedAllocatorElement *pop()
         {
             MONAD_VM_DEBUG_ASSERT(!empty());
             auto *ptr = elements;
             elements = ptr->next;
+            if constexpr (zeroed) {
+                ptr->next = nullptr;
+                ptr->idx = 0;
```
goodlyrottenapple:
```suggestion
                std::memset(ptr, 0, sizeof(CachedAllocatorElement));
```

nit: not sure c++ technically guarantees the bit pattern will be zero when setting a pointer to `nullptr`?

goodlyrottenapple (approved):

PR #1989, opened 2025-12-18
use future.get() to propagate exceptions

andreaslyn:


andreaslyn:
Close in favor of https://github.com/category-labs/monad/pull/2425

Issue #1888, opened 2025-11-07
Remember partial literal information


andreaslyn:
The x86 emitter is currently only tracking whole-word literal stack elements. It is possible to also keep track of literal values when only parts of the stack element are known. For example the BYTE instruction is known to clear the upper 3 words of the destination stack element. Also when performing a bitwise-and of a stack element and the literal `1`, the emitter should be able to deduce that the upper 3 words are cleared in the destination stack element.

A potential solution for this is to add an array of optional word values to the `StackElem` class,

```
class StackElem
{
    ...
private:
    ...
    std::array<std::optional<uint64_t>, 4> partial_literal_;
}
``` 

If a stack element `StackElem e` has non-null `e.literal_`, then we should additionally have `e.partial_literal_[i].value() == e.literal_.value()[i]` for  `0 <= i < 4`.

The partial literal information can be used to optimize instructions, by omitting instructions when they are known to be no-operations. If a partial literal word is known to be zero, then it is a no-operation with used as a bitwise-or operand.

PR #1991, opened 2025-12-18
Infer bounds for stack element

laurenthuberdeau:
## Context

For https://github.com/category-labs/monad/issues/1888

A possible performance improvement comes from the observation that some instructions produce values that will never exceed a certain bound. `IsZero`, `Byte`, `Clz` are examples of such instructions, whose result always fit in 1, 8 and 9 bits, respectively. The current implementation of the compiler treats the results of these operations as unknown 256-bit values (stored either in a register or in memory), but instructions that consume these operands could avoid doing the full 256-bit operation since we know that the upper N bits are all 0. Non-arithmetic operations also produce results that only use some of the 256 bits. For example, `CallDataSize`, which is frequently used in the function selector dispatch, produces a value that fits in 32-bit, and `Caller`’s output is a 160-bit address.

This PR implements a local abstract interpretation over the bit width of stack elements, so that future PRs can use that information to generate faster and more compact machine code.

Examples of code generation optimizations:

- Instructions using `test_high_bits192`, such as `byte` and shifts, could avoid testing the upper 192 bits.
- Instructions using `GENERAL_BIN_INSTR`, such as `add`, can avoid doing `op_with_carry` over the full 256-bit range.
- Logical operations, comparison operations and `CLZ` can skip some comparisons when the result is known to be 0/false.
- More accurate “fast `MUL`” logic to fallback to runtime implementation less often if we know that an unknown stack element is small enough.

## Testing

Ran the fuzzer, found and fixed an issue with `SUB` and then it found nothing while running for 10-15 minutes. One thing the fuzzer doesn't test is that the bounds are as small as possible (i.e. the fuzzer would be happy if `compute_result_bound` always returned 256).

### DIFF
--- a/category/vm/compiler/ir/bound_inference.hpp
+++ b/category/vm/compiler/ir/bound_inference.hpp
@@ -1,0 +1,339 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/core/runtime/uint256.hpp>
+#include <category/vm/compiler/ir/instruction.hpp>
+#include <category/vm/compiler/ir/x86.hpp>
+#include <category/vm/compiler/ir/x86/emitter.hpp>
+
+#include <evmc/evmc.h>
+
+#include <cstdint>
+
+namespace monad::vm::compiler::bound_inference
+{
+    using namespace monad::vm::compiler;
+    using namespace monad::vm::compiler::native;
+
+    // Compute bounds on the result of an instruction.
+    // It may not compute the tightest possible bound for instructions that are
+    // constant folded, i.e. composed only of literals.
+    std::uint32_t compute_result_bound(Instruction instr, Emitter &emit)
+    {
+        auto &stack = emit.get_stack();
+        auto const top_index = stack.top_index();
+        using enum OpCode;
+        if (instr.opcode() == Add) {
+            return std::min(
+                256u,
+                std::max(
+                    stack.get(top_index)->bit_upper_bound(),
+                    stack.get(top_index - 1)->bit_upper_bound()) +
+                    1);
+        }
+        else if (instr.opcode() == Mul) {
+            return std::min(
+                256u,
+                stack.get(top_index)->bit_upper_bound() +
+                    stack.get(top_index - 1)->bit_upper_bound());
+        }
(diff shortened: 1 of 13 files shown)


category/vm/compiler/ir/x86.cpp
```diff
@@ -331,9 +333,15 @@ namespace
             MONAD_VM_DEBUG_ASSERT(
                 remaining_base_gas >= instr.static_gas_cost());
             remaining_base_gas -= instr.static_gas_cost();
+            auto const result_bound = compute_result_bound(emit, instr);
             emit_instr<traits>(emit, instr, remaining_base_gas);
+            if (instr.stack_increase() > 0) {
```
laurenthuberdeau: Nit: `instr.stack_increase()` is always 0 or 1, but type allows up to 256 while `compute_result_bound` assumes there's up to 1 result per instruction.

category/vm/compiler/ir/bound_inference.hpp
```diff
@@ -0,0 +1,293 @@
...
+        case TStore:
+            return 256; // no result pushed
+        case MCopy:
+            return 256; // no result pushed
+        case Push:
+            return instr.index() * 8; // number of bits pushed
+        case Dup:
+            return 256; // same as the duplicated value
+        case Swap:
+            return 256; // same as the swapped value
```
Baltoli: Needs updating I guess - comment is correct but implementation isn't

category/vm/compiler/ir/bound_inference.hpp
```diff
@@ -0,0 +1,293 @@
...
+            return std::max(
+                       stack.get(top_index)->bit_upper_bound(),
+                       stack.get(top_index - 1)->bit_upper_bound()) +
+                   1;
+        case Mul:
+            return stack.get(top_index)->bit_upper_bound() *
+                   stack.get(top_index - 1)->bit_upper_bound();
+
+            // Max bound corresponds to case where denominator = 1.
+            return stack.get(top_index)->bit_upper_bound();
```
Baltoli: dead code + the multiply / add bug Andreas flagged

category/vm/compiler/ir/bound_inference.hpp
```diff
@@ -0,0 +1,293 @@
...
+            auto const signed_shift = instr.opcode() == Sar && val_bound == 256;
+            if (!signed_shift && shift->literal()) {
+                // Unsigned shift with a literal shift amount
+                auto const shift_lit = shift->literal()->value;
+                if (shift_lit >= 256) {
+                    return 0; // All bits shifted out of range
+                }
+                else {
+                    // Difference with SHL: subtract the shift amount
+                    return val_bound - static_cast<std::uint32_t>(shift_lit);
```
Baltoli, line 169: Can this underflow? I think it's not a safety issue if it does, but it'll pessimise to 256 rather than 0 on clamping.

laurenthuberdeau: Good catch!

category/vm/compiler/ir/bound_inference.hpp
laurenthuberdeau: Fixed

category/vm/compiler/ir/bound_inference.hpp
laurenthuberdeau: Fixed

Baltoli:
Marking as draft rather than closing as it's plausible that we want to revisit some version of this later.

PR #1993, opened 2025-12-19
`monad-ethereum-test` improvements

goodlyrottenapple:
This PR:
* removes the confusingly named ethereum_test.{hpp,cpp} files
* adds `--blockchain-tests` and `--transaction-tests` flags for manually specifying the path to test files, disabling the hard coded tests if either flag is specified
* extends support to all monad forks
* fixes bug wherein specifying no fork would trigger an assertion. The behaviour now should be the originally intended one, i.e. if no `--fork` argument is passed, all tests will run regardless of revision
* adds aliases for forks which aren't labeled with their canoncial name, e.g. `Paris` is refered to as `Merge` in the Ethereum JSON tests. This way, we can call `--fork paris` and it will pick up the right tests instead of having to do `--fork merge`


test/ethereum_test/include/revision_map.hpp
```diff
@@ -0,0 +1,71 @@
...
+inline std::unordered_map<std::string, std::string> const revision_map_keys =
+    [] {
+        std::unordered_map<std::string, std::string> m;
+
+        for (auto const &[k, v] : revision_map) {
+            m.emplace(k, k);
+        }
+
+        return m;
+    }();
```
goodlyrottenapple: This is a bit of a hack because the CLI library does not like the mapping to an `std::variant` so this map is used as a way to normalise the command line argument to the canoncial name, which is then looked up in the original map.

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -86,19 +91,85 @@
...
+    {
+        if constexpr (is_evm_trait_v<traits>) {
+            auto const sender_account = state.recent_account(sender);
+            auto const icode = state.get_code(sender)->intercode();
+            return ::monad::validate_transaction(
+                traits::evm_rev(),
+                tx,
+                sender_account,
+                {icode->code(), icode->size()});
+        }
```
goodlyrottenapple, line 146: Had to unfold this definition because `EthereumMainnet{}.validate_transaction(..)` would call the incorrect version of `get_revision`.

test/ethereum_test/src/transaction_test.cpp
```diff
@@ -168,12 +175,19 @@ void TransactionTest::TestBody()
...
-        GTEST_SKIP() << "no test cases found for revision="
-                     << revision_.value();
+        if (std::holds_alternative<evmc_revision>(revision_.value())) {
+            GTEST_SKIP() << "no test cases found for revision="
+                         << std::get<evmc_revision>(revision_.value());
+        }
+        else {
+            GTEST_SKIP() << "no test cases found for revision="
+                         << std::get<monad_revision>(revision_.value());
+        }
```
goodlyrottenapple: bit ugly, not sure if there is a more canonical way to get a string from the revisions

test/ethereum_test/src/from_json.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <nlohmann/json_fwd.hpp>
+
+#include <gtest/gtest.h>
+
+#include <cstdint>
+
+namespace nlohmann
+{
+
+    void
```
goodlyrottenapple, line 39: This function was originally in `ethereum_test.cpp`. Not sure whether it should conform to the nlohmann JSON library convention as we are doing asserts within it, but this won't be exported elsewhere in the codebase so should just be a matter of preference/convention.

test/ethereum_test/src/main.cpp
```diff
@@ -53,20 +55,40 @@ int main(int argc, char *argv[])
...
         ->transform(
-            CLI::CheckedTransformer(test::revision_map, CLI::ignore_case));
+            CLI::CheckedTransformer(test::revision_map_keys, CLI::ignore_case));
+
+    app.callback([&]() {
+        if (fork_name) {
+            auto it = test::revision_map.find(*fork_name);
+            if (it == test::revision_map.end()) {
+                throw CLI::ValidationError("Invalid fork name");
+            }
```
Baltoli:
```suggestion
            if (!test::revision_map.contains(*fork_name)) {
                throw CLI::ValidationError("Invalid fork name");
            }
```

test/ethereum_test/src/transaction_test.cpp
```diff
@@ -113,14 +109,25 @@ void process_transaction(Transaction const &txn, nlohmann::json const &expected)
...
+    if (std::holds_alternative<evmc_revision>(revision)) {
+        auto const rev = std::get<evmc_revision>(revision);
+        MONAD_ASSERT(rev != EVMC_CONSTANTINOPLE);
+        SWITCH_EVM_TRAITS(process_transaction, txn, expected);
+    }
+    else {
+        auto const rev = std::get<monad_revision>(revision);
+        SWITCH_MONAD_TRAITS(process_transaction, txn, expected);
+    }
     MONAD_ASSERT(false);
```
Baltoli:
```suggestion
```
should be dead code now I think - compiler should be able to tell both of the branches return

test/ethereum_test/src/transaction_test.cpp
Baltoli: You could de-duplicate these branches by visiting with a lambda that has a deduced argument type: https://gcc.godbolt.org/z/E8WPxKGb1

test/ethereum_test/src/main.cpp
goodlyrottenapple: I switched to `revision = test::revision_map.at(*fork_name);` because we know `fork_name` is validated/canonical at that point.

test/ethereum_test/src/transaction_test.cpp
goodlyrottenapple: ah thanks!

test/ethereum_test/src/main.cpp
```diff
@@ -53,20 +55,35 @@ int main(int argc, char *argv[])
...
+        }
+    });
+
     app.add_option("--txn", txn_index, "Index of transaction to run");
     app.add_flag("--trace_calls", trace_calls, "Enable call tracing");
+    app.add_option(
+           "--tests",
+           tests_path,
+           "Path to blockchain tests, overrides the hard-coded tests path.")
+        ->check(CLI::ExistingPath);
```
goodlyrottenapple, line 88: I added the `--tests` flag which sets the blockchain tests folder and disables all baked in paths including the transaction tests. However, perhaps we want to instead have two flags `--blockchain-tests` and `--transaction-tests` to specify either kind of tests manually?

test/ethereum_test/src/main.cpp
```diff
@@ -53,20 +55,41 @@ int main(int argc, char *argv[])
...
     unsigned sleep_seconds = 0;
 
     CLI::App app{"monad ethereum tests runner"};
     app.add_option("--log_level", log_level, "Logging level")
         ->transform(CLI::CheckedTransformer(log_level_map, CLI::ignore_case));
-    app.add_option("--fork", revision, "Fork to run unit tests for")
+    app.add_option("--fork", fork_name, "Fork to run unit tests for")
         ->transform(
-            CLI::CheckedTransformer(test::revision_map, CLI::ignore_case));
+            CLI::CheckedTransformer(test::revision_map_keys, CLI::ignore_case));
```
cmcl:
```suggestion
            CLI::IsMember(std::views::keys(test::revision_map) | std::ranges::to<std::set>(), CLI::ignore_case));
```

Then `revision_map_keys` can be removed.

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -86,19 +91,79 @@
...
 MONAD_ANONYMOUS_NAMESPACE_BEGIN
 
-struct EthereumMainnetRev : EthereumMainnet
+using BOOST_OUTCOME_V2_NAMESPACE::success;
+
+template <Traits traits>
+struct TraitsMainnet : MonadChain
 {
-    evmc_revision const rev;
+    TraitsMainnet() {}
```
cmcl: Clang tidy nit:
```suggestion
    TraitsMainnet() = default;
```

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -345,61 +479,281 @@ Result<BlockExecOutput> BlockchainTest::execute(
...
+    }
+
+    if (has_post_state) {
+        validate_post_state(j_contents.at("postState"), db_post_state);
+    }
+    LOG_DEBUG("post_state: {}", db_post_state.dump());
+}
+
+void BlockchainTest::process_test(
+    std::variant<evmc_revision, monad_revision> const revision,
```
cmcl:
```suggestion
    std::variant<evmc_revision, monad_revision> const &revision,
```
Alternatively, can also just remove the `const` qualifier.

test/ethereum_test/src/transaction_test.cpp
```diff
@@ -113,14 +80,24 @@ void process_transaction(Transaction const &txn, nlohmann::json const &expected)
 }
 
 void process_transaction(
-    evmc_revision const rev, Transaction const &txn,
-    nlohmann::json const &expected)
+    std::variant<evmc_revision, monad_revision> const revision,
```
cmcl:
```suggestion
    std::variant<evmc_revision, monad_revision> const &revision,
```

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -345,61 +479,281 @@ Result<BlockExecOutput> BlockchainTest::execute(
...
-    SWITCH_EVM_TRAITS(
-        execute_and_record, block, db, vm, block_hash_buffer, enable_tracing);
-    MONAD_ASSERT(false);
-}
+    InMemoryMachine machine;
+    mpt::Db db{machine};
+    db_t tdb{db};
+    vm::VM vm;
+    {
+        auto const genesisJson = j_contents.at("genesisBlockHeader");
```
cmcl:
```suggestion
        auto const &genesisJson = j_contents.at("genesisBlockHeader");
```
clang-tidy: performance-unnecessary-copy-initialization (copy-constructed from const reference)

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -440,274 +794,54 @@ void BlockchainTest::TestBody()
...
+    bool enable_tracing)
+{
+    namespace fs = std::filesystem;
+    MONAD_ASSERT(fs::exists(root) && fs::is_directory(root));
 
-        bool const has_post_state = j_contents.contains("postState");
-        bool const has_post_state_hash = j_contents.contains("postStateHash");
-        MONAD_DEBUG_ASSERT(has_post_state || has_post_state_hash);
+    for (auto const &entry : fs::recursive_directory_iterator{root}) {
+        auto const path = entry.path();
```
cmcl:
```suggestion
        auto const &path = entry.path();
```

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -440,274 +794,54 @@ void BlockchainTest::TestBody()
...
-                }
-            }
-        }
+void register_blockchain_tests_path(
+    std::filesystem::path const &root,
+    std::optional<std::variant<evmc_revision, monad_revision>> const &revision,
+    bool enable_tracing)
+{
+    namespace fs = std::filesystem;
+    MONAD_ASSERT(fs::exists(root) && fs::is_directory(root));
```
cmcl: I wonder if it is useful to support running single JSON fixtures?  Some directories in the execution spec test fixtures have only one file under a given sub-directory but it isn't universal.

cmcl (commented):
Partially through review: mostly commenting on `blockchain_test.cpp` so far.

test/ethereum_test/src/main.cpp
goodlyrottenapple: Oh nice! Thanks for this :)

test/ethereum_test/src/blockchain_test.cpp
goodlyrottenapple: Good idea. the `{blockchain,transaction}-tests` flags now both support either a single file or a directory

test/ethereum_test/include/blockchain_test.hpp
```diff
@@ -45,37 +46,43 @@ MONAD_NAMESPACE_END
...
 
-    static Result<std::vector<Receipt>> execute_dispatch(
-        evmc_revision, Block &, test::db_t &, vm::VM &, BlockHashBuffer const &,
-        bool);
+    template <Traits traits>
+    static void process_test(std::string const &, nlohmann::json const &, bool);
 
-    static void
-    validate_post_state(nlohmann::json const &json, nlohmann::json const &db);
+    static void process_test(
```
cmcl: There's a bit of inconsistency between `BlockchainTest` and `TransactionTest` here.  `process_test` is a static member but the analogous `process_transaction` is a defined within the anonymous monad namespace.

cmcl (dismissed):
The rest looks good to me.  My final nit was about the discrepancy between `BlockchainTest` and `TransactionTest`.  A potential solution is to make `TransactionTest::process_transaction` a static member but I won't block merging on it.

cmcl (approved):

test/ethereum_test/include/blockchain_test.hpp
```diff
@@ -45,37 +46,22 @@ MONAD_NAMESPACE_END
 
 MONAD_TEST_NAMESPACE_BEGIN
 
+using db_t = monad::TrieDb;
```
dhil: Is this used anywhere now?

test/ethereum_test/exclude/paris.cmake
```diff
@@ -17,4 +17,11 @@ set(merge_excluded_tests
     "BlockchainTests.InvalidBlocks/bcMultiChainTest/UncleFromSideChain.json" # Uncle
     "TransactionTests.ttWrongRLP/RLPIncorrectByteEncoding01.json"
     "TransactionTests.ttWrongRLP/RLPIncorrectByteEncoding127.json"
+
+    # Stricter validation of base fee
+    "BlockchainTests.london/validation/test_invalid_header.json"
+
+    # EIP-7610
+    "BlockchainTests.paris/eip7610_create_collision/test_init_collision_create_opcode.json"
+    "BlockchainTests.paris/eip7610_create_collision/test_init_collision_create_tx.json"
```
goodlyrottenapple, line 26: The recently bumped version of ethereum tests seems to have changed the network name from `Merge` to `Paris`. This wasn't caught on master, but because this PR accepts both network names and maps both to `EVMC_PARIS`, the above tests run on this branch and predictably fail.

cmcl: Just want to check I understand what is meant by this comment.  `master` doesn't support `--fork paris` so it won't fail these tests because they aren't run?  Whereas this PR adds support for `--fork paris` and thus will fail without excluding these tests?

goodlyrottenapple: Yes, that's correct. The change from "Merge" to "Paris" happened a while ago I think, but it's only been picked up when I added the additional mapping from "Paris" to "EVMC_PARIS".

PR #1997, opened 2025-12-19
[mpt] Abstract out hash function from Merkle Compute

kkuehlz:
Introduces a the MerkleHasher trait which is plumbed through as a template parameter for the different Compute implementations. This PR is fully backwards compatible with the current implementation. Every Compute implementation uses the Keccak256Hasher. This PR prepares the compute module for pluggable hash functions, namely to support the coming BSTORE subtrie, which will use blake3.

The trait allows us to reuse all the existing code for merkelization while customizing the hash function used. The blake3 addition will look roughly like the following:

```
struct Blake3Hasher
{
    static_assert(BLAKE3_OUT_LEN == HASH_SIZE);

    static void hash(unsigned char const *in, size_t len, unsigned char *out)
    {
        blake3_hasher hasher;
        blake3_hasher_init(&hasher);
        blake3_hasher_update(&hasher, in, len);
        blake3_hasher_finalize(&hasher, out, HASH_SIZE);
    }
};

using BlockStorageMerkleCompute =
        MerkleComputeBase<Blake3Hasher, ComputeBlockStorageLeaf>>;
```

### DIFF
--- a/category/mpt/compute.hpp
+++ b/category/mpt/compute.hpp
@@ -33,17 +33,18 @@ MONAD_MPT_NAMESPACE_BEGIN
 
 namespace detail
 {
+    template <MerkleHasher Hasher>
     struct InternalMerkleState
     {
-        unsigned char buffer[KECCAK256_SIZE];
+        unsigned char buffer[HASH_SIZE];
         unsigned len{0};
 
-        void keccak_inplace_to_root_hash()
+        void hash_inplace_to_root_hash()
         {
-            MONAD_DEBUG_ASSERT(len <= KECCAK256_SIZE);
-            if (len < KECCAK256_SIZE) {
-                keccak256(buffer, len, buffer);
-                len = KECCAK256_SIZE;
+            MONAD_DEBUG_ASSERT(len <= HASH_SIZE);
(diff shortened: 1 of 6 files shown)


category/mpt/compute.hpp
```diff
@@ -57,9 +58,37 @@ encode_16_children(std::span<ChildData>, std::span<unsigned char> result);
...
+
+    MONAD_DEBUG_ASSERT(path.data_size() <= HASH_SIZE);
+
+    unsigned char path_arr[max_compact_encode_size];
+    auto const first = compact_encode(path_arr, path, has_value);
+    MONAD_ASSERT(first.size() <= max_compact_encode_size);
+    // leaf and hashed node ref requires rlp encoding,
+    // rlp encoded but unhashed branch node ref doesn't
+    bool const need_encode_second = has_value || second.size() >= HASH_SIZE;
+    auto const concat_len =
```
marcograss, line 76: would it be possible to check second.size() similarly to first.size() with MONAD_ASSERT?

also in theory concat_len could integer overflow although unlikely because the data we are summing sizes is backed by memory, however we can explore options like checked arithmetic

PR #2001, opened 2025-12-23
[vm] Facility to lazily construct `SharedIntercode`

### LINKED ISSUE #2000: Add a raw bytecode version of try_insert_varcode
Add a raw bytecode version of `Compiler::try_insert_varcode` with the signature
```
SharedVarcode try_insert_varcode_raw(
    evmc::bytes32 const &code_hash, std::span<std::uint8_t const> code);
```
The purpose is to avoid calling `make_shared_intercode` when the code is already cached in the `LruWeightCache`.

Also a corresponding function to `VarcodeCache`
```
SharedVarcode try_set_raw(
    evmc::bytes32 const &code_hash, std::span<std::uint8_t const> code);
```
One way to implement `VarcodeCache::try_set_raw` is to lookup in the LRU weight cache with first, and call `LruWeightCache::try_insert` only if not found. This does allow for a race condition where `LruWeightCache::try_insert` is called multiple times on the same code, if two threads are executing `VarcodeCache::try_set_raw` at the same time.

dhil:
This patch adds the method `VarcodeCache::try_set_raw` (and corresponding methods in `VM` and `Compiler`) which given a bytecode checks whether a suitable `SharedIntercode` object exists already in the cache before constructing it.

Resolves #2000.

### DIFF
--- a/test/vm/unit/monad_vm_interface_tests.cpp
+++ b/test/vm/unit/monad_vm_interface_tests.cpp
@@ -196,6 +196,30 @@ namespace
     };
 }
 
+TEST(MonadVmInterface, VarcodeCacheEmptyCode)
+{
+    static uint32_t const bytecode_cache_weight = 3;
+    static uint32_t const warm_cache_kb = 2 * bytecode_cache_weight;
+    static uint32_t const max_cache_kb = warm_cache_kb;
+
+    VarcodeCache cache{max_cache_kb, warm_cache_kb};
+
+    uint8_t *const p = nullptr;
+    std::span<uint8_t const> empty_code{p, 0};
+    auto code_hash = std::bit_cast<evmc::bytes32>(
+        ethash::keccak256(empty_code.data(), empty_code.size()));
+
+    auto vcode0 = cache.try_set(code_hash, make_shared_intercode(empty_code));
+
+    ASSERT_EQ(vcode0, cache.get(code_hash));
+    ASSERT_EQ(vcode0->intercode()->size(), 0);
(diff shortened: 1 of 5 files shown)


test/vm/unit/monad_vm_interface_tests.cpp
```diff
@@ -260,6 +262,33 @@ TEST(MonadVmInterface, VarcodeCache)
...
     ASSERT_EQ(vcode1, cache.get(hash1).value());
     ASSERT_FALSE(cache.get(hash0).has_value());
+    ASSERT_EQ(vcode2, cache.try_set_raw(hash2, bytecode2));
+
+    auto [bytecode3, hash3] = make_bytecode(3);
+    ASSERT_EQ(
+        VarcodeCache::code_size_to_cache_weight(
+            static_cast<uint32_t>(bytecode3.size())),
+        bytecode_cache_weight);
+    auto icode3 = make_shared_intercode(bytecode3);
```
andreaslyn: Is this dead code?

andreaslyn (approved):

test/vm/unit/monad_vm_interface_tests.cpp
dhil: Yes, I think I originally intended to compare the vcode3 bytecode against the icode3 bytecode. I will remove the icode3 binding.

andreaslyn (approved):

goodlyrottenapple (approved):

andreaslyn (approved):

PR #2009, opened 2025-12-30
Switch to VM memory pool

andreaslyn:


category/execution/ethereum/block_hash_history_test.cpp
```diff
@@ -95,14 +95,18 @@ namespace
             call_tracer, tx_context, block_hash_buffer, state};
 
         bytes32_t const calldata = enc(block_number);
+        auto msg_memory = state.vm().message_memory_ref();
```
andreaslyn, line 98: At some point this memory ref structure will become part of the message structure, but it seems that this is better to do after we have migrated away from evmc.

.github/workflows/test-vm.yml
```diff
@@ -76,7 +76,7 @@ jobs:
               with:
                 repository: category-labs/evmone
                 token: ${{ steps.generate_token.outputs.token }}
-                ref: v0.18.0-category
+                ref: andreaslyn/memory-pool
```
andreaslyn: TODO: update here and below when the evmone branch is merged.

test/vm/unit/monad_vm_interface_tests.cpp
```diff
@@ -745,3 +744,21 @@ TEST(MonadVmInterface, execute_bytecode)
...
+    {
+        auto mem1 = vm.message_memory_ref();
+        ASSERT_EQ(mem1.get()[capacity - 1], 0);
+        mem1.get()[capacity - 1] = 1;
+        {
+            auto mem2 = vm.message_memory_ref();
+            ASSERT_EQ(mem2.get()[capacity - 1], 0);
+        }
+        ASSERT_EQ(mem1.get()[capacity - 1], 1);
+        mem1.get()[capacity - 1] = 0;
```
goodlyrottenapple, line 769: why the final assignment `mem1.get()[capacity - 1] = 0;`?

andreaslyn: Under normal circumstances, it is an invariant that memory is cleared before it is deallocated, but since the test has finished at this point, it is fine to not clear the memory. Let me add a comment explaining this.

category/vm/runtime/types.hpp
```diff
@@ -112,33 +113,23 @@ namespace monad::vm::runtime
         // pointer into the `parent_handle` memory.
         std::uint8_t *parent_handle;
```
goodlyrottenapple:
```suggestion
        std::uint8_t *const parent_handle;
```
Now that these are initialised in the constructor, perhaps best make it clear that the parent memory pointer is constant (same with `parent_capacity`)?

category/vm/runtime/types.hpp
```diff
@@ -197,21 +188,19 @@ namespace monad::vm::runtime
         [[gnu::always_inline]]
         void dealloc()
```
goodlyrottenapple: perhaps a better name for this method would now be `zero_out_and_release()` or something similar, since it is basically a) the same as release + the extra step of zeroing out the data/parent_handle b) we no longer actually dealloc the mempool data handle here.

test/vm/unit/runtime/memory_tests.cpp
```diff
@@ -359,42 +364,29 @@ TEST_F(RuntimeTest, ExpandMemory)
...
-    ASSERT_EQ(EvmMemoryAllocatorMeta::cache_list.size(), 0);
-
-    ctx_.gas_remaining = 1'000'000;
-    ctx_.expand_memory(Bin<29>::unsafe_from(Memory::initial_capacity + 1));
-
-    ctx_.memory.dealloc();
-    ctx_.memory.data_handle = nullptr; // mark deallocated
-
-    ASSERT_EQ(EvmMemoryAllocatorMeta::cache_list.size(), 1);
-}
```
goodlyrottenapple, line 385: do we no longer need such a test with the new mem-pool?

category/execution/monad/execute_system_transaction.cpp
```diff
@@ -171,7 +173,9 @@ Result<void> ExecuteSystemTransaction<traits>::execute(State &state)
     state.set_nonce(sender_, nonce + 1);
 
     state.push();
-    call_tracer_.on_enter(to_message());
+    auto msg_memory = state.vm().message_memory_ref();
+    call_tracer_.on_enter(
+        to_message(msg_memory, state.vm().message_memory_capacity()));
```
goodlyrottenapple: Nit: do we need to grab and pass memory to the staking syscall contract?

test/vm/unit/runtime/memory_tests.cpp
andreaslyn: There is a separate unit test file for the memory pool [monad_vm_memory_pool_tests.cpp](https://github.com/category-labs/monad/pull/2009/files#diff-0923cae98070cd2cc2ba8b6a91195334666631ec7a21066357a2b2bd9cc4a874)

category/execution/monad/execute_system_transaction.cpp
andreaslyn: Right that should not be necessary.

test/vm/unit/monad_vm_memory_pool_tests.cpp
Baltoli, line 1: Nice

category/vm/memory_pool.cpp
```diff
@@ -0,0 +1,103 @@
...
+        if (old_head == &empty_head_) {
+            void *const p = std::aligned_alloc(32, alloc_capacity_);
+            MONAD_VM_ASSERT(p);
+            runtime::non_temporal_bzero(p, alloc_capacity_);
+            return reinterpret_cast<std::uint8_t *>(p);
+        }
+
+        // This clears the memory buffer:
+        std::memset(
+            static_cast<void *>(&old_head->next), 0, sizeof(old_head->next));
```
goodlyrottenapple:
```suggestion
        std::memset(
            static_cast<void *>(old_head), 0, sizeof(Node));
```
to appease clang-tidy?

PR #2020, opened 2026-01-14
Fix LLVM compiler build

laurenthuberdeau:
https://github.com/category-labs/monad/pull/1986 changed the `Context::expand_memory(Bin<30> min_size)` signature to `Context::expand_memory(Bin<29> min_size)`, which broke the LLVM compiler. This PR fixes the type error, and adjusts the LLVM compiler with the reduced memory size limit.


Baltoli (approved):
Looks good; please squash down commits before merging

PR #2027, opened 2026-01-19
Deduct cold storage cost before accessing trie in SSTORE

Baltoli:
This is a minor optimisation that prevents cheap trie accesses via deferred gas accounting.


category/vm/runtime/storage.hpp
```diff
@@ -87,7 +90,7 @@ namespace monad::vm::runtime
 
         if constexpr (traits::eip_2929_active()) {
             if (access_status == EVMC_ACCESS_COLD) {
-                gas_used += traits::cold_storage_cost() + min_gas;
+                gas_used += min_gas;
```
andreaslyn: Can this be moved up as well?

Baltoli: Good catch, it can indeed

PR #2037, opened 2026-01-23
Skip reserve checks for init-selfdestruct contracts

aa755:
# Problem
This PR solves https://github.com/category-labs/category-external/issues/133

> If I create a contract under an address that has been previously sent a balance, then self destruct that contract, the entire transaction reverts.
TX 1: send address 0xFOO... 1 gwei
TX 2 create contract at 0xFOO... and self destruct it in the constructor. Reverts.

>You can repro this by doing:
>Simulate the create/destruct with debug_traceCall.

>curl -X POST $ETH_RPC_URL -H "Content-Type: application/json" --data '{"method":"debug_traceCall","params":[{"from":".....YOUR_ADDDRESS_HERE.....", "data":"0x60256000600073608c4c81f4c5b60a56d05d3ab7fc2034ffd841683c602560006000f060005560016025f3","gas":"0x2dc6c0"}, "latest",{}],"id":1,"jsonrpc":"2.0"}' | jq .
send any amount of funds to the second create call's "to" address. 1gwei works. You must do this from a different address than the simulated "from" so that you don't change the nonces.
Simulate the call again, and it will revert.

# Diagnosis
In the above example, no contract was actually ever deployed at `0xfoo`, because `SELFDESTRUCT` was called during construction: `code_hash` was always `NULL_HASH`.
Note that as long as a contract does get deployed in a transaction, even if SELFDESTRUCT happens in the same transaction where the contract was deployed, the current implementation already does the right (permissive) thing because actual [account nullification](https://github.com/category-labs/monad/blob/d22abd656c70a598d67fa1756db471921ddb06ba/category/execution/ethereum/state3/state.cpp#L446) is deferred until the end of the transaction ([destruct_suicides](https://github.com/category-labs/monad/blob/d22abd656c70a598d67fa1756db471921ddb06ba/category/execution/ethereum/execute_transaction.cpp#L375) in [execute_final](https://github.com/category-labs/monad/blob/d22abd656c70a598d67fa1756db471921ddb06ba/category/execution/ethereum/execute_transaction.cpp#L448)), which happens after reserve balance checks.

So, to solve the problem we mainly need to tweak the implementation to make more permissive the case where `SELFDESTRUCT` happens during construction. This PR does a very minor tweak: it doesn't add any new state/tracking/marking and just uses the existing info.

# Design/Justification 

## Correctness criteria
First, we can recall the main correctness criteria for this isSC (is smart contract) exemption (from reverts): we need to ensure that nobody has a private key for accounts exempted this way. (Coq proofs needed the assumption that no transactions were sent by accounts exempted under this isSC exemption). We should aim to choose a maximally permissive check satisfying this criteria. Over time, we have been making this more and more permissive, and this PR is another one in that direction. The correctness criteria is satisfied by the [current design](https://github.com/category-labs/monad/pull/1916) because SC addresses are derived from hashes of code/nonces and thus having a corresponding private key is cryptographically hard.

## Design
In reserve balance checks, ac.[is_destructed()](https://github.com/category-labs/monad/blob/d22abd656c70a598d67fa1756db471921ddb06ba/category/execution/ethereum/state3/account_substate.hpp#L50-L53) returns true iff SELFDESTRUCT was done on the account ac during the transaction. If ac.[is_destructed()](https://github.com/category-labs/monad/blob/d22abd656c70a598d67fa1756db471921ddb06ba/category/execution/ethereum/state3/account_substate.hpp#L50-L53) is true, before exempting ac, we need to ensure that ac can never be an EOA. It suffices to check that `ac.incarnation` is equal to the current block/tx index (suggestion made by codex-cli originally).

## Justification
We need to prove that `ac.is_destructed() == true /\ ac.incarnation == current_block_tx` implies nobody has a private key for the address of `ac`.
Because `ac.is_destructed() = true`, when that `SELFDESTRUCT` happened, the code was running as `ac`. 
We consider the two cases: whether `ac.code` indicated delegation (started with the delegation marker) at the beginning of the execution of that specific call frame.

### Case 1. `ac.code` indicated delegation
Whether an account's code indicates delegation is a property that does not change after a transaction starts executing, because in this codebase the delegation marker is only written during the pre-execution EIP7702 auth processing stage (there is no precompile/opcode that can set it during contract execution). 
So, at the end of auth processing, `ac.code` was already indicating delegation. At that time, the account must already have existed and had an incarnation. 
That incarnation can only be LAST_TX (if the account was created during auth processing), or something strictly less than the current block/tx (the account was created before this transaction). 
Since then, the incarnation must have remained the same.
During normal transaction execution (as opposed to eth_call overrides), the only place we upgrade an account's incarnation is when CREATE(2) targets an address that has empty code (and 0 nonce, any balance), but in this case, the account's code has a delegation marker.

So this case is impossible, because it contradicts the assumption `ac.incarnation == current_block_tx`.
(If this PR didn't also check `ac.incarnation == current_block_tx`, it would have been possible that the self-destructed account was a delegated EOA. EVM semantics seems to allow SELFDESTRUCT to be called when the code is being run on behalf of a delegated EOA and in that case, [the EOA's balance gets transferred but the code/nonce fields remain the same](https://chatgpt.com/share/6973013f-5314-8003-bcc3-6c0924c793f1))

### Case 2. `ac.code` did not indicate delegation
There are 3 possibilities:
1. the call frame was running init code for `ac`.
2. the call frame was running already-deployed code at `ac`.
3. the call frame was running already-deployed code at some other address (e.g. `DELEGATECALL`).

In each case, the code executes in `ac`'s context, so `ac` must be a smart contract account, thus its address is derived from hashing nonce/initcode (and other items); thus, nobody can have a private key for it.

### DIFF
--- a/category/execution/ethereum/state3/state.cpp
+++ b/category/execution/ethereum/state3/state.cpp
@@ -239,6 +239,21 @@ bytes32_t State::get_code_hash(Address const &address)
     return NULL_HASH;
 }
 
+bool State::is_destructed(Address const &address)
+{
+    auto const &account_state = recent_account_state(address);
+    return account_state.is_destructed();
+}
+
+bool State::is_current_incarnation(Address const &address)
+{
+    auto const &account = recent_account(address);
+    if (MONAD_LIKELY(account.has_value())) {
+        return account.value().incarnation == incarnation_;
+    }
+    return false;
+}
+
 bytes32_t State::get_storage(Address const &address, bytes32_t const &key)
 {
     auto const it = current_.find(address);
@@ -439,7 +454,10 @@ State::selfdestruct(Address const &address, Address const &beneficiary)
         }
     }
 
-    return {account_state.destruct(), initial_balance};
+    bool const inserted = account_state.destruct();
+    // Recompute reserve-balance status after setting the destructed flag.
+    rb_.on_debit(address);
+    return {inserted, initial_balance};
 }
 
 EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
--- a/category/execution/monad/reserve_balance.cpp
+++ b/category/execution/monad/reserve_balance.cpp
@@ -61,6 +61,9 @@ bool dipped_into_reserve(
     MONAD_ASSERT(i < ctx.authorities.size());
     MONAD_ASSERT(ctx.senders.size() == ctx.authorities.size());
 
+    static constexpr bool allow_init_selfdestruct_exemption =
+        traits::monad_rev() >= MONAD_NEXT;
+
     uint256_t const gas_fees =
         uint256_t{tx.gas_limit} * gas_price<traits>(tx, base_fee_per_gas);
     auto const &orig = state.original();
@@ -89,6 +92,11 @@ bool dipped_into_reserve(
             if (!effective_is_delegated) {
                 continue;
             }
+        }
+        else if (
+            allow_init_selfdestruct_exemption && state.is_destructed(addr) &&
+            state.is_current_incarnation(addr)) {
+            continue;
         }
 
         // Check if dipped into reserve
@@ -214,6 +222,14 @@ void ReserveBalance::update_violation_status(Address const &address)
     }
 
     auto &violation_threshold = violation_thresholds_[address];
+    if (allow_init_selfdestruct_exemption_ && state_->is_destructed(address) &&
+        state_->is_current_incarnation(address)) {
+        // Contracts that selfdestruct during init never get a code hash.
+        violation_threshold = uint256_t{0};
+        failed_.erase(address);
+        return;
+    }
+
     if (!violation_threshold.has_value()) {
         if (!subject_account(address)) {
             violation_threshold = uint256_t{0};
@@ -312,6 +328,7 @@ void ReserveBalance::init_from_tx(
     if constexpr (tracking_disabled) {
         tracking_enabled_ = false;
         use_recent_code_hash_ = false;
+        allow_init_selfdestruct_exemption_ = false;
         sender_ = {};
         sender_gas_fees_ = 0;
         sender_can_dip_ = false;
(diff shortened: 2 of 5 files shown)


category/execution/ethereum/state3/state.cpp
```diff
@@ -444,7 +459,10 @@ State::selfdestruct(Address const &address, Address const &beneficiary)
         }
     }
 
-    return {account_state.destruct(), initial_balance};
+    bool const inserted = account_state.destruct();
+    // Recompute reserve-balance status after setting the destructed flag.
+    rb_.on_debit(address);
```
0xSqualo, line 459: Is the unconditional `on_debit` call intentionally left unguarded for pre-MONAD_NEXT revisions, given that it triggers `update_violation_status` for addresses that previously were never evaluated there (even though it should be idempotent)?

aa755: Yes, this is intentional.  Pre-MONAD_NEXT, this `on_debit` should be a no-op. As you mention, `on_debit` is idempotent. The `subtract_from_balance(address, account.value().balance);` above already did `on_debit` and since then, nothing affecting the result of `on_debit` changed in  pre-MONAD_NEXT, as that version does not look at `is_destructed` (IIUC).

mkolosick (dismissed):

category/execution/monad/reserve_balance.cpp
```diff
@@ -214,6 +222,14 @@ void ReserveBalance::update_violation_status(Address const &address)
     }
 
     auto &violation_threshold = violation_thresholds_[address];
+    if (allow_init_selfdestruct_exemption_ && state_->is_destructed(address) &&
```
aa755, line 225: need to guard it on MONAD_NEXT

jhunsaker (approved):

category/execution/monad/reserve_balance.cpp
```diff
@@ -214,6 +222,14 @@ void ReserveBalance::update_violation_status(Address const &address)
...
+        state_->is_current_incarnation(address)) {
```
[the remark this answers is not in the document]
aa755: here, the account cannot be in a delegated state, as explained in the PR description. so either it is an SC or has null code hash. in the former case, we exempted it anyway in [on_set_code](https://github.com/category-labs/monad/blob/25a64a3c73d09801c9e2b453193932e5ecde96f6/category/execution/monad/reserve_balance.cpp#L304) or [initial threshold computation](https://github.com/category-labs/monad/blob/25a64a3c73d09801c9e2b453193932e5ecde96f6/category/execution/monad/reserve_balance.cpp#L208) if the code was installed before this tx, so adding the `state_->get_code_hash(address) == NULL_HASH)` conjunct would not actually change the reserve balance check results.

magedm (approved):

PR #2038, opened 2026-01-23
Pull out state overrides from `eth_call_impl`

Baltoli:


category/rpc/monad_executor.cpp
```diff
@@ -235,6 +235,69 @@ namespace
...
+             state_overrides.override_sets) {
+            // This would avoid seg-fault on storage override for
+            // non-existing accounts
+            if (MONAD_UNLIKELY(!state.account_exists(address))) {
+                state.create_contract(address);
+            }
+
+            if (state_delta.balance.has_value()) {
+                uint256_t const balance = state_delta.balance.value();
+                uint256_t const pessimistic_balance =
```
dhil: Nit: I recall discussing the name `pessimistic_balance` with @aa755 about a month ago. I think the name induced some confusion, because it is not a pessimistic estimate of the balance, but rather the exact balance at this point in time. I think the `pessimistic_` came from the fact that it used to read out `state.get_balance_pessimistic` (or something like that; I think I might be responsible for this particular name binding here). 

Since we are editing this code now, then maybe we can sneak in a name change here. I suggest:
* `balance` => `new_balance`
* `pessimistic_balance` => `current_balance`

Then the statement below would read
```
 state.add_to_balance(address, new_balance - current_balance)
= (expand `state.add_to_balance`)
 current_balance + new_balance - current_balance
```

Baltoli: Makes sense, thanks

PR #2039, opened 2026-01-26
Include remaining account balance in selfdestruct call tracer frame

laurenthuberdeau:
## Context

Other EVM implementation include the balance of the self destructed account (that is transferred to the account passed as argument to the `SELFDESTRUCT` instruction) in the value field of selfdestruct call tracer frames. This PR makes the monad execution client do the same.

### DIFF
--- a/category/execution/ethereum/evmc_host.hpp
+++ b/category/execution/ethereum/evmc_host.hpp
@@ -140,8 +140,11 @@ struct EvmcHost final : public EvmcHostBase
         Address const &address, Address const &beneficiary) noexcept override
     {
         try {
-            call_tracer_.on_self_destruct(address, beneficiary);
-            return state_.selfdestruct<traits>(address, beneficiary);
+            auto const [result, transferred_balance] =
+                state_.selfdestruct<traits>(address, beneficiary);
+            call_tracer_.on_self_destruct(
+                address, beneficiary, transferred_balance);
+            return result;
         }
         catch (...) {
             capture_current_exception();
--- a/category/execution/ethereum/state2/test/test_state.cpp
+++ b/category/execution/ethereum/state2/test/test_state.cpp
@@ -301,15 +301,23 @@ TYPED_TEST(InMemoryStateTraitsTest, selfdestruct)
     s.create_contract(b);
     s.add_to_balance(b, 28'000);
 
-    EXPECT_TRUE(s.selfdestruct<typename TestFixture::Trait>(a, c));
+    EXPECT_EQ(
+        s.selfdestruct<typename TestFixture::Trait>(a, c),
+        std::make_pair(true, 18'000));
     EXPECT_EQ(s.get_balance(a), 0);
     EXPECT_EQ(s.get_balance(c), 56'000);
-    EXPECT_FALSE(s.selfdestruct<typename TestFixture::Trait>(a, c));
+    EXPECT_EQ(
+        s.selfdestruct<typename TestFixture::Trait>(a, c),
+        std::make_pair(false, 0));
 
-    EXPECT_TRUE(s.selfdestruct<typename TestFixture::Trait>(b, c));
+    EXPECT_EQ(
+        s.selfdestruct<typename TestFixture::Trait>(b, c),
+        std::make_pair(true, 28'000));
     EXPECT_EQ(s.get_balance(b), 0);
     EXPECT_EQ(s.get_balance(c), 84'000);
-    EXPECT_FALSE(s.selfdestruct<typename TestFixture::Trait>(b, c));
+    EXPECT_EQ(
+        s.selfdestruct<typename TestFixture::Trait>(b, c),
+        std::make_pair(false, 0));
 
     s.destruct_suicides<typename TestFixture::Trait>();
     if constexpr (TestFixture::Trait::evm_rev() >= EVMC_CANCUN) {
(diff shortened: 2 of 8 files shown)


category/execution/ethereum/state3/state.cpp
```diff
@@ -401,26 +401,30 @@ State::access_storage(Address const &address, bytes32_t const &key)
...
+std::pair<bool, uint256_t>
+State::selfdestruct(Address const &address, Address const &beneficiary)
 {
     auto &account_state = current_account_state(address);
     auto &account = account_state.account_;
     MONAD_ASSERT(account.has_value());
+    uint256_t transferred_balance = 0;
 
     if constexpr (traits::evm_rev() < EVMC_CANCUN) {
+        transferred_balance = account.value().balance;
```
dhil: Looks like this is duplicated in both branches, and thus can be hoisted to the initialization time, and consequently the binding can be made `const`.

laurenthuberdeau: Not quite, because `transferred_balance = 0` when `traits::evm_rev() >= EVMC_CANCUN && address == beneficiary && account->incarnation != incarnation_`. I considered moving the return statement in the if blocks, but that made the function harder to read because of multiple exit points.

category/execution/ethereum/state2/test/test_state.cpp
```diff
@@ -408,7 +424,12 @@ TYPED_TEST(InMemoryStateTraitsTest, selfdestruct_self_separate_tx)
 
     State s{bs, Incarnation{1, 1}};
 
-    EXPECT_TRUE(s.selfdestruct<typename TestFixture::Trait>(a, a));
+    // Balance is burned when self-destructing with the beneficiary being self
+
+    EXPECT_EQ(
```
Baltoli, line 430: Can you confirm with Geth what should happen here? It's not immediately obvious to me that including zero in the frame is the right thing to do (as much as it's a niche edge case).

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -466,12 +466,88 @@ TYPED_TEST(TraitsTest, selfdestruct_logs)
...
+TYPED_TEST(TraitsTest, selfdestruct_logs_value)
+{
+    InMemoryMachine machine;
+    mpt::Db db{machine};
+    TrieDb tdb{db};
+    vm::VM vm;
+
+    // selfdestruct(ADDR_A): PUSH2 0x0100, SELFDESTRUCT
+    static_assert(ADDR_A == 0x0000000000000000000000000000000000000100_address);
+    auto const code = evmc::from_hex("0x610100FF").value();
```
Baltoli: This is a precompile address - worth changing for the avoidance of doubt as this test is setting up something funny by deploying to 0x100. Maybe a bigger issue with this test file post-Osaka?

category/execution/ethereum/evmc_host.hpp
```diff
@@ -140,8 +140,11 @@ struct EvmcHost final : public EvmcHostBase
         Address const &address, Address const &beneficiary) noexcept override
     {
         try {
-            call_tracer_.on_self_destruct(address, beneficiary);
-            return state_.selfdestruct<traits>(address, beneficiary);
+            auto [result, transferred_balance] =
```
Baltoli:
```suggestion
            auto const [result, transferred_balance] =
```

category/execution/ethereum/state2/test/test_state.cpp
laurenthuberdeau: Good catch! They seem to include the balance of the self destructed account, even when there is no transfer.

For Geth:
1. opSelfdestruct passes the initial account balance to the tracer: https://github.com/ethereum/go-ethereum/blob/31d5d82ce595f8bdb519d0062c92ddc361a471ec/core/vm/instructions.go#L894

For Reth:
1. Self destruct calls ENTRY::account_destroyed with balance = account.balance: https://github.com/bluealloy/revm/blob/540ec38bec80f8ff17f423c975ef2985960899c9/crates/context/src/journal/inner.rs#L635
2. account_destroyed creates a JournalEntry::AccountDestroyed with the had_balance = balance: https://github.com/bluealloy/revm/blob/540ec38bec80f8ff17f423c975ef2985960899c9/crates/context/interface/src/journaled_state/entry.rs#L241
3. inspect_selfdestruct matches on JournalEntry::AccountDestroyed and passes had_balance to the tracer: https://github.com/bluealloy/revm/blob/a31fa98efe2208af326cf70f59d374165e0df363/crates/inspector/src/handler.rs#L288

dhil (approved):

category/execution/ethereum/state2/test/test_state.cpp
```diff
@@ -408,7 +424,12 @@ TYPED_TEST(InMemoryStateTraitsTest, selfdestruct_self_separate_tx)
 
     State s{bs, Incarnation{1, 1}};
 
-    EXPECT_TRUE(s.selfdestruct<typename TestFixture::Trait>(a, a));
+    // Balance is burned when self-destructing with the beneficiary being self.
+    // But the trace contains the initial balance, not the balance actually
+    // transferred.
+    EXPECT_EQ(
+        s.selfdestruct<typename TestFixture::Trait>(a, a),
+        std::make_pair(true, 18'000));
```
laurenthuberdeau, line 432: Last force push changed this line from `std::make_pair(true, TestFixture::Trait::evm_rev() < EVMC_CANCUN ? 18'000 : 0)` to `std::make_pair(true, 18'000)` because the trace contains the balance prior to the selfdestruct, regardless of whether the balance is burned or not (and so it's always 18'000).

PR #2043, opened 2026-01-30
cached/incremental reserve balance checks

aa755:
Now that reserve balance checks can be done several times during a transaction using a precompile (previously the CHECKRESERVEBALANCE opcode), it is important to do it in an incremental/cached way, as suggested by Piotr. Before this PR, the check was done by [iterating over all changed accounts](https://github.com/category-labs/monad/blob/f5c1bec66612a9acc8f76a7c44681a231dd24f28/category/execution/monad/reserve_balance.cpp#L66). In this PR, the check is done by just checking whether the cached set of "accounts in reserve balance violation" is empty. This set is maintained in `ReserveBalance::failed_`. It is a private field of the `ReserveBalance` class, but the class provides hooks (methods) that are called on changes to parts of `State` relevant to reserve balance: balance and code of accounts, so that `ReserveBalance::failed_` can be updated. Additionally, `ReserveBalance::on_pop_reject()` is called on `State::pop_reject`.

## Balance changes
The method `ReserveBalance::update_violation_status` is the main workhorse which updates the violation status of the given account. It is roughly equivalent to the [loop body of the previous implementation](https://github.com/category-labs/monad/blob/f5c1bec66612a9acc8f76a7c44681a231dd24f28/category/execution/monad/reserve_balance.cpp#L67-L125): it computes the violation threshold if not already cached and then checks if the balance lower.

### `State::subtract_from_balance`
Here, we just add a call to `ReserveBalance::update_violation_status`.

### `State::add_to_balance`
We know that adding to an account's balance can only cause the account to go from being in violation to being not in violation, but not the other direction. Thus, we only call `ReserveBalance::update_violation_status` if the account was in violation.

## `State::set_code`
If the new code is a delegation marker, we recompute the violation threshold. Otherwise, [just like the previous implementation](https://github.com/category-labs/monad/blob/f5c1bec66612a9acc8f76a7c44681a231dd24f28/category/execution/monad/reserve_balance.cpp#L76), such accounts are exempt and thus we remove the account from the violation set.

## `State::pop_reject`

This happens when the changes to the current call frame are discarded, e.g. due to `REVERT`. `State` already maintains the set of addresses that were updated in this call frame. Fortunately, `State::current_` is guaranteed to have the version of the account just before the call frame, so it merely "pops" the `VersionedStack`. In each element in the versioned stack (`AccountState`), we now store a bool indicating whether the account was in violation of reserve balance. Thus we use that bool to decide whether to add/remove that account from `ReserveBalance::failed_`.

### DIFF
--- a/category/execution/ethereum/test/test_monad_chain.cpp
+++ b/category/execution/ethereum/test/test_monad_chain.cpp
@@ -320,6 +320,22 @@ TYPED_TEST(MonadTraitsTest, revert_transaction_no_dip_gas_fee_with_value_false)
     }
 }
 
+TYPED_TEST(MonadTraitsTest, reserve_balance_checks_disabled_before_monad_four)
+{
+    if constexpr (TestFixture::Trait::monad_rev() < MONAD_FOUR) {
+        // For revisions before MONAD_FOUR, reserve-balance tracking must stay
+        // disabled. If it is accidentally enabled, this case hits
+        // sender_gas_fees > reserve for a non-dipping sender and throws.
+        run_revert_transaction_test<typename TestFixture::Trait>(
+            (1 << IsDelegated), // not allowed to dip into reserve
+            20, // initial balance (MON)
+            11, // gas fee (MON), strictly greater than reserve (10 MON)
+            0, // value (MON)
+            false // expected should_revert
+        );
+    }
+}
+
 TYPED_TEST(MonadTraitsTest, revert_transaction_dip_false)
 {
     run_revert_transaction_test<typename TestFixture::Trait>(


category/execution/ethereum/execute_transaction.cpp
```diff
@@ -220,6 +221,31 @@ template <Traits traits>
...
+        if (sender_code_hash != NULL_HASH) {
+            vm::SharedIntercode const intercode =
+                state.read_code(sender_code_hash)->intercode();
+            sender_is_delegated = monad::vm::evm::is_delegated(
+                {intercode->code(), intercode->size()});
+        }
+
+        bool const sender_can_dip = can_sender_dip_into_reserve<traits>(
+            sender_, host.i_, sender_is_delegated, host.chain_ctx_);
+        uint256_t const max_reserve = get_max_reserve<traits>(sender_);
```
aa755: we are using the sender's threshold for all accounts. this will not work when the threshold becomes configurable per user.  some new plumbing may be needed to get access to `traits` in `rb_reserve_cap`

category/execution/monad/reserve_balance.cpp
```diff
@@ -52,84 +52,190 @@ MONAD_ANONYMOUS_NAMESPACE_BEGIN
...
+}
+
+void ReserveBalance::update_violation_status(
+    Address const &address, AccountState &account_state)
+{
+    if (!tracking_enabled_) {
+        return;
+    }
+
+    AccountState &acct_state = account_state;
```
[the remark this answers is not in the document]
aa755: fixed

category/execution/ethereum/state3/state.hpp
```diff
@@ -216,6 +226,20 @@ class State
...
+
+    template <Traits traits>
+    void init_reserve_balance_context(
+        Address const &sender, Transaction const &tx,
+        std::optional<uint256_t> const &base_fee_per_gas, uint64_t i,
+        ChainContext<traits> const &ctx)
+    {
+        if constexpr (is_monad_trait_v<traits>) {
+            rb_.init_from_tx<traits>(sender, tx, base_fee_per_gas, i, ctx);
+            rb_.on_debit(sender, current_account_state(sender));
```
mkolosick: Why call `on_debit` here? Shouldn't it be handled entirely in `subtract_from_balance`? Is it because this is needed to establish the initial state of the sender being in violation or not?

category/execution/monad/reserve_balance.cpp
```diff
@@ -52,84 +52,207 @@ MONAD_ANONYMOUS_NAMESPACE_BEGIN
 
 template <Traits traits>
 bool dipped_into_reserve(
-    Address const &sender, Transaction const &tx,
-    uint256_t const &base_fee_per_gas, uint64_t const i,
+    Address const &, Transaction const &, uint256_t const &, uint64_t const i,
```
mkolosick: These are no longer necessary. I've pushed a commit with this and some other cleanup stuff [here](https://github.com/category-labs/monad/commit/e46dc778e375f585ca5f773c754fb66e035b4fb1) if you want to cherry-pick.

category/execution/ethereum/state3/state.cpp
```diff
@@ -337,13 +363,16 @@ void State::subtract_from_balance(
...
 }
 
 void State::set_code_hash(Address const &address, bytes32_t const &hash)
 {
-    auto &account = current_account(address);
+    auto &account_state = current_account_state(address);
+    auto &account = account_state.account_;
     MONAD_ASSERT(account.has_value());
     account.value().code_hash = hash;
+    // NOTE: Production code should use set_code instead of set_code_hash.
```
aa755:
```suggestion
    // NOTE: Production code must use set_code instead of set_code_hash.
```

category/execution/ethereum/state3/state.hpp
```diff
@@ -220,6 +230,24 @@ class State
...
+    {
+        if constexpr (is_monad_trait_v<traits>) {
+            rb_.init_from_tx<traits>(sender, tx, base_fee_per_gas, i, ctx);
+        }
+    }
+
+private:
+    bool check_account_min_balance(
+        OriginalAccountState &, std::optional<Account> const &,
+        uint256_t const &);
```
aa755:
```suggestion
```

category/execution/ethereum/state3/state.hpp
aa755: That was in old code when ReserveBalance was being init after `irrevocable_changes` (which debits the sender) and after EIP 7702 processing, to optimize the implenentation of `ReserveBalance::on_code_change`, so that it can assume that the code changes never install a delegation marker. Later I thought the optimization was not worth the risk of these assumptions being violated in future code changes.

category/execution/monad/reserve_balance.cpp
aa755: Thanks. Those changes look great. Can I squash those changes to the 1 commit in this PR? Or do you prefer a separate commit?

mkolosick: Squash them, no need for a separate commit.

PR #2047, opened 2026-02-03
Avoid potential UB on empty modexp precompile input

Baltoli:
If `input` is empty here, then the call to `std::memcpy` is undefined should `input.data()` return `nullptr`.

### DIFF
--- a/category/execution/ethereum/precompiles_impl.cpp
+++ b/category/execution/ethereum/precompiles_impl.cpp
@@ -261,6 +261,10 @@ constexpr uint64_t expmod_min_gas()
 static uint256_t
 uint256_load_partial_be(byte_string_view const input, size_t const len)
 {
+    if (MONAD_UNLIKELY(input.empty())) {
+        return 0;
+    }
+
     uint256_t result{};
     MONAD_VM_ASSERT(32 >= len);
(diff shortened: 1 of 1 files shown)


goodlyrottenapple (approved):

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -261,6 +261,10 @@ constexpr uint64_t expmod_min_gas()
 static uint256_t
 uint256_load_partial_be(byte_string_view const input, size_t const len)
 {
+    if (MONAD_UNLIKELY(input.empty())) {
```
0xSqualo, line 264: Generally looks okay. @Baltoli for awarness that the [`sha256` precompile](https://github.com/category-labs/monad/blob/cbbd825b4a5202143bbd06c471ade93015cb55dc/category/execution/ethereum/precompiles_impl.cpp#L414) checks for `input.data() == nullptr` (instead of `input.empty()`), if there's a preference.

Baltoli: Thanks - will leave it as is to avoid invalidating reviews as I don't think there's a meaningful difference

PR #2048, opened 2026-02-03
Make a single debug_empty chain ctx for testing

goodlyrottenapple:
Followup cleanup PR to #1979

### DIFF
--- a/category/execution/monad/chain/monad_chain.cpp
+++ b/category/execution/monad/chain/monad_chain.cpp
@@ -31,6 +31,18 @@
 #include <category/execution/monad/reserve_balance.hpp>
 #include <category/execution/monad/system_sender.hpp>
 #include <category/execution/monad/validate_monad_transaction.hpp>
+#include <category/vm/evm/explicit_traits.hpp>
+
+namespace
+{
+    using namespace monad;
+
+    static ankerl::unordered_dense::segmented_set<Address> const
+        empty_senders_and_authorities{};
+    static std::vector<Address> const empty_senders{Address{0}};
+    static std::vector<std::vector<std::optional<Address>>> const
+        empty_authorities{{}};
+}
 
 MONAD_NAMESPACE_BEGIN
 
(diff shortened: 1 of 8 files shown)


category/execution/monad/chain/monad_chain.cpp
```diff
@@ -61,4 +62,23 @@ Result<void> MonadChain::validate_transaction(
...
 
+template <typename T>
+    requires is_monad_trait_v<T>
+ChainContext<T> ChainContext<T>::debug_empty()
+{
+    static ankerl::unordered_dense::segmented_set<Address> const
+        empty_senders_and_authorities{};
+    static std::vector<Address> const empty_senders{Address{0}};
+    static std::vector<std::vector<std::optional<Address>>> const
+        empty_authorities{{}};
```
Baltoli: Very minor nitpick - this will stamp out a copy of each static for each `T` this is instantiated with. because they're empty constants you could stick them all at static file scope and reference the same shared statics from this template.

category/execution/monad/chain/monad_chain.hpp
```diff
@@ -50,6 +50,7 @@ struct ChainContext<T>
         &senders_and_authorities;
     std::vector<Address> const &senders;
     std::vector<std::vector<std::optional<Address>>> const &authorities;
+    static ChainContext<T> debug_empty();
```
Baltoli, line 56: Maybe add a comment with a warning only to call this from test code?

category/execution/monad/chain/monad_chain.cpp
goodlyrottenapple: Ah good point thanks

Baltoli (approved):
Looks good!

dhil (approved):
Looks good to me, though I don't like the name `debug_empty` much, as I don't think it has anything to do with debugging per see. I'd prefer simply `empty`.

PR #2049, opened 2026-02-04
Remove Chain's virtual `static_validate_header` and `validate_transaction` methods

goodlyrottenapple:
This PR remove the virtual methods above and Instead dispatches via the trait versions of these functions. This should complete the work of using traits everywhere in execution and switching on timestamp/block only in the core loop.


category/execution/ethereum/validate_transaction_error.hpp
```diff
@@ -0,0 +1,70 @@
+// Copyright (C) 2025 Category Labs, Inc.
```
goodlyrottenapple:
```suggestion
// Copyright (C) 2026 Category Labs, Inc.
```

Should we update the year for new files?

category/rpc/monad_executor.cpp
```diff
@@ -323,8 +323,8 @@ namespace
             // a subroutine. Solving this issue by manually setting account to
             // be EOA for validation
             state.set_code(sender, {});
-            BOOST_OUTCOME_TRY(
-                validate_transaction<traits>(enriched_txn, sender, state));
+            BOOST_OUTCOME_TRY(validate_ethereum_transaction<traits>(
```
goodlyrottenapple, line 288: This seems somewhat surprising? Should we be calling `validate_transaction` which now includes the logic of `validate_monad_transaction` for monad revisions (I tried this, but the `insufficient_balance` test started failing).

cmd/monad/runloop_monad.cpp
```diff
@@ -187,7 +187,6 @@ Result<BlockExecOutput> propose_block(
 
     // Block input validation
     BOOST_OUTCOME_TRY(static_validate_consensus_header(consensus_header));
-    BOOST_OUTCOME_TRY(chain.static_validate_header(block.header));
```
goodlyrottenapple, line 190: The logic from `EthereumMainnet{}.static_validate_header` has been merged into the trait version of `static_validate_header` which gets called via `static_validate_block<traits>` below.

category/execution/ethereum/test/test_validation.cpp
```diff
@@ -322,16 +329,19 @@ TYPED_TEST(TraitsTest, invalid_gas_limit)
...
         .extra_data = {0x00, 0x01, 0x02}};
 
-    auto const result = EthereumMainnet{}.static_validate_header(header);
+    auto const result =
+        static_validate_header<typename TestFixture::Trait>(header);
     ASSERT_TRUE(result.has_error());
-    EXPECT_EQ(result.error(), BlockError::WrongDaoExtraData);
+    if constexpr (is_evm_trait_v<typename TestFixture::Trait>) {
+        EXPECT_EQ(result.error(), BlockError::WrongDaoExtraData);
+    }
```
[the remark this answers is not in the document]
goodlyrottenapple: The test fails validation for different reason in monad traits. Question is whether we want to:
a) leave as is
b) focus only on etherum behaviour by restricting to `EvmTraitsTest`
c) add an else case for monad which specifies the alternate validation error
d) make the header pass validation for monad traits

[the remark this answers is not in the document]
dhil: Is case `c` the matter of adding a single `else`, or do we need an `else` per monad revision?

category/rpc/monad_executor.cpp
dhil: I think this was always using the `validate_transaction` from the `ethereum` module, so I think it is correct to use `validate_ethereum_transaction` here for now.

dhil: I am guessing the test failure that you observe when switching to `validate_monad_transaction` may be due to reserve balance.

category/execution/ethereum/validate_transaction_error.hpp
Baltoli: Yep, if new file or wholesale changes then do "2025-26" (checked w/ Peter)

category/execution/ethereum/validate_transaction.hpp
```diff
@@ -19,74 +19,89 @@
 #include <category/core/int.hpp>
 #include <category/core/result.hpp>
 #include <category/execution/ethereum/core/address.hpp>
+#include <category/execution/ethereum/core/transaction.hpp>
+#include <category/execution/ethereum/state3/state.hpp>
+#include <category/execution/ethereum/transaction_gas.hpp>
+#include <category/execution/ethereum/validate_transaction_error.hpp>
 #include <category/vm/code.hpp>
 #include <category/vm/evm/traits.hpp>
```
Baltoli, line 28:
```suggestion
#include <category/vm/evm/delegation.hpp>
#include <category/vm/evm/traits.hpp>
```

category/execution/ethereum/test/test_validation.cpp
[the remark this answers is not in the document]
Baltoli: I'd just make it an EVM traits test for now. Ellis will be removing Homestead imminently so all this will go away soon anyway.

category/execution/ethereum/validate_block.cpp
```diff
@@ -75,6 +76,16 @@ bytes32_t compute_ommers_hash(std::vector<BlockHeader> const &ommers)
 template <Traits traits>
 Result<void> static_validate_header(BlockHeader const &header)
 {
+    if constexpr (is_evm_trait_v<traits>) {
```
Baltoli, line 85: There's a subtle way in which this introduces a bug that doesn't really matter - if for some reason we were trying to run non-mainnet Ethereum blocks with Homestead block numbers, this check would incorrectly get fired.

However given that a) we don't support non-mainnet Ethereum chains, and b) @eclark0426 is removing support for this revision soon anyway, just stick a comment in about EVMC_HOMESTEAD so a grep picks it up.

category/execution/monad/chain/monad_chain.cpp
```diff
@@ -31,6 +31,7 @@
 #include <category/execution/monad/reserve_balance.hpp>
 #include <category/execution/monad/system_sender.hpp>
 #include <category/execution/monad/validate_monad_transaction.hpp>
+#include <category/vm/evm/switch_traits.hpp>
```
Baltoli: Why do you need this?

category/rpc/monad_executor.cpp
Baltoli: Yeah this wasn't ever hitting the virtual dispatch with Monad-specific logic. Worth understanding properly though

cmd/monad/runloop_ethereum.cpp
```diff
@@ -89,7 +89,6 @@ Result<void> process_ethereum_block(
     auto const block_begin = std::chrono::steady_clock::now();
 
     // Block input validation
-    BOOST_OUTCOME_TRY(chain.static_validate_header(block.header));
```
Baltoli, line 92: How come this gets deleted with no replacement?

cmd/monad/runloop_monad_ethblocks.cpp
```diff
@@ -125,7 +125,6 @@ Result<void> process_monad_block(
     auto const block_begin = std::chrono::steady_clock::now();
 
     // Block input validation
-    BOOST_OUTCOME_TRY(chain.static_validate_header(block.header));
```
Baltoli, line 128: Same here

cmd/monad/runloop_ethereum.cpp
goodlyrottenapple: The logic from `EthereumMainnet{}.static_validate_header` has been merged into the trait version of `static_validate_header` which gets called via `static_validate_block<traits>` below.

category/rpc/monad_executor.cpp
goodlyrottenapple: Ok, I've had a look at this and it is indeed related to the reserve balance validation. The monad version of validate_transaction ignores the `TransactionError::InsufficientBalance` returned by `validate_ethereum_transaction` and will return success instead. 

This causes divergent behaviour in one of the rpc tests which tests for insufficient balance. Currently, the test fails at transaction validation because the ethereum validation check throws an insufficient balance. However if we switch to the monad version, validation will actually pass. Then, only later at execution time, does the check for balance fail and `EVMC_INSUFFICIENT_BALANCE` is returned instead of the `EVMC_REJECTED` code. 

Not really sure what the intended behaviour is in the RPC client? It would seem to me that either response is correct, but there might be some behaviour somewhere that depends on `EVMC_REJECTED` being returned together with the message `"insufficient balance"`, as is currently the case?

dhil: Currently the desired behaviours are different for the various simulation calls. I think `eth_call` is expected to use ethereum-style validation, whereas `debug_trace{Block,Transaction}` is expected to use monad-style validation.

Changing either behaviour is likely to break some downstream tools, so I think, at least for now, the plan is to continue to keep things as-is.

category/execution/monad/validate_monad_transaction.cpp
```diff
@@ -28,21 +29,21 @@
 
 MONAD_NAMESPACE_BEGIN
 
-Result<void> validate_monad_transaction(
-    monad_revision const monad_rev, evmc_revision const rev,
+template <Traits traits>
```
dhil, line 32: I suppose morally this should be constrained by `is_monad_trait_v`, as we invoke `traits::monad_rev()` which is part of the subinterface `MonadTraits`.

category/execution/monad/validate_monad_transaction.cpp
```diff
@@ -51,16 +52,14 @@ Result<void> validate_monad_transaction(
             return MonadTransactionError::SystemTransactionSenderIsAuthority;
         }
     }
-    else if (monad_rev >= MONAD_ZERO) {
-        return res;
-    }
     else {
-        MONAD_ABORT("invalid revision");
```
dhil, line 58: This one is a bit interesting. Supposedly it was meant to guard for an ill-formed `monad_rev` instance. First, is it ever the case that we construct an instance of `monad_rev` based on user input? I think we construct them only based on block numbers and timestamps. 

Second, even if someone manage to forge a bad instance, e.g. using a negative number, then I believe it would still be implementation-defined whether this abort would fire, as I believe the compiler is free to pick whichever integral implementation it likes for a plain enum.

What I am trying to ask is whether this `MONAD_ABORT` was always useless, and hence its removal is justified or is it the case that we should still do some revision sanitation in refactored code?

category/execution/monad/validate_monad_transaction.cpp
goodlyrottenapple: morally yes, but i don't think this is possible, since we define `validate_monad_transaction` across two TUs both here and in `category/execution/ethereum/validate_transaction.hpp`. 
This should not be an issue in practice here, since the compiler will throw an error if we try to instantiate this version with an evm trait. If we are worried about the converse, we could add a `static_assert(is_evm_trait_v<traits>);` to the `category/execution/ethereum/validate_transaction.hpp` version.

category/execution/monad/validate_monad_transaction.cpp
goodlyrottenapple: Given all the calls to the current `validate_monad_transaction` (AFAICS) are guarded either via `traits::monad_rev()` or `get_monad_revision`, it would seem that it is safe to push the responsibility of revision validation onto these functions.

category/execution/monad/validate_monad_transaction.cpp
dhil: I think adding the static assert in the ethereum one is a freebie, that has the potential to save some headache.

category/execution/monad/validate_monad_transaction.cpp
dhil: I think so too. I was confused about why it used to be there, as I think it might not have been doing anything useful.

category/execution/monad/validate_monad_transaction.cpp
goodlyrottenapple: Added. I think `scripts/check-trait-instantiations.py` should also catch any potential overlaps, but the static_assert costs us nothing and as you said, might save some headache...

dhil (dismissed):

category/execution/ethereum/validate_transaction_error.cpp
```diff
@@ -0,0 +1,69 @@
+// Copyright (C) 2025 Category Labs, Inc.
```
cmcl:
```suggestion
// Copyright (C) 2025-26 Category Labs, Inc.
```

cmcl (dismissed):

category/execution/ethereum/validate_transaction_error.cpp
goodlyrottenapple: ah good spot thanks

PR #2051, opened 2026-02-04
Fix LLVM build and blockchain tests after MIP3

brett-monad:
Update LLVM emitter and runtime to work with MIP3 memory changes:
- Add explicit template instantiations for LLVM runtime functions
- Fix memory expansion traits usage in LLVM emitter
- Update blockchain test assertions for new memory behavior
- Correct template instantiation calls in llvm.cpp

### DIFF
--- a/category/vm/llvm/emitter.hpp
+++ b/category/vm/llvm/emitter.hpp
@@ -36,9 +36,10 @@
 namespace monad::vm::runtime
 {
 
+    template <Traits traits>
     inline void context_expand_memory(Context *ctx, Bin<29> min_size)
     {
-        ctx->expand_memory(min_size);
+        ctx->expand_memory<traits>(min_size);
     }
 
     inline void llvm_runtime_debug(
@@ -1473,7 +1474,7 @@ namespace monad::vm::llvm
             if (context_expand_memory_f == nullptr) {
                 auto f = declare_symbol(
                     "context_expand_memory",
-                    (void *)(context_expand_memory),
+                    (void *)(context_expand_memory<traits>),
                     llvm.void_ty,
                     {
                         llvm.ptr_ty(context_ty),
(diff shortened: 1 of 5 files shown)


laurenthuberdeau (approved):

category/vm/llvm/llvm.cpp
```diff
@@ -105,13 +106,17 @@ namespace monad::vm::llvm
         return ptr;
     }
 
+    template <Traits traits>
+    static evmc::Result copy_to_evmc_result_impl(runtime::Context &ctx)
+    {
+        auto result = ctx.copy_to_evmc_result<traits>();
+        return result;
```
Baltoli, line 113:
```suggestion
        return ctx.copy_to_evmc_result<traits>();
```

category/vm/llvm/emitter.hpp
```diff
@@ -1514,7 +1515,7 @@ namespace monad::vm::llvm
             llvm.insert_at(entry);
 
             uint256_t const max_offset =
-                (uint64_t{1} << 29) - static_cast<uint32_t>(width);
+                (uint64_t{1} << 31) - static_cast<uint32_t>(width);
```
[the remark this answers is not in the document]
Baltoli: Is copilot right here? The `Bin` sizes did change in MIP3, so convince yourself of correctness here

category/vm/llvm/llvm.cpp
```diff
@@ -120,7 +125,9 @@ namespace monad::vm::llvm
 
         monad::vm::llvm::execute(*llvm, ctx, evm_stack);
 
-        return ctx.copy_to_evmc_result();
+        SWITCH_EVM_TRAITS(copy_to_evmc_result_impl, ctx);
+
+        std::unreachable();
```
Baltoli, line 130: This should just be a `MONAD_ASSERT(false);` - we want to guarantee that an error fires if we hit this

PR #2056, opened 2026-02-09
More premissive licence check allowing "Copyright (C) 2025-26 ..."

goodlyrottenapple:
This version of the licence check was produced by Claude version `claude-sonnet-4-5-20250929`

The PR also adds missing licence headers in `.S` files.

### DIFF
--- a/.github/workflows/license.sh
+++ b/.github/workflows/license.sh
@@ -1,43 +1,72 @@
-C_LICENSE_HEADER="\
-// Copyright (C) 2025 Category Labs, Inc.
-//
-// This program is free software: you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
-//
-// This program is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-//
-// You should have received a copy of the GNU General Public License
-// along with this program.  If not, see <http://www.gnu.org/licenses/>."
+#!/bin/bash
 
-PYTHON_LICENSE_HEADER="\
-# Copyright (C) 2025 Category Labs, Inc.
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>."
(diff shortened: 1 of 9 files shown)


.github/workflows/license.sh
```diff
@@ -1,43 +1,59 @@
...
+    local patterns=("$@")
+    local line_num=1
+
+    for pattern in "${patterns[@]}"; do
+        if ! sed -n "${line_num}p" "$file" | grep -qE "$pattern"; then
+            return 1
+        fi
+        line_num=$((line_num + 1))
+    done
+
```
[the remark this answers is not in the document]
Baltoli: I think this is correct

.github/workflows/license.sh
```diff
@@ -1,43 +1,59 @@
...
+    shift
+    local patterns=("$@")
+    local line_num=1
+
+    for pattern in "${patterns[@]}"; do
+        if ! sed -n "${line_num}p" "$file" | grep -qE "$pattern"; then
+            return 1
+        fi
+        line_num=$((line_num + 1))
+    done
```
Baltoli: Is there a phrasing of this that avoids one `sed` / `grep` per line? It makes the license check take 10x longer (but only 1.4s -> 14s so I probably wouldn't block the PR on it if it's awkward / confusing to fix)

goodlyrottenapple: Got it down to 5s

.github/workflows/license.sh
```diff
@@ -1,43 +1,70 @@
...
+# Function to check if a file has valid license header
+check_license() {
+    local file="$1"
+    shift
+    local patterns=("$@")
+    local num_lines=${#patterns[@]}
+    local total_lines=$(wc -l < "$file")
+
+    # Read the first num_lines (+1 for potential blank line) into an array
+    mapfile -t lines < <(head -n "$((num_lines + 1))" "$file")
```
dhil: I don't think the `wc -l` is necessary here. Similarly, I think the `head -n` is unnecessary. You can use `mapfile -n N` to read at most `N` lines, e.g.
```suggestion

    # Read the first num_lines (+1 for potential blank line) into an array
    mapfile -n $((num_lines + 1)) -t lines < "$file"
```

dhil: With this change, plus the below change, the execution time is sub 2s on my machine.

.github/workflows/license.sh
```diff
@@ -1,43 +1,70 @@
...
+            return 1
+        fi
+    done
+
+    # If file has content beyond the header, check for blank line
+    if [ "$total_lines" -gt "$num_lines" ]; then
+        if [[ "${lines[$num_lines]}" != "" ]]; then
+            return 1
+        fi
+    fi
```
dhil: With the above change this can be simplified to:
```suggestion
    if [[ "${lines[$num_lines]}" != "" ]]; then
            return 1
    fi
```

.github/workflows/license.sh
dhil: I think we can probably shave off the remaining overhead by using `diff -I="Copyright (C) 2025[[-26]] Category Labs, Inc."`, which should ignore the difference `2025` and `2025-26`.

goodlyrottenapple: Thanks for the suggestions, in the end I changed the logic to only use regex for the first line and direct string comparison for the rest of the header. We are down to 2s in CI and sub second on my local machine

dhil (approved):

PR #2057, opened 2026-02-09
Implement block overrides for `eth_simulatev1`

Baltoli:
This is preparatory work that isn't live yet; these overrides are required for `eth_simulatev1` support. Rather than overriding the state of particular accounts, they apply to block-level metadata. Additionally extract the state and block override code into a new header / TU pair to facilitate unit tests.

Geth spec for these overrides: https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#block-overrides. Note that this PR doesn't implement overriding withdrawals (as we don't have them in Monad, and it would be a lot of extra complexity), but does allow overriding blob base fees as that's very little extra code.

Developed with Claude Opus 4.6; this is mechanical code with a clear API and existing examples of similar work in the repo already.

### DIFF
--- a/category/rpc/overrides.cpp
+++ b/category/rpc/overrides.cpp
@@ -1,0 +1,245 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


category/rpc/overrides.cpp
```diff
@@ -0,0 +1,246 @@
...
+#include <category/rpc/overrides.hpp>
+
+#include <intx/intx.hpp>
+
+#include <cstdint>
+#include <cstring>
+
+using namespace monad;
+
+using StateOverrideObj = monad_state_override::monad_state_override_object;
```
dhil: Is it used only once use now? If so, then I suggest we simply inline it below.

PR #2060, opened 2026-02-10
[rpc] `eth_simulateV1` 

Baltoli:
Depends on:
- https://github.com/category-labs/monad/pull/2057
- https://github.com/category-labs/monad/pull/1926

(commits should be rebased out when those PRs are merged)

### DIFF
--- a/category/rpc/monad_executor.cpp
+++ b/category/rpc/monad_executor.cpp
@@ -727,10 +727,6 @@ namespace
         uint64_t const gas_limit, size_t const max_calls,
         bool emit_native_transfer_logs)
     {
-        // TODO(dhil): Pass `emit_native_transfer_logs` through to the EvmcHost
-        // instantiation.
-        (void)emit_native_transfer_logs;
-
         // TODO(dhil): Decide on the default timestamp increment.
         static constexpr uint64_t DEFAULT_TIMESTAMP_INCREMENT = 1;
 
@@ -865,7 +861,8 @@ namespace
                         block_metrics,
                         call_tracers,
                         state_tracers,
-                        chain_context));
+                        chain_context,
+                        emit_native_transfer_logs));
 
                 // NOTE(dhil): Synthetic blocks are free, so we don't update
                 // `gas_consumed_so_far`.
@@ -982,7 +979,8 @@ namespace
                     block_metrics,
                     call_tracers,
                     state_tracers,
-                    chain_context));
+                    chain_context,
+                    emit_native_transfer_logs));
 
             // Receipts have cumulative gas_used (YP eq. 22), so
             // the last receipt's value is the total for the block.
--- a/category/rpc/monad_executor.cpp
+++ b/category/rpc/monad_executor.cpp
@@ -24,6 +24,7 @@
 #include <category/core/fiber/fiber_thread_pool.hpp>
 #include <category/core/fiber/priority_pool.hpp>
 #include <category/core/hex.hpp>
+#include <category/core/int.hpp>
 #include <category/core/keccak.hpp>
 #include <category/core/likely.h>
 #include <category/core/log.hpp>
@@ -41,6 +42,7 @@
 #include <category/execution/ethereum/core/rlp/bytes_rlp.hpp>
 #include <category/execution/ethereum/core/rlp/transaction_rlp.hpp>
 #include <category/execution/ethereum/core/transaction.hpp>
+#include <category/execution/ethereum/core/withdrawal.hpp>
 #include <category/execution/ethereum/db/trie_rodb.hpp>
 #include <category/execution/ethereum/db/util.hpp>
 #include <category/execution/ethereum/evmc_host.hpp>
@@ -56,6 +58,7 @@
 #include <category/execution/ethereum/trace/tracer_config.h>
 #include <category/execution/ethereum/tx_context.hpp>
 #include <category/execution/ethereum/types/incarnation.hpp>
+#include <category/execution/ethereum/validate_block.hpp>
 #include <category/execution/ethereum/validate_transaction.hpp>
 #include <category/execution/ethereum/validate_transaction_error.hpp>
 #include <category/execution/monad/chain/monad_chain.hpp>
@@ -65,6 +68,8 @@
 #include <category/execution/monad/reserve_balance.hpp>
 #include <category/mpt/db.hpp>
 #include <category/mpt/ondisk_db_config.hpp>
+#include <category/rpc/chain_context_buffer.hpp>
+#include <category/rpc/eth_simulate_block_hash_buffer.hpp>
 #include <category/rpc/lazy_block_hash.hpp>
 #include <category/vm/evm/switch_traits.hpp>
 #include <category/vm/evm/traits.hpp>
@@ -74,7 +79,6 @@
 #include <boost/outcome/try.hpp>
 #include <boost/scope_exit.hpp>
 
-#include <array>
 #include <atomic>
 #include <chrono>
 #include <cstdint>
@@ -82,6 +86,7 @@
 #include <cstring>
 #include <filesystem>
 #include <format>
+#include <limits>
 #include <memory>
 #include <optional>
 #include <span>
@@ -106,6 +111,8 @@ namespace
     char const *const UNEXPECTED_EXCEPTION_ERR_MSG = "unexpected error";
     char const *const EXCEED_QUEUE_SIZE_ERR_MSG =
         "failure to submit eth_call to thread pool: queue size exceeded";
+    char const *const ETH_SIMULATE_EXCEED_QUEUE_SIZE_ERR_MSG =
+        "failure to submit eth_simulateV1 to thread pool: queue size exceeded";
     char const *const TIMEOUT_ERR_MSG =
         "failure to execute eth_call: queuing time exceeded timeout threshold";
     char const *const PRESTATE_TRACER_SUPPORT_ERR_MSG =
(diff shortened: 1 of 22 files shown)


Baltoli:
@claude Please do a detailed re-review of this PR, taking into account resolved comments from previous rounds

Baltoli:
@claude Please do a final detailed re-review of this PR, taking into account resolved comments from previous reviews and responses

Baltoli:
_From Codex 5.5._

# Codex Adversarial Review — PR #2060

## Findings

### 🔴 [critical] Simulated block hashes are not chained into later blocks

**Location:** `category/rpc/monad_executor.cpp:759-775`

Each simulated header is built without a `parent_hash`, and the same `LazyBlockHash` anchored at the original `block_number` is reused for every simulated block. A later simulated transaction that calls `BLOCKHASH` for an earlier simulated block can fall through to `LazyBlockHash::get` for a block number newer than its anchor, hitting an assertion. Even when it does not crash, `parentHash` and block-hash history for simulated blocks is wrong.

**Recommendation:** Maintain a per-simulation block-hash chain — compute each simulated block hash, set it as the next `header.parent_hash`, and serve simulated hashes to `BLOCKHASH` / EIP-2935 before falling back to the DB. Add a multi-block `BLOCKHASH` test.

---

### 🟠 [high] First simulated blocks ignore real parent reserve context

**Location:** `category/rpc/monad_executor.cpp:687-689`

`eth_simulate` starts `ChainContextBuffer` empty, so the first simulated block sees empty parent and grandparent sender / authority sets. Monad reserve checks explicitly reject reserve dipping when the sender appeared in the parent or grandparent; this path can therefore report success for a simulation that production block execution would reject.

**Recommendation:** Seed the context buffer from the real base block and its parent before simulating, mirroring the trace path's DB recovery. If the needed block ids are unavailable, extend the API to pass them or reject the simulation instead of using empty context.

---

### 🟡 [medium] `baseFeePerGas` is emitted as decimal with a hex prefix

**Location:** `category/rpc/monad_executor.cpp:472-473`

`store_output_header` formats `baseFeePerGas` with `intx::to_string(value)` but prefixes it with `0x`. `intx` defaults to decimal, so a base fee like `1000000000` is returned as `0x1000000000`, which clients parse as a much larger hex value.

**Recommendation:** Format `uint256` JSON quantities with base 16, e.g. `intx::to_string(value, 16)`, and add coverage for base fees above `9`.

---

### 🟡 [medium] Log indexes are duplicated within a transaction

**Location:** `category/rpc/monad_executor.cpp:607-620`

The log serialization loop uses `tx_idx` as `logIndex` for every log in that transaction. Any transaction that emits multiple logs — including the new native-transfer logging path — returns duplicate `logIndex` values and non-cumulative indexes across the block, breaking clients that identify logs by `blockHash` / `logIndex`.

**Recommendation:** Track a cumulative block log index and increment it for every emitted log. Add a test with multiple logs from one transaction and logs across multiple transactions.

---

category/rpc/monad_executor.cpp
```diff
@@ -538,6 +540,631 @@ namespace
...
+                // override.
+                if (gap > 1) {
+                    previous_header.timestamp +=
+                        gap * default_timestamp_increment;
+                }
+
+                MONAD_ASSERT_THROW(
+                    previous_header.timestamp <= *bo.time,
+                    "block timestamps must be strictly increasing");
+                previous_header.timestamp = *bo.time;
```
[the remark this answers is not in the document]
dhil: Fixed.

category/rpc/monad_executor.cpp
```diff
@@ -538,6 +540,631 @@ namespace
...
+                // timestamps before validating the possible block timestamp
+                // override.
+                if (gap > 1) {
+                    previous_header.timestamp +=
+                        gap * default_timestamp_increment;
+                }
+
+                MONAD_ASSERT_THROW(
+                    previous_header.timestamp <= *bo.time,
+                    "block timestamps must be strictly increasing");
```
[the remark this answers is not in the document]
dhil: Fixed.

category/rpc/monad_executor.cpp
```diff
@@ -1491,3 +2307,119 @@ void monad_executor_run_transactions(
...
+    uint8_t const *const rlp_block_id, size_t rlp_block_id_len,
+    uint8_t const *const rlp_parent_block_id,
+    size_t const rlp_parent_block_id_len,
+    uint8_t const *const rlp_grandparent_block_id,
+    size_t const rlp_grandparent_block_id_len,
+    struct monad_state_override_vec const *const state_overrides,
+    struct monad_block_override_vec const *const block_overrides,
+    bool emit_native_transfer_logs,
+    void (*complete)(monad_executor_result *, void *user), void *user)
+{
```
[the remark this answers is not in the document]
dhil: Good suggestion. Fixed.

category/rpc/overrides.hpp
```diff
@@ -52,4 +66,17 @@ struct monad_block_override
     std::optional<monad::bytes32_t> prev_randao{std::nullopt};
     std::optional<monad::uint256_t> base_fee_per_gas{std::nullopt};
     std::optional<monad::uint256_t> blob_base_fee{std::nullopt};
```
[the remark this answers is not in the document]
dhil: Fixed.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +642,294 @@ pub async fn eth_trace_block_or_transaction(
...
+    eth_call_executor: &EthCallExecutor,
+    overrides: &[(&BlockOverride, &StateOverrideSet)],
+) -> SimulateResult {
+    assert_eq!(calls.len(), overrides.len());
+
+    let mut rlp_encoded_senders = vec![];
+    senders.encode(&mut rlp_encoded_senders);
+
+    let mut rlp_encoded_txns = vec![];
+    calls.encode(&mut rlp_encoded_txns);
```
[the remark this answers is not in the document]
dhil: I don't think this suggestion is correct. The templated invocation in `monad_executor_eth_simulate_submit` is parametrized by `rlp::decode_transaction` which internally dispatches to either `rlp::decode_transaction_legacy` or `rlp::decode_transaction_eip2718`. I've added a test that submits legacy transactions and checks they execute successfully.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +642,294 @@ pub async fn eth_trace_block_or_transaction(
...
+    block_number: u64,
+    block_id: Option<[u8; 32]>,
+    parent_id: Option<[u8; 32]>,
+    grandparent_id: Option<[u8; 32]>,
+    emit_native_transfer_logs: bool,
+    eth_call_executor: &EthCallExecutor,
+    overrides: &[(&BlockOverride, &StateOverrideSet)],
+) -> SimulateResult {
+    assert_eq!(calls.len(), overrides.len());
+
```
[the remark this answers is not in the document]
dhil: Fixed.

dhil:
@claude Please do another detailed re-review of this PR, taking into account unresolved and resolved comments from previous reviews and responses. Please consider the PR commit by commit; we want each commit be sound and coherent.

dhil:
P3 Mutually-exclusive state-override fields not enforced. Wont fix.

dhil:
@claude please review this PR again. This time focus on identifying opportunities for data races, deadlocks, livelocks, or other parallel and concurrency-related bugs introduced by this PR.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -100,6 +100,55 @@ impl Drop for EthCallExecutor {
     }
 }
 
+struct MonadExecutorResult {
+    c_handle: *mut ffi::monad_executor_result,
```
andr-dev: I would recommend using the https://doc.rust-lang.org/beta/std/ptr/struct.NonNull.html type here. The `from_c_handle` fn would then produce an `Option<Self>` allowing the caller to choose how to handle the error.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -100,6 +100,55 @@ impl Drop for EthCallExecutor {
...
+
+    fn status_code(&self) -> i32 {
+        unsafe { (*self.c_handle).status_code }
+    }
+
+    fn encoded_trace(&self) -> Vec<u8> {
+        let output_data_len = unsafe { (*self.c_handle).encoded_trace_len };
+        let output_data = if output_data_len != 0 {
+            unsafe { std::slice::from_raw_parts((*self.c_handle).encoded_trace, output_data_len) }
+                .to_vec()
```
andr-dev: I don't remember there being any issues with `std::slice::from_raw_parts` taking a length of 0, is this just an optimization?

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -100,6 +100,55 @@ impl Drop for EthCallExecutor {
...
+        output_data
+    }
+
+    fn message(&self) -> String {
+        let cstr_msg = unsafe { CStr::from_ptr((*self.c_handle).message.cast()) };
+        let message = match cstr_msg.to_str() {
+            Ok(str) => String::from(str),
+            Err(_) => String::from("execution error: message invalid utf-8"),
+        };
+        message
```
andr-dev:
```rust
cstr_msg.to_str().map(String::from).unwrap_or_else(|| String::from("execution error: message invalid utf-8"))
```

You might even be able to do
```rust
String::from(
  cstr_msg.to_str().unwrap_or("execution error: message invalid utf-8")
)
```

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -100,6 +100,55 @@ impl Drop for EthCallExecutor {
...
+    }
+}
+
+impl Drop for MonadExecutorResult {
+    fn drop(&mut self) {
+        unsafe {
+            ffi::monad_executor_result_release(self.c_handle);
+        }
+    }
+}
```
andr-dev: I would avoid implementing traits like `Into` and `From` for an FFI type because you usually want to be very explicit with these structs to avoid using them incorrectly.

Here, it is even more important to do so because `MonadExecutorResult` is a bit of an anti-pattern in that it is responsible for releasing the FFI object when it didn't allocate it. In the context of the ethcall interface it can't be avoided because of the send/recv pattern so an explicit call to `MonadExecutorResult::from_c_handle` on the sent `ffi::monad_executor_result` makes it clearer this conversion is happening.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +691,328 @@ pub async fn eth_trace_block_or_transaction(
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub prev_randao: Option<B256>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub base_fee_per_gas: Option<U256>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub withdrawals: Vec<Withdrawal>,
+}
+
+struct CStateOverrideVec {
+    c_handle: *mut ffi::monad_state_override_vec,
```
andr-dev: Can use `NonNull` for the types down here as well. I'd actually be in favor of unwrapping on the result of these `NonNull::new(...)` calls to panic if any of the `_create` fns produce a null pointer.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +691,328 @@ pub async fn eth_trace_block_or_transaction(
...
+
+impl CStateOverrideVec {
+    fn new(size: usize) -> Self {
+        let c_handle = unsafe { ffi::monad_state_override_vec_create(size) };
+        Self { c_handle }
+    }
+
+    fn as_mut_ptr(&self) -> *mut ffi::monad_state_override_vec {
+        self.c_handle
+    }
```
andr-dev: Instead of using this exposed pointer object below, can we add helpers that wrap the functionality? For example:
```rust
fn set_override_balance(&mut self, at: usize, addr: ABC, balance: DEF) {
  let balance_bytes = balance.to_be_bytes_vec();

  unsafe {
    ffi::set_override_balance_at(
      self.c_handle.as_mut_ptr(),
      at,
      addr.as_ptr(),
      addr.len(),
      balance_bytes.as_ptr(),
      balance_bytes.len(),
    );
  }
}
```
We'll probably still want this `as_mut_ptr` for the `monad_executor_eth_simulate_submit` fn because `EthCallExecutor` doesn't hide it away, but I can resolve that in a later change. Let's also make `as_mut_ptr` take a `&mut self`.

kkuehlz (dismissed):
simulate + related DS

rust/crates/monad-ethcall/src/lib.rs
dhil: According to the docs the pointer must be non-null (even if length is zero) https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#ffi-handling-null-pointers.

rust/crates/monad-ethcall/src/lib.rs
dhil: We can wrap everything -- but it feels like over-abstracting to me, because we use these utilities in one local-only place.

I agree that the `&mut self` annotation is morally correct. I'll add that.

rust/crates/monad-ethcall/src/lib.rs
andr-dev: If so, can we check the nullness of `self.c_handle.as_ptr().encoded_trace` itself?

rust/crates/monad-ethcall/src/lib.rs
andr-dev: Agreed on it only being used once, I prefer wrapping it more so to keep the main function clean. It's much easier to reason about once you're convinced the ffi is correct separately and then you only see clean function calls. If you prefer, we can keep it as is and I can wrap it in a subsequent change.

rust/crates/monad-ethcall/src/lib.rs
dhil: Good point!

rust/crates/monad-ethcall/src/lib.rs
dhil: I've implemented this change in the latest commit.

dhil:
@Baltoli are you happy with the `ChainContextBuffer` TU? I had to introduce a concept to preserve the `requires` clause, because clang fails to resolve `K` in the out-of-line definition otherwise. An alternative would be to add the `age < K` as a static assertion inside the body of `get`.

Baltoli:
> are you happy with the ChainContextBuffer TU?

yes, looks good to me!

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -100,6 +101,57 @@ impl Drop for EthCallExecutor {
...
+            Some(
+                unsafe {
+                    std::slice::from_raw_parts(
+                        (*self.c_handle.as_ptr()).encoded_trace,
+                        output_data_len,
+                    )
+                }
+                .to_vec(),
+            )
+        }
```
andr-dev: As is written now, the `if output_data_len == 0` branch is always hit before we check to see if `encoded_trace` is null. In the null case, I think `output_data_len` is always set to 0 so this method will always produce `Some(x)`. I think we can change this to:
```rust
fn encoded_trace(&self) -> Option<Box<[u8]>> {
  let this = unsafe { self.c_handle.as_ref() };

  if this.encoded_trace.is_null() {
    return None;
  }

  Some(Box::from(unsafe {
    std::slice::from_raw_parts(this.encoded_trace, this.encoded_trace_len)
  }))
}
```

I also changed the return value here to `Box<[u8]>`. This stylistic choice is a bit more polarizing since most people like using `Vec<u8>` everywhere but I really like to distinguish between growable containers and heap allocated byte arrays. It's similar to `std::vector<uint8_t>` vs `std::unique_ptr<std::byte[]>` or `std::unique_ptr<uint8_t[]>`. I would also change `SuccessSimulateResult` to produce `Box<[u8]>` as well.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +694,423 @@ pub async fn eth_trace_block_or_transaction(
...
+            ffi::monad_state_override_vec_destroy(self.c_handle.as_ptr());
+        }
+    }
+}
+struct CBlockOverrideVec {
+    c_handle: NonNull<ffi::monad_block_override_vec>,
+}
+
+impl CBlockOverrideVec {
+    fn new(size: usize) -> Option<Self> {
```
andr-dev: nit: to match rust vec naming more closely, can we call this `with_capacity`? Same for `CStateOverrideVec::new`. Apologies for not mentioning this in my earlier review.

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +694,423 @@ pub async fn eth_trace_block_or_transaction(
...
+            warn!("failed to create state override vector");
+
+            return SimulateResult::Failure(FailureSimulateResult {
+                error_code: EthCallResult::OtherError,
+                message: "internal eth_simulate_v1 error: failed to create state override vector"
+                    .to_string(),
+                data: None,
+            });
+        }
+    };
```
andr-dev, line 958: super nit: If you want to make these blocks even shorter, you can do something like:
```rust
let Some(mut state_overrides) = CStateOverrideVec::new(calls.len()) else {
  warn!("failed to create state override vector");

  return SimulateResult::Failure(FailureSimulateResult {
    error_code: EthCallResult::OtherError,
    message: "internal eth_simulate_v1 error: failed to create state override vector".to_string(),
    data: None,
  });
};
```

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +694,423 @@ pub async fn eth_trace_block_or_transaction(
...
+                }
+                Some(StorageOverride::StateDiff(override_state_diff)) => {
+                    for (k, v) in override_state_diff {
+                        state_overrides.set_state_diff_at(i, addr, k, v)
+                    }
+                }
+                None => {}
+            }
+        }
+
```
andr-dev, line 1008: This section looks so clean now with the helpers :)

For future proofing, I like to do this pattern where you decompose the struct to show that every field has been used, i.e.
```rust
let BlockOverride {
  number,
  time,
  gas_limit,
  fee_recipient,
  prev_randao,
  base_fee_per_gas
} = block_override;

if let Some(number) = number {
  ...
}

...
```
This way, if we ever add a field to `BlockOverride`, it'll fail to compile unless you add the field to that destructuring block. If any fields aren't used, clippy warnings. Same can be done for `obj` in the inner state_override loop.
```rust
for (
  addr,
  StateOverrideObject { balance, nonce, code, storage_override}
) in state_override.iter() {
  ...
}
```
Might also make sense to rename `state_override` to `state_overrides` since it's a `StateOverrideSet`, contains multiple state overrides objects.

rust/crates/monad-ethcall/src/lib.rs
```diff
@@ -624,6 +694,423 @@ pub async fn eth_trace_block_or_transaction(
...
+        None => {
+            warn!("callback from eth_simulate_v1 failed: result pointer is null");
+
+            return SimulateResult::Failure(FailureSimulateResult {
+                error_code: EthCallResult::OtherError,
+                message: "internal eth_simulate_v1 error: result pointer is null".to_string(),
+                data: None,
+            });
+        }
+    };
```
andr-dev: Can we break this into two separate statements?
```rust
let result_raw = match recv.await {
  ...
};

let Some(result) = MonadExecutorResult::from_c_handle(result_raw) else {
  ...
};
```

rust/crates/monad-ethcall/src/lib.rs
dhil: I intended to use `Option::None` to indicate failure. The execution client can return `output_data_len = 0` with `encoded_trace = nullptr`, because we cannot allocate and serialise a zero-sized container. That's why I structured it the other way.

I think, the problem with your approach is that we conflate the cases of "output is zero-sized" and "allocation failed in the execution client".

andr-dev: Ahh I see, ok in that case we can do:
```rust
fn encoded_trace(&self) -> Result<Option<Box<[u8]>>, ()> {
  let this = unsafe { self.c_handle.as_ref() };

  if this.encoded_trace_len == 0 {
    return Ok(None);
  }

  if this.encoded_trace.is_null() {
    // Non-zero len with nullptr indicates an allocation failure.
    return Err(());
  }

  Ok(Some(Box::from(unsafe {
    std::slice::from_raw_parts(this.encoded_trace, this.encoded_trace_len)
  })))
}
```
It's not super idiomatic to have a Result with `()` as error type but I think it's better than producing a None in an error case. `SuccessSimulateResult` would then have an `Option<Box<[u8]>>`.

dhil: I decided to use `Result<Box<[u8]>, ()>` and allocate a box for the empty array in the `encoded_trace_len == 0` case. I don't think there is any point in optimising for this case, as it ought to not occur in practice.

andr-dev (approved):
rust looks amazing, nice work!

kkuehlz (approved):

Chen-Yifan (approved):

PR #2061, opened 2026-02-10
Remove Homestead support

eclark0426:


category/vm/compiler/ir/basic_blocks.hpp
Baltoli: This whole function can have its trait parameter removed now; it was only the selfdestruct case that we needed it for.

category/vm/runtime/selfdestruct.cpp
```diff
@@ -39,29 +41,26 @@ namespace monad::vm::runtime
             auto const access_status =
                 ctx->host->access_account(ctx->context, &address);
             if (access_status == EVMC_ACCESS_COLD) {
-                // The minimum gas for SELFDESTRUCT is 0, so we have to account
-                // for the extra 100 gas for accessing a warm account here.
+                // +100 for the warm account access cost.
                 ctx->deduct_gas(traits::cold_account_cost() + 100);
```
Baltoli, line 44: Maybe this can be refactored now? Not sure if it's cleaner to fold the minimum gas for selfdestruct into the table or to leave it out

category/vm/evm/opcodes.hpp
Baltoli, line 1: Consider tightening the assertion in `previous_evm_revision`:
```cpp
static_assert(rev > constants::EARLIEST_SUPPORTED_EVM_FORK);
```

eclark0426: I had initially had this in the Frontier commit; James preferred having it just assert that the revision is greater than 0, as an assertion that the math is correct

Baltoli: Sure, fine by me

Baltoli (approved):

goodlyrottenapple (approved):

jhunsaker (approved):

PR #2064, opened 2026-02-11
Remove intx usage from 6 files

brett-monad:
Replace intx big-endian operations with std::byteswap and unaligned_load/store for primitive types. Remove unused intx includes where no intx APIs are used.

### DIFF
--- a/category/core/unaligned.hpp
+++ b/category/core/unaligned.hpp
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include <category/core/config.hpp>
+
 #include <algorithm>
 #include <array>
 #include <bit>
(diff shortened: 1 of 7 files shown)


category/core/unaligned.hpp
```diff
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include <category/core/config.hpp>
```
m-alvarez, line 18: Why is this include now necessary?

goodlyrottenapple: direct import of `MONAD_NAMESPACE_*` macros. Best to not rely on transitive dependencies to include it, I guess?

PR #2065, opened 2026-02-11
cached/incremental reserve balance checks: design2 

aa755:
Now that reserve balance checks can be done several times during a transaction using a precompile (previously the CHECKRESERVEBALANCE opcode), it is important to do it in an incremental/cached way, as suggested by Piotr. Before this PR, the check was done by [iterating over all changed accounts](https://github.com/category-labs/monad/blob/f5c1bec66612a9acc8f76a7c44681a231dd24f28/category/execution/monad/reserve_balance.cpp#L66). In this PR, the check is done by just checking whether the cached set of "accounts in reserve balance violation" is empty. This set is maintained in `ReserveBalance::failed_`. It is a private field of the `ReserveBalance` class, but the class provides hooks (methods) that are called on changes to parts of `State` relevant to reserve balance: balance and code of accounts, so that `ReserveBalance::failed_` can be updated. Additionally, `ReserveBalance::on_pop_reject()` is called on `State::pop_reject`.

## Balance changes
The method `ReserveBalance::update_violation_status` is the main workhorse which updates the violation status of the given account. It is roughly equivalent to the [loop body of the previous implementation](https://github.com/category-labs/monad/blob/f5c1bec66612a9acc8f76a7c44681a231dd24f28/category/execution/monad/reserve_balance.cpp#L67-L125): it computes the violation threshold if not already cached and then checks if the balance lower.

### `State::subtract_from_balance`
Here, we just add a call to `ReserveBalance::update_violation_status`.

### `State::add_to_balance`
We know that adding to an account's balance can only cause the account to go from being in violation to being not in violation, but not the other direction. Thus, we only call `ReserveBalance::update_violation_status` if the account was in violation.

## `State::set_code`
If the new code is a delegation marker, we recompute the violation threshold. Otherwise, [just like the previous implementation](https://github.com/category-labs/monad/blob/f5c1bec66612a9acc8f76a7c44681a231dd24f28/category/execution/monad/reserve_balance.cpp#L76), such accounts are exempt and thus we remove the account from the violation set.

## `State::pop_reject`

This happens when the changes to the current call frame are discarded, e.g. due to `REVERT`. `State` already maintains the set of addresses that were updated in this call frame. We recompute the violation status of all of them

## Correctness
We are assuming that the above methods are the *only* places where the balances of an account or whether the account is delegated or isSC changes until the point reserve balance checks can still happen.
These assumptions seem to be valid now but will need to be maintained in the future.
I will start doing Coq proofs of the functions added here once this is merged.

co-authored with codex-cli.

### DIFF
--- a/category/execution/ethereum/evm_test.cpp
+++ b/category/execution/ethereum/evm_test.cpp
@@ -53,6 +53,24 @@ using db_t = TrieDb;
 
 #define PUSH3(x) 0x62, (((x) >> 16) & 0xFF), (((x) >> 8) & 0xFF), ((x) & 0xFF)
 
+namespace
+{
+    template <Traits traits>
+    void init_rb_for_test(
+        State &state, EvmcHost<traits> &host, Address const &sender)
+    {
+        if constexpr (is_monad_trait_v<traits>) {
+            init_reserve_balance_context<traits>(
+                state,
+                sender,
+                host.tx_,
+                host.base_fee_per_gas_,
+                host.i_,
+                host.chain_ctx_);
+        }
+    }
+}
+
 TYPED_TEST(TraitsTest, create_with_insufficient)
 {
     InMemoryMachine machine;
@@ -102,6 +120,7 @@ TYPED_TEST(TraitsTest, create_with_insufficient)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
     auto const result = create<typename TestFixture::Trait>(&h, s, m);
 
     EXPECT_EQ(result.status_code, EVMC_INSUFFICIENT_BALANCE);
@@ -163,6 +182,7 @@ TYPED_TEST(TraitsTest, create_insufficient_balance_nonce_bump)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
 
     auto const result = create<typename TestFixture::Trait>(&h, s, m);
 
@@ -242,6 +262,7 @@ TYPED_TEST(TraitsTest, eip684_existing_code)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
     auto const result = create<typename TestFixture::Trait>(&h, s, m);
     EXPECT_EQ(result.status_code, EVMC_INVALID_INSTRUCTION);
 }
@@ -301,6 +322,7 @@ TYPED_TEST(TraitsTest, create_nonce_out_of_range)
     uint256_t const v{70'000'000};
     intx::be::store(m.value.bytes, v);
 
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
     auto const result = create<typename TestFixture::Trait>(&h, s, m);
 
     EXPECT_FALSE(s.account_exists(new_addr));
@@ -347,6 +369,7 @@ TYPED_TEST(TraitsTest, static_precompile_execution)
                  .account = {std::nullopt, Account{.balance = 15'000}}}}},
         Code{},
         BlockHeader{});
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{from});
 
     static constexpr char data[] = "hello world";
     static constexpr auto data_size = sizeof(data);
@@ -414,6 +437,7 @@ TYPED_TEST(TraitsTest, out_of_gas_static_precompile_execution)
                  .account = {std::nullopt, Account{.balance = 15'000}}}}},
         Code{},
         BlockHeader{});
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{from});
 
     static constexpr char data[] = "hello world";
     static constexpr auto data_size = sizeof(data);
@@ -512,6 +536,7 @@ TYPED_TEST(TraitsTest, create_op_max_initcode_size)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{from});
 
     // Initcode fits inside size limit
     if constexpr (
@@ -632,6 +657,7 @@ TYPED_TEST(TraitsTest, create2_op_max_initcode_size)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{from});
 
     // Initcode fits inside size limit
     if constexpr (
@@ -888,6 +914,7 @@ TYPED_TEST(TraitsTest, create_inside_delegated_call)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
 
     if constexpr (TestFixture::Trait::evm_rev() >= EVMC_PRAGUE) {
         auto const result = h.call(m);
@@ -1017,6 +1044,7 @@ TYPED_TEST(TraitsTest, create2_inside_delegated_call_via_delegatecall)
         base_fee,
         0,
         chain_ctx};
+    init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
 
     if constexpr (TestFixture::Trait::evm_rev() >= EVMC_PRAGUE) {
         auto const result = h.call(m);
@@ -1132,6 +1160,7 @@ TYPED_TEST(TraitsTest, nested_call_to_delegated_precompile)
             base_fee,
             0,
             chain_ctx};
+        init_rb_for_test<typename TestFixture::Trait>(s, h, Address{m.sender});
 
         auto const result = h.call(m);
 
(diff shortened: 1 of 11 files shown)


category/execution/monad/reserve_balance.cpp
```diff
@@ -51,94 +51,178 @@ unsigned monad_default_max_reserve_balance_mon(enum monad_revision)
...
+    tracking_enabled_ = true;
+    sender_ = sender;
+    sender_gas_fees_ = uint256_t{tx.gas_limit} *
+                       gas_price<traits>(tx, base_fee_per_gas.value_or(0));
+    sender_can_dip_ = sender_can_dip;
+    get_max_reserve_ = [](Address const &addr) {
+        return get_max_reserve<traits>(addr);
+    };
+    failed_.clear();
+    violation_thresholds_.clear();
```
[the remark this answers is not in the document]
aa755: done.

category/execution/ethereum/evm_test.cpp
```diff
@@ -90,8 +132,7 @@ TYPED_TEST(TraitsTest, create_with_insufficient)
     BlockHashBufferFinalized const block_hash_buffer;
     NoopCallTracer call_tracer;
     Transaction tx{};
-    auto const chain_ctx =
-        ChainContext<typename TestFixture::Trait>::debug_empty();
+    auto const chain_ctx = empty_chain_ctx<typename TestFixture::Trait>();
```
goodlyrottenapple: I think this reversal to empty_chain_ctx might not be intentional?

aa755: Great catch. I apologize for the sloppy conflict resolution with [65750a6a6](https://github.com/category-labs/monad/commit/65750a6a6). I will be more careful with git conflicts.

category/execution/ethereum/state3/state.cpp
```diff
@@ -337,13 +363,16 @@ void State::subtract_from_balance(
...
 }
 
 void State::set_code_hash(Address const &address, bytes32_t const &hash)
 {
-    auto &account = current_account(address);
+    auto &account_state = current_account_state(address);
+    auto &account = account_state.account_;
     MONAD_ASSERT(account.has_value());
     account.value().code_hash = hash;
+    // NOTE: Production code must use set_code instead of set_code_hash.
```
0xSqualo: Should this be disabled in release builds or flag enabled to avoid potential future confusion?

aa755: I had tried to make `set_code_hash` private. But some tests were calling it. I plan rewrite those tests to call `set_code` instead, but that can go in a separate PR.

category/execution/ethereum/evm_test.cpp
goodlyrottenapple: no worries :)

category/execution/monad/reserve_balance.hpp
```diff
@@ -23,15 +23,64 @@
...
 struct Transaction;
 
+class ReserveBalance
+{
+    using FailedSet = ankerl::unordered_dense::segmented_set<Address>;
+    using ViolationThresholdMap = ankerl::unordered_dense::segmented_map<
+        Address, std::optional<uint256_t>>;
+
+    State *state_;
+    bool tracking_enabled_{false};
```
goodlyrottenapple, line 48: I'm wondering why `tracking_enabled_` and `get_max_reserve_` are runtime values. From what I can see, they are set based purely on the monad revision in `init_from_tx`, so was wondering if we could replace their uses with `if constexpr ...` pattern instead? but perhaps I am missing some context whereby we do want `tracking_enabled_` to be dynamically set at runtime?

mkolosick: IIRC there is work being done to enable changing accounts reserve balance so I believe `get_max_reserve_` is in advance of that change.

aa755: I will try to make `tracking_enabled_` not a runtime value. The main issue is that traits are not available in all contexts.

category/execution/ethereum/state3/state.cpp
0xSqualo: Got it.

Could [potentially use `NDEBUG`](https://github.com/category-labs/monad/blob/3e5a6c63674bc3241181aa84e6a3da7f25839e91/category/core/assert.h#L118-L126) as a temporary measure to `static_assert` here if it accidentally gets used in the production build as an intermediate measure.

category/execution/monad/reserve_balance.hpp
goodlyrottenapple: Ah i see, we would have to trait-ify quite a lot more of the state class. @mkolosick w.r.t. `get_max_reserve_` is there a plan for C api that needs the function ptr? Just thinking of the recent refactor of the chain ctx where we also originally had a function ptr being passed around.

category/execution/ethereum/state3/state.cpp
aa755: I removed `State::set_code_hash`: https://github.com/category-labs/monad/pull/2068

category/execution/monad/reserve_balance.hpp
mkolosick: `get_max_reserve_` shouldn't be needed from C anywhere. Since we switched to a precompile for the reserve balance check we actually don't need to change `evmc_host`.

The closure capturing the trait is a pattern we use in the compiler as well so I'm not opposed to it for now.

category/execution/ethereum/state3/state.cpp
```diff
@@ -404,13 +433,13 @@ State::selfdestruct(Address const &address, Address const &beneficiary)
 
     if constexpr (traits::evm_rev() < EVMC_CANCUN) {
         add_to_balance(beneficiary, account.value().balance);
```
aa755, line 432: This was a pre-existing "issue" but may be worth improving, perhaps in a different PR:
- add_to_balance may insert an element to `State::current_`, e.g. if the beneficiary account was unchanged so far in this transaction.
- in general, insertion/deletion in a container may invalidate the iterators/references to the container elements obtained before the insertion/deletion, because the container may move elements. So continuing to use `account_state`/`account`, e.g. in the later line can be dangerous in general. However, currently, it is safe because the map we use does not seem to move elements on insertion.  Perhaps better to not rely on that implementation detail of the map.

category/execution/monad/reserve_balance.hpp
aa755: The diff with plumbing traits everywhere is turning out to be huge. `tracking_enabled_` is just one bit for the whole transaction `State`, so it may not be too bad.

category/execution/ethereum/state3/state.cpp
aa755: The library officially [promises](https://github.com/martinus/unordered_dense/blob/main/README.md) that references (but not iterators) are stable on insert.

category/execution/monad/reserve_balance.cpp
```diff
@@ -51,94 +51,197 @@ unsigned monad_default_max_reserve_balance_mon(enum monad_revision)
...
+    return !failed_.empty();
+}
+
+bool ReserveBalance::failed_contains(Address const &address) const
+{
+    return failed_.contains(address);
+}
+
+bool ReserveBalance::subject_account(Address const &address)
+{
```
aa755, line 185: IIUC, it is possible that here, `address` is `0x1000`, the address of the staking contract. IIUC, its balance can decrease during execution of the transaction. Although it is unlikely that the balance decreases below 10MON, perhaps it is better to explicitly carve out an exception for the staking contract.
```suggestion
{
    if (is_staking_contract(address))
       return false;
```

category/execution/ethereum/state3/state.hpp
```diff
@@ -24,6 +24,7 @@
 #include <category/execution/ethereum/state3/account_state.hpp>
 #include <category/execution/ethereum/state3/version_stack.hpp>
 #include <category/execution/ethereum/types/incarnation.hpp>
+#include <category/execution/monad/reserve_balance.hpp>
```
aa755, line 28: @goodlyrottenapple made a great observation here that this breaks separation of monad specific features from vanilla ethereum. i.e. category/execution/ethereum source files don't import stuff from category/execution/monad/. 
This should be fixed, but as the diff would be large, perhaps it can be done in a subsequent cleanup PR. Made an issue for it: https://github.com/category-labs/monad/issues/2085

category/execution/monad/reserve_balance.cpp
```diff
@@ -51,94 +52,204 @@ unsigned monad_default_max_reserve_balance_mon(enum monad_revision)
 MONAD_ANONYMOUS_NAMESPACE_BEGIN
 
 template <Traits traits>
```
mkolosick, line 54: The trait is now unused and can be removed

category/execution/ethereum/state3/state.hpp
jhunsaker: yeah this a reason to pull this out of this class

aa755: Not fully sure what you meant by "this".  What do you want to pulled out? The `State::rb_` field?

jhunsaker: yes, there is no reserve balance for ethereum

jhunsaker (approved):

PR #2068, opened 2026-02-13
removed set_code_hash

aa755:
Calling `set_code_hash` arbitrarily may break invariants of the State class, e.g. that the account's code's hash is the same. Clients should call `set_code` instead to update the state in a consistent way.
Already, it was only being called in testcases, which I modified to use `set_code` instead.

co-authored with codex-cli

### DIFF
--- a/category/execution/ethereum/state2/test/test_state.cpp
+++ b/category/execution/ethereum/state2/test/test_state.cpp
@@ -19,6 +19,7 @@
 #include <category/core/blake3.hpp>
 #include <category/core/byte_string.hpp>
 #include <category/core/bytes.hpp>
+#include <category/core/keccak.hpp>
 #include <category/execution/ethereum/core/account.hpp>
 #include <category/execution/ethereum/core/address.hpp>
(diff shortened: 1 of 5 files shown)


category/execution/ethereum/state2/test/test_state.cpp
```diff
@@ -272,15 +273,15 @@ TEST_F(InMemoryStateTest, get_code_hash)
...
 
-TEST_F(InMemoryStateTest, set_code_hash)
+TEST_F(InMemoryStateTest, set_code_sets_code_hash)
 {
     BlockState bs{this->tdb, this->vm};
 
     State s{bs, Incarnation{1, 1}};
     s.create_contract(b);
-    s.set_code_hash(b, hash1);
+    s.set_code(b, code1);
```
aa755, line 282: `hash(code1)` is probably not equal to `hash1`. Was `hash1` just a random/arbitrary value or was it testing something specific that other values dont?

jhunsaker: I think it was just random

PR #2077, opened 2026-02-18
Enforce read permissions in CI workflows

andr-dev:


.github/workflows/code-quality.yml
```diff
@@ -1,5 +1,8 @@
 name: Code Quality Checks
 
+permissions:
+  contents: read
+
```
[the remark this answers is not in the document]
andr-dev: Seems like build re-ran and pushed to cache successfully, other LLM saying that as long as you don't restrict `actions: read` then this is fine.

.github/workflows/docker-base.yaml
```diff
@@ -31,5 +31,4 @@ jobs:
           file: docker/Dockerfile
           ulimit: "memlock=-1:-1"
           target: base
-          cache-from: type=gha,scope=base
-          cache-to: type=gha,scope=base,mode=max
+          cache-to: type=gha,scope=base,mode=min
```
andr-dev: This eliminates caching intermediate docker layers.

jhunsaker: mode=min is default

jhunsaker: cache-from is removed because it's the base?

PR #2086, opened 2026-02-19
Adding dippedIntoReserve to reserve balance precompile

mkolosick:


category/execution/monad/reserve_balance.cpp
```diff
@@ -51,94 +52,204 @@ unsigned monad_default_max_reserve_balance_mon(enum monad_revision)
 MONAD_ANONYMOUS_NAMESPACE_BEGIN
 
 template <Traits traits>
-bool dipped_into_reserve(
-    Address const &sender, Transaction const &tx,
-    uint256_t const &base_fee_per_gas, uint64_t const i,
-    ChainContext<traits> const &ctx, State &state)
+bool dipped_into_reserve(State &state)
```
dhil: The template parameter (line 54) appears to be unused. It should either be used or removed.

category/execution/monad/reserve_balance/reserve_balance_error.hpp
```diff
@@ -34,6 +34,7 @@ enum class ReserveBalanceError
 {
     Success = 0,
     MethodNotSupported,
+    ValueNonZero,
```
es3n1n, line 37: In reserve_balance_error.cpp there's a value mapping, this probably should be added there https://github.com/category-labs/monad/blob/d06dc1941399eb4947ea5e5f6135b6f49e830d48/category/execution/monad/reserve_balance/reserve_balance_error.cpp#L32-L44

category/execution/monad/reserve_balance/reserve_balance_contract.cpp
```diff
@@ -27,10 +28,25 @@
...
+        abi_encode_selector("dippedIntoReserve()");
+};
+
+static_assert(PrecompileSelector::DIPPED_INTO_RESERVE == 0x3a61584e);
+
 //
 // Gas Costs
 //
 
+constexpr uint64_t DIPPED_INTO_RESERVE_OP_COST = 100; // warm sload coast
```
es3n1n:
```suggestion
constexpr uint64_t DIPPED_INTO_RESERVE_OP_COST = 100; // warm sload cost
```

category/execution/monad/reserve_balance/reserve_balance_contract_test.cpp
```diff
@@ -89,6 +108,249 @@ struct ReserveBalanceEvm : public ReserveBalanceTest
...
+        STOP,
+        JUMPDEST,
+        0xFE,
+    });
+}
+
+void add_revert_check(std::vector<uint8_t> &code)
+{
+    // auto const *s =
+    // intx::as_bytes(abi_encode_selector("dippedIntoReserve()"));
```
dhil:
```suggestion
```

category/execution/monad/reserve_balance/reserve_balance_contract_test.cpp
```diff
@@ -89,6 +108,249 @@ struct ReserveBalanceEvm : public ReserveBalanceTest
...
+    uint64_t const initial_balance_mon, uint64_t const value_mon,
+    bool should_revert)
+{
+    constexpr uint256_t GAS_FEE = 4 * 1000000000000000000ULL;
+    constexpr uint256_t BASE_FEE_PER_GAS = 10;
+    constexpr uint256_t GAS_LIMIT = GAS_FEE / BASE_FEE_PER_GAS;
+    constexpr Address BUNDLER{0xbbbbbbbb};
+    constexpr Address ENTRYPOINT{0xeeeeeeee};
+    constexpr Address EOA{0xaaaaaaaa};
+    constexpr Address SCW{0xcccccccc};
```
dhil: I'd slab some `static`s on here:
```suggestion
    static constexpr uint256_t GAS_FEE = 4 * 1000000000000000000ULL;
    static constexpr uint256_t BASE_FEE_PER_GAS = 10;
    static constexpr uint256_t GAS_LIMIT = GAS_FEE / BASE_FEE_PER_GAS;
    static constexpr Address BUNDLER{0xbbbbbbbb};
    static constexpr Address ENTRYPOINT{0xeeeeeeee};
    static constexpr Address EOA{0xaaaaaaaa};
    static constexpr Address SCW{0xcccccccc};
```

category/execution/monad/reserve_balance/reserve_balance_contract_test.cpp
```diff
@@ -120,3 +382,33 @@ TEST_F(ReserveBalanceEvm, precompile_fallback)
...
 }
+
+TYPED_TEST(MonadTraitsTest, reverttransaction_no_dip)
+{
+    constexpr bool should_revert = [] {
+        if (TestFixture::Trait::monad_rev() >= MONAD_NEXT) {
+            return false;
+        }
+        else {
+            return false;
```
dhil: `if X then false else false` I suppose one of these were intended to be `true`.

mkolosick: So the old test distinguished between not needing to revert and the contract failing because the opcode didn't exist in prior versions. The tests now use a chain generic call to the precompile that returns false for reverting if the contract is not present.

So these should both be false (the precompile says we don't need to revert or the contract isn't there and we get false). Perhaps there is a way to re-distinguish these cases nonetheless let me think about it.

mkolosick: Updated the tests to distinguish the cases better

category/execution/monad/reserve_balance/reserve_balance_contract_test.cpp
```diff
@@ -89,6 +115,258 @@ struct ReserveBalanceEvm : public ReserveBalanceTest
...
+
+        auto const *a = intx::as_bytes(SCW);
+        // set EOA to delegate to SCW
+        state.set_code(
+            EOA,
+            byte_string_view{std::initializer_list<uint8_t>{
+                0xef,  0x01,  0x00,  a[0],  a[1],  a[2],  a[3],  a[4],
+                a[5],  a[6],  a[7],  a[8],  a[9],  a[10], a[11], a[12],
+                a[13], a[14], a[15], a[16], a[17], a[18], a[19],
+            }});
```
dhil: I think we can afford to raise the level abstraction here, e.g.
```suggestion
      evmc::bytes const delegate_code =
            evmc::from_hex(std::format("0xef0100{}", evmc::hex(SCW))).value();
        state.set_code(EOA, byte_string_view{delegate_code});
```

category/execution/monad/reserve_balance/reserve_balance_contract_test.cpp
```diff
@@ -89,6 +115,258 @@ struct ReserveBalanceEvm : public ReserveBalanceTest
...
+    auto const *g = intx::as_bytes(gas_fee);
+    code.append_range(std::initializer_list<uint8_t>{
+        PUSH0, PUSH0, PUSH0, PUSH0, PUSH0, PUSH20, v[0],   v[1],  v[2],  v[3],
+        v[4],  v[5],  v[6],  v[7],  v[8],  v[9],   v[10],  v[11], v[12], v[13],
+        v[14], v[15], v[16], v[17], v[18], v[19],  PUSH32, g[31], g[30], g[29],
+        g[28], g[27], g[26], g[25], g[24], g[23],  g[22],  g[21], g[20], g[19],
+        g[18], g[17], g[16], g[15], g[14], g[13],  g[12],  g[11], g[10], g[9],
+        g[8],  g[7],  g[6],  g[5],  g[4],  g[3],   g[2],   g[1],  g[0],  CALL,
+    });
+}
```
dhil, line 225: I have some utilities to make this sort of thing manageable. I will try to wrap it up soon, so we can start using it in our tests.

category/execution/monad/reserve_balance/reserve_balance_contract.cpp
```diff
@@ -27,10 +28,25 @@
 
 MONAD_ANONYMOUS_NAMESPACE_BEGIN
 
+////////////////////////
```
dhil:
```suggestion
```

dhil (dismissed):

PR #2087, opened 2026-02-19
[core/runtime] Make uint256 library header only

cmcl:
## Summary

Move remaining `uint256_t` helpers from `uint256.cpp` to `uint256.hpp` as inline functions to obtain header only library for uint256_t. This PR moves `signextend`, `from_bytes` and `countr_zero`.  I provide some benchmarking results for `signextend`.  For the latter two, there is no associated benchmark(s) but I manually compared the before/after disassembly for some of the occurrences and didn't see anything concerning.  I also had Claude perform an analysis of the disassembly and it claimed inlining shrunk the callers but I haven't confirmed the specifics yet.  I am happy to do a deeper dive if it is felt such a thing is warranted.

To summarise (including Claude's findings for completeness; I have more details if it is of interest):

- **`signextend`** &mdash; moved to header; interpreter seq delta improves 18&ndash;74% (standalone) and 35&ndash;68% (arithmetic combinations)
- **`from_bytes`** &mdash; completely eliminated 25 occurrences from `scan_from` template instantiations; shrinking each by ~624 B
- **`countr_zero`** &mdash; completely eliminated 6 occurrences; `mul_optimized` and `div_optimized` callers shrink 10&ndash;18%

## `signextend`

I ran the following experiment for `signextend`:

`./build/test/vm/micro_benchmarks/vm-micro-benchmarks --format md --seq-filter="^SIGNEXTEND\n"`

Benchmarks show consistent improvements, closing the gap between the subject sequence and the baseline. There was a bit more variance in the baseline for Standalone results but the arithmetic combinations were fairly consistent so the seq delta in those cases is a good indicator of performance gain.

### Benchmark Results

All benchmarks use the `BYTE/SIGNEXTEND` opcode group. The tables show **Seq Delta (ns)** (incremental cost of `SIGNEXTEND` above the baseline sequence, i.e. lower is better) for both **Before** and **After** the PR change.

#### Standalone SIGNEXTEND

| Input    | Mode        | Metric     | Before (ns) | After (ns) |     Change |
| :----    | :---        | :-----     | ----------: | ---------: |     -----: |
| random   | interpreter | throughput |       9.358 |      7.634 | **-18.4%** |
| random   | interpreter | latency    |       1.648 |      0.426 | **-74.1%** |
| constant | interpreter | throughput |       7.702 |      2.287 | **-70.3%** |
| constant | interpreter | latency    |       5.231 |      2.617 | **-50.0%** |
| random   | compiler    | throughput |       0.644 |      0.623 |      -3.3% |
| random   | compiler    | latency    |       4.610 |      4.593 |      -0.4% |
| constant | compiler    | throughput |       0.636 |      0.572 |     -10.1% |
| constant | compiler    | latency    |       4.584 |      4.605 |      +0.5% |

#### SIGNEXTEND + Arithmetic (constant input, interpreter)

Average Seq Delta across all 15 binary operations (ADD, MUL, SUB, DIV, SDIV, MOD, SMOD, LT, GT, SLT, SGT, EQ, AND, OR, XOR):

| Benchmark             | Metric     | Before (ns) | After (ns) | Change     |
| :--------             | :-----     | ----------: | ---------: | -----:     |
| SIGNEXTEND; op        | throughput |       14.20 |       6.38 | **-55.1%** |
| SIGNEXTEND; op        | latency    |       15.48 |       8.35 | **-46.1%** |
| SIGNEXTEND; SWAP1; op | throughput |       16.61 |       6.95 | **-58.1%** |
| SIGNEXTEND; SWAP1; op | latency    |       16.50 |      10.00 | **-39.4%** |

#### SIGNEXTEND + Arithmetic (constant input, throughput, interpreter) &mdash; per-operation

| Sequence         | Before (ns) | After (ns) | Change     |
| :-------         | ----------: | ---------: | -----:     |
| SIGNEXTEND; ADD  |       11.82 |       4.97 | **-57.9%** |
| SIGNEXTEND; MUL  |       15.63 |       6.51 | **-58.3%** |
| SIGNEXTEND; SUB  |       12.71 |       4.69 | **-63.1%** |
| SIGNEXTEND; DIV  |       17.09 |       9.15 | **-46.5%** |
| SIGNEXTEND; SDIV |       22.01 |      13.86 | **-37.0%** |
| SIGNEXTEND; MOD  |       16.85 |       9.10 | **-46.0%** |
| SIGNEXTEND; SMOD |       20.96 |      13.50 | **-35.6%** |
| SIGNEXTEND; LT   |       13.03 |       4.52 | **-65.3%** |
| SIGNEXTEND; GT   |       13.05 |       4.21 | **-67.7%** |
| SIGNEXTEND; SLT  |       13.49 |       4.89 | **-63.8%** |
| SIGNEXTEND; SGT  |       13.70 |       4.72 | **-65.5%** |
| SIGNEXTEND; EQ   |       13.22 |       4.59 | **-65.3%** |
| SIGNEXTEND; AND  |        9.79 |       3.39 | **-65.3%** |
| SIGNEXTEND; OR   |        9.79 |       3.79 | **-61.3%** |
| SIGNEXTEND; XOR  |        9.83 |       3.77 | **-61.7%** |

### DIFF
--- a/category/core/runtime/uint256.hpp
+++ b/category/core/runtime/uint256.hpp
@@ -38,14 +38,15 @@
 
 // GCC's overeager SLP vectorizer sometimes pessimizes code. For functions that
 // are particularly sensitive about this (such as multiplication), the
-// vectorizer can be turned off with the MONAD_VM_NO_VECTORIZE pragma.
-// Since optimization pragmas are not applied to inlined functions, any function
-// that is annotated as MONAD_VM_NO_VECTORIZE should either be marked as
-// noinline or only called from other MONAD_VM_NO_VECTORIZE functions.
+// vectorizer can be turned off with the MONAD_VM_NO_VECTORIZE pragma.  Since
+// optimization pragmas are not applied to functions that are inlined at call
+// sites, any function that is annotated as MONAD_VM_NO_VECTORIZE should either
+// also be marked as noinline, or only called from other MONAD_VM_NO_VECTORIZE
+// functions (thus inheriting the caller's attributes).
 #if defined(__GNUC__) && !defined(__clang__)
-    #define MONAD_VM_NO_VECTORIZE __attribute__((optimize("no-tree-vectorize")))
+    #define MONAD_NO_VECTORIZE __attribute__((optimize("no-tree-vectorize")))
 #else
-    #define MONAD_VM_NO_VECTORIZE
+    #define MONAD_NO_VECTORIZE
 #endif
 
 namespace monad::vm::runtime
@@ -1058,7 +1059,7 @@ namespace monad::vm::runtime
      * corresponds to full precision multiplication
      */
     template <size_t R, size_t M, size_t N>
-    MONAD_VM_NO_VECTORIZE [[gnu::always_inline]]
+    MONAD_NO_VECTORIZE [[gnu::always_inline]]
     inline constexpr words_t<R>
     truncating_mul(words_t<M> const &x, words_t<N> const &y) noexcept
         requires(0 < R && 0 < M && 0 < N && R <= M + N)
@@ -1071,7 +1072,7 @@ namespace monad::vm::runtime
         }
     }
 
-    MONAD_VM_NO_VECTORIZE [[gnu::always_inline]]
+    MONAD_NO_VECTORIZE [[gnu::always_inline]]
     inline constexpr uint256_t
     truncating_mul(uint256_t const &x, uint256_t const &y) noexcept
     {
@@ -1079,7 +1080,7 @@ namespace monad::vm::runtime
             truncating_mul<uint256_t::num_words>(x.as_words(), y.as_words())};
     }
 
-    MONAD_VM_NO_VECTORIZE
+    MONAD_NO_VECTORIZE
     [[gnu::noinline]]
     inline constexpr uint256_t
     operator*(uint256_t const &lhs, uint256_t const &rhs) noexcept
@@ -1314,7 +1315,7 @@ namespace monad::vm::runtime
         return uint256_t{udivrem(sum, mod.as_words()).rem};
     }
 
-    MONAD_VM_NO_VECTORIZE
+    MONAD_NO_VECTORIZE
     [[gnu::noinline]]
     inline constexpr uint256_t mulmod(
         uint256_t const &u, uint256_t const &v, uint256_t const &mod) noexcept
@@ -1368,7 +1369,7 @@ namespace monad::vm::runtime
         return (~diff & (x < y)) | (x_neg & ~y_neg);
     }
 
-    MONAD_VM_NO_VECTORIZE
+    MONAD_NO_VECTORIZE
     [[gnu::noinline]] inline constexpr uint256_t
     exp(uint256_t base, uint256_t const &exponent) noexcept
     {


goodlyrottenapple (approved):

category/core/runtime/uint256.hpp
m-alvarez: Maybe rename to `MONAD_NO_VECTORIZE`

m-alvarez, line 52: Maybe move to `monad::runtime` or just `monad` as @kjcamann suggested

cmcl: If we go with `monad`, should I move the library out of the `vm` subdirectory into `category/core`?

cmcl: Also to note this change will touch a lot of other files.  May be better as a separate refactoring PR?

kjcamann: Yeah I'll do that as a separate PR

kjcamann (approved):
LGTM

category/core/runtime/uint256.hpp
```diff
@@ -1017,7 +1058,7 @@ namespace monad::vm::runtime
      * corresponds to full precision multiplication
      */
     template <size_t R, size_t M, size_t N>
-    MONAD_VM_NO_VECTORIZE [[gnu::always_inline]]
+    MONAD_NO_VECTORIZE [[gnu::always_inline]]
```
dhil, line 1062: Is comment near the start of the file incorrect or is this annotation incorrect (or redundant)?
> Since optimization pragmas are not applied to **inlined** functions, any function that is annotated as **MONAD_NO_VECTORIZE** should either be marked as **noinline** or only called from other MONAD_NO_VECTORIZE functions.

cmcl: Good question/catch.  Maybe @m-alvarez has an idea?  Should this function and the other `truncating_mul` be annotated with `[[gnu::noinline]]` instead?

andreaslyn: [[gnu::always_inline]] together with MONAD_NO_VECTORIZE is correct as long as all callers have MONAD_NO_VECTORIZE attribute. The functions with both [[gnu::always_inline]] and MONAD_NO_VECTORIZE are good candidates for a monad::uint256_t::internal namespace.

dhil: Thanks for the clarification. Probably worth updating the comment near the top of the file to clarify this matter. I suppose the comment is talking about C++ `inline` qualifier.

cmcl: Does that mean `from_string` should be annotated with `MONAD_NO_VECTORIZE` since it calls `truncating_mul`?  It was unclear to me which aspect was incorrect

m-alvarez: @cmcl I'm inclined to say not in this case. `from_string` is not meant to be performance-sensitive (in fact performance is known to be quite bad) so it's unlikely that the SLP vectorizer's pessimization makes a difference. We'll revisit this if and when we need a faster `from_string`.

andreaslyn: The problem is that gcc and clang do not respect function attributes of inlined/unfolded functions. So it effectively means that the `MONAD_NO_VECTORIZE` "optimization" is not performed on `truncating_mul` in `from_string`.

andreaslyn: Looks like this is already documented in the comment to `MONAD_NO_VECTORIZE`

cmcl: I reworded the comment slightly

andreaslyn (approved):

dhil (approved):

PR #2090, opened 2026-02-20
[zkVM] Add RISC-V cross-compilation target

goodlyrottenapple:
Add a standalone CMake project (`zkvm/`) that builds `libmonad-zkvm.a` for bare-metal RISC-V targets (planned ZisK and SP1 zkVM backends). This allows the Monad EVM interpreter to run inside zkVMs.

Key changes:
- RISC-V toolchain file (rv64ima, lp64, medany, -nostdlib++)
- zkVM-abstracted allocator (zkvm_aligned_alloc) and exit (zkvm_exit) interfaces with planned ZisK/SP1 backends
- Splitting out any platform-specific code from `{X}.hpp` into `{X}/intrinsics.hpp`
- Shadowing of `category/` headers with zkvm specific ones under `zkvm/category/`
- Shared `cmake/compile_options.cmake` extracted from root CMakeLists.txt and a new `cmake/zkvm.cmake` specific options for the zkVM targets
- `zkvm-core` bare-metal shims for libc (malloc/calloc/free), libstdc++ (operator new/delete, std::terminate), and assert handlers
- minimal CI build of `zkvm/` code targeting RISC-V

Co-authored with Claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,23 @@ You can also run the full test suite in parallel with:
 CTEST_PARALLEL_LEVEL=$(nproc) ctest
 ```
 
+## Compiling zkVM binary
+
+To compile monad as a guest program for various zkVMs, such as ZisK or SP1, we need to use a riscv64 cross-compiler. The easiest way to do this is to use the [riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain) which includes newlib. The cmake build extracts only the needed libc objects (setjmp/longjmp) from the unmodified newlib; malloc and syscalls are weakly linked by the zkVM frameworks.
+
+```shell
+cmake -B build-zkvm -S zkvm/guest \
+   -DCMAKE_TOOLCHAIN_FILE=$PWD/category/core/toolchains/riscv64-elf.cmake \
+   -DRISCV_TOOLCHAIN_DIR="path/to/riscv_gcc" \
+   -DCMAKE_BUILD_TYPE=Release -GNinja
+```
+
+We can then build the static library:
+
+```shell
+cmake --build build-zkvm --target monad-zkvm --parallel
+```
+
 ## A tour of execution
 
 To understand how the source code is organized, you should start by reading
--- a/category/core/throw.hpp
+++ b/category/core/throw.hpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2025 Category Labs, Inc.
+// Copyright (C) 2025-26 Category Labs, Inc.
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -15,18 +15,7 @@
 
 #pragma once
 
-#include <category/core/thread_local.h>
-#include <category/vm/runtime/cached_allocator.hpp>
-
-namespace monad::vm::runtime
-{
-    struct EvmStackAllocatorMeta
-    {
-        using base_type = uint256_t;
-        static constexpr size_t size = 1024;
-        static constexpr size_t alignment = 32;
-        static MONAD_THREAD_LOCAL CachedAllocatorList cache_list;
-    };
-
-    using EvmStackAllocator = CachedAllocator<EvmStackAllocatorMeta>;
-}
+// Throw a C++ exception. The zkVM mirror replaces this with
+// MONAD_ASSERT(false) since the bare-metal zkVM build is compiled with
+// -fno-exceptions.
+#define MONAD_THROW(exc, msg) throw exc(msg)
--- a/category/core/toolchains/riscv64-elf.cmake
+++ b/category/core/toolchains/riscv64-elf.cmake
@@ -1,0 +1,43 @@
+# Copyright (C) 2025 Category Labs, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_SYSTEM_PROCESSOR riscv64)
+set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+
+# Configurable toolchain directory — pass -DRISCV_TOOLCHAIN_DIR=<path> to cmake.
+set(RISCV_TOOLCHAIN_DIR "/opt/riscv" CACHE PATH "RISC-V toolchain directory")
+list(APPEND CMAKE_TRY_COMPILE_PLATFORM_VARIABLES RISCV_TOOLCHAIN_DIR)
+
+if(EXISTS "${RISCV_TOOLCHAIN_DIR}/bin/riscv64-unknown-elf-gcc")
+    set(RISCV_PREFIX "riscv64-unknown-elf-")
+else()
+    message(FATAL_ERROR "No riscv64 gcc found in ${RISCV_TOOLCHAIN_DIR}/bin/")
+endif()
+
+set(CMAKE_C_COMPILER "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}gcc")
+set(CMAKE_CXX_COMPILER "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}g++")
+set(CMAKE_AR "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}ar")
+set(CMAKE_RANLIB "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}ranlib")
+
+set(CMAKE_C_FLAGS_INIT
+    "-march=rv64ima -mabi=lp64 -mcmodel=medany -nostartfiles -nostdlib -ffunction-sections -fdata-sections")
+set(CMAKE_CXX_FLAGS_INIT
+    "-march=rv64ima -mabi=lp64 -mcmodel=medany -nostartfiles -nostdlib++ -fno-exceptions -fno-rtti -ffunction-sections -fdata-sections")
+set(CMAKE_ASM_FLAGS_INIT "-march=rv64ima -mabi=lp64")
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
--- a/category/vm/runtime/exit.cpp
+++ b/category/vm/runtime/exit.cpp
@@ -13,15 +13,14 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include <category/vm/runtime/exit.hpp>
 #include <category/vm/runtime/types.hpp>
-
-extern "C" void monad_vm_runtime_exit [[noreturn]] (void *);
 
 extern "C" void monad_vm_runtime_context_out_of_gas_exit
     [[noreturn]] (monad::vm::runtime::Context *const ctx)
 {
     ctx->result.status = monad::vm::runtime::StatusCode::OutOfGas;
-    monad_vm_runtime_exit(ctx->exit_stack_ptr);
+    monad::vm::runtime::exit(ctx->exit_stack_ptr);
 }
 
 namespace monad::vm::runtime
@@ -30,12 +29,12 @@ namespace monad::vm::runtime
     {
         is_stack_unwinding_active = true;
         result.status = StatusCode::Error;
-        monad_vm_runtime_exit(exit_stack_ptr);
+        ::monad::vm::runtime::exit(exit_stack_ptr);
     }
 
     void Context::exit [[noreturn]] (StatusCode const code) noexcept
     {
         result.status = code;
-        monad_vm_runtime_exit(exit_stack_ptr);
+        ::monad::vm::runtime::exit(exit_stack_ptr);
     }
 }
--- a/category/vm/runtime/exit.hpp
+++ b/category/vm/runtime/exit.hpp
@@ -1,0 +1,38 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Host unwind primitive. On the host build the actual implementation is
+// the hand-rolled assembly in exit.S; the inline `exit` wrapper here
+// gives call sites a single typed entry point so the zkVM mirror can
+// supply an inline setjmp/longjmp variant without any extern call.
+
+#pragma once
+
+namespace monad::vm::runtime
+{
+    using exit_stack_ptr_t = void *;
+}
+
+extern "C" void monad_vm_runtime_exit
+    [[noreturn]] (monad::vm::runtime::exit_stack_ptr_t);
+
+namespace monad::vm::runtime
+{
+    [[gnu::always_inline, noreturn]]
+    inline void exit(exit_stack_ptr_t p)
+    {
+        monad_vm_runtime_exit(p);
+    }
+}
--- a/category/vm/runtime/transmute.hpp
+++ b/category/vm/runtime/transmute.hpp
@@ -20,39 +20,16 @@
 #include <category/core/bytes.hpp>
 #include <category/core/int.hpp>
 #include <category/core/runtime/uint256.hpp>
+#include <category/vm/runtime/transmute/intrinsics.hpp>
 
 #include <evmc/evmc.h>
 #include <evmc/evmc.hpp>
-
-#include <immintrin.h>
-
-// Load `load_size` bytes from `src_buffer` and clear the remaining upper bytes
-// of the result. It is required that `load_size <= 32`. If `load_size <= 0`
-// then zero is returned.
-extern "C" __m256i
-monad_vm_runtime_load_bounded_le(uint8_t const *src_buffer, int64_t load_size);
-
-// Note: monad_vm_runtime_load_bounded_le_raw uses non-standard
-// calling convention. See transmute.S. Use the
-// monad_vm_runtime_load_bounded_le function for a version
-// using standard calling convention.
-extern "C" __m256i monad_vm_runtime_load_bounded_le_raw();
 
 namespace monad::vm::runtime
 {
     static_assert(sizeof(evmc_address) == 20);
     static_assert(sizeof(bytes32_t) == 32);
     static_assert(sizeof(uint256_t) == 32);
-
-    [[gnu::always_inline]]
-    inline uint256_t
-    uint256_load_bounded_le(uint8_t const *const bytes, int64_t const max_len)
-    {
-        if (MONAD_LIKELY(max_len >= 32)) {
-            return load_le_unsafe<uint256_t>(bytes);
-        }
-        return uint256_t{monad_vm_runtime_load_bounded_le(bytes, max_len)};
-    }
 
     [[gnu::always_inline]]
     inline uint256_t
--- a/category/vm/runtime/transmute/intrinsics.hpp
+++ b/category/vm/runtime/transmute/intrinsics.hpp
@@ -1,0 +1,51 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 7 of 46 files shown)


category/vm/runtime/transmute.hpp
```diff
@@ -45,10 +47,15 @@ namespace monad::vm::runtime
...
+#if defined(MONAD_ZKVM)
+        (void)max_len;
+        return uint256_t::load_le_unsafe(bytes);
+#else
         if (MONAD_VM_LIKELY(max_len >= 32)) {
             return uint256_t::load_le_unsafe(bytes);
         }
         return uint256_t{monad_vm_runtime_load_bounded_le(bytes, max_len)};
+#endif
     }
```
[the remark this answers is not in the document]
goodlyrottenapple: ~~I assumed this is just an optimisation and it is always safe to call `uint256_t::load_le_unsafe`?~~

cmake/compile_options.cmake
```diff
@@ -0,0 +1,56 @@
...
+    # This fires on constexpr-guarded AVX2 code paths that are never instantiated
+    # on non-x86 targets.
+    target_compile_options(
+      ${target} PRIVATE $<$<CXX_COMPILER_ID:GNU>:-Wno-template-body>)
+
+    # GCC 15+ warns when reference parameters are forwarded through musttail
+    # calls. The interpreter's threaded dispatch passes references to objects
+    # that live in earlier stack frames, not the current one, so this is safe.
+    target_compile_options(
+      ${target} PRIVATE $<$<CXX_COMPILER_ID:GNU>:-Wno-maybe-musttail-local-addr>)
```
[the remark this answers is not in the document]
goodlyrottenapple: only ever expect to use the gcc cross-compiler here, so don't think this is necessary

zkvm/category/core/runtime/uint256.hpp
```diff
@@ -0,0 +1,1371 @@
...
+#include <bit>
+#include <climits>
+#include <cstdint>
+#include <cstring>
+#include <format>
+#include <limits>
+#include <stdexcept>
+#include <type_traits>
+
+namespace monad
```
dhil: I suppose we ought to use the `MONAD_NAMESPACE_BEGIN` and `MONAD_NAMESPACE_END` macros.

zkvm/category/core/runtime/uint256.hpp
```diff
@@ -0,0 +1,1371 @@
...
+                    result[I + M - 1],
+                    // Output
+                    result[I + M],
+                    result[I + M - 1]);
+            }
+            else if constexpr (I + M < R + 1) {
+                result[I + M - 1] += c_lo;
+            }
+        }
+    }
```
dhil: This is dead code. 
```suggestion
```

goodlyrottenapple: whoop sorry must have re-added during the re-shuffle...

cmake/zkvm.cmake
```diff
@@ -0,0 +1,107 @@
...
+    foreach(lib ${ARGN})
+        list(REMOVE_ITEM _libs "${lib}")
+    endforeach()
+    set_property(TARGET ${target} PROPERTY LINK_LIBRARIES ${_libs})
+endfunction()
+
+# Replace each named host source with its zkVM mirror counterpart.
+# Each argument is a repo-root-relative path; the mirror must exist at
+# ${MONAD_ROOT}/zkvm/${path}.
+function(monad_zkvm_shadow_sources target)
```
goodlyrottenapple: I've included this to be able to shadow cpp files, but for now all the changes are contained in headers, which can just be included before the regular `category/` ones. Can delete for now if we don't end up needing it.

zkvm/category/core/throw.h
```diff
@@ -0,0 +1,22 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+// zkVM mirror: exceptions are disabled (-fno-exceptions) on the bare-metal
+// zkVM build, so MONAD_THROW falls back to MONAD_ASSERT(false).
+#include <category/core/assert.h>
+
+#define MONAD_THROW(exc, msg) MONAD_ASSERT(false)
```
goodlyrottenapple:
```suggestion
#define MONAD_THROW(exc, msg) MONAD_ABORT()
```

README.md
```diff
@@ -116,6 +116,23 @@ You can also run the full test suite in parallel with:
...
 ```
 
+## Compiling zkVM binary
+
+To compile monad as a guest program for various zkVMs, such as ZisK or SP1, we need to use a riscv64 cross-compiler. The easiest way to do this is to use a nix package which builds a GCC 15 riscv64 cross-compiler targeting rv64ima / lp64 (soft-float) with newlib headers. The cmake build extracts only the needed libc objects (setjmp/longjmp) from the unmodified newlib; malloc and syscalls are weakly linked by the zkVM frameworks.
+
+```shell
+cmake -B build-zkvm -S zkvm \
+   -DCMAKE_TOOLCHAIN_FILE=$PWD/category/core/toolchains/riscv64-elf-toolchain.cmake \
+   -DRISCV_TOOLCHAIN_DIR=$(nix-build ./riscv64-none-elf.nix --no-out-link) \
```
mkolosick: This and the above directions should remove the nix directions.

mkolosick:
`build-zkvm` should be added to `.gitignore`

zkvm/core/zkvm_exit.h
```diff
@@ -0,0 +1,48 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef ZKVM_EXIT_H
+#define ZKVM_EXIT_H
+
+#if defined(MONAD_ZKVM_SP1)
+[[noreturn]] void syscall_halt(unsigned char exit_code);
```
mkolosick: Should this be moved into the `extern "C"` block down below so its name doesn't get mangled?

zkvm/core/libstdcxx.cpp
```diff
@@ -0,0 +1,78 @@
...
+// entrypoint crate (ziskos for ZisK, sp1-zkvm for SP1) at link time.
+
+#include <zkvm/core/zkvm_exit.h>
+
+#include <cstddef>
+
+extern "C" void *sys_alloc_aligned(std::size_t bytes, std::size_t align);
+
+// operator new / delete
+[[gnu::always_inline]] static inline void *alloc_or_exit(std::size_t size)
```
mkolosick, line 24: Passing on copilot review comments:
> operator new currently treats size == 0 as fatal (zkvm_exit(1)), but the C++ standard requires ::operator new(0) to return a non-null pointer (often by allocating 1 byte). This can cause unexpected termination from valid library/container code paths; consider normalizing size to at least 1 instead of exiting.

zkvm/category/vm/runtime/transmute/intrinsics.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <category/core/likely.h>
+#include <category/core/runtime/uint256.hpp>
+
+#include <cstring>
+
+namespace monad::vm::runtime
+{
+    [[gnu::always_inline]]
+    inline uint256_t
+    uint256_load_bounded_le(uint8_t const *const bytes, int64_t const max_len)
```
mkolosick, line 30: > max_len is an int64_t but is passed directly to std::memcpy as the byte count; if max_len <= 0 this underflows to a huge size_t and will corrupt memory. This helper should explicitly handle max_len <= 0 (return 0) and clamp the copy length to <= 32.

zkvm/core/zkvm_exit.h
goodlyrottenapple: Ah this file is a bit performative. It's more of a documentation for how the call should be implemented. I've now removed the body so that it follows the other zkvm_*.h header files convention, where each zkvm is supposed to provide implementations for these functions.

zkvm/guest/CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+    OUTPUT_VARIABLE _libc_a
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+set(_newlib_dir "${CMAKE_CURRENT_BINARY_DIR}/_libc_objs")
+file(MAKE_DIRECTORY "${_newlib_dir}")
+
+if(EXISTS "${_libc_a}")
+    foreach(_obj libc_a-setjmp.o)
+        execute_process(
```
mkolosick, line 122: > [P3] execute_process(${CMAKE_AR} x ...) without RESULT_VARIABLE silently swallows non-zero exit codes from ar. If the newlib archive ever changes its internal object name (e.g. away from libc_a-setjmp.o), the extract command fails quietly and the subsequent file(GLOB) catches the empty directory — but only because libc_a-setjmp.o happens to be the only member extracted. The foreach(_obj libc_a-setjmp.o) over a one-element list suggests this was meant to grow (the comment says "setjmp/longjmp objects" plural). Capture the result and fail fast on the ar invocation itself.

jhunsaker:
@claude

docker/Dockerfile
```diff
@@ -120,6 +120,46 @@ RUN cd src/rust && cargo check --all-targets --all-features
 RUN cd src/rust && cargo doc --all-features --no-deps
 RUN cd src/rust && cargo test --release --doc --all-features
 
+FROM base AS base_riscv_gnu_toolchain
+
+RUN apt-get update && apt-get install -y autoconf automake autotools-dev curl python3 python3-pip python3-tomli libmpc-dev libmpfr-dev libgmp-dev gawk build-essential bison flex texinfo gperf libtool patchutils bc zlib1g-dev libexpat-dev ninja-build git cmake libglib2.0-dev libslirp-dev libncurses-dev
```
Baltoli: Aesthetic nit - can we pull these into a dependency install script as the original Dockerfiles do? `install-zk-deps.sh` or something

docker/Dockerfile
```diff
@@ -120,6 +120,46 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
+    
+RUN make -j$(nproc)
+
+ENV PATH="/root/riscv/bin:${PATH}"
+
+WORKDIR /
+
+FROM base_riscv_gnu_toolchain AS base_zisk
+
+RUN apt-get update && apt-get install -y xz-utils jq curl build-essential qemu-system libomp-dev libgmp-dev nlohmann-json3-dev protobuf-compiler uuid-dev libgrpc++-dev libsecp256k1-dev libsodium-dev libpqxx-dev nasm libopenmpi-dev openmpi-bin openmpi-common libclang-dev clang
```
Baltoli: as above

category/core/throw.hpp
Baltoli, line 1: Should be `throw.hpp` because the macro expands to a C++ `throw` - this won't compile in C

category/vm/runtime/transmute/intrinsics.hpp
```diff
@@ -0,0 +1,50 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Host (AVX2) implementation of the runtime transmute primitive
+// uint256_load_bounded_le, backed by hand-rolled x86 assembly in
+// transmute.S.
+
+#pragma once
+
+#include <category/core/likely.h>
+#include <category/core/runtime/uint256.hpp>
```
Baltoli, line 24: needs `int.hpp` for IWYU on `load_le_unsafe` I think

category/vm/runtime/exit.cpp
```diff
@@ -30,12 +29,12 @@ namespace monad::vm::runtime
     {
         is_stack_unwinding_active = true;
         result.status = StatusCode::Error;
-        monad_vm_runtime_exit(exit_stack_ptr);
+        ::monad::vm::runtime::exit(exit_stack_ptr);
```
Baltoli, line 32: Presumably qualified because there's another `exit` with an ambiguous overload if you don't?

category/vm/runtime/exit.hpp
```diff
@@ -0,0 +1,38 @@
...
+// Host unwind primitive. On the host build the actual implementation is
+// the hand-rolled assembly in exit.S; the inline `exit` wrapper here
+// gives call sites a single typed entry point so the zkVM mirror can
+// supply an inline setjmp/longjmp variant without any extern call.
+
+#pragma once
+
+namespace monad::vm::runtime
+{
+    using exit_stack_ptr_t = void *;
```
Baltoli, line 25: How come we need this typedef? I'm going commit by commit so maybe missing something later on

category/vm/runtime/exit.cpp
goodlyrottenapple: yes

category/vm/runtime/exit.hpp
goodlyrottenapple: see the shadowed `zkvm/category/vm/runtime/exit.hpp`. logjmp uses a different ptr type (`std::jmp_buf *`)

Baltoli: Makes sense

cmake/compile_options.cmake
Baltoli, line 1: If CI isn't shouting about this, it should be - this should have a license (I can alter the check in a separate PR; haven't checked but I bet it globs for `CMakeLists.txt` and not for `*.cmake`).

docker/Dockerfile
```diff
@@ -120,6 +120,56 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
+
+ENV PATH="/root/riscv/bin:${PATH}"
+
+WORKDIR /
+
+FROM base_riscv_gnu_toolchain AS base_zisk
+
+COPY scripts/ubuntu-build/ /opt
+RUN apt-get update && \
+    /opt/install-zisk-deps.sh && \
```
mkolosick: @Baltoli how do we feel about this? Or do we want to abstract these `apt-get` blocks?

Baltoli: Seems fine to me in the scope of this PR; it matches the existing examples and we can clean it up separately if needs be.

category/core/toolchains/riscv64-elf.cmake
Baltoli, line 1: Naming nit - could probably just be `riscv64-elf.cmake`?

category/core/toolchains/riscv64-elf-toolchain.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_SYSTEM_PROCESSOR riscv64)
+set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+
+# Configurable toolchain directory — pass -DRISCV_TOOLCHAIN_DIR=<path> to cmake.
+set(RISCV_TOOLCHAIN_DIR "/opt/riscv" CACHE PATH "RISC-V toolchain directory")
+list(APPEND CMAKE_TRY_COMPILE_PLATFORM_VARIABLES RISCV_TOOLCHAIN_DIR)
+
+# Auto-detect target prefix (riscv64-none-elf- for nix, riscv64-unknown-elf- for ZisK).
```
Baltoli: We scrapped the Nix version of this build, right? Do we still need both toolchain choices?

category/core/toolchains/riscv64-elf.cmake
```diff
@@ -0,0 +1,45 @@
...
+    set(RISCV_PREFIX "riscv64-none-elf-")
+elseif(EXISTS "${RISCV_TOOLCHAIN_DIR}/bin/riscv64-unknown-elf-gcc")
+    set(RISCV_PREFIX "riscv64-unknown-elf-")
+else()
+    message(FATAL_ERROR "No riscv64 gcc found in ${RISCV_TOOLCHAIN_DIR}/bin/")
+endif()
+
+set(CMAKE_C_COMPILER "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}gcc")
+set(CMAKE_CXX_COMPILER "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}g++")
+set(CMAKE_AR "${RISCV_TOOLCHAIN_DIR}/bin/${RISCV_PREFIX}ar")
```
Baltoli, line 32: Opus 4.8 says:
>   3. Minor: `CMAKE_AR` is set but not `CMAKE_RANLIB`. CI works (CMake finds a usable ranlib), but for a hermetic cross static-lib build it'd be tidier to set the cross ranlib too. And `CMAKE_FIND_ROOT_PATH_MODE_{LIBRARY,INCLUDE} ONLY` is set with no `CMAKE_FIND_ROOT_PATH`/`CMAKE_SYSROOT` — which works only because nothing uses find_library/find_path for target deps (newlib's `libc.a` is located via `gcc -print-file-name`, not CMake find). It's effectively a "find nothing on the host" lockdown, which is defensible, just unusual.

zkvm/category/core/runtime/uint256/intrinsics.hpp
```diff
@@ -0,0 +1,44 @@
...
+    using portable::addc;
+    using portable::div;
+    using portable::mulx;
+    using portable::shld;
+    using portable::shrd;
+    using portable::subb;
+    using portable::truncating_mul;
+
+    [[gnu::always_inline]] constexpr inline uint64_t
+    force(uint64_t expr) noexcept
```
Baltoli: Tiny nitpick: `uint64_t const expr`

zkvm/category/core/runtime/non_temporal_memory.hpp
```diff
@@ -0,0 +1,33 @@
...
+{
+    inline void non_temporal_bzero(void *dest, size_t n)
+    {
+        std::memset(dest, 0, n);
+    }
+
+    inline void non_temporal_memcpy(void *dest, void *src, size_t n)
+    {
+        std::memcpy(dest, src, n);
+    }
```
Baltoli, line 32: `const` here too

zkvm/category/vm/interpreter/debug.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <evmc/evmc.h>
+
+namespace monad::vm::interpreter
+{
+#ifdef MONAD_VM_INTERPRETER_DEBUG
+    constexpr auto debug_enabled = true;
+#else
+    constexpr auto debug_enabled = false;
+#endif
```
Baltoli: Any need to actually keep this? Debug will always be disabled in ZKVM mode, right

docker/Dockerfile
```diff
@@ -120,6 +120,56 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    rm -rf /opt
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --profile minimal
+
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+RUN curl -L https://raw.githubusercontent.com/0xPolygonHermez/zisk/main/ziskup/ziskup | bash -s -- --nokey
+
+ENV PATH="/root/.zisk/bin:${PATH}"
```
Baltoli: Maybe move this out to a separate PR later? Along with the dependencies - does anything actually use this stage?

zkvm/category/vm/interpreter/trampoline.hpp
```diff
@@ -0,0 +1,46 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// zkVM mirror: the trampoline is implemented inline with setjmp/longjmp
+// instead of via the host's hand-rolled assembly (entry.S). The jmp_buf
+// lives on the caller's stack frame thanks to always_inline; longjmp
```
Baltoli: always_inline is forbidden below

docker/Dockerfile
mkolosick: Yeah that makes sense. I'll shift it over to #2288 unless we find it a better home.

Baltoli: I think that's a reasonable home for it, especially if you can cut the size of #2288 down a bit by pulling bits out

zkvm/category/vm/interpreter/trampoline.hpp
mkolosick: Not sure what to do with this one. I think the current code is fine, the documentation is just incorrect.

I think I will just change the documentation to leave a note that in the future it is potentially worth looking into how to get the stack sharing.

Baltoli: Yeah I think that's fine, agree that the code's OK as is - just noting the drift

Baltoli (dismissed):

docker/Dockerfile
```diff
@@ -120,6 +120,39 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
+COPY scripts/ubuntu-build/ /opt
+RUN apt-get update && \
+    /opt/install-riscv-deps.sh && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    rm -rf /opt
+
+RUN git clone https://github.com/riscv-collab/riscv-gnu-toolchain /riscv-gnu-toolchain
+WORKDIR /riscv-gnu-toolchain
+RUN git checkout 98c8ec4cb520443f16cdd36280b65fba3a5c5f54
```
eclark0426: Mostly just wondering, is there a reason for using a specific untagged commit, not a released tag?

zkvm/core/libstdcxx.cpp
```diff
@@ -0,0 +1,78 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Minimal C++ runtime stubs for bare-metal zkVM.
+//
+// sys_alloc_aligned(bytes, align) is supplied by the backend's Rust
```
eclark0426: `sys_alloc_aligned` is declared in multiple cpp files. Would it make sense to pull into a header so that the redeclaration isn't needed, and any future Rust-provided functions can also just be declared once in the same header?

kjcamann (dismissed):

docker/Dockerfile
Baltoli: Yeah they release tagged nightly builds; we should pick a recent one and pin there @mkolosick

Baltoli (approved):

eclark0426 (approved):

PR #2092, opened 2026-02-20
Fix a potential for race condition in execute_block_transactions

andreaslyn:
The line `txn_exec_finished.fetch_add(1, std::memory_order::relaxed);` does not generally guarantee that earlier memory operations have finished. This causes the potential for a race condition where `results[i]` is used after move. The solution presented here is to remove `txn_exec_finished` and instead use the `promises` to signal when transactions have finished.


kjcamann (approved):
LGTM. This has been dead code since 29f1c15 thanks for taking the time to remove and double-check

PR #2093, opened 2026-02-23
Refactor ExecutionEventRecorder into reusable pieces

kjcamann:
This PR was originally created to support the op-code-level EVM tracer from last year. The current need to make these changes is to refactor the event recording machinery so that it can be used in other contexts: `eth_call`, Rust testing harnesses, etc. The Rust testing harness is the immediate need.

Prior to this commit, `ExecutionEventRecorder` was a giant class that acted as `category/execution`'s interface to event recording, and it also "owned" the event ring (it was responsible for owning the open file). This commit moves the ownership responsibility into a separate class,`OwnedEventRing`.

In addition to separating the event ring and recorder, the old recording interface is split into two different classes: `EventRecorder` contains most of the actual machinery of recording, and `ExecutionEventRecorder` uses it (via private inheritance), adding in the small bit of block context data that is execution-specific.


kjcamann:
@claude review this

category/core/event/event_recorder_test.cpp
```diff
@@ -481,3 +483,131 @@ TEST_F(EventRecorderDefaultFixture, LargePayloads)
...
+    // copied up the maximum truncation size from this smaller buffer first,
+    // the library won't try to access the giant buffer
+    truncated.reserve(EventRecorder::RECORD_ERROR_TRUNCATED_SIZE);
+    for (unsigned i = 0; i < EventRecorder::RECORD_ERROR_TRUNCATED_SIZE; ++i) {
+        truncated.push_back(static_cast<uint8_t>(i));
+    }
+
+    auto ex_recorder = EventRecorder::from_event_ring(&event_ring_);
+    ASSERT_TRUE(ex_recorder);
+    std::unique_ptr<EventRecorder> recorder = std::move(*ex_recorder);
```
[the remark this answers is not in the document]
kjcamann: This has enough coverage in other places (the recording that happens in `ethereum-tests`)

cmd/monad/event.cpp
```diff
@@ -341,24 +301,121 @@ int init_execution_event_recorder(EventRingConfig const ring_config)
...
+            "could not dup(2) ring file {} fd: {} {}",
+            ring_path,
+            strerror(saved_errno),
+            saved_errno);
+        (void)unlink(ring_path);
+        monad_event_ring_unmap(&event_ring);
+        return saved_errno;
+    }
+    owned_event_ring =
+        std::make_unique<OwnedEventRing>(owned_fd, ring_path, event_ring);
```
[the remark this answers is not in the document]
kjcamann: A thrown exception here will terminate the process, which will drop the lock.

category/execution/ethereum/event/exec_event_recorder.hpp
```diff
@@ -23,74 +23,55 @@
...
-    ~ExecutionEventRecorder();
+    static std::expected<std::unique_ptr<ExecutionEventRecorder>, std::errc>
+    from_event_ring(monad_event_ring const *const ring)
+    {
+        auto t = std::make_unique<ExecutionEventRecorder>();
+        if (int const r = monad_event_ring_init_recorder(ring, &t->recorder_)) {
+            return std::unexpected(std::errc{r});
+        }
+        return t;
+    }
```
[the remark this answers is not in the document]
kjcamann: This changes in the next PR

maxkozlovsky (commented):
Nice cleanup, no correctness issues in the
usage by the current code, however some api changes introduce potential
problems. 

- `test_exec_event_recorder.cpp` deletion leaves **no unit test constructing an
  `ExecutionEventRecorder`**. The new core tests cover the generic base; the
  subclass is now untested. 
- `ExecutionEventRecorder` holds raw pointers into a mapping owned by a separate
  global (`exec_event_recorder.cpp:24-25`); safe only because of definition order.
  Pre-PR one object owned both. Could `from_event_ring` take a
  `shared_ptr<OwnedEventRing>` and drop the borrowing overload? 
- `MONAD_ASSERT(!g_exec_event_ring, "recorder initialized twice?")`
  (`cmd/monad/event.cpp:396`) checks the ring but names the recorder
- The `friend std::make_unique<X>()` definition (`event_recorder.hpp:105`,
  `exec_event_recorder.hpp:98`) doesn't restrict or help with anything as far
   as I can tell.
- `--exec-event-ring=name:0:0` now silently gets the defaults
  (`cmd/monad/event.cpp:246-251`); the parser already fills them, so these branches
  only fire on an explicit zero, which previously got a clear `ERANGE`.
- `record_{block,txn}_marker_event` changed `void` → `uint64_t` (`:88`, `:91`,
  `:204-220`); no caller uses it, no `[[nodiscard]]`, no comments documenting the value.
- New tests don't cover `MONAD_EVENT_RECORD_ERROR_OVERFLOW_EXPIRE`, whose threshold
  input this PR changed from `exec_ring_.payload_buf_mask` to
  `recorder_.payload_buf_mask`. 
- `event_recorder_test.cpp:528` and `:599-601` `memcmp` `payload_peek(...)` against
  `vlt_event.payload` — same address, so those two can't fail. 
- `event_recorder.hpp:79`, `:136-166`:  the comments refer to types outside of 
   the core module where it was moved. Comment can be shortened significantly, any
   examples can be moved into documentation sections and should be using generic types
   rather than monad transactions.

category/core/event/event_recorder.hpp
```diff
@@ -0,0 +1,251 @@
...
+        size_t original_payload_size);
+
+    friend constexpr std::unique_ptr<EventRecorder>
+    std::make_unique<EventRecorder>();
+
+    EventRecorder() noexcept = default;
+};
+
+template <
+    typename T, typename EventEnum,
```
dhil, line 112: I'm inclined to believe that, for safety, we should constrain this `T` to guard against overaligned data. It seems to me to be an invariant that the event payload is 16-byte aligned, but it is not asserted here. 

Perhaps we can just do with
```c++
static_assert(
    alignof(T) <= MONAD_EVENT_PAYLOAD_ALIGN,
    "T requires stricter alignment than event payload buffer guarantees");
```

Further, to ensure the `reinterpret_cast` at the end is sound, we should guard against overflow, e.g.
```c++
static_assert(
    sizeof(monad_event_record_error) % alignof(T) == 0,
    "RECORD_ERROR header offset must preserve T alignment");
```

kjcamann:
Thanks for the review @maxkozlovsky!

I changed some things as a result of this, but it's rebased so might hard to see. I'll explain what I did for each one.

> Nice cleanup, no correctness issues in the usage by the current code, however some api changes introduce potential problems.
> 
> * `test_exec_event_recorder.cpp` deletion leaves **no unit test constructing an
>   `ExecutionEventRecorder`**. The new core tests cover the generic base; the
>   subclass is now untested.

I'm OK with this if everyone else is. This functionality _is_ covered in testing, but by the `ethereum_test` framework, since it records execution events and does some checking of them.

> * `ExecutionEventRecorder` holds raw pointers into a mapping owned by a separate
>   global (`exec_event_recorder.cpp:24-25`); safe only because of definition order.
>   Pre-PR one object owned both. Could `from_event_ring` take a
>   `shared_ptr<OwnedEventRing>` and drop the borrowing overload?

This PR is first in a series of three that will get rid of the big globals. I thought about doing it as one giant PR, but it would be harder to review. In the interim, some of the interface gets a little uglier (like you're mentioning) but it all goes away in the PR. I am hoping to merge them both before a release is cut.

> * `MONAD_ASSERT(!g_exec_event_ring, "recorder initialized twice?")`
>   (`cmd/monad/event.cpp:396`) checks the ring but names the recorder

I think I left it because the name of the function itself is named after the recorder, even though it does both things. In the next PR, this function disappears too.

> * The `friend std::make_unique<X>()` definition (`event_recorder.hpp:105`,
>   `exec_event_recorder.hpp:98`) doesn't restrict or help with anything as far
>   as I can tell.

Without it, I get this:

```
/usr/include/c++/15/bits/unique_ptr.h:1084:30: error: 'constexpr monad::EventRecorder::EventRecorder()' is protected within this context
 1084 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
```

But this is also gone in the next PR.

> * `--exec-event-ring=name:0:0` now silently gets the defaults
>   (`cmd/monad/event.cpp:246-251`); the parser already fills them, so these branches
>   only fire on an explicit zero, which previously got a clear `ERANGE`.

I wasn't sure what to do with this one. You're right it's a little weird. Any thoughts on what it should do?

> * `record_{block,txn}_marker_event` changed `void` → `uint64_t` (`:88`, `:91`,
>   `:204-220`); no caller uses it, no `[[nodiscard]]`, no comments documenting the value.

Thank you for mentioning this. It grew out out of the old opcode tracer work. I added a comment documenting it, which is:

```c
/// Record a block-level event with no payload in one step; this returns
/// the event sequence number in case the caller wants to refer to the
/// event location, e.g., in a related data stream
uint64_t record_block_marker_event(monad_exec_event_type);
```

Unfortunately even that probably isn't very clear, because it requires the reader to know a little more about how the sequence numbers are sometimes used. Here's a longer explanation that I didn't go with, because it seemed too verbose to me:

```c
    // Return the sequence number of the marker event. The normal recording
    // process (i.e., for non-marker events) has three steps: (1) reserve an
    // event descriptor, then (2) initialize payload, then (3) commit. Because
    // control passes back to the caller after reservation, the caller has a
    // chance to inspect the event's sequence number.
    //
    // Marker events have no payload, so a marker event recording function
    // performs all recording steps, and just returns the sequence number.
    //
    // Note: (sequence number % ring_size) gives the memory location in the
    // descriptor ring where the event is located; sequence numbers are used to
    // "tie" related events across multiple rings together.
```

It shouldn't be [[nodiscard]] because the user might not care, it's not possible to know. It was added because the old EVM opcode tracer (where this refactor originally came from) did use it. It copied the transaction performance start marker event into all opcode trace events associated with that same transaction, so you could easily seek to the position in the execution event stream where the associated transaction occurred. Even though we're not merging that right now, it's a reasonable thing to want to give the caller.

> * New tests don't cover `MONAD_EVENT_RECORD_ERROR_OVERFLOW_EXPIRE`, whose threshold
>   input this PR changed from `exec_ring_.payload_buf_mask` to
>   `recorder_.payload_buf_mask`.

To be fair, they didn't before either. I agree that I should add one, but since this refactoring is taking me longer to do that I thought, do you mind if I add it later?

> * `event_recorder_test.cpp:528` and `:599-601` `memcmp` `payload_peek(...)` against
>   `vlt_event.payload` — same address, so those two can't fail.

The first one was unnecessary so I removed it. The second one, I think, is more subtle and is necessary. Here, `vlt_event.payload` is not pointing at the base peek address. This is related to how "record error events" are recorded (the truncated payload appears after an error header). I added more explanation of these to `event_recorder.md`.

> * `event_recorder.hpp:79`, `:136-166`:  the comments refer to types outside of
>   the core module where it was moved. Comment can be shortened significantly, any
>   examples can be moved into documentation sections and should be using generic types
>   rather than monad transactions.

This was the biggest change I made. The comment is very long, and it made me realize there is no good high level overview of how recording works, so I was trying to explain it _there_, which is the wrong place.

I added a lot of documentation to `event_recorder.md` explaining how it works in general, and removed that long comment.

category/core/event/event_recorder.hpp
kjcamann: Thanks for the review @dhil

For your first point, I added this in the rebase:

```cpp
static_assert(
        alignof(T) <= MONAD_EVENT_PAYLOAD_ALIGN,
        "Payload type `T` requires greater alignment than payload buffer "
        "allocator guarantees");
```

For your second point, I think we're OK though because `struct monad_event_record_error` structure be alignof(16) because the first field is, and then the `T` after it is good right? The definition is:

```c
struct monad_event_record_error
{
    alignas(16) enum monad_event_record_error_type
        error_type;                  ///< Kind of recording error that occurred
    uint16_t dropped_event_type;     ///< What kind of event was discarded
    uint32_t truncated_payload_size; ///< Size of truncated trailing payload
    uint64_t requested_payload_size; ///< Untruncated size of event payload
};
```

dhil: Yes, I believe it is OK under the current (implicit) assumptions. My point was to make them explicit to prevent a poor future programmer from running into a potential hazard.

cmd/monad/event.cpp
```diff
@@ -341,24 +301,121 @@ int init_execution_event_recorder(EventRingConfig const ring_config)
...
+// Parse a configuration string, which has the form
+//
+//   <ring-name-or-path>[:<descriptor-shift>:<buf-shift>]
+//
+// A shift can be empty, e.g., <descriptor-shift> in `my-file::30`, in which
+// case the default value is used
+std::expected<EventRingConfig, std::string>
+try_parse_event_ring_config(std::string_view const s)
+{
+    std::vector<std::string_view> tokens;
```
dhil, line 358: I'd write this function without `std::vector`, using only views to parse `s`. I suppose it is not critical here, because this is commandline tool, and the user can only mess with themselves by inputting a malformed config string with loads of colons.

dhil (dismissed):
This looks good to me now.

category/core/event/event_recorder.hpp
kjcamann: I rebased and added this:

```cpp
// In the event of a recording error, we'll record the `T` instance at
// the next address after a `struct monad_event_record_error` object;
// this ensures we won't under-align the `T{}`
static_assert(alignof(monad_event_record_error) >= alignof(T));
```

cmd/monad/event.cpp
kjcamann: @maxkozlovsky also didn't like something about this code (the way it does input validation and the default setting). To be honest I never liked it either; CLI11 boxes you in for how custom parsing is done, and I think the way this flows could be cleaner.

Is it fine if we table this though? It didn't really change much in this refactoring PR: it was not great before, it just moved around.

The reason I say this is that in my EVM opcode tracer branch, there are other kinds of rings whose configuration gets parsed. So that's the natural place to "rethink" it and clean it up. And I _do_ intend on revisiting that, even if we never actually release it in prod (I want to manually run it in a few places sometimes).

dhil: Yes absolutely fine by me.

dhil (approved):

maxkozlovsky (approved):

PR #2098, opened 2026-02-24
Inline/replace silkpre precompile gas cost functions

goodlyrottenapple:
This PR is preparatory work for removing/in-lining the silkpre precompiles library. This has been a goal for some time due to silkpre no longer being actively maintained. The current main motivation for this change, however, is due to efforts to cross-compile monad as a guest program in ZisK/SP1 zkVMs. In a followup PR (together with #2090), the calls to precompiles in a zkVM will be re-routed to ZisK/SP1/RISC0/etc. implementations via the following [interface](https://github.com/eth-act/zkvm-standards/blob/main/standards/c-interface-accelerators/zkvm_accelerators.h).

### DIFF
--- a/category/execution/ethereum/precompiles.hpp
+++ b/category/execution/ethereum/precompiles.hpp
@@ -25,6 +25,8 @@
 #include <evmc/evmc.h>
 #include <evmc/evmc.hpp>
 
+#include <bit>
+#include <cstring>
 #include <optional>
 
 MONAD_NAMESPACE_BEGIN
@@ -51,10 +53,8 @@ using precompiled_gas_cost_fn = std::optional<uint64_t>(byte_string_view);
 template <Traits traits>
 uint64_t ecrecover_gas_cost(byte_string_view);
 
-template <Traits traits>
 uint64_t sha256_gas_cost(byte_string_view);
 
-template <Traits traits>
 uint64_t ripemd160_gas_cost(byte_string_view);
 
 uint64_t identity_gas_cost(byte_string_view);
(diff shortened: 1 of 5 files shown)


category/execution/ethereum/precompiles.hpp
```diff
@@ -68,9 +68,38 @@ uint64_t ecadd_gas_cost(byte_string_view);
...
 template <Traits traits>
 uint64_t snarkv_gas_cost(byte_string_view);
 
+[[gnu::always_inline]] inline uint64_t
+blake2bf_gas_cost_ethereum(byte_string_view const input)
+{
+    // EIP-152
+    if (input.size() < 4) {
+        // blake2bf call will fail anyway
+        return 0;
```
Baltoli: We already have a mechanism for returning errors here via `std::optional` - see `expmod_gas_cost`

goodlyrottenapple: So there is a potential mismatch in the error code returned. If we set `blake2bf_gas_cost_ethereum` to return a null value, then we will potentially get an `EVMC_FAILURE` instead of the current `EVMC_PRECOMPILE_FAILURE`. Not sure if this would matter in practice?

goodlyrottenapple: If the exact error code is not important I would suggest perhaps returning `EVMC_PRECOMPILE_FAILURE` in all instances

category/execution/ethereum/precompiles.hpp
```diff
@@ -68,11 +68,40 @@ uint64_t ecadd_gas_cost(byte_string_view);
...
 uint64_t snarkv_gas_cost(byte_string_view);
 
+[[gnu::always_inline]] inline std::optional<uint64_t>
+blake2bf_gas_cost_ethereum(byte_string_view const input)
+{
+    // EIP-152, if the input isn't exactly 213 bytes
+    // the call will fail at runtime anyway
+    if (input.size() != 213) {
+        return std::nullopt;
+    }
```
goodlyrottenapple: I tightened the check as there is no point calculating the cost if the input isn't exactly 213 bytes anyway

category/execution/ethereum/precompiles.hpp
```diff
@@ -68,11 +68,40 @@ uint64_t ecadd_gas_cost(byte_string_view);
...
+
 template <Traits traits>
 uint64_t snarkv_gas_cost(byte_string_view);
 
+[[gnu::always_inline]] inline std::optional<uint64_t>
+blake2bf_gas_cost_ethereum(byte_string_view const input)
+{
+    // EIP-152, if the input isn't exactly 213 bytes
+    // the call will fail at runtime anyway
+    if (input.size() != 213) {
```
0xSqualo: @goodlyrottenapple for awareness technically this is an execution change, since some inputs will now error differently; malformed inputs with `len >= 4` but `len != 213` used to sometimes return `EVMC_OUT_OF_GAS` because gas was still parsed from the first 4 bytes and checked before execution failed, whereas now they fail immediately as `EVMC_PRECOMPILE_FAILURE`. 

It seems here, most other precompiles here still follow the “compute gas first, reject malformed input in execute” pattern generally, may be worth unifying this across all of them if the shift is to reject sooner.

goodlyrottenapple: Ah that is a good point. Perhaps best to do this shift across the board in a separate PR.

PR #2100, opened 2026-02-24
Remove State::set_balance and make selfdestruct access balances using get_balance/add/subtract only

magedm:
While updating PR 1966 (relaxed merge of balance), it occurred to me that State::set_balance shouldn't exist.
- Balance changes should always be expressed as deltas (add_to_balance/subtract_from_balance), not as absolute values. In contrast, set_balance allows direct overwrites.
- The only caller is selfdestruct, which uses set_balance(address, 0) to zero out the selfdestructing contract and can be replaced with get_balance and subtract.
- Removing set_balance and using get_balance and add/subtract instead reduces the surface area for improperly annotating balance validation requirements.

This PR:
- Replaces set_balance in selfdestruct with get_balance and subtract_from_balance.
- Removes set_balance

### DIFF
--- a/category/execution/ethereum/state3/state.cpp
+++ b/category/execution/ethereum/state3/state.cpp
@@ -359,27 +359,6 @@ void State::subtract_from_balance(
     rb_.on_debit(address);
 }
 
-void State::set_balance(Address const &address, uint256_t const &balance)
-{
-    auto &account_state = current_account_state(address);
-    auto &account = account_state.account_;
-    if (MONAD_UNLIKELY(!account.has_value())) {
-        account = Account{.incarnation = incarnation_};
-    }
-
-    uint256_t const previous_balance = account->balance;
-    account->balance = balance;
-    account_state.touch();
-    original_account_state(address).set_validate_exact_balance();
-
-    if (balance > previous_balance) {
-        rb_.on_credit(address);
-    }
-    else {
-        rb_.on_debit(address);
-    }
-}
-
 evmc_storage_status State::set_storage(
     Address const &address, bytes32_t const &key, bytes32_t const &value)
 {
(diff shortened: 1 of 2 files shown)


jhunsaker (approved):

category/execution/ethereum/state3/state.cpp
```diff
@@ -439,25 +418,27 @@ std::pair<bool, uint256_t>
 State::selfdestruct(Address const &address, Address const &beneficiary)
 {
     auto &account_state = current_account_state(address);
-    auto &account = account_state.account_;
-    MONAD_ASSERT(account.has_value());
-    auto const initial_balance = account.value().balance;
+    uint256_t const balance = get_balance(address);
```
aa755, line 421: This is correct but we can improve the performance .
Having `get_balance` here does exact validation for `address` even if we never need to use the balance, e.g. in case of Cancun + with `address==beneficiary` and non-current incarnation
So we can optimize more by moving this `get_balance` inside the branches where we actually use the returned value.

dhil: It seems `balance` is unconditionally used to construct the return value, which I believe is used to emit the transfer event.

category/execution/ethereum/state3/state.cpp
```diff
@@ -439,25 +418,27 @@ std::pair<bool, uint256_t>
...
     }
     else {
-        if (address != beneficiary || account->incarnation == incarnation_) {
-            add_to_balance(beneficiary, account.value().balance);
-            set_balance(address, 0);
+        if (address != beneficiary || is_current_incarnation(address)) {
+            if (address != beneficiary) {
+                add_to_balance(beneficiary, balance);
+            }
+            subtract_from_balance(address, balance);
```
dhil, line 434: In my opinion, the `if constexpr` disproportionately complicates the legibility of the code here.

I think this code (including the `if constexpr` above the highlighted potion) is observably equivalent to the following slightly simpler code:
```c++
if (traits::evm_rev() < EVMC_CANCUN || address != beneficiary || is_current_incarnation(address)) {
  if (address != beneficiary) {
     add_to_balance(beneficiary, balance);
  }
  subtract_from_balance(address, balance);
}
```
It may trade off some minor efficiency control in favour of simplicity, at least at `-O0`, but at `-O2` it ought to partially reduce the `if` condition expression at compile-time anyways.

EDIT: GitHub is rendering the diff strangely here; I had highlighted the entire +change. I am suggesting we can eliminate the duplicated source code, by relying a little bit on the compiler's grace.

magedm: Initially the two conditions were combined in this PR but feedback was to keep the conditions separate.

PR #2103, opened 2026-02-25
Add support for hive rpc conformance suite

es3n1n:
This PR adds support for the https://github.com/ethereum/hive rpc conformance suite. This suite requires importing specific predefined blocks and defines a custom revision schedule.

Notes:
- I had to add support for eip-7002, eip-7251 because they're relied on in their blocks and otherwise the verification will fail. ~Currently the code that handles them is gated behind a virtual function in `Chain` that returns `true` only for the `HiveNet`. This solution is somewhat not ideal because of the overhead that virtual function adds in a hot path, but I'm not sure what the best way to implement this is, though that call is done only on `trats::evm_rev() >= EVMC_PRAGUE` and I imagine on a perfectly predicted branch this wouldn't waste a lot of cycles~
  - (@Baltoli - replaced by a trait-ed implementation now)
- The scheduling / other chain config values are hardcoded to the values defined in json files (genesis alloc, chain config, evm revision schedule) in the [ethereum/execution-apis](https://github.com/ethereum/execution-apis/tree/main/tests) repo. Geth supports just straight up loading these values directly, which we can implement as well, but I tried to keep the number of changes small and I don't see much value in doing so

Rust part: https://github.com/category-labs/monad-bft/pull/2820


eclark0426 (dismissed):

category/execution/ethereum/execute_block.cpp
```diff
@@ -107,6 +111,113 @@ void set_beacon_root(State &state, BlockHeader const &header)
     }
 }
 
+template <Traits traits>
```
goodlyrottenapple: My only slight nit here would be to move this into a separate file as this code is not used in production. Perhaps mirror the `category/execution/monad/execute_system_transaction.cpp` file path?

goodlyrottenapple (dismissed):

category/execution/ethereum/chain/hive_net_alloc.hpp
```diff
@@ -0,0 +1,116 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <category/core/config.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+inline constexpr char const *HIVE_NET_ALLOC = R"(
```
Baltoli: Can we add a comment explaining / referencing where this comes from?

category/execution/ethereum/chain/hive_net.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    return 3503995874084926;
+}
+
+evmc_revision HiveNet::get_revision(
+    uint64_t const block_number, uint64_t const timestamp) const
+{
+    if (block_number >= 36) {
+        if (timestamp >= 450) {
+            return EVMC_PRAGUE;
```
Baltoli, line 36: Again here it would be good to link to the specification of these

category/execution/ethereum/process_requests.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Address const &contract_address, ChainContext<traits> const &chain_ctx)
+{
+    // as per eip-7002, eip-7251
+    constexpr auto SYSTEM_ADDRESS =
+        0xfffffffffffffffffffffffffffffffffffffffe_address;
+    if (!state.account_exists(contract_address)) {
+        return;
+    }
+
+    evmc_tx_context tx_context = {
```
Baltoli: nitpick - missing `const` here and elsewhere

category/execution/ethereum/execute_block.cpp
```diff
@@ -354,6 +359,20 @@ Result<std::vector<Receipt>> execute_block(
         process_withdrawal(state, block.withdrawals);
     }
 
+    if constexpr (
+        traits::evm_rev() >= EVMC_PRAGUE && !is_monad_trait_v<traits>) {
```
Baltoli: A further nit here on usage of the trait system; we'd probably prefer to introduce a new trait like `eip_7002_enabled` and branch on that rather than mentioning `EVMC_PRAGUE` explicitly. We prefer not to leak `is_monad_trait_v` into files in the `ethereum/` subtree.

category/execution/ethereum/process_requests.cpp
```diff
@@ -0,0 +1,147 @@
...
+        0,
+        chain_ctx};
+
+    auto const hash = state.get_code_hash(contract_address);
+    auto const code = state.read_code(hash);
+
+    // as per EIP-7002/EIP-7251,
+    //   system call failures are non-fatal and must not revert the block
+    [[maybe_unused]] auto result =
+        block_state.vm().template execute<traits>(host, &msg, hash, code);
```
Baltoli: What's the reason for accessing the VM directly rather than using `execute_call_message`?

Baltoli: https://github.com/category-labs/monad/blob/a29508657800b695110190721da333e8cff3b085/category/execution/ethereum/evm.cpp#L306

Baltoli: (I may be missing something in the context of how the request system calls are defined)

category/execution/ethereum/process_requests.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <category/vm/evm/explicit_traits.hpp>
+#include <category/vm/evm/traits.hpp>
+
+#include <intx/intx.hpp>
+
+MONAD_ANONYMOUS_NAMESPACE_BEGIN
+
+template <Traits traits>
+void system_call(
+    Chain const &chain, State &state, BlockState &block_state,
```
Baltoli: Also re. my comment on accessing the VM directly, it's a bit of an antipattern to need to pass in a `State` and a `BlockState` at the same timed - if you could use `execute_call_message` then you wouldn't need the block state to get at the VM

category/execution/ethereum/execute_block.cpp
es3n1n: Good point, moved this behind the `eip_7002_enabled` consteval function in traits. Thanks

category/execution/ethereum/process_requests.cpp
es3n1n: Added a comment with explanation why vm execute is used directly

category/execution/ethereum/process_requests.cpp
```diff
@@ -0,0 +1,198 @@
+// Copyright (C) 2025 Category Labs, Inc.
```
goodlyrottenapple:
```suggestion
// Copyright (C) 2025-26 Category Labs, Inc.
```

category/execution/ethereum/process_requests.hpp
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2025 Category Labs, Inc.
```
goodlyrottenapple:
```suggestion
// Copyright (C) 2025-26 Category Labs, Inc.
```

goodlyrottenapple: same for the new chain config files.

Baltoli:
@claude review

category/execution/ethereum/chain/hive_net.hpp
```diff
@@ -0,0 +1,32 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <category/execution/ethereum/chain/ethereum_mainnet.hpp>
+
+MONAD_NAMESPACE_BEGIN
+
+struct HiveNet : EthereumMainnet
```
Baltoli: This now overrides everything from `EthereumMainnet`, so should directly inherit from `Chain` directly I think

category/execution/ethereum/chain/genesis_state.cpp
```diff
@@ -40,19 +44,52 @@ void load_genesis_state(GenesisState const &genesis, TrieDb &db)
...
-        deltas.emplace(addr, StateDelta{.account = {std::nullopt, account}});
+        account.balance = intx::from_string<uint256_t>(value["wei_balance"]);
+
+        if (auto const it = value.find("nonce"); it != value.end()) {
+            account.nonce = std::stoull(it->get<std::string>(), nullptr, 0);
+        }
+
+        if (auto const it = value.find("code"); it != value.end()) {
+            auto const code = it->get<std::string>();
+            auto const code_bytes = evmc::from_hex(code).value();
```
Baltoli: We migrated away from `evmc::from_hex` as part of general cleanup - there's now a `monad::from_hex` instead

category/execution/ethereum/chain/genesis_state.cpp
```diff
@@ -40,19 +44,52 @@ void load_genesis_state(GenesisState const &genesis, TrieDb &db)
...
+
+        StateDelta state_delta{.account = {std::nullopt, account}};
+        if (auto const it = value.find("storage"); it != value.end()) {
+            for (auto const &storage_item : it->items()) {
+                auto const slot =
+                    evmc::from_hex<bytes32_t>(storage_item.key()).value();
+                auto const slot_value =
+                    evmc::from_hex<bytes32_t>(
+                        storage_item.value().get<std::string>())
+                        .value();
```
Baltoli: Same here

category/execution/ethereum/chain/hive_net.cpp
```diff
@@ -0,0 +1,79 @@
...
+    return EVMC_HOMESTEAD;
+}
+
+GenesisState HiveNet::get_genesis_state() const
+{
+    BlockHeader header;
+    header.difficulty = 0x20000;
+    header.gas_limit = 0x23f3e20;
+    intx::be::unsafe::store<uint64_t>(header.nonce.data(), 0x0);
+    header.extra_data = evmc::from_hex("0x68697665636861696e").value();
```
Baltoli: Again

Baltoli (dismissed):

Baltoli:
Claude comments above confirmed as addressed by @es3n1n

category/execution/ethereum/process_requests.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+MONAD_NAMESPACE_BEGIN
+
+template <Traits traits>
+Result<bytes32_t> process_requests(
+    Chain const &chain, State &state, BlockHashBuffer const &block_hash_buffer,
+    BlockHeader const &header, ChainContext<traits> const &chain_ctx)
+{
+    // EIP-7002
+    // Code is set in the hive genesis.
```
jhunsaker: this file isn't "hive" specific, right?

category/execution/ethereum/db/block_db.cpp
```diff
@@ -23,22 +23,67 @@
...
 #include <filesystem>
+#include <fstream>
 #include <string>
 #include <string_view>
 
 MONAD_NAMESPACE_BEGIN
 
-BlockDb::BlockDb(std::filesystem::path const &dir)
+BlockDb::BlockDb(
+    std::filesystem::path const &dir, std::filesystem::path const &rlp_path)
```
jhunsaker: should these just be virtual functions and this rlp code be in a derived class?

category/execution/ethereum/process_requests.cpp
es3n1n: sorry, yes, these comments were stale

category/execution/ethereum/chain/genesis_state.cpp
```diff
@@ -40,19 +43,52 @@ void load_genesis_state(GenesisState const &genesis, TrieDb &db)
...
-        deltas.emplace(addr, StateDelta{.account = {std::nullopt, account}});
+        account.balance = intx::from_string<uint256_t>(value["wei_balance"]);
+
+        if (auto const it = value.find("nonce"); it != value.end()) {
+            account.nonce = std::stoull(it->get<std::string>(), nullptr, 0);
+        }
+
+        if (auto const it = value.find("code"); it != value.end()) {
+            auto const code = it->get<std::string>();
+            // Genesis alloc is trusted input; .value() is safe.
```
dhil, line 64: I'd be inclined to put a `MONAD_ASSERT(code.has_value())` here, even though, it is coming from a trusted source.

dhil (dismissed):

dhil (approved):

Baltoli (approved):

PR #2115, opened 2026-03-03
Cleanup monad-cxx

andr-dev:


rust/crates/monad-cxx/src/error.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) 2025 Category Labs, Inc.
```
dhil: I am not sure know how pedantic we are about these headers, but the CI allows you to write the following for new files:
```suggestion
// Copyright (C) 2025-26 Category Labs, Inc.
```

rust/crates/monad-cxx/src/ffi.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[allow(non_camel_case_types)]
+pub(crate) type c_write_fn = extern "C" fn(*const monad_log, usize);
+
+#[allow(non_camel_case_types)]
+pub(crate) type c_flush_fn = extern "C" fn(usize);
+
+extern "C" {
+    pub(crate) fn monad_log_handler_create(
+        handler: *mut *mut std::ffi::c_void,
```
kjcamann: This was `*const *mut` in the original, why did it become `*mut *mut`?

rust/crates/monad-cxx/src/ffi.rs
```diff
@@ -0,0 +1,85 @@
...
+        name: *const std::ffi::c_char,
+        write: c_write_fn,
+        flush: Option<c_flush_fn>,
+        user: usize,
+    ) -> std::ffi::c_int;
+
+    pub(crate) fn monad_log_handler_destroy(handler: *mut std::ffi::c_void);
+
+    pub(crate) fn monad_log_init(
+        handlers: *mut *mut std::ffi::c_void,
```
kjcamann: These were both `*const` before. This just an array of opaque handles which aren't changed by this function.

rust/crates/monad-cxx/src/ffi.rs
andr-dev: in `category/core/log_ffi.h`:
```rust
int monad_log_handler_create(
    struct monad_log_handler**, ...
```
Thus, the FFI should use `*mut *mut`. Honestly, this should also just be bindgened.

rust/crates/monad-cxx/build.rs
```diff
@@ -13,7 +13,13 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use std::path::PathBuf;
+
+const INCLUDES: &[(&str, &[&str])] = &[("../../../", &["category/core/log_ffi.h"])];
+
 fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=wrapper.h");
```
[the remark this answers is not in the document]
andr-dev: Seems like a hallucination, `rerun-if-changed` docs clearly state directories are supported: https://docs.rs/cargo-build/latest/cargo_build/fn.rerun_if_changed.html

PR #2117, opened 2026-03-05
[core/execution] Replace intx API usage with project-owned implementa…

brett-monad:
…tions

- Add monad::to_big_endian for unsigned_integral types, implemented using std::byteswap for builtins and manual word reversal for intx multi-word types, replacing the using ::intx::to_big_endian declaration
- Qualify all remaining intx::to_big_endian calls explicitly, then migrate them to monad::to_big_endian throughout
- Replace intx::as_bytes with reinterpret_cast at all call sites


andreaslyn (approved):

category/core/int.hpp
```diff
@@ -50,6 +53,61 @@ inline constexpr uint256_t UINT256_MAX = std::numeric_limits<uint256_t>::max();
...
+        std::endian::native == std::endian::little,
+        "to_big_endian assumes little-endian platform");
+    return std::byteswap(x);
+}
+
+namespace detail
+{
+    template <unsigned_integral T>
+        requires(!std::unsigned_integral<T>)
+    [[nodiscard]] constexpr T to_big_endian_intx(T const x) noexcept
```
m-alvarez, line 69: This should probably be renamed to something like `to_big_endian_words` or similar

category/core/test/to_big_endian_test.cpp
```diff
@@ -0,0 +1,186 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
```
cmcl, line 1: Nit: Should this just be 2026?  Since it's a new file?

category/execution/ethereum/execute_block.cpp
cmcl, line 99: Accessed transitively because the `#include` was removed.

category/execution/ethereum/execute_block.cpp
```diff
@@ -52,7 +52,6 @@
 #include <boost/fiber/future/promise.hpp>
 #include <boost/outcome/try.hpp>
 #include <evmc/evmc.h>
```
cmcl, line 54:
```suggestion
#include <evmc/evmc.h>
#include <category/core/int.hpp>
```

cmcl (approved):

category/core/test/to_big_endian_test.cpp
goodlyrottenapple: I was the first one to add a new file in 2026 and this was the agreed-upon format. The licence check script will only allow `Copyright (C) 2025` or `Copyright (C) 2025-2X` format.

cmcl: Ah, I had seen that discussion but misremembered the details.

PR #2133, opened 2026-03-11
Fix compaction progress reset during statesync that set compaction head back to 0

Chen-Yifan:
This is fixed by doing upsert on top of the latest version instead of from a fresh state. Finalized update has to change to `incarnation=true`.

### DIFF
--- a/category/mpt/test/db_test.cpp
+++ b/category/mpt/test/db_test.cpp
@@ -845,24 +845,22 @@ TEST(DbTest, history_length_adjustment_never_under_min)
     Db db{machine, config};
     Node::SharedPtr root{};
 
-    constexpr unsigned nkeys = 1000;
-
-    // prepare updates with 8KB size value
-    std::deque<monad::byte_string> bytes_alloc;
-    std::deque<Update> updates_alloc;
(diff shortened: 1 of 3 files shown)


category/mpt/copy_trie.cpp
```diff
@@ -282,6 +282,9 @@ Node::SharedPtr copy_trie_to_dest(
     Node::SharedPtr dest_root, NibblesView const dest_prefix,
     uint64_t const dest_version, bool const write_root)
 {
+    auto const expected_compact_offset =
+        dest_root ? compact_offset_pair::deserialize(dest_root->value())
+                  : compact_offset_pair::deserialize(src_root->value());
```
es3n1n, line 277: the `compact_offset_pair::deserialize` call here will do `MONAD_ASSERT(bytes.size() == 2 * sizeof(uint32_t));` for non on-disk aux, which by the looks of it, will always fail

Chen-Yifan: Thanks, updated `copy_trie_to_dest()` to assert on disk aux, since that's the intended usage in the Db.

maxkozlovsky (approved):

maxkozlovsky (approved):

PR #2141, opened 2026-03-16
Improve debug output of validate_post_state

### LINKED ISSUE #2121: Improve debugging output of blockchain_test
When invoking blockchain tests with:

```
monad-ethereum-test --fork MONAD_NINE --record-exec-events --blockchain-tests ~/fixtures/
```

I would love to see some more details of what went wrong in case of test failing:

1. Comparison against expected post state in case of wrong merkle root. Right now, when there is a mismatch in post state, the test fails with `wrong merkle root` and when asserting on the post state expectations, it doesn't use the state associated with that merkle root, leaving me with no clue as to where the mismatch is.
2. Include unhashed account addresses. I include them in parenthesis in addition to db keys
3. Include unhashed storage slots alongside db keys - wherever possible, in parenthesis as above
4. Include the storage items from the db, which are absent in the expectations JSON. Unfortunately, unhashed slots are not available in this case. I considered brute force pre-imaging them, assuming in the spec tests they usually are just consecutive integers, but have so far refrained from it.

I tentatively hacked this together in [a branch for reference](https://github.com/category-labs/monad/compare/main...pdobacz:monad:pd/debug-monad-eth-test?expand=1). It seemed to work fine, I'm unsure though of the `db_post_state` change. For my cases it looked OK.

pdobacz:
Closes #2121

### DIFF
--- a/test/ethereum_test/src/blockchain_test.cpp
+++ b/test/ethereum_test/src/blockchain_test.cpp
@@ -33,6 +33,7 @@
 #include <category/execution/ethereum/chain/ethereum_mainnet.hpp>
 #include <category/execution/ethereum/core/address.hpp>
 #include <category/execution/ethereum/core/block.hpp>
+#include <category/execution/ethereum/core/fmt/address_fmt.hpp>
 #include <category/execution/ethereum/core/fmt/bytes_fmt.hpp>
 #include <category/execution/ethereum/core/receipt.hpp>
 #include <category/execution/ethereum/core/rlp/block_rlp.hpp>
@@ -59,6 +60,7 @@
 #include <category/mpt/nibbles_view.hpp>
 #include <category/vm/evm/switch_traits.hpp>
 #include <category/vm/evm/traits.hpp>
+#include <category/vm/utils/evmc_utils.hpp>
 
 #include <monad/test/config.hpp>
 
@@ -86,6 +88,7 @@
 #include <filesystem>
 #include <fstream>
 #include <optional>
+#include <set>
 #include <span>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -236,10 +239,15 @@ void validate_post_state(nlohmann::json const &json, nlohmann::json const &db)
     for (auto const &[addr, j_account] : json.items()) {
         nlohmann::json const addr_json = addr;
         auto const addr_bytes = addr_json.get<Address>();
+        auto const addr_hex = monad::vm::utils::hex_string(addr_bytes);
```
Baltoli: This is now unused, I think
```suggestion
```

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -671,6 +693,11 @@ void process_test(
             }
         }
         else {
+            // On unexpected error persist post state so that
+            // potential discrepancies in state are reported
```
dhil: I am somewhat confused about this comment.

I suppose what is happening here is that if the test failed, and it was not expected to fail, then we serialize the db to json such that we can dump it for inspection.

```suggestion
            // Error case: if this test failed unexpectedly, then serialize the db state such that we can dump it for inspection.
```

test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -253,25 +260,40 @@ void validate_post_state(nlohmann::json const &json, nlohmann::json const &db)
...
+                EXPECT_EQ(
+                    db_storage.at(db_storage_key).at("value"), expected_value)
+                    << fmt::format("{} ({})", db_storage_key, key_bytes);
+            }
+        }
+        for (auto const &[key, db_value] : db_account.at("storage").items()) {
+            auto const db_storage_value_bytes =
+                db_value.at("value").get<bytes32_t>();
+            // Above loop checked all the keys in j_account's storage,
+            // so if it contained db_storage_key, value was checked.
```
dhil: Nit:
```suggestion
            // The previous loop has already checked that for all key-value pairs of j_account whether there exist a corresponding pair in the db. So, it remains to check whether every db key has a corresponding entry in j_account.
```

PR #2142, opened 2026-03-16
Wait for predecessor tx in execute_system_transaction on early error

andreaslyn:


category/execution/monad/execute_system_transaction_test.cpp
```diff
@@ -282,3 +284,82 @@ TEST(SystemTransaction, statediff_trace_staking_epoch_change)
...
+
+    BlockHeader const header{};
+
+    NoopCallTracer noop_call_tracer;
+    trace::StateTracer noop_state_tracer = std::monostate{};
+
+    auto const tx = Transaction{.type = TransactionType::eip7702};
+
+    boost::fibers::promise<void> promise;
+    promise.set_value();
```
aa755, line 308: I agree with Copilot's [comment](https://github.com/category-labs/monad/pull/2142#discussion_r2940244123) on the testcase. The testcase passes even without the code fix. Ideally, it should fail without the fix. Perhaps this PR can be split into 2 commits. The first commit should add a testcase that fails for current code. The second commit should do the code fix. 
I asked codex-cli to do that and it produced the following 2 commits. An issue with the first commit is that the test case is sensitive to timing and thus may be flaky. But I don't see how to make it completely independent of timing, without changing the code just to write the testcase.  We can increase the 50ms wait but then we also don't want to slow down the CI.

- [tweaked testcase demonstrating failure](https://github.com/category-labs/monad/commit/b31e83fdad9d49eabf786ca2ddcdba9928157cd2)
- [your code fix](https://github.com/category-labs/monad/commit/b35eb71652d98db9cd0d418bfd727d7849c75ba5)

andreaslyn: Changing the tests as suggested by copilot will potentially guard against future regressions, but the suggestion is introducing significant complexity, which I do not think is worth it. I believe that the code changes are already covered by the two test cases, because it is not the responsibility of these units tests to verify whether `future::wait` blocks as expected.

PR #2143, opened 2026-03-17
Remove intx library from category/

brett-monad:
Replace all intx usage in category/ with monad-native types:
- Add monad::vm::runtime::uint512_t (uint512.hpp)
- Refactor uint128_t to match uint256_t idioms (static from_string, UDL in main namespace)
- Replace intx type aliases in int.hpp with vm::runtime types
- Replace intx be_load/be_store, bswap, to_big_endian, as_bytes, to_string/from_string with monad wrappers
- Replace intx arithmetic and bitwise operations with monad::vm::runtime equivalents
- Fix int_fmt.hpp: explicit uint256_t/uint512_t/uint128_t specializations replacing intx templates
- Fix precompiles_bls12: replace intx 512-bit load+compare with std::memcmp
- Remove intx link targets from category/vm/* and redundant category/execution CMakeLists
- Add fmt/quill formatters and to_string for uint128_t
- Assert base range in uint512_t::to_string to guard out-of-bounds indexing
- Restore intx cross-checks for exp, sdivrem, addmod, mulmod, slt in uint256_tests
- Use be_load in receipt.cpp set_3_bits instead of reinterpret_cast + to_big_endian
- Saturate max_gas_cost on overflow instead of wrapping
- Drop redundant monad:: namespace qualifiers from call sites inside namespace monad
- Clean up stale intx comments

Remaining: validate_transaction.cpp uses intx::uint256 at silkpre boundary (third_party change needed).


category/core/runtime/uint128.hpp
Baltoli, line 1: Can we confirm that this implementation is necessary? My understanding from talking to @m-alvarez previously is that we could get away with the compiler-supplied `uint128` types

category/core/runtime/uint512.hpp
Baltoli, line 1: Likewise here - I thought the only usage of 512-bit arithmetic was to perform an overflow check? It doesn't seem worth implementing this type if we don't need it

scripts/check-clang-tidy-changed.sh
Baltoli, line 1: If this is useful to you it could probably go in a follow-up PR rather than here

category/execution/monad/staking/fuzzer/staking_contract_model.cpp
```diff
@@ -580,7 +574,7 @@ namespace monad::staking::test
         auto const input = encoder.encode_final();
         auto res = dispatch<traits>(input, sender, value);
         if (res.has_value()) {
-            auto const rew = intx::be::load<uint256_t>(value.bytes);
+            auto const rew = be_load<uint256_t>(value.bytes);
```
Baltoli: A general point here following my other comments is that if we don't need to implement the fully-general machinery for `uint{128,512}_t`, this function might be more naturally expressed as `uint256::load_be`. This might be an easier change to make once the tests are all passing and you can refactor with confidence

Baltoli: c.f. for similar instances as well, this is just one case

category/execution/ethereum/validate_transaction.cpp
```diff
@@ -214,10 +215,25 @@ template <Traits traits>
...
+        (tx.value > UINT256_MAX - gas_cost) ? UINT256_MAX : tx.value + gas_cost;
     if (tx.type == TransactionType::eip4844) {
-        v0 += tx.max_fee_per_blob_gas * get_total_blob_gas(tx);
+        uint64_t const total_blob_gas = get_total_blob_gas(tx);
+        uint256_t const blob_cost =
+            (total_blob_gas != 0 &&
+             tx.max_fee_per_blob_gas > UINT256_MAX / total_blob_gas)
+                ? UINT256_MAX
+                : tx.max_fee_per_blob_gas * total_blob_gas;
+        v0 = (blob_cost > UINT256_MAX - v0) ? UINT256_MAX : v0 + blob_cost;
```
Baltoli: In fact this is what I believed to be the only place where we used a 512-bit operation, and you've refactored it to not require one - why do we need the 512-bit implementation?

category/execution/ethereum/precompiles_bls12.cpp
```diff
@@ -76,9 +77,8 @@ namespace bls12
         static_assert(sizeof(blst_fp) == 48);
         static constexpr std::size_t fp_encoded_offset = 16;
 
-        auto const integer_value = intx::be::unsafe::load<intx::uint512>(in);
-
-        if (MONAD_UNLIKELY(integer_value >= BASE_FIELD_MODULUS)) {
+        if (MONAD_UNLIKELY(
+                std::memcmp(in, BASE_FIELD_MODULUS_BYTES.data(), 64) >= 0)) {
```
Baltoli, line 98: Could use a comment here to explain the logic - I think the implementation is indeed what you want, but it's not immediately obvious on looking at the new code that this is just a 384-bit `>=`

category/execution/ethereum/validate_transaction.hpp
```diff
@@ -49,15 +53,36 @@ Result<void> validate_transaction(
...
-    // YP (70)
-    uint512_t v0 = tx.value + max_gas_cost(tx.gas_limit, tx.max_fee_per_gas);
+    // YP (70): total cost = value + gas_cost (+ blob_fee).
+    // If gas_limit * max_fee_per_gas overflows uint256, no balance can cover
+    // it — return InsufficientBalance. Also guard value + gas_cost overflow.
+    if (MONAD_UNLIKELY(
+            tx.max_fee_per_gas != 0 &&
+            uint256_t{tx.gas_limit} > UINT256_MAX / tx.max_fee_per_gas)) {
+        return TransactionError::InsufficientBalance;
+    }
```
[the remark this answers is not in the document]
brett-monad: they differ for historical reasons.  no reason to change for this commit

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,219 @@
...
+        result.lo = static_cast<uint64_t>(r);
+        result.hi = static_cast<uint64_t>(r >> 64);
+        return result;
+    }
+
+    [[nodiscard]] constexpr uint128_t byteswap(uint128_t x) noexcept
+    {
+        uint128_t result;
+        result.lo = __builtin_bswap64(x.hi);
+        result.hi = __builtin_bswap64(x.lo);
```
Baltoli:
```suggestion
        result.lo = std::byteswap(x.hi);
        result.hi = std::byteswap(x.lo);
```

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,219 @@
...
+    };
+
+    static_assert(sizeof(uint128_t) == 16);
+    static_assert(alignof(uint128_t) == 8);
+    static_assert(std::is_trivially_copyable_v<uint128_t>);
+
+    [[nodiscard]] constexpr bool operator==(uint128_t a, uint128_t b) noexcept
+    {
+        return a.lo == b.lo && a.hi == b.hi;
+    }
```
Baltoli:
```suggestion
    [[nodiscard]] constexpr bool operator==(uint128_t a, uint128_t b) noexcept = default;
```

Baltoli: I wonder whether `operator<=>` can be defaulted with the correct behaviour - otherwise it makes sense to provide if we're providing `operator==`.

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    [[nodiscard]] constexpr uint128_t
+    operator*(uint128_t a, uint128_t b) noexcept
+    {
+        auto const r = static_cast<unsigned __int128>(a) *
+                       static_cast<unsigned __int128>(b);
+        uint128_t result;
+        result.lo = static_cast<uint64_t>(r);
+        result.hi = static_cast<uint64_t>(r >> 64);
+        return result;
```
Baltoli: Nitpick - I'd prefer to see:
```suggestion
    return {
        .lo = static_cast<uint64_t>(r),
        .hi = static_cast<uint64_t>(r >> 64),
    };
```
as a general rule. No big impact here but I think it's cleaner and less error-prone. Same comment elsewhere

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,219 @@
...
+                    ++p;
+                }
+            }
+            return result;
+        }
+    };
+
+    static_assert(sizeof(uint128_t) == 16);
+    static_assert(alignof(uint128_t) == 8);
+    static_assert(std::is_trivially_copyable_v<uint128_t>);
```
Baltoli: I think we should also have some kind of static assert that makes sure we can round-trip between this type and the compiler intrinsic. Claude suggests something like:
```cpp
static_assert([] {
    unsigned __int128 native = (static_cast<unsigned __int128>(2) << 64) | 1;
    auto s = std::bit_cast<uint128_t>(native);
    return s.lo == 1 && s.hi == 2;
}());
```
with an appropriate comment.

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,219 @@
...
+    static constexpr bool is_specialized = true;
+    static constexpr bool is_integer = true;
+    static constexpr bool is_signed = false;
+    static constexpr int digits = 128;
+
+    static constexpr monad::vm::runtime::uint128_t max() noexcept
+    {
+        monad::vm::runtime::uint128_t r;
+        r.lo = UINT64_MAX;
+        r.hi = UINT64_MAX;
```
Baltoli:
```suggestion
        r.lo = std::numeric_limits<uint64_t>::max();
        r.hi = std::numeric_limits<uint64_t>::max();
```

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    [[nodiscard]] consteval uint128_t operator""_u128(char const *s)
+    {
+        return uint128_t::from_string(s);
+    }
+
+} // namespace monad::vm::runtime
+
+template <>
+struct std::numeric_limits<monad::vm::runtime::uint128_t>
```
Baltoli: None of them are important, but it's worth filling in the full specialisation here as `uint256_t` does, just for completeness.

category/core/runtime/uint256.hpp
```diff
@@ -222,14 +226,14 @@ namespace monad::vm::runtime
...
-    typedef __int128 int128_t;
+    typedef unsigned __int128 u128;
+    typedef __int128 i128;
 
     [[gnu::always_inline]]
     constexpr inline div_result<uint64_t>
     div_constexpr(uint64_t u_hi, uint64_t u_lo, uint64_t const v) noexcept
     {
-        auto const u = (static_cast<uint128_t>(u_hi) << 64) | u_lo;
+        auto const u = (static_cast<u128>(u_hi) << 64) | u_lo;
```
Baltoli: Maybe add a constructor to `uint128_t` that constructs from a hi / lo pair to avoid shifting and masking outside of library code

category/core/runtime/uint256.hpp
```diff
@@ -644,6 +648,20 @@ namespace monad::vm::runtime
             return load_le_unsafe(bytes).to_be();
         }
 
+        [[gnu::always_inline]]
+        static inline uint256_t be_load(uint8_t const *src) noexcept
+        {
+            return load_be_unsafe(src);
+        }
```
Baltoli: This should be `load_be` to match the existing code, right? Is there an issue with overload resolution that requires the name change?

category/core/runtime/uint256.hpp
```diff
@@ -644,6 +648,20 @@ namespace monad::vm::runtime
...
+
+        template <typename Src>
+            requires requires(Src s) { s.bytes; }
```
Baltoli: Can you use the `FixedBytes` concept in `core/hex.hpp`?

category/core/runtime/uint256.hpp
```diff
@@ -222,14 +226,14 @@ namespace monad::vm::runtime
         }
     };
 
-    typedef unsigned __int128 uint128_t;
-    typedef __int128 int128_t;
+    typedef unsigned __int128 u128;
+    typedef __int128 i128;
```
Baltoli: I think we can dispense with these typedefs now that you've implemented 128 bit integers - you'll need to change the Knuth division code as well.

Baltoli: Or, if @m-alvarez prefers to keep the implementation of that algorithm using native intrinsics, shift the typedefs to local `using uint128_native_t = ...` inside that function

category/core/runtime/uint256.hpp
```diff
@@ -1115,7 +1180,7 @@ namespace monad::vm::runtime
 
         for (int i = static_cast<int>(m - n); i >= 0; i--) {
             auto const ix = static_cast<size_t>(i);
-            uint128_t q_hat;
+            u128 q_hat;
```
Baltoli, line 1174: See above re. typedefs over native intrinsics

category/core/int.hpp
```diff
@@ -16,40 +16,157 @@
...
+#include <type_traits>
+
+static_assert(
+    std::endian::native == std::endian::little,
+    "monad integer utilities require a little-endian platform");
 
 MONAD_NAMESPACE_BEGIN
 
-using uint128_t = ::intx::uint128;
+using uint128_t = vm::runtime::uint128_t;
```
Baltoli: Seeing this typedef, I think the `vm` namespace should get dropped from the integer code because it's in `core/` now.

category/core/int.hpp
Baltoli, line 1: Why do we need the free function overloads added in this file? Is that an intx feature we were missing and use elsewhere in the code?

category/execution/ethereum/chain/genesis_state.cpp
```diff
@@ -13,21 +13,26 @@
...
 #include <category/execution/ethereum/core/receipt.hpp>
 #include <category/execution/ethereum/core/transaction.hpp>
+#include <category/execution/ethereum/core/withdrawal.hpp>
 #include <category/execution/ethereum/db/trie_db.hpp>
+#include <category/execution/ethereum/state2/state_deltas.hpp>
 #include <category/execution/ethereum/trace/call_frame.hpp>
 
-#include <evmc/evmc.hpp>
 #include <nlohmann/json.hpp>
+#include <nlohmann/json_fwd.hpp>
```
Baltoli: Why do we need `_fwd` and the full header? Surely `json.hpp` is sufficient

category/execution/ethereum/core/contract/test_abi_encode.cpp
```diff
@@ -14,18 +14,23 @@
...
 #include <category/execution/ethereum/core/address.hpp>
 #include <category/execution/ethereum/core/contract/abi_encode.hpp>
 #include <category/execution/ethereum/core/contract/big_endian.hpp>
 
-#include <evmc/evmc.hpp>
+// clang-format off
 #include <gtest/gtest.h>
-#include <intx/intx.hpp>
+#include <evmc/evmc.hpp>
+// clang-format on
```
Baltoli: Why a formatter suppression?

category/execution/ethereum/core/test/test_block_rlp.cpp
```diff
@@ -753,7 +757,7 @@ TEST(Rlp_Block, DecodeEncodeShanghai)
         0xfc, 0xe5, 0xed, 0xbc, 0x8e, 0x2a, 0x86, 0x97, 0xc1, 0x53, 0x31, 0x67,
         0x7e, 0x6e, 0xbf, 0x0b, 0x82, 0x27, 0x10};
 
-    auto const encoded_block_copy = encoded_block;
+    auto const &encoded_block_copy = encoded_block;
```
Baltoli: Nit: I presume clang-tidy has done this, but this is no longer a copy

category/execution/ethereum/core/base_ctypes.h
```diff
@@ -50,9 +50,9 @@
     using monad_c_bytes32 = std::array<std::uint8_t, 32>;
 #endif
 
-#if MONAD_CXX_CTYPES_USE_INTX
-    #include <intx/intx.hpp>
-    using monad_c_uint256_ne = intx::uint256;
+#if MONAD_CXX_CTYPES_USE_MONAD_INTS
+    #include <category/core/int.hpp>
+    using monad_c_uint256_ne = monad::uint256_t;
```
Baltoli, line 55: Let's loop in @kjcamann here to check what we want to do - preserve the enable / disable even now that we have our own integer types?

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -561,13 +574,14 @@ PrecompileResult p256_verify_execute(byte_string_view const input)
         return empty_result;
     }
 
-    Integer h(input.data(), 32);
-    Integer r(input.data() + 32, 32);
-    Integer s(input.data() + 64, 32);
-    Integer qx(input.data() + 96, 32);
-    Integer qy(input.data() + 128, 32);
+    Integer const h(
+        input.data(), 32); // NOLINT(bugprone-suspicious-stringview-data-usage)
```
Baltoli: Why is this suspicious?

category/execution/ethereum/validate_transaction.cpp
```diff
@@ -162,16 +162,23 @@ Result<void> static_validate_transaction(
...
+    static_assert(std::is_trivially_copyable_v<uint256_t>);
+    static_assert(std::is_trivially_copyable_v<intx::uint256>);
     // EIP-2
     if (MONAD_UNLIKELY(!silkpre::is_valid_signature(
-            tx.sc.r, tx.sc.s, traits::evm_rev() >= EVMC_HOMESTEAD))) {
+            std::bit_cast<intx::uint256>(tx.sc.r),
+            std::bit_cast<intx::uint256>(tx.sc.s),
+            traits::evm_rev() >= EVMC_HOMESTEAD))) {
         return TransactionError::InvalidSignature;
     }
```
Baltoli, line 201: Let's add a TODO here to remove this use of silkpre - probably out of scope for this PR

category/execution/ethereum/validate_transaction.cpp
```diff
@@ -211,6 +218,31 @@ Result<void> validate_transaction(
     return validate_ethereum_transaction<traits>(tx, sender, state);
 }
 
-EXPLICIT_EVM_TRAITS(validate_transaction);
+// EXPLICIT_EVM_TRAITS cannot be used here because validate_transaction has two
+// overloads (3-param inline and 5-param). Explicit instantiation requires the
+// full signature to disambiguate.
```
Baltoli: Why did this PR need to add a new overload?

Baltoli (dismissed):

category/execution/ethereum/validate_transaction.cpp
brett-monad: this was a byproduct of the refactor process.  fixed

category/core/runtime/uint128.hpp
brett-monad: unable to use the names as uint128_t is a non-aggregate.  added brace initializer to support this style instead

category/execution/ethereum/precompiles_impl.cpp
brett-monad: clang-tidy marks it as suspicious because input.data() isn't null-terminated and clang-tidy doesn't realize it is being used as a buffer.  The calls after that (e.g. input.data() + 32) use pointer arithmetic so clang-tidy assumes they are ok

category/core/int.hpp
```diff
@@ -16,40 +16,153 @@
...
+    T const be = bswap(x);
+    std::memcpy(dst, &be, sizeof(be));
+}
+
+// Store an integer as big-endian bytes into a new value of type DstT
+// (which must have a .bytes member of matching size).
+template <typename DstT, typename SrcT>
+[[nodiscard]] inline DstT be_store_as(SrcT const x) noexcept
+{
+    static_assert(sizeof(DstT::bytes) == sizeof(SrcT));
```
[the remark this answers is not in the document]
brett-monad: fine as written

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,248 @@
...
+                    r *= 10;
+                    if (r > uint128_max - digit) {
+                        throw std::out_of_range(s);
+                    }
+                    r += digit;
+                    result.lo = static_cast<uint64_t>(r);
+                    result.hi = static_cast<uint64_t>(r >> 64);
+                    ++p;
+                }
+            }
```
goodlyrottenapple: The intermediate `unsigned __int128 r` could be pulled outside of the loop right?

```suggestion
                unsigned __int128 r{};
                while (*p != '\0') {
                    if (*p < '0' || *p > '9') {
                        throw std::invalid_argument(s);
                    }
                    auto const digit = static_cast<uint8_t>(*p - '0');
                    if (r > max_before_mul10) {
                        throw std::out_of_range(s);
                    }
                    r *= 10;
                    if (r > uint128_max - digit) {
                        throw std::out_of_range(s);
                    }
                    r += digit;
                    ++p;
                }
                result.lo = static_cast<uint64_t>(r);
                result.hi = static_cast<uint64_t>(r >> 64);
```

category/execution/ethereum/core/contract/checked_math.cpp
```diff
@@ -29,29 +36,28 @@ MONAD_NAMESPACE_BEGIN
...
-    return res.value;
+    return x - y;
 }
 
 Result<uint256_t> checked_mul(uint256_t const &x, uint256_t const &y)
 {
-    uint512_t const z = intx::umul(x, y);
-    if (MONAD_UNLIKELY((z > UINT256_MAX))) {
+    if (MONAD_UNLIKELY(
+            x != 0 && y > std::numeric_limits<uint256_t>::max() / x)) {
```
goodlyrottenapple: would it perhaps be better to use `truncating_mul<2 * uint256_t::num_words>` to avoid the division here?

category/execution/ethereum/core/contract/checked_math.cpp
```diff
@@ -29,29 +36,28 @@ MONAD_NAMESPACE_BEGIN
 
 Result<uint256_t> checked_add(uint256_t const &x, uint256_t const &y)
 {
-    auto const res = intx::addc(x, y);
-    if (MONAD_UNLIKELY(res.carry)) {
+    uint256_t const sum = x + y;
```
goodlyrottenapple: why not keep `addc`?

category/execution/ethereum/core/signature.cpp
```diff
@@ -33,7 +35,7 @@ void SignatureAndChain::from_v(uint256_t const &v)
         auto tmp = v - 35;
         if (tmp & 1u) {
             y_parity = 1;
-            tmp ^= 1u;
```
goodlyrottenapple, line 36: this change can be reverted, as `operator⁼` has now been added

category/execution/ethereum/transaction_gas.hpp
```diff
@@ -52,10 +52,15 @@ uint256_t calculate_txn_award(
...
+inline uint256_t
 max_gas_cost(uint64_t const gas_limit, uint256_t max_fee_per_gas) noexcept
 {
-    return intx::umul(uint256_t{gas_limit}, max_fee_per_gas);
+    if (max_fee_per_gas != 0 &&
+        uint256_t{gas_limit} >
+            std::numeric_limits<uint256_t>::max() / max_fee_per_gas) {
+        return std::numeric_limits<uint256_t>::max();
+    }
+    return uint256_t{gas_limit} * max_fee_per_gas;
```
goodlyrottenapple: perhaps use `truncating_mul` here as well?

category/execution/ethereum/transaction_gas.hpp
```diff
@@ -52,10 +52,15 @@ uint256_t calculate_txn_award(
     Transaction const &, uint256_t const &base_fee_per_gas,
     uint64_t gas_used) noexcept;
 
-inline intx::uint512
+inline uint256_t
 max_gas_cost(uint64_t const gas_limit, uint256_t max_fee_per_gas) noexcept
```
goodlyrottenapple:
```suggestion
inline Result<uint256_t>
max_gas_cost(uint64_t const gas_limit, uint256_t max_fee_per_gas) noexcept
```

Could return `TransactionError::InsufficientBalance` on overflow to avoid repeated check [here](https://github.com/category-labs/monad/pull/2143/changes#diff-c780631aed6e183b7e219a84733fecc63457af8bf4b86d36f8e23b0d2f4546efR166-R170) and instead just use `BOOST_OUTCOME_TRY`?

category/execution/ethereum/validate_transaction.hpp
```diff
@@ -54,10 +58,35 @@ template <Traits traits>
...
+                    std::numeric_limits<uint256_t>::max() / total_blob_gas)) {
+            return TransactionError::InsufficientBalance;
+        }
+        uint256_t const blob_cost = tx.max_fee_per_blob_gas * total_blob_gas;
+        if (MONAD_UNLIKELY(
+                blob_cost > std::numeric_limits<uint256_t>::max() - v0)) {
+            return TransactionError::InsufficientBalance;
+        }
+        v0 += blob_cost;
     }
```
goodlyrottenapple:
```suggestion
Result<uint256_t> const v0_r = [&]() -> Result<uint256_t> {
        BOOST_OUTCOME_TRY(
            uint256_t const gas_cost,
            max_gas_cost(tx.gas_limit, tx.max_fee_per_gas));

        BOOST_OUTCOME_TRY(
            uint256_t const gas_cost_value, checked_add(tx.value, gas_cost));
        if (tx.type == TransactionType::eip4844) {
            BOOST_OUTCOME_TRY(
                uint256_t const blob_cost,
                checked_mul(tx.max_fee_per_blob_gas, get_total_blob_gas(tx)));
            return checked_add(gas_cost_value, blob_cost);
        }
        return gas_cost_value;
    }();

    if (!v0_r) {
        return TransactionError::InsufficientBalance;
    }
    auto const v0 = v0_r.assume_value();
```

category/core/int.hpp
brett-monad: did you have particular ones in mind?  I think everything in there is used somewhere

category/core/runtime/uint256.hpp
brett-monad: based on the input type I think this is correct but I may be misunderstanding the comment

category/core/runtime/uint256.hpp
brett-monad: did you mean a constructor for u128?

category/core/runtime/uint256.hpp
Baltoli: I just mean that the name is backwards - why is there both `load_be` and `be_load`?

Baltoli: It also seems fishy that this isn't an "unsafe" function but it just dispatches immediately to an unsafe function without any extra checking

category/core/runtime/uint256.hpp
Baltoli: I guess if `u128` is the native type here (on second look) then I'd probably prefer to see `static_cast<u128>(uint128_t(hi, lo))` than manual shifting

brett-monad: fixed

category/core/runtime/uint256.hpp
brett-monad: ok, I think this is fixed now

Baltoli:
@claude review

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -126,10 +127,10 @@ static evmone::test::TestState initial_state()
...
     // Genesis account with some large balance, but sufficiently small
     // so that token supply will not overflow uint256.
-    init[genesis_address] = {
-        .balance = std::numeric_limits<intx::uint256>::max() / 2,
-        .storage = {},
-        .code = {}};
+    // TestAccount::balance is intx::uint256 at the evmone boundary (external
+    // library);
+    init[genesis_address] = evmone::test::TestAccount{
+        .balance = ~intx::uint256{0} / 2, .storage = {}, .code = {}};
```
goodlyrottenapple: why not keep `std::numeric_limits<intx::uint256>::max() / 2`? seems clearer

category/execution/ethereum/validate_transaction.hpp
```diff
@@ -54,11 +59,27 @@ template <Traits traits>
...
+                    uint256_t{get_total_blob_gas(tx)},
+                    tx.max_fee_per_blob_gas));
+            return checked_add(base_cost, blob_cost);
+        }
+        return base_cost;
+    }();
+    if (MONAD_UNLIKELY(!v0_r)) {
+        return TransactionError::InsufficientBalance;
     }
+    auto const v0 = v0_r.assume_value();
```
goodlyrottenapple:
```suggestion
    auto const &v0 = v0_r.assume_value();
```

category/execution/ethereum/core/contract/checked_math.cpp
```diff
@@ -29,29 +36,34 @@ MONAD_NAMESPACE_BEGIN
 
 Result<uint256_t> checked_add(uint256_t const &x, uint256_t const &y)
 {
-    auto const res = intx::addc(x, y);
-    if (MONAD_UNLIKELY(res.carry)) {
+    // addc captures the carry directly; avoids a separate 256-bit comparison.
```
goodlyrottenapple, line 42: We could mark all these functions as noexcept right?

category/execution/ethereum/transaction_gas.hpp
```diff
@@ -52,10 +54,10 @@ uint256_t calculate_txn_award(
     Transaction const &, uint256_t const &base_fee_per_gas,
     uint64_t gas_used) noexcept;
 
-inline intx::uint512
-max_gas_cost(uint64_t const gas_limit, uint256_t max_fee_per_gas) noexcept
+inline Result<uint256_t>
+max_gas_cost(uint64_t const gas_limit, uint256_t max_fee_per_gas)
```
goodlyrottenapple: Could keep noexcept if `checked_*` are marked noexcept

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,330 @@
...
+#include <climits>
+#include <compare>
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+
+namespace monad
```
jhunsaker: use the namespace macros

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,330 @@
...
+    [[nodiscard]] constexpr uint128_t
+    operator*(uint128_t a, uint128_t b) noexcept
+    {
+        auto const r = static_cast<unsigned __int128>(a) *
+                       static_cast<unsigned __int128>(b);
+        return {static_cast<uint64_t>(r), static_cast<uint64_t>(r >> 64)};
+    }
+
+    [[nodiscard]] constexpr uint128_t
+    operator/(uint128_t a, uint64_t b) noexcept
```
jhunsaker: const params (throughout)

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,330 @@
...
+            // Step 1: 64/64 for the high word.
+            uint64_t const q_hi = a.hi / b;
+            uint64_t const r_hi = a.hi % b;
+            // Step 2: hardware 128/64 for (r_hi:a.lo) / b.
+            // r_hi = a.hi % b < b guarantees the quotient fits in 64 bits.
+            // Separate input/output constraints so r_hi can be const;
+            // the remainder output is discarded.
+            uint64_t q_lo;
+            uint64_t rem_unused;
+            asm("div %[v]"
```
jhunsaker: is asm needed? how will this work for other archs like risc-v?

goodlyrottenapple: for now we are keeping intx for the riscv build, there is too much x86 assembly in the uint256 library atm to be portable anyway.

category/core/runtime/uint128.hpp
```diff
@@ -0,0 +1,314 @@
...
+        }
+        return result;
+    }
+};
+
+// The following five asserts together guarantee that std::bit_cast between
+// uint128_t and unsigned __int128 is a lossless round-trip in both
+// directions, and that lo maps to the low 64 bits and hi to the high 64
+// bits of the native type.
+static_assert(alignof(uint128_t) == 8, "unexpected alignment of uint128_t");
```
jhunsaker: these static assert comments seems unnecessary as they are just restating the assertion

this is like assert(x == 4, "x must be 4") - the comment isn't necessary

brett-monad: There were a couple of obvious restatements that I fixed.  Left others that added information as to why, e.g. "from_string not supported for this type " or "monad integer utilities require a little-endian platform".

category/core/int.hpp
```diff
@@ -16,40 +16,132 @@
...
+    T x;
+    std::memcpy(&x, src, sizeof(x));
+    return bswap(x);
+}
+
+// Load a big-endian encoded integer from a struct with a .bytes member.
+template <typename T, FixedBytes Src>
+[[nodiscard]] inline T load_be(Src const &src) noexcept
+{
+    return load_be_unsafe<T>(src.bytes);
```
goodlyrottenapple, line 119:
```suggestion
    static_assert(sizeof(src.bytes) == sizeof(T));
    return load_be_unsafe<T>(src.bytes);
```

category/execution/ethereum/evmc_host_test.cpp
```diff
@@ -106,9 +107,9 @@ TYPED_TEST(TraitsTest, get_tx_context)
         .block_gas_limit = 50'000,
         .block_prev_randao = evmc::uint256be{10'000'000u},
     };
-    intx::be::store(ctx.chain_id.bytes, uint256_t{chain_id});
-    intx::be::store(ctx.tx_gas_price.bytes, gas_cost);
-    intx::be::store(ctx.block_base_fee.bytes, base_fee_per_gas);
+    store_be(ctx.chain_id.bytes, uint256_t{chain_id});
+    store_be(ctx.tx_gas_price.bytes, gas_cost);
+    store_be(ctx.block_base_fee.bytes, base_fee_per_gas);
```
goodlyrottenapple, line 112:
```suggestion
    ctx.chain_id = store_be_as<evmc::uint256be>(uint256_t{chain_id});
    ctx.tx_gas_price = store_be_as<evmc::uint256be>(gas_cost);
    ctx.block_base_fee = store_be_as<evmc::uint256be>(base_fee_per_gas);
```

category/execution/monad/execute_system_transaction.cpp
```diff
@@ -169,7 +179,7 @@ evmc_message ExecuteSystemTransaction<traits>::to_message() const
         .memory = nullptr,
         .memory_capacity = 0,
     };
-    intx::be::store(msg.value.bytes, tx_.value);
+    store_be(msg.value.bytes, tx_.value);
```
goodlyrottenapple, line 182: could initialise value inline via `.value = store_be_as<evmc::uint256be>(tx_.value)`

category/core/runtime/uint256.hpp
es3n1n, line 287: should this also require `sizeof(T) <= sizeof(uint64_t)` to prevent silent truncation? no caller does this right now, but it's kind of a footgun for the future (same for the variadic constructor above)

test/vm/unit/uint128_tests.cpp
```diff
@@ -0,0 +1,310 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
```
goodlyrottenapple, line 1: These and the uint256 tests should probably live in `category/core/runtime`, but I think that can be a cleanup for a followup PR

test/vm/unit/runtime/create_tests.cpp
```diff
@@ -29,8 +29,8 @@ using namespace monad::vm;
 using namespace monad::vm::runtime;
 using namespace monad::vm::compiler::test;
 
-constexpr vm::runtime::uint256_t prog = 0x63FFFFFFFF6000526004601CF3_u256;
-constexpr Address result_addr = Address{uint8_t{0x42}};
+constexpr uint256_t prog = 0x63FFFFFFFF6000526004601CF3_u256;
+constexpr evmc_address result_addr = {0x42};
```
goodlyrottenapple:
```suggestion
constexpr Address result_addr = {0x42};
```

I think this should be a `monad::Address` as it's being passed into `uint256_from_address`

category/vm/interpreter/instructions_fwd.hpp
```diff
@@ -26,432 +26,432 @@ namespace monad::vm::interpreter
     // Arithmetic
     template <Traits traits>
     MONAD_VM_INSTRUCTION_CALL void
-    add(runtime::Context &, Intercode const &, runtime::uint256_t const *,
-        runtime::uint256_t *, int64_t, uint8_t const *);
+    add(runtime::Context &, Intercode const &, uint256_t const *, uint256_t *,
+        std::int64_t, std::uint8_t const *);
```
goodlyrottenapple: Recently merged https://github.com/category-labs/monad/pull/2207 removed `std::` prefix for integer types for consistency

category/vm/interpreter/instruction_table.hpp
```diff
@@ -379,8 +379,8 @@ namespace monad::vm::interpreter
     template <uint8_t Opcode, Traits traits, typename... FnArgs>
     [[gnu::always_inline]] inline void checked_runtime_call(
         void (*f)(FnArgs...), runtime::Context &ctx, Intercode const &analysis,
-        runtime::uint256_t const *stack_bottom, runtime::uint256_t *stack_top,
-        int64_t &gas_remaining, uint8_t const *)
+        uint256_t const *stack_bottom, uint256_t *stack_top,
+        std::int64_t &gas_remaining, std::uint8_t const *)
```
goodlyrottenapple:
```suggestion
        int64_t &gas_remaining, uint8_t const *)
```

remove across the whole file

category/vm/interpreter/call_runtime.hpp
```diff
@@ -24,8 +24,8 @@ namespace monad::vm::interpreter
     template <typename... FnArgs>
     [[gnu::always_inline]]
     inline void call_runtime(
-        void (*f)(FnArgs...), runtime::Context &ctx,
-        runtime::uint256_t *&stack_top, int64_t &gas_remaining)
+        void (*f)(FnArgs...), runtime::Context &ctx, uint256_t *&stack_top,
+        std::int64_t &gas_remaining)
```
goodlyrottenapple:
```suggestion
        int64_t &gas_remaining)
```

category/execution/ethereum/execute_block.cpp
```diff
@@ -43,28 +43,31 @@
...
 #include <memory>
 #include <optional>
+#include <span>
 #include <utility>
 #include <vector>
 
 MONAD_ANONYMOUS_NAMESPACE_BEGIN
 
+using evmc::literals::operator""_address;
+
```
goodlyrottenapple:
```suggestion
using namespace monad::literals;
```

category/execution/ethereum/core/signature.hpp
```diff
@@ -36,7 +38,7 @@ struct SignatureAndChain
      */
     constexpr bool has_upper_s() const
     {
-        using namespace intx::literals;
+        using monad::literals::operator""_u256;
```
goodlyrottenapple:
```suggestion
        using namespace ::monad::literals;
```

category/execution/ethereum/core/test/test_account_rlp.cpp
```diff
@@ -28,8 +28,6 @@ using namespace monad::rlp;
 
 TEST(Rlp_Account, Encode)
 {
-    using namespace intx;
```
goodlyrottenapple, line 31: keep `using namespace monad::literals;` for `operator""_bytes32`

test/vm/unit/uint128_tests.cpp
brett-monad: created https://github.com/category-labs/monad/issues/2239 to address

category/vm/interpreter/instruction_table.hpp
```diff
@@ -405,8 +405,8 @@ namespace monad::vm::interpreter
     }
 #else
     [[gnu::always_inline]] inline void fuzz_tstore_stack(
-        runtime::Context const &, runtime::uint256_t const *,
-        runtime::uint256_t const *, uint64_t const)
+        runtime::Context const &, uint256_t const *, uint256_t const *,
+        uint64_t)
```
goodlyrottenapple:
```suggestion
        uint64_t const)
```

category/vm/interpreter/instruction_table.hpp
```diff
@@ -1783,18 +1783,17 @@ namespace monad::vm::interpreter
...
         fuzz_tstore_stack(ctx, stack_bottom, stack_top, analysis.size());
         ctx.gas_remaining = gas_remaining;
         ctx.exit(Success);
     }
 
     MONAD_VM_INSTRUCTION_CALL inline void invalid(
-        runtime::Context &ctx, Intercode const &, runtime::uint256_t const *,
-        runtime::uint256_t *, int64_t const gas_remaining, uint8_t const *)
+        runtime::Context &ctx, Intercode const &, uint256_t const *,
+        uint256_t *, int64_t gas_remaining, uint8_t const *)
```
goodlyrottenapple:
```suggestion
        uint256_t *, int64_t const gas_remaining, uint8_t const *)
```

category/execution/ethereum/block_reward.cpp
```diff
@@ -59,8 +56,9 @@ constexpr uint256_t calculate_block_reward(
     size_t const ommers_size)
 {
     MONAD_ASSERT(
-        intx::umul(ommer_reward, uint256_t{ommers_size}) <=
-        std::numeric_limits<uint256_t>::max() - reward);
+        ommers_size == 0 ||
+        ommer_reward <=
+            (std::numeric_limits<uint256_t>::max() - reward) / ommers_size);
```
goodlyrottenapple, line 62: Since this is not a debug assert, keeping mul instead of division seems more efficient?

brett-monad: Rewriting as a 256x256 mul is probably similar to a 256x64 bit div.  also code is run once per block so not on a hot path

goodlyrottenapple: ah i missed the div is smaller!

goodlyrottenapple (dismissed):

goodlyrottenapple (dismissed):

Baltoli (dismissed):

category/execution/ethereum/chain/genesis_state.cpp
```diff
@@ -53,7 +55,7 @@ void load_genesis_state(GenesisState const &genesis, TrieDb &db)
         auto const &value = item.value();
 
         Account account{};
-        account.balance = intx::from_string<uint256_t>(value["wei_balance"]);
+        account.balance = uint256_t::from_string(value["wei_balance"]);
```
jhunsaker, line 58: I think it's weird to have types which could be built-in C types to have member functions like from_string

category/execution/ethereum/chain/hive_net.cpp
```diff
@@ -72,7 +73,7 @@ GenesisState HiveNet::get_genesis_state() const
     BlockHeader header;
     header.difficulty = 0x20000;
     header.gas_limit = 0x23f3e20;
-    intx::be::unsafe::store<uint64_t>(header.nonce.data(), 0x0);
+    monad::store_be(header.nonce.data(), uint64_t{0x0});
```
jhunsaker: why is the monad namespace here

category/execution/ethereum/core/contract/checked_math.cpp
```diff
@@ -27,34 +34,42 @@
...
 }
 
-Result<uint256_t> checked_mul(uint256_t const &x, uint256_t const &y)
+Result<uint256_t> checked_mul(uint256_t const &x, uint256_t const &y) noexcept
 {
-    uint512_t const z = intx::umul(x, y);
-    if (MONAD_UNLIKELY((z > UINT256_MAX))) {
+    // Compute the full 512-bit product; overflow iff any upper word is set.
+    auto const prod =
+        truncating_mul<2 * uint256_t::num_words>(x.as_words(), y.as_words());
```
jhunsaker, line 64: similarly here - this seems like a std::numeric_limits type thing

category/execution/ethereum/core/contract/storage_array.hpp
```diff
@@ -39,7 +47,7 @@ class StorageArray
         : state_{state}
         , address_{address}
         , length_{StorageVariable<u64_be>(state, address, slot)}
-        , start_index_{intx::be::load<uint256_t>(slot) + 1}
+        , start_index_{uint256_t::load_be(slot.bytes) + 1}
```
jhunsaker, line 50: again

jhunsaker (commented):
my primary objection is that I think uint256_t should be treated like a built-in language C type like uint64_t rather than a C++ class with intermixed usage of free and member functions

category/execution/ethereum/chain/genesis_state.cpp
brett-monad: planning follow up PR to clean up uint128 and uint256 interface

category/execution/ethereum/core/contract/checked_math.cpp
brett-monad: planning follow up PR to clean up uint128 and uint256 interface

Baltoli (approved):

goodlyrottenapple (approved):

PR #2145, opened 2026-03-18
Fix geth precompile failing tests validation

goodlyrottenapple:
We were previously not checking if a non-success status code was actually returned

### DIFF
--- a/category/execution/ethereum/precompiles_test.cpp
+++ b/category/execution/ethereum/precompiles_test.cpp
@@ -33,7 +33,9 @@
 #include <filesystem>
 #include <fstream>
 #include <iterator>
+#include <limits>
 #include <string_view>
+#include <variant>
 #include <vector>
 
 #include "test_resource_data.h"
@@ -125,12 +127,15 @@ namespace
                                   "fffe5bfeffffffff00000001"})
             .value();
 
+    struct evmc_some_error
+    {
+    };
+
     struct test_case
     {
         std::string name;
         evmc::bytes input;
-        std::optional<evmc::bytes> expected;
-        std::optional<evmc_status_code> expected_failure;
+        std::variant<evmc::bytes, evmc_some_error, evmc_status_code> expected;
         int64_t gas;
         std::optional<int64_t> gas_offset;
     };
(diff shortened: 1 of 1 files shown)


category/execution/ethereum/precompiles_test.cpp
```diff
@@ -305,17 +311,31 @@ namespace
...
+                if (test_case.expected_failure) {
+                    MONAD_ASSERT(!test_case.expected);
+                    if (std::holds_alternative<evmc_status_code>(
+                            *test_case.expected_failure)) {
+                        // expecting a specific failure code
+                        EXPECT_EQ(
+                            result.status_code,
+                            std::get<evmc_status_code>(
+                                *test_case.expected_failure))
+                            << suite_name << " test case " << test_case.name;
```
Baltoli: Nitpick - you can unpack with `std::get_if` rather than do the two-step `holds_alternative` / `get` dance

category/execution/ethereum/precompiles_test.cpp
```diff
@@ -130,7 +132,8 @@ namespace
         std::string name;
         evmc::bytes input;
         std::optional<evmc::bytes> expected;
-        std::optional<evmc_status_code> expected_failure;
+        std::optional<std::variant<std::monostate, evmc_status_code>>
```
dhil: Can we simplify this to a single variant type, e.g.
```c++
std::variant<evmc::bytes, evmc_status_code> expected_outcome
```
That is, we expect either a concrete result or some status code, which can be a failure code. Then I think all the `if (test_case.expected) { .. }` below can be simplified to

```c++
if (evmc::bytes const *expected_value = std::get_if<evmc::bytes>(&test_case.expected_outcome)) {
  ...
} else {
  evmc_status_code expected = std::get<evmc_status_code>(test_case.expected_outcome);
  EXPECT_EQ(result.status_code, expected) << ...;
}
```

Or would will this have too many cascading changes throughout?

goodlyrottenapple: Good point! I amended to 
```
std::variant<evmc::bytes, evmc_some_error, evmc_status_code> expected
```

where `evmc_some_error` is just a more descriptive replacement for monostate. We need `evmc_some_error` for the `fail-*.json` tests, where we don't actually know which error code will be returned, and just want to assert that `EVMC_SUCCESS` was definitely **not** returned.

This actually highlighted two missing expected values in the `ECRECOVER_TEST_CASES`

dhil: Excellent!

PR #2147, opened 2026-03-18
[mpt] Refactor trie npending management and deduplicate upward propag…

maxkozlovsky:
…ation

Structured npending management:
- Extract maybe_expire_or_compact_child() to consolidate child dispatch logic that was duplicated across dispatch_updates_impl_, mismatch_handler_, expire_(), and compact_()
- Extract erase_child_from_parent() helpers for both UpdateTNode and UpdateExpireBase parents

Harden npending lifecycle:
- Add child_done() method with assertion that npending > 0
- Add destructor assertion that npending == 0 to catch leaks
- Use std::exchange in move constructor to zero out moved-from npending

Defer ExpireTNode creation:
- Restructure expire_() to take (parent, branch, index, node) instead of a pre-created ExpireTNode. The tnode is only created when dispatching to children, after the early-stop check. This is more consistent with how upsert_() works.
- De-template ExpireTNode constructor and make() to accept UpdateExpireBase* directly, since the type is runtime-checked anyway.

Deduplicate upward propagation:
- Extract propagate_upward() to replace 3 duplicated while-loops that walk up the tnode tree processing completed compact/expire ancestors before reaching an UpdateTNode for upward_update().

### DIFF
--- a/category/mpt/trie.cpp
+++ b/category/mpt/trie.cpp
@@ -38,6 +38,7 @@
 #include <algorithm>
 #include <bit>
 #include <cassert>
+#include <concepts>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


category/mpt/trie.cpp
```diff
@@ -1030,122 +1055,78 @@ void mismatch_handler_(
...
+                    *sm,
+                    *parent,
+                    branch,
+                    index,
+                    std::move(read_node),
+                    INVALID_OFFSET,
+                    false);
+                propagate_upward(*aux, *sm, parent);
             },
             node_offset}; // NOLINT(clang-analyzer*.Malloc)
```
Chen-Yifan: nit: can remove the NOLINT

category/mpt/trie.cpp
```diff
@@ -1298,20 +1253,17 @@ void compact_(
...
-        else {
-            --tnode->npending;
-        }
+        auto child_ptr = node.move_next(j);
+        maybe_expire_or_compact_child(
+            aux,
+            sm,
+            *tnode,
+            j,
+            0u,
```
Chen-Yifan: perhaps add comment `// unused` around the 0 dead value.

PR #2161, opened 2026-03-25
Cut out duplicated code from `vm/`

Baltoli:


test/vm/vm/test_vm.hpp
```diff
@@ -52,9 +52,7 @@ class BlockchainTestVM : public evmc_vm
     };
 
     template <typename V>
-    using CodeMap = std::unordered_map<
-        evmc::bytes32, V, monad::vm::utils::Hash32Hash,
-        monad::vm::utils::Bytes32Equal>;
```
dhil, line 57: Are the default parameters for the hash function and equality function equivalent to the previously handwritten ones?

PR #2164, opened 2026-03-26
Rationalise use of `bytes32_t`

Baltoli:


category/vm/runtime/transmute.hpp
```diff
@@ -59,7 +60,7 @@ namespace monad::vm::runtime
     }
 
     [[gnu::always_inline]]
-    inline evmc::bytes32 bytes32_from_uint256(uint256_t const &x)
+    inline bytes32_t bytes32_from_uint256(uint256_t const &x)
     {
         evmc_bytes32 ret;
```
goodlyrottenapple: nit:
```suggestion
        bytes32_t ret;
```

category/vm/runtime/context.cpp
```diff
@@ -98,8 +99,8 @@ namespace monad::vm::runtime
                     .depth = msg->depth,
                     .recipient = msg->recipient,
                     .sender = msg->sender,
-                    .value = msg->value,
-                    .create2_salt = msg->create2_salt,
+                    .value = static_cast<bytes32_t>(msg->value),
+                    .create2_salt = static_cast<bytes32_t>(msg->create2_salt),
```
dhil, line 103: Shouldn't these be `static_cast<evmc::bytes32>` as the definition of `Environment` (type of field `env`) is:
```c++
 struct Environment
    {
        ...
        evmc::bytes32 value;
        evmc::bytes32 create2_salt;
        ...
     }
```

goodlyrottenapple: those types have been changed as well: https://github.com/category-labs/monad/pull/2164/changes#diff-938a85cf5ff8ad63326dfd79348091a7eea55948a056b80defe8b28a8239da9eR57

PR #2170, opened 2026-04-02
[zkVM] Partial trie DB

goodlyrottenapple:
~~Depends on https://github.com/category-labs/monad/pull/2222~~
~~Depends on #2240~~

This adds a new simple implementation of a "partial" MPT, i.e. some child nodes only contain a keccak hash of the node instead of the actual subtree. This Db back-end will be used by the zkVM guest program compiled to RISCV.
This PR introduces the core data-structure as well as the decoder from reth's RLP encoded execution witness. The witness is encoded as an RLP list and has the following form:

| Index | Field | Type | Description |
|-------|-------|------|-------------|
| 0 | `block_rlp` | `string` | RLP-encoded block |
| 1 | `pre_state_root` | `bytes32` | State root before execution (for debugging only, will likely be removed) |
| 2 | `post_state_root` | `bytes32` | State root after execution |
| 3 | `[node...]` | `list<string>` | Flat, unordered trie node preimages; `keccak256(node_i) == key` |
| 4 | `[code...]` | `list<string>` | Contract bytecodes; `keccak256(code_i) == code_hash` |
| 5 | `[key...]` | `list<string>` | 20-byte addresses or 32-byte slots (un-hashed preimages) |
| 6 | `[header...]` | `list<string>` | RLP-encoded ancestor headers |

All the overridden virtual methods (besides `state_root`) currently return dummy values and will be properly implemented in followup PR(s)

### DIFF
--- a/category/execution/ethereum/db/partial_trie_db.cpp
+++ b/category/execution/ethereum/db/partial_trie_db.cpp
@@ -1,0 +1,454 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/core/byte_string.hpp>
+#include <category/core/bytes.hpp>
+#include <category/core/cases.hpp>
+#include <category/core/config.hpp>
+#include <category/core/keccak.hpp>
+#include <category/execution/ethereum/core/rlp/account_rlp.hpp>
+#include <category/execution/ethereum/core/rlp/bytes_rlp.hpp>
+#include <category/execution/ethereum/db/partial_trie_db.hpp>
+#include <category/execution/ethereum/db/util.hpp>
+#include <category/execution/ethereum/rlp/decode.hpp>
+#include <category/mpt/merkle/compact_encode.hpp>
+#include <category/mpt/nibbles_view.hpp>
+#include <category/vm/code.hpp>
+
+#include <ankerl/unordered_dense.h>
+
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <memory>
+#include <span>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace
+{
+
+    // Forward declaration for mutual recursion.
+    template <typename T>
+    Result<ChildRef<T>> decode_child_ref(
+        rlp::RlpType ty, byte_string_view &enc, mpt::NibblesView path,
+        NodeIndex const &nodes);
+
+    template <typename T>
+    byte_string encode_child_ref(ChildRef<T> const &child);
+
+    // ensures enc.empty() if successful
+    template <typename T>
+    Result<PartialNode<T>> decode_partial_node(
+        byte_string_view &enc, mpt::NibblesView curr_path,
+        NodeIndex const &nodes)
+    {
+        if (MONAD_UNLIKELY(enc.empty())) {
+            return rlp::DecodeError::InputTooShort;
+        }
+
+        {
+            // Make a copy of enc in case we need to backtrack in the branch
+            // case below
+            byte_string_view short_node_enc{enc};
+            BOOST_OUTCOME_TRY(
+                auto key_enc, rlp::parse_metadata(short_node_enc));
+            BOOST_OUTCOME_TRY(
+                auto val_enc, rlp::parse_metadata(short_node_enc));
+
+            if (short_node_enc.empty()) {
+                enc = short_node_enc;
+                // Extension/Leaf node
+                if (MONAD_UNLIKELY(key_enc.first != rlp::RlpType::String)) {
+                    return rlp::DecodeError::TypeUnexpected;
+                }
+
+                BOOST_OUTCOME_TRY(
+                    auto decoded_pair, mpt::compact_decode(key_enc.second));
+                auto const &[decoded_path, is_leaf] = decoded_pair;
+                mpt::Nibbles full_path =
+                    mpt::concat(curr_path, mpt::NibblesView{decoded_path});
+
+                if (MONAD_UNLIKELY(full_path.nibble_size() > 64)) {
+                    return rlp::DecodeError::PathTooLong;
+                }
+
+                if (is_leaf) {
+                    if (MONAD_UNLIKELY(val_enc.first != rlp::RlpType::String)) {
+                        return rlp::DecodeError::TypeUnexpected;
+                    }
+                    BOOST_OUTCOME_TRY(
+                        auto value, T::decode(val_enc.second, nodes));
+                    MONAD_DEBUG_ASSERT(val_enc.second.empty());
+                    return PartialNode<T>{
+                        LeafData<T>{std::move(decoded_path), std::move(value)}};
+                }
+                else {
+                    BOOST_OUTCOME_TRY(
+                        auto child,
+                        decode_child_ref<T>(
+                            val_enc.first,
+                            val_enc.second,
+                            mpt::NibblesView{full_path},
+                            nodes));
+                    MONAD_DEBUG_ASSERT(val_enc.second.empty());
+                    return PartialNode<T>{ExtensionData<T>{
+                        std::move(decoded_path), std::move(child)}};
+                }
+            }
+        }
+
+        // Branch node: re-parse from the beginning since we consumed k and v
+        // above
+        BranchData<T> branch{};
+
+        if (MONAD_UNLIKELY(curr_path.nibble_size() + 1 > 64)) {
+            return rlp::DecodeError::PathTooLong;
+        }
+
+        for (unsigned i = 0; i < 16; ++i) {
+            BOOST_OUTCOME_TRY(auto child_enc, rlp::parse_metadata(enc));
+            mpt::Nibbles child_path =
+                mpt::concat(curr_path, static_cast<unsigned char>(i));
+            BOOST_OUTCOME_TRY(
+                auto child,
+                decode_child_ref<T>(
+                    child_enc.first,
+                    child_enc.second,
+                    mpt::NibblesView{child_path},
+                    nodes));
+            branch.children[i] = std::move(child);
+        }
+        BOOST_OUTCOME_TRY(auto v_enc, rlp::decode_string(enc));
+        if (!v_enc.empty()) {
+            BOOST_OUTCOME_TRY(auto value, T::decode(v_enc, nodes));
+            MONAD_DEBUG_ASSERT(v_enc.empty());
+            branch.value = std::move(value);
+        }
+        if (MONAD_UNLIKELY(!enc.empty())) {
+            return rlp::DecodeError::InputTooLong;
+        }
+        return PartialNode<T>{std::move(branch)};
+    }
+
+    // ensures enc.empty() if successful
+    template <typename T>
+    Result<ChildRef<T>> decode_child_ref(
+        rlp::RlpType ty, byte_string_view &enc, mpt::NibblesView path,
+        NodeIndex const &nodes)
+    {
+        if (ty == rlp::RlpType::List) {
+            // Embedded node (inline RLP list). Ethereum requires the *full* RLP
+            // encoding of an inline node to be < 32 bytes; since `enc` is the
+            // list payload (header already stripped), its size must be <= 30.
+            if (MONAD_UNLIKELY(enc.size() > 30)) {
+                return rlp::DecodeError::InputTooLong;
+            }
+            // Manual expansion of BOOST_OUTCOME_TRY to avoid a named
+            // PartialNode<T> local that triggers a GCC false-positive
+            // -Wmaybe-uninitialized on the unique_ptr inside AccountLeafValue.
+            auto node_result_ = decode_partial_node<T>(enc, path, nodes);
+            if (MONAD_UNLIKELY(node_result_.has_failure())) {
+                return std::move(node_result_).as_failure();
+            }
+            MONAD_DEBUG_ASSERT(enc.empty());
+            return std::make_unique<PartialNode<T>>(
+                std::move(node_result_).assume_value());
+        }
+        else {
+            // String: either empty (null slot) or 32-byte hash reference
+            if (enc.empty()) {
+                return nullptr;
+            }
+            if (MONAD_UNLIKELY(enc.size() < 32)) {
+                return rlp::DecodeError::InputTooShort;
+            }
+            if (MONAD_UNLIKELY(enc.size() > 32)) {
+                return rlp::DecodeError::InputTooLong;
+            }
+            bytes32_t hash{};
+            std::memcpy(hash.bytes, enc.data(), 32);
+            enc = enc.substr(32);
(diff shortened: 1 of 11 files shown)


category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -0,0 +1,419 @@
...
+    std::vector<Receipt> const &, std::vector<std::vector<CallFrame>> const &,
+    std::vector<Address> const &, std::vector<Transaction> const &,
+    std::vector<BlockHeader> const &,
+    std::optional<std::vector<Withdrawal>> const &)
+{
+}
+
+void PartialTrieDb::apply_inserts_and_updates(StateDeltas const &) {}
+
+void PartialTrieDb::apply_deletions(StateDeltas const &) {}
```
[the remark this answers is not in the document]
goodlyrottenapple: will be implemented in a followup PR

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -0,0 +1,419 @@
...
+}
+
+std::optional<Account> PartialTrieDb::read_account(Address const &)
+{
+    return std::nullopt;
+}
+
+bytes32_t
+PartialTrieDb::read_storage(Address const &, Incarnation, bytes32_t const &)
+{
```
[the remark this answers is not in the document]
goodlyrottenapple: will be implemented in a followup PR

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    PartialTrieDb db;
+    db.root_ = std::move(root_node);
+    db.codes_ = std::move(code_index);
+
+    return db;
+}
+
+std::optional<Account> PartialTrieDb::read_account(Address const &)
+{
```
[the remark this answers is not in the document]
goodlyrottenapple: will be implemented in a followup PR

category/execution/ethereum/db/test/test_partial_trie_db.cpp
```diff
@@ -0,0 +1,251 @@
...
+{
+    // Seed with a fixed value so the test is deterministic.
+    std::mt19937_64 rng{0xc0ffee'dead'beef'13ULL};
+    std::uniform_int_distribution<unsigned int> byte_dist{0, 255};
+
+    NodeIndex nodes;
+
+    for (int i = 0; i < 10000; ++i) {
+        bytes32_t val{};
+        std::generate(std::begin(val.bytes), std::end(val.bytes), [&] {
```
dhil, line 173: I think we should roundtrip the two extreme cases manually (unless the fixed seed already exercises them), i.e. all `0x00`s and `0xFF`s.

dhil (dismissed):

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +56,108 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
-        length = enc[0] - 0xc0;
-    }
-    else {
-        size_t const length_of_length = enc[0] - 0xf7;
-
-        if (MONAD_UNLIKELY(i + length_of_length >= enc.size())) {
-            return DecodeError::InputTooShort;
-        }
+    RlpType type;
+    return parse_metadata_raw(enc, type);
```
Baltoli: I see from the control flow that the call to `parse_metadata_raw` should never produce `RlpType::List` here, but maybe it's worth an assert + comment to note explicitly that we know from L141 that the input was a string? Similarly for the list case below

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +56,108 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
-        end = i + 1;
-    }
-    else if (enc[0] < 0xb8) // [0x80, 0xb7]
-    {
-        ++i;
-        uint8_t const length = enc[0] - 0x80;
-        end = i + length;
-    }
-    else // [0xb8, 0xbf]
+namespace
```
Baltoli: Preferable to put this in `detail::` rather than an anonymous namespace as it's in a header file

category/execution/ethereum/db/util.hpp
```diff
@@ -167,6 +167,8 @@ load_header(mpt::Node::SharedPtr, mpt::Db &, BlockHeader const &);
 
 mpt::Nibbles proposal_prefix(bytes32_t const &);
 
+bytes32_t to_bytes32(mpt::Nibbles const &);
```
Baltoli: This new public API will be used in a follow up PR?

category/execution/ethereum/db/partial_trie_db.hpp
```diff
@@ -0,0 +1,196 @@
...
+    using Variant =
+        std::variant<BranchData<V>, ExtensionData<V>, LeafData<V>, HashStub>;
+
+    Variant v;
+
+    PartialNode() = default;
+
+    template <class T>
+        requires std::constructible_from<Variant, T>
+    PartialNode(T &&x)
```
Baltoli:
```suggestion
    explicit PartialNode(T &&x)
```

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -0,0 +1,444 @@
...
+            branch.value = std::move(value);
+        }
+        if (MONAD_UNLIKELY(!enc.empty())) {
+            return rlp::DecodeError::InputTooLong;
+        }
+        return branch;
+    }
+
+    // ensures enc.empty() if successful
+    template <class T>
```
Baltoli: Very nitpicky:
```suggestion
    template <typename T>
```

Baltoli: Just for stylistic consistency. There's no functional effect here

category/execution/ethereum/db/partial_trie_db.hpp
```diff
@@ -0,0 +1,196 @@
...
+MONAD_NAMESPACE_BEGIN
+
+/// An unresolved subtree whose contents are absent from the witness.
+struct HashStub
+{
+    bytes32_t hash;
+};
+
+template <typename V>
+struct PartialNode;
```
Baltoli: Perhaps overkill but a concept for `PartialTrieLeaf` would be nice documentation:
```cpp
  template <typename T>
  concept PartialTrieLeaf = requires(byte_string_view &enc, NodeIndex const &nodes, T const &v) {
      { T::decode(enc, nodes) } -> std::same_as<Result<T>>;
      { T::encode(v) } -> std::same_as<byte_string>;
  };
```

Baltoli: Then, e.g.:
```cpp
template <PartialTrieLeaf V>
struct BranchData
```

category/execution/ethereum/db/partial_trie_db.hpp
```diff
@@ -0,0 +1,196 @@
...
+        requires std::constructible_from<Variant, T>
+    PartialNode(T &&x)
+        : v(std::forward<T>(x))
+    {
+    }
+};
+
+using NodeIndex = ankerl::unordered_dense::map<bytes32_t, byte_string>;
+
+struct StorageLeafValue : bytes32_t
```
Baltoli: I'd probably prefer to implement the leaf values via explicit composition rather than inheritance - there are fewer slicing footguns if you do that, at the expense of more verbosity at usage sites (fine I think as constrained to this module)

category/execution/ethereum/db/partial_trie_db.hpp
```diff
@@ -0,0 +1,196 @@
...
+    AccountTrie root_;
+    CodeIndex codes_;
+    uint64_t block_number_{0};
+    BlockHeader last_committed_header_{};
+    bytes32_t receipts_root_{};
+    bytes32_t transactions_root_{};
+    std::optional<bytes32_t> withdrawals_root_{};
+
+public:
+    PartialTrieDb() = default;
```
Baltoli: Is there an interface requirement for a `Db` to be default-constructible? Would be good to tighten this and make `from_reth_witness` the only way to obtain one of these objects if possible

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +56,108 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
+{
+    [[gnu::always_inline]] inline constexpr Result<byte_string_view>
+    parse_metadata_raw(byte_string_view &enc, RlpType &type)
```
Baltoli: This function is kind of tricky to follow - is it possible to massage a little bit to a flatter structure with the branches on enc[0] all at the same level?

category/execution/ethereum/rlp/execution_witness.hpp
```diff
@@ -0,0 +1,56 @@
...
+    byte_string_view block_rlp;
+    bytes32_t pre_state_root;
+    bytes32_t post_state_root;
+    byte_string_view encoded_nodes;
+    byte_string_view encoded_codes;
+    byte_string_view encoded_headers;
+};
+
+Result<ExecutionWitness>
+parse_execution_witness(std::span<uint8_t const> witness_bytes);
```
Baltoli: Why not just take a byte_string_view directly? It seems like there's some back and forth between that and span<uint8_t>?

category/execution/ethereum/rlp/decode.hpp
Baltoli: I also don't love the by-ref-param return - why not just return a pair?

goodlyrottenapple: The second return value by ref seemed to be a convention in the rlp decode code, e.g. https://github.com/category-labs/monad/blob/121b0d3ffcfcbb8eae637064c8900fb2eb6ada10/category/execution/ethereum/core/rlp/account_rlp.cpp#L46 returns both the account and the storage root. Happy to switch to a pair...

category/execution/ethereum/db/util.hpp
goodlyrottenapple: good catch, i will remove for now. I don't think i will actually need it...

category/execution/ethereum/rlp/decode.hpp
goodlyrottenapple: I added a debug assert just before the return in both instances where parse_metadata_raw is used

goodlyrottenapple:
@claude review plz

goodlyrottenapple:
Addressing clade's review:

> #### 1. `compact_buf[33]` is safe today but fragile

added checks on the length of the path which throw a new `rlp::DecodeError::PathTooLong`.

> #### 2. `BranchData` value encoding is missing the RLP string wrapper

don't think this needs further comments.

> #### 3. `decode_partial_node` short-node detection is clever but deserves a comment

added a slight clarification about making a copy of enc to be able to backtrack in the else branch. the else brach aready has a comment which i think is fairly descriptive

> #### 5. `encode_child_ref` re-hashing vs the original preimage

plan on adding more comprehensive test in a followup PR which test constructing a PartialTrieDb with a bunch of accounts and then serialisingand deserialising. THis will be easier in the followup because we will have upsert/lookup functions

> #### 6. Minor: `commit()` signature doesn't match `Db` interface

addressed in a followup PR

> #### 7. Minor: `from_reth_witness` eagerly compiles all code

The witness should only contain code that will be executed so it makes sense to load it eagerly.. there is no upside to lazy loading afaics

> #### 8. Test coverage observations
Added more tests for error paths in decode_child_ref tests for a non-trivial triedb encode/decode will be in the followup as noted above...

category/execution/ethereum/rlp/decode.hpp
Baltoli: If it's consistent then fine, no issue

Baltoli (dismissed):

Baltoli (dismissed):

category/execution/ethereum/db/partial_trie_db.hpp
```diff
@@ -0,0 +1,204 @@
...
+/// An unresolved subtree whose contents are absent from the witness.
+struct HashStub
+{
+    bytes32_t hash;
+};
+
+using NodeIndex = ankerl::unordered_dense::map<bytes32_t, byte_string>;
+
+template <typename T>
+concept PartialTrieLeaf =
```
anne-laure-s: Nit: I would call this `PartialTrieValue`, `PartialLeafValue` or `LeafValue` to reflect the value inside the leaf, and not the leaf itself

anne-laure-s (dismissed):
The code seems fine to me.

As a reviewer, I would have prefered the unrelated changes (ie. nibbles_view.hpp or decode.hpp) in other commits or PRs

anne-laure-s (dismissed):

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -0,0 +1,456 @@
...
+PartialTrieDb::read_storage(Address const &, Incarnation, bytes32_t const &)
+{
+    return {};
+}
+
+vm::SharedIntercode PartialTrieDb::read_code(bytes32_t const &code_hash)
+{
+    auto it = codes_.find(code_hash);
+    if (it == codes_.end()) {
+        return {};
```
cmcl: Probably addressable in a follow-up PR but:
```suggestion
        return vm::make_shared_intercode({});
```

matches `TrieDB::read_code`, and prevents `BlockState::read_code` from raising an assert on `NULL_HASH`  lookup.

category/mpt/merkle/compact_encode.hpp
```diff
@@ -58,4 +60,51 @@ compact_encode_len(unsigned const si, unsigned const ei)
...
+    size_t const nibble_count = (enc.size() - 1) * 2 + static_cast<size_t>(odd);
+
+    // A non-terminating (extension) node with an empty path is structurally
+    // invalid — compact_encode asserts against it, so reject here to keep
+    // decode/encode symmetric.
+    if (MONAD_UNLIKELY(nibble_count == 0 && !terminating)) {
+        return std::nullopt;
+    }
+
+    Nibbles result{nibble_count};
```
cmcl, line 103: A malformed input to `compact_decode` can exceed 255 nibbles after decoding and cause an abort here when constructing  `Nibbles`.  This is not caught by the existing `PathTooLong` tests, which only cover paths that fit in Nibbles.  I annotated `test_partial_trie_db.cpp` with a minimal reproducer.  The fix needs both a bounds check here and a way to preserve `PathTooLong`, since `decode_partial_node` currently treats every `std::nullopt` from `compact_decode` as `InputTooShort`.

category/execution/ethereum/db/test/test_partial_trie_db.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+    byte_string const leaf_rlp = rlp::encode_list2(
+        rlp::encode_string2(compact_sv),
+        rlp::encode_string2(
+            byte_string_view{encoded_val.data(), encoded_val.size()}));
+
+    auto result = try_decode_node(leaf_rlp);
+    ASSERT_TRUE(result.has_error());
+    EXPECT_EQ(result.error(), rlp::DecodeError::PathTooLong);
+}
```
cmcl, line 389: see comment on `compact_encode.hpp`:

```suggestion
}

TEST(PartialTrieDb, PathTooLongOver255CompactPath)
{
    byte_string compact_key(129, 0x00);
    compact_key[0] = 0x20; // leaf node, even path length

    AccountLeafValue const val{.account = {.nonce = 1}};
    byte_string const encoded_val = AccountLeafValue::encode(val);

    byte_string const leaf_rlp = rlp::encode_list2(
        rlp::encode_string2(
            byte_string_view{compact_key.data(), compact_key.size()}),
        rlp::encode_string2(
            byte_string_view{encoded_val.data(), encoded_val.size()}));

    ASSERT_DEATH((void)try_decode_node(leaf_rlp), "");
}
```

Generated by Codex GPT 5.4

cmcl (approved):

Baltoli (approved):

PR #2180, opened 2026-04-08
Rewrite EXPLICIT_*_TRAITS_MEMBER macros for clang-21 compatibility

Baltoli:
## Summary

- Replaces the variable template + `decltype` pattern in `EXPLICIT_*_TRAITS_MEMBER` macros with explicit instantiation of a local function template using the member pointer as an NTTP
- The old pattern put `&Class::member<traits>` in a variable template initializer, which [temp.spec.general]/6 specifically does NOT exempt from access checking in explicit instantiation declarations — clang-21 enforces this
- The new pattern places the member pointer in the NTTP of an explicit instantiation declaration, where access checking IS relaxed per the standard
- No changes needed to any class headers — the fix is entirely in `explicit_traits.hpp`

## Test plan

- [x] Full build passes (GCC 15, RelWithDebInfo)
- [x] Full build passes (Clang 19, Debug + MONAD_COMPILER_TESTING)
- [x] clang-tidy clean (138/138 files)
- [x] Trait instantiation check passes (no overlaps)

### DIFF
--- a/category/execution/monad/reserve_balance/reserve_balance_contract.hpp
+++ b/category/execution/monad/reserve_balance/reserve_balance_contract.hpp
@@ -23,6 +23,7 @@
 #include <category/execution/ethereum/core/contract/big_endian.hpp>
 #include <category/execution/ethereum/state3/state.hpp>
 #include <category/execution/ethereum/trace/call_tracer.hpp>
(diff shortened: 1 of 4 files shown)


Baltoli:
@claude review

PR #2185, opened 2026-04-09
Improve triedb rust driver FFI and error handling

andr-dev:


rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -102,6 +104,12 @@ int triedb_read(
     triedb *db, uint8_t const *const key, uint8_t const key_len_nibbles,
     uint8_t const **value, uint64_t const block_id)
 {
+    *value = nullptr;
+
+    if (db == nullptr || value == nullptr) {
+        return -3;
```
0xSqualo: Is the `nullptr` assignment intentional here @andr-dev ?

andr-dev: Haha yeah this looks a bit silly, was working on it a bit later into the night 😅 . Fixed!

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -14,19 +14,57 @@
...
+    /// Traversal finished early (error or early exit)
+    FinishedEarly,
+}
+
+impl TraverseCallbackKind {
+    /// Convert from C enum to Rust enum
+    pub(crate) fn from_c(kind: triedb_async_traverse_callback) -> Option<Self> {
+        #[allow(non_upper_case_globals)]
+        match kind {
+            triedb_async_traverse_callback_triedb_async_traverse_callback_value => {
```
maxkozlovsky: the repeated name looks strange, may be we should fix it in the future.

rust/crates/monad-triedb/src/lib.rs
```diff
@@ -139,48 +147,55 @@ unsafe extern "C" fn traverse_callback(
...
-        let mut lock = traverse_context.data.lock().expect("mutex poisoned");
-        let _ = traverse_context
-            .sender
-            .send(Some(std::mem::take(&mut *lock)));
-        // traverse_context is freed here, because we don't call Box::into_raw
+    let context = context as *mut TraverseContext;
+
+    let Some(op_kind) = TraverseCallbackKind::from_c(op_kind) else {
+        error!(
+            "traverse_callback: unexpected op_kind value: {}",
```
maxkozlovsky, line 158: Might be better to signal the error explicitly to the caller instead of dropping the callback.

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -14,19 +14,57 @@
...
+    Value,
+    /// Traversal completed normally
+    FinishedNormally,
+    /// Traversal finished early (error or early exit)
+    FinishedEarly,
+}
+
+impl TraverseCallbackKind {
+    /// Convert from C enum to Rust enum
+    pub(crate) fn from_c(kind: triedb_async_traverse_callback) -> Option<Self> {
```
maxkozlovsky: may be more idiomatic to implement this as TryFrom trait. Since it is standard trait, the reader already knows what to expect and how the errors are reported without reading the body of implementation.

rust/crates/monad-triedb/src/lib.rs
```diff
@@ -139,48 +147,55 @@ unsafe extern "C" fn traverse_callback(
...
+    };
 
-    let key = unsafe { std::slice::from_raw_parts(key_ptr, key_len).to_vec() };
-    let value = unsafe { std::slice::from_raw_parts(value_ptr, value_len).to_vec() };
+    match op_kind {
+        TraverseCallbackKind::FinishedEarly => {
+            let ctx = unsafe { Box::from_raw(context) };
+            let _ = ctx.sender.send(None);
+        }
+        TraverseCallbackKind::FinishedNormally => {
```
maxkozlovsky, line 170: Could be single match statement instead of if + match, 
```
match from_c(op_kind) {
Some(FinishedEarly) => ...
...
None => 
}
```

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -63,23 +63,25 @@ struct triedb
 int triedb_open(
     char const *dbdirpath, triedb **db, uint64_t const node_lru_max_mem)
 {
-    if (db == nullptr || *db != nullptr) {
+    if (dbdirpath == nullptr || db == nullptr || *db != nullptr) {
```
maxkozlovsky, line 68: The error handling is inconsistent (returns -1 here, -3 in another function) and uses magic constants. Would be nice to define symbolic names.

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -63,23 +63,25 @@ struct triedb
...
-        std::error_code ec;
+    else if (!ec) {
         for (auto const &file :
              std::filesystem::directory_iterator(dbdirpath, ec)) {
             paths.emplace_back(file.path());
         }
-        if (ec) {
-            return -2;
-        }
+    }
```
maxkozlovsky, line 83: ec is silently convered to -2 without log message. The caller will not be able to figure out what exactly went wrong here in case of an error.

maxkozlovsky (commented):
Left some comments. Recommed fixing silent failure and error code inconsistency, the rest are suggestions.

rust/crates/monad-triedb/src/lib.rs
andr-dev: Dropping the callback signals through an error on the sender side, I haven't looked at how it is used in `monad-bft` so didn't want to change that just yet.

rust/crates/monad-triedb/src/ffi.rs
andr-dev: I dislike `TryFrom` for ffi/bindgen types because `triedb_async_traverse_callback` is actually a type alias to `c_uint` so you'd actually end up allowing a `TraverseCallbackKind` to be created from all types of things without explicitly stating `TraverseCallbackKind::from_c`.

rust/crates/monad-triedb/src/lib.rs
andr-dev: Yep, although I think the let with return is a bit cleaner. It's not an if statement :)

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
andr-dev: Agreed, I think generally speaking we should come up with a nice system for handling errors across all of our FFI boudaries.

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
andr-dev: Chatted offline, agreed to print the error in a log message.

maxkozlovsky (dismissed):

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
Chen-Yifan: echoing on this, we can improve the clarity of the return int using a enum error type in C then bindgen to import to Rust

Chen-Yifan (dismissed):

Chen-Yifan (approved):

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
andr-dev: Agreed offline to do this in the next commit, we'll make a custom error type, nicely wrap all the ffi functions with their errors, and try to keep as many unsafe blocks in the `ffi.rs` module.

maxkozlovsky (approved):

PR #2190, opened 2026-04-10
DbCache: Support reading finalized state older than latest finalized block

Chen-Yifan:
- Fix `Proposals::try_read` assert-fail when reading older than the latest 
   finalized block. Although this scenario doesn't occur in the current                                                     
    execution runloop, `DbCache` should support any read that `mpt::Db` can.                                                         
- Replace the assertion with explicit checks before the chain walk, set correct 
   value to `truncated` so DbCache looks up in LRU cache or fall through to 
   the trie correctly.
- The added test lines in `proposal_basics` work only after the change.


Chen-Yifan:
The bug reported by @claude  is invalid: all proposals earlier or equal to the finalized block are already erased from the proposal_map_. Reading of block <= last_finalized_block_ should skip the loop into proposal chain.

maxkozlovsky (approved):

magedm (approved):
Looks good.

PR #2194, opened 2026-04-10
[event] round up payload size before checking for overflow

kjcamann:
This fixes a bug where certain large payload sizes could trigger `MONAD_ASSERT(event != nullptr)`. If a payload size were below 4 GiB before rounding up for alignment, but equal to 4 GiB after rounding, then:

- The pre-alignment check in exec_event_recorder.hpp would pass
- The post-alignment check in monad_event_recorder_reserve would fail and return nullptr

This commit makes both checks the same, preventing `event == nullptr` as intended. This is not severe since it is not possible to manufacture such a large payload with the current gas model.

While here, add some drive-by `MONAD_UNLIKELY` notations.

### DIFF
--- a/category/execution/ethereum/event/exec_event_recorder.hpp
+++ b/category/execution/ethereum/event/exec_event_recorder.hpp
@@ -27,6 +27,7 @@
 #include <category/core/config.hpp>
 #include <category/core/event/event_recorder.h>
 #include <category/core/event/event_ring.h>
+#include <category/core/mem/align.h>
 #include <category/execution/ethereum/event/exec_event_ctypes.h>
 
 #include <array>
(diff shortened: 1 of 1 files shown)


category/execution/ethereum/event/exec_event_recorder.hpp
```diff
@@ -206,8 +207,15 @@ ReservedExecEvent<T> ExecutionEventRecorder::reserve_block_event(
...
+    // payload_size is the true size requested, and alloc_payload_size is the
+    // aligned size actually allocated from the event ring. The former is
+    // recorded in the event descriptor, but the latter must be used for the
+    // various size overflow checks.
     size_t const payload_size = (size(trailing_bufs) + ... + sizeof(T));
-    if (payload_size > std::numeric_limits<uint32_t>::max()) [[unlikely]] {
+    size_t const alloc_payload_size =
+        monad_round_size_to_align(payload_size, MONAD_EVENT_PAYLOAD_ALIGN);
+    if (alloc_payload_size > std::numeric_limits<uint32_t>::max())
+        [[unlikely]] {
```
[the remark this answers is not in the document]
kjcamann: I will do this in PR https://github.com/category-labs/monad/pull/2093

Baltoli, line 218: Does this code deliberately avoid pulling in `MONAD_UNLIKELY`?

kjcamann: I wasn't really trying to _avoid_ pulling it in, but `MONAD_UNLIKELY` is ugly for a number of reasons (more parentheses, it's a macro, etc.) We've had the language standard `[[unlikely]]` attribute since C++20 if you're going to stay in C++.

`[[unlikely]]` is not supported in C, so you still need `likely.h` in C translation units and in header files with inline functions that can be included by C. Most of my code is in C code and uses `likely.h`, but this header is C++ only.

kjcamann: Changed to `MONAD_UNLIKELY` for consistency with the rest of the code

Baltoli (approved):

jhunsaker (approved):

PR #2201, opened 2026-04-15
[vm/utils] clang tidy: missing `const`

dhil:
Clang tidy flagged this missing `const` in `evm-as/builder.hpp` in an unrelated patch (https://github.com/category-labs/monad/pull/2060). I am submitting the fix as a separate patch to avoid cluttering the other patch.

### DIFF
--- a/category/vm/utils/evm-as/builder.hpp
+++ b/category/vm/utils/evm-as/builder.hpp
@@ -123,7 +123,7 @@ namespace monad::vm::utils::evm_as
 
         EvmBuilder &push(runtime::uint256_t const &imm) noexcept
         {
-            size_t n = byte_width(imm);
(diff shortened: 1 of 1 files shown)


Baltoli (approved):

Baltoli:
I presume that somehow the header wasn't being included in any TUs checked by clang-tidy prior to this test? If #2060 builds test inputs with evm-as then that makes sense to me

dhil:
> I presume that somehow the header wasn't being included in any TUs checked by clang-tidy prior to this test? If #2060 builds test inputs with evm-as then that makes sense to me

Yes, I think your hypothesis is correct. I am using `evm-as` there to build smart contracts that are being used in the simulation tests.

goodlyrottenapple (approved):

PR #2205, opened 2026-04-15
[zkVM] Implement read and commit functionality in PartialTrieDb

goodlyrottenapple:


test/ethereum_test/src/blockchain_test.cpp
```diff
@@ -301,9 +303,177 @@ void validate_post_state(nlohmann::json const &json, nlohmann::json const &db)
...
+
+// Verifies that a PartialTrieDb pre-state matches the TrieDb pre-state.
+//
+// tdb_pre_state is the output of tdb.to_json() captured immediately before the
+// block was executed, i.e. the TrieDb view of the same pre-state the witness
+// encodes. Every account and storage slot in the TrieDb is looked up via
+// read_account / read_storage on the PartialTrieDb and compared.
+//
+// Returns true if verification passes.
+bool verify_witness(
```
goodlyrottenapple, line 367: This code is only for debugging and shouldn't be called in CI, but might nonetheless be useful to keep behind a flag for diagnosing witness mismatches.

category/execution/ethereum/db/partial_trie_db.hpp
```diff
@@ -168,9 +165,19 @@ class PartialTrieDb final : public Db
 
     std::optional<Account> read_account(Address const &) override;
 
+    /// Like read_account, but returns std::nullopt (outer) when the lookup
+    /// hits an unresolved HashStub.
+    std::optional<std::optional<Account>>
+    read_account_maybe(Address const &addr);
```
goodlyrottenapple, line 171: This method, together with `read_storage_maybe`, are only needed for debug and should **not** be called in zkvm execution. They are used in `verify_witness`. If we decide to drop that method, we should remove these too.

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +269,320 @@ namespace
...
+                    return branch->value ? &*branch->value : nullptr;
+                }
+                unsigned const nibble = remaining.get(0);
+                remaining = remaining.substr(1);
+                node = branch->children[nibble].get();
+            }
+            else {
+                // HashStub: subtree not present in witness
+                MONAD_ASSERT(false);
+            }
```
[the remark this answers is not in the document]
goodlyrottenapple: no this is intentional. a bad witness must abort the computation, especially inside a zkvm

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +269,320 @@ namespace
...
+        }
+        return nullptr; // null child — key absent
+    }
+
+    unsigned common_prefix_length(mpt::NibblesView a, mpt::NibblesView b)
```
Baltoli:
```suggestion
    unsigned common_prefix_length(mpt::NibblesView const a, mpt::NibblesView const b)
```

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +269,320 @@ namespace
...
+    template <bool override = true, typename T>
+    T &trie_get_or_upsert(ChildRef<T> &root, mpt::NibblesView key, T value)
+    {
+        if (!root) {
+            root = std::make_unique<PartialNode<T>>(
+                LeafData<T>{mpt::Nibbles{key}, std::move(value)});
+            return std::get<LeafData<T>>(root->v).value;
+        }
+
+        if (auto *leaf = std::get_if<LeafData<T>>(&root->v)) {
```
Baltoli: Weakly I'd prefer to split this function out into smaller cases - you could have an overload set of:
```cpp
T &trie_get_or_upsert(LeafData<T>&, mpt::NibblesView key, T value) { ... }
T &trie_get_or_upsert(ExtensionData<T>&, mpt::NibblesView key, T value) { ... }
```
etc., and dispatch with `std::visit` rather than `get_if`?

Baltoli: Might need some thought to get the recursion scheme right, but it would make the individual cases in the recursion simpler if we could do it

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +269,320 @@ namespace
...
+            if (mpt::NibblesView{leaf->path} == key) {
+                root.reset();
+                return true;
+            }
+            return false;
+        }
+        else if (auto *ext = std::get_if<ExtensionData<T>>(&root->v)) {
```
Baltoli: Same comment here about breaking the recursion scheme up

category/execution/ethereum/db/partial_trie_db.cpp
goodlyrottenapple: Can do this, however, may be a bit ugly because we have to also pass the `root` ref to the overloaded functions, since we may be overriding the value stored inside, e.g. if a leave turns into a branch after insert, etc... Should not be a perf issue however if we mark all the new functions as always inline

goodlyrottenapple: Actually, I found a nice way to merge half of both leaf and ext code into a single function this way.

category/execution/ethereum/db/partial_trie_db.cpp
goodlyrottenapple: I switched to std::visit and pulled out some common functionality into a helper function but didn't hink it made sense to split the case bodies into separate functions. Let me know what you think.

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +270,349 @@ namespace
...
+        return n;
+    }
+
+    // Insert or update a leaf in the trie when override = true.
+    // Get or insert when override = false.
+    template <bool override = true, typename T>
+    T &trie_get_or_upsert(ChildRef<T> &root, mpt::NibblesView key, T value);
+
+    template <bool override, typename T>
+    [[gnu::always_inline]] inline T &trie_get_or_upsert_at_new_branch(
```
anne-laure-s, line 325: It could be useful to make the invariant `common_prefix <= key.nibble_size()` explicit or documented

anne-laure-s (dismissed):

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +270,351 @@ namespace
...
+                    }
+                    if (child_count == 0 && has_value) {
+                        root = std::make_unique<PartialNode<T>>(LeafData<T>{
+                            mpt::Nibbles{}, std::move(*branch.value)});
+                        return DeleteResult::Compressed;
+                    }
+                    if (child_count == 1 && !has_value) {
+                        auto &child = branch.children[single_idx];
+                        MONAD_ASSERT(
+                            !std::holds_alternative<HashStub>(child->v));
```
cmcl, line 596: Can witness generation produce a sparse account trie where deletion leaves a branch with exactly one remaining child and that child is a `HashStub`? In that case we are not traversing into the unresolved subtree; we only need to preserve its known hash by replacing the branch with an extension to the hash stub. The current assertion aborts under that scenario even though the post-delete state root is still computable. The regression test below captures the failure.

```c++
namespace
{
    bytes32_t hash_rlp(byte_string const &rlp)
    {
        return to_bytes(keccak256(byte_string_view{rlp.data(), rlp.size()}));
    }

    byte_string encode_hash_ref(bytes32_t const &hash)
    {
        return rlp::encode_string2(byte_string_view{hash.bytes, 32});
    }

    byte_string
    make_account_leaf_node(mpt::NibblesView leaf_path, Account const &acct)
    {
        unsigned char compact_buf[33];
        auto const compact_sv = mpt::compact_encode(
            compact_buf, leaf_path, /*terminating=*/true);

        byte_string const encoded_acct =
            AccountLeafValue::encode(AccountLeafValue{.account = acct});
        return rlp::encode_list2(
            rlp::encode_string2(compact_sv),
            rlp::encode_string2(byte_string_view{
                encoded_acct.data(), encoded_acct.size()}));
    }

    byte_string make_branch_with_two_children(
        unsigned slot1,
        bytes32_t const &child1_hash,
        unsigned slot2,
        bytes32_t const &child2_hash)
    {
        byte_string body;
        for (unsigned i = 0; i < 16; ++i) {
            if (i == slot1) {
                body += encode_hash_ref(child1_hash);
            }
            else if (i == slot2) {
                body += encode_hash_ref(child2_hash);
            }
            else {
                body += rlp::EMPTY_STRING;
            }
        }
        body += rlp::EMPTY_STRING; // value slot
        return rlp::encode_list2(body);
    }

    byte_string make_extension_to_hash_stub(
        unsigned slot, bytes32_t const &hash)
    {
        mpt::Nibbles path{1};
        path.set(0, static_cast<unsigned char>(slot));

        unsigned char compact_buf[2];
        auto const compact_sv = mpt::compact_encode(
            compact_buf, mpt::NibblesView{path}, /*terminating=*/false);
        return rlp::encode_list2(
            rlp::encode_string2(compact_sv), encode_hash_ref(hash));
    }

    Result<PartialTrieDb> db_from_rlp_nodes(
        bytes32_t const &root_hash, std::initializer_list<byte_string> nodes)
    {
        byte_string encoded_nodes;
        for (auto const &node : nodes) {
            encoded_nodes += rlp::encode_string2(
                byte_string_view{node.data(), node.size()});
        }

        return PartialTrieDb::from_witness(
            root_hash,
            byte_string_view{encoded_nodes.data(), encoded_nodes.size()},
            {});
    }

    int delete_account_with_hash_stub_sibling_exit_code()
    {
        Account const acct{.balance = 1, .nonce = 1};

        auto const key_hash = keccak256(ADDR_X.bytes);
        mpt::NibblesView const key{key_hash};
        unsigned const leaf_slot = key.get(0);
        unsigned const stub_slot = (leaf_slot + 1) % 16;
        constexpr auto stub_hash =
            0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_bytes32;

        byte_string const leaf_rlp =
            make_account_leaf_node(key.substr(1), acct);
        byte_string const root_rlp = make_branch_with_two_children(
            leaf_slot, hash_rlp(leaf_rlp), stub_slot, stub_hash);

        auto result =
            db_from_rlp_nodes(hash_rlp(root_rlp), {root_rlp, leaf_rlp});
        if (result.has_error()) {
            return 1;
        }

        auto db = std::move(result.value());
        commit_with(
            db,
            BlockHeader{.number = 1},
            StateDeltas{
                {ADDR_X, StateDelta{.account = {acct, std::nullopt}}}});

        bytes32_t const expected_root =
            hash_rlp(make_extension_to_hash_stub(stub_slot, stub_hash));
        return db.state_root() == expected_root ? 0 : 2;
    }

} // namespace

TEST(PartialTrieDb, Commit_DeleteAccountWithHashStubSibling_CompressesToExtension)
{
    EXPECT_EXIT(
        std::exit(delete_account_with_hash_stub_sibling_exit_code()),
        ::testing::ExitedWithCode(0),
        "");
}
```

goodlyrottenapple: Good spot. We could indeed relax that requirement when the hash-stub points to a branch node. And in fact with the constraint in place a truly minimal witness would get rejected. On the other hand, having the assert there makes it easier to hunt down bugs in the witness gen, because we hit a hard assert instead of a state root mismatch due to incorrect tree structure (which is harder to trace), so my inclination would be to perhaps leave it in for now?

cmcl: I haven't looked at [witness generation](https://github.com/category-labs/monad/pull/2256) in depth yet. To clarify: correct witness generation prevents the above scenario by emitting the surviving sibling node, not only its hash, whenever deletion may trigger branch compression.  So the child is decoded as a concrete node rather than a generic `HashStub`? If so, happy for the assert to remain. It might be worth adding a short comment near the assert documenting that invariant.

goodlyrottenapple: Yes that is correct. We would get incorrect behaviour if the single surviving child was a leaf or an extension (and we failed to emit it in the witness), so early on I wanted to ensure that we always emit the single surviving child unconditionally. Will add a comment explaining this here.

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +270,368 @@ namespace
...
+                        root, std::move(branch), cp, key, std::move(value));
+                },
+                [&](BranchData<T> &branch) -> T & {
+                    return trie_get_or_upsert_at_branch<override>(
+                        branch, key, std::move(value));
+                },
+                [](HashStub &) -> T & {
+                    // HashStub — witness incomplete
+                    MONAD_ASSERT(false);
+                    std::unreachable();
```
dhil: Can we replace this with
```suggestion
                    MONAD_ABORT("Incomplete witness);
```
? I think `MONAD_ABORT` should propagate `noreturn` by virtue of invoking `abort()` in tail position, thus this lambda should type check successfully.

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -269,6 +270,368 @@ namespace
...
+                        }
+
+                        return DeleteResult::Compressed;
+                    }
+                    return DeleteResult::Unchanged;
+                },
+                [](HashStub &) -> DeleteResult {
+                    // HashStub — witness incomplete
+                    MONAD_ASSERT(false);
+                    std::unreachable();
```
dhil: Same thing here, if it works
```suggestion
                    MONAD_ABORT("Incomplete witness);
```

dhil (approved):

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -446,9 +813,71 @@ void PartialTrieDb::set_block_and_prefix(
 }
 
 void PartialTrieDb::commit(
-    bytes32_t const &, CommitBuilder &, BlockHeader const &,
-    std::unique_ptr<StateDeltas>, std::function<void(BlockHeader &)>)
+    bytes32_t const &, CommitBuilder &, BlockHeader const &header,
```
cmcl, line 816: Ignoring `CommitBuilder` looks intentional, but that means `codes_` can become out of date with respect to commits. Are `PartialTrieDb`s only used per-block and never used for post-commit reads?

goodlyrottenapple: yeah partialtriedb is only meant to be used to execute a single block. atm, all we care about is that the computed state root hash at the end of executing the block after a commit() is the correct one.

cmcl (approved):

Baltoli (approved):

PR #2209, opened 2026-04-16
[zkvm] Precompiles refactor

dhil:


dhil:
The diff is kind of pointless here as the target has diverted from `main`. I'll retarget to main instead, though, I am not seeking to get this patch included in `main` at the moment.

dhil:
@goodlyrottenapple The actual patch starts on commit https://github.com/category-labs/monad/pull/2209/commits/fa1e9a533f8beb236ecebcf08e67f7443c653526. Each precompile refactor has its own commit, so it is probably best reviewed on a commit-by-commit basis.

(the base of this patch is #2090)

goodlyrottenapple:
Might it be worth rolling most of the new `*_execute` functions into a templated generic one?

```cpp
enum class PrecompileImplOpts
{
    FreeOnNull,
    SuccessOnNull,
};

template <typename>
struct impl_out_size;

template <size_t N>
struct impl_out_size<
    PrecompileImplResult (*)(byte_string_view, std::span<uint8_t, N>)>
{
    static constexpr size_t value = N;
};

template <auto Impl, PrecompileImplOpts... Opts>
static PrecompileResult alloc_and_execute(byte_string_view const input)
{
    constexpr bool free_on_null =
        ((Opts == PrecompileImplOpts::FreeOnNull) || ...);
    constexpr bool success_on_null =
        ((Opts == PrecompileImplOpts::SuccessOnNull) || ...);

    constexpr size_t N = impl_out_size<decltype(Impl)>::value;
    auto *const out = static_cast<uint8_t *>(std::malloc(N));
    MONAD_ASSERT(out != nullptr);
    auto const result = Impl(input, std::span<uint8_t, N>{out, N});

    if (result.data == nullptr) {
        MONAD_DEBUG_ASSERT(result.size == 0);
        if constexpr (free_on_null) {
            std::free(out);
        }
        if constexpr (success_on_null) {
            return {EVMC_SUCCESS, nullptr, 0};
        }
        return PrecompileResult::failure();
    }
    return {EVMC_SUCCESS, result.data, result.size};
}
```

then in cases, have e.g.:

```cpp
CASE(0x02, sha256_gas_cost, alloc_and_execute<sha256_impl>);
...
CASE(
    0x09, blake2bf_gas_cost<traits>,
    alloc_and_execute<blake2bf_impl, PrecompileImplOpts::FreeOnNull>);
...
CASE(
    0x0100, p256_verify_gas_cost,
    alloc_and_execute<
        p256_verify_impl, PrecompileImplOpts::FreeOnNull,
        PrecompileImplOpts::SuccessOnNull>
...

**Comment by dhil** [comment]
> Might it be worth rolling most of the new `*_execute` functions into a templated generic one?
> 
> ```c++
> enum class PrecompileImplOpts
> {
>     FreeOnNull,
>     SuccessOnNull,
> };
> 
> template <typename>
> struct impl_out_size;
> 
> template <size_t N>
> struct impl_out_size<
>     PrecompileImplResult (*)(byte_string_view, std::span<uint8_t, N>)>
> {
>     static constexpr size_t value = N;
> };
> 
> template <auto Impl, PrecompileImplOpts... Opts>
> static PrecompileResult alloc_and_execute(byte_string_view const input)
> {
>     constexpr bool free_on_null =
>         ((Opts == PrecompileImplOpts::FreeOnNull) || ...);
>     constexpr bool success_on_null =
>         ((Opts == PrecompileImplOpts::SuccessOnNull) || ...);
> 
>     constexpr size_t N = impl_out_size<decltype(Impl)>::value;
>     auto *const out = static_cast<uint8_t *>(std::malloc(N));
>     MONAD_ASSERT(out != nullptr);
>     auto const result = Impl(input, std::span<uint8_t, N>{out, N});
> 
>     if (result.data == nullptr) {
>         MONAD_DEBUG_ASSERT(result.size == 0);
>         if constexpr (free_on_null) {
>             std::free(out);
>         }
>         if constexpr (success_on_null) {
>             return {EVMC_SUCCESS, nullptr, 0};
>         }
>         return PrecompileResult::failure();
>     }
>     return {EVMC_SUCCESS, result.data, result.size};
> }
> ```
> 
> then in cases, have e.g.:
> 
> ```c++
> CASE(0x02, sha256_gas_cost, alloc_and_execute<sha256_impl>);
> ...
> CASE(
>     0x09, blake2bf_gas_cost<traits>,
>     alloc_and_execute<blake2bf_impl, PrecompileImplOpts::FreeOnNull>);
> ...
> CASE(
>     0x0100, p256_verify_gas_cost,
>     alloc_and_execute<
>         p256_verify_impl, PrecompileImplOpts::FreeOnNull,
>         PrecompileImplOpts::SuccessOnNull>
> ...
> ```

Yes. I think this might be a good way to cut down the boilerplate. I will look into this.

**Comment by mkolosick** [comment]
I dug more into the precompiles I hadn't looked at yet, and I think that template generic'ing is actually premature here. I think there are more opportunities to follow the pattern in `ecrecover_execute` of pulling the common error handling/parsing up into the `_execute` layer *once we've removed silkpre*. I started doing this for the other precompiles but too many of them are tied up to internal silkpre implementation at the moment, so I pulled back on that for now.

In light of that I'm just cleaning up the commits and rebasing my other changes and then I would be good with the commit.

File: zkvm/core/zkvm_exit.h [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef ZKVM_EXIT_H
+#define ZKVM_EXIT_H
+
+#if defined(MONAD_ZKVM_SP1)
+[[noreturn]] void syscall_halt(unsigned char exit_code);
+#endif
```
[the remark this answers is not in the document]
mkolosick: This is from a different pr.

zkvm/core/libstdcxx.cpp
```diff
@@ -0,0 +1,78 @@
...
+// operator new / delete
+[[gnu::always_inline]] static inline void *alloc_or_exit(std::size_t size)
+{
+    if (size == 0) {
+        zkvm_exit(1);
+    }
+    void *ptr = sys_alloc_aligned(size, 16);
+    if (!ptr) {
+        zkvm_exit(1);
+    }
```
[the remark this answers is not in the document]
mkolosick: applies to a different pr

zkvm/category/vm/runtime/transmute/intrinsics.hpp
```diff
@@ -0,0 +1,39 @@
...
+    [[gnu::always_inline]]
+    inline uint256_t
+    uint256_load_bounded_le(uint8_t const *const bytes, int64_t const max_len)
+    {
+        if (MONAD_LIKELY(max_len >= 32)) {
+            return uint256_t::load_le_unsafe(bytes);
+        }
+        uint256_t v{0};
+        std::memcpy(v.as_bytes(), bytes, max_len);
+        return v;
```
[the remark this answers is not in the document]
mkolosick: applies to another pr

zkvm/core/libstdcxx.cpp
```diff
@@ -0,0 +1,78 @@
...
+    return ptr;
+}
```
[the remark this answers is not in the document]
mkolosick: relevant to different pr

zkvm/CMakeLists.txt
```diff
@@ -0,0 +1,170 @@
...
+            WORKING_DIRECTORY "${_newlib_dir}"
+        )
+    endforeach()
+    file(GLOB _libc_objs "${_newlib_dir}/*.o")
+    if(NOT _libc_objs)
+        message(FATAL_ERROR "Failed to extract setjmp/longjmp objects from ${_libc_a}")
+    endif()
+else()
+    message(FATAL_ERROR "libc.a not found via compiler query; setjmp/longjmp will be missing")
+endif()
```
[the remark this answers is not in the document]
mkolosick: applies to different pr

category/execution/CMakeLists.txt
```diff
@@ -323,6 +332,29 @@ target_compile_definitions(
...
+  PUBLIC blst::blst
+  PUBLIC c-kzg-4844
+  PRIVATE PkgConfig::crypto++
+  PUBLIC ethash::keccak
+  PUBLIC evmc
+  PUBLIC immer
+  PUBLIC intx::intx
+  PUBLIC nlohmann_json::nlohmann_json
+  PUBLIC silkpre
+  PUBLIC unordered_dense)
```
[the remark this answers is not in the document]
mkolosick: This is due to the separation between the zkvm and regular client.

category/execution/ethereum/precompiles.cpp
```diff
@@ -188,4 +189,246 @@ check_call_precompile(State &, CallTracerBase &, evmc_message const &msg)
...
+static PrecompileResult
+from_impl_result(PrecompileImplResult result, uint8_t *out)
+{
+    auto const [data, size] = result;
+    if (data == nullptr) {
+        MONAD_DEBUG_ASSERT(size == 0);
+        std::free(out);
+        return PrecompileResult::failure();
+    }
+    return {EVMC_SUCCESS, data, size};
```
goodlyrottenapple, line 205: Potential optimization for functions like ecrecover which always return success, but may return an empty response, in which case it's better to free the `data` buffer immediately.

```suggestion
    if (size == 0) {
        std::free(out);
        return {EVMC_SUCCESS, nullptr, 0};
    }
    return {EVMC_SUCCESS, data, size};
```

goodlyrottenapple (dismissed):

0xSqualo (dismissed):

goodlyrottenapple (dismissed):

.github/workflows/license.sh
```diff
@@ -97,6 +97,10 @@ for file in $(git ls-files -- '*.rs' '*.h' '*.hpp' '*.c' '*.cpp' '*.S'); do
         continue
     fi
 
+    if [ "$file" == "zkvm/core/zkvm_accelerators.h" ]; then
+        continue
+    fi
```
Baltoli: https://github.com/eth-act/zkvm-standards/tree/main#license The standards are Apache / MIT-licensed. I think we should follow the same convention as the rest of the repo and carve this out into a third-party directory with attribution.

zkvm/guest/CMakeLists.txt
```diff
@@ -70,12 +70,35 @@ add_library(evmc::evmc ALIAS evmc)
...
 # ethash — only keccak, no ethash or global_context (needs std::mutex)
 set(ETHASH_BUILD_ETHASH OFF CACHE BOOL "" FORCE)
 set(ETHASH_TESTING OFF CACHE BOOL "" FORCE)
 set(ETHASH_INSTALL_CMAKE_CONFIG OFF CACHE BOOL "" FORCE)
 add_subdirectory("${THIRD_PARTY_DIR}/ethash" ethash)
 
+# blst (BLS12-381)
+set(BLST_SOURCE_DIR "${THIRD_PARTY_DIR}/blst")
+
+enable_language(ASM)
```
Baltoli: Redundant - the `project()` call enables ASM

zkvm/guest/CMakeLists.txt
```diff
@@ -70,12 +70,35 @@ add_library(evmc::evmc ALIAS evmc)
...
+
+add_library(blst STATIC
+    "${BLST_SOURCE_DIR}/src/server.c")
+
+target_include_directories(blst PUBLIC ${BLST_SOURCE_DIR}/bindings)
+
+# The compilation options and definitions match what build.sh would do; if you
+# upgrade libblst, ensure this is still the case
```
Baltoli: It's not clear to me that we're building BLST correctly for a non-x86 architecture: https://github.com/supranational/blst/blob/54e6e55674722fc2797ebb4bbb71b26d881eb4b8/build.sh#L107-L109

Do we need to disable its assembly support to get it to actually work (not just compile) for RISC-V?

cmake/precompile_deps.cmake
```diff
@@ -0,0 +1,65 @@
...
+
+# nlohmann_json
+add_subdirectory("${THIRD_PARTY_DIR}/nlohmann_json" "${CMAKE_CURRENT_BINARY_DIR}/_nlohmann_json" SYSTEM)
+
+# unordered_dense
+add_subdirectory("${THIRD_PARTY_DIR}/unordered_dense" "${CMAKE_CURRENT_BINARY_DIR}/_unordered_dense")
+
+# asmjit
+set(ASMJIT_STATIC ON)
+add_subdirectory("${THIRD_PARTY_DIR}/asmjit" "${CMAKE_CURRENT_BINARY_DIR}/_asmjit")
```
Baltoli: I don't fully understand why these dependencies have been moved out - they're unrelated to precompiles. It might well be the case that they don't work on the RISC-V toolchain, but if that's the case we should clean them up properly rather than lumping them into the precompiles build (presumably as the current only place where we do non-X86 compilation?)

cmake/precompile_deps.cmake
```diff
@@ -0,0 +1,65 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Shared setup for third-party libraries needed by the precompile
+# implementations.  Included by both the root CMakeLists.txt (full build)
+# and category/zkvm/CMakeLists.txt (standalone zkvm build).
```
Baltoli: `category/zkvm` isn't the correct path, right? The directory structure has evolved since this was written

category/execution/CMakeLists.txt
```diff
@@ -312,13 +306,28 @@ add_library(
...
+
 target_include_directories(
   monad_execution_ethereum
   PUBLIC ${CATEGORY_MAIN_DIR}
   PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/test
   PRIVATE ${Boost_INCLUDE_DIRS}
   PUBLIC ${cthash_SOURCE_DIR}/include
-  PUBLIC ${silkpre_SOURCE_DIR}/lib)
+  PUBLIC ${silkpre_SOURCE_DIR}/lib
+  PUBLIC ${monad_crypto_SOURCE_DIR}/src)
```
Baltoli: What defines this variable?

zkvm/category/execution/ethereum/precompiles_impl.hpp
```diff
@@ -0,0 +1,652 @@
...
+        .status_code = EVMC_SUCCESS,
+        .obuf = obuf,
+        .output_size = buf_size,
+    };
+}
+
+bool init_trusted_setup()
+{
+    return true;
+}
```
Baltoli: These are non-inline definitions in a header file - works now but if anything else includes this file it'll fail to link

Baltoli:
As an aside, let's squash this when merging - certainly from looking at the first build system commit, there's a reasonable amount of stuff that gets changed / corrected later, so I don't think it's a sensible by-commit merge.

zkvm/category/execution/ethereum/precompiles_bls12.hpp
```diff
@@ -0,0 +1,78 @@
...
+        std::array<uint8_t, 64> buf{};
+        for (std::size_t i = 0; i < sizeof(BASE_FIELD_MODULUS.bytes); ++i) {
+            buf[16 + i] = BASE_FIELD_MODULUS.bytes[i];
+        }
+        return buf;
+    }();
+    static_assert(BASE_FIELD_MODULUS_BYTES.size() == 64);
+
+    template <typename Group>
+    uint16_t msm_discount(uint64_t);
```
Baltoli, line 60: Are these discount tables linked in properly?

zkvm/category/execution/ethereum/precompiles_impl.hpp
```diff
@@ -0,0 +1,631 @@
...
+        raw_fp_to_evm(zkvm192, evm256);
+        raw_fp_to_evm(zkvm192 + 48, evm256 + 64);
+        raw_fp_to_evm(zkvm192 + 96, evm256 + 128);
+        raw_fp_to_evm(zkvm192 + 144, evm256 + 192);
+    }
+}
+
+MONAD_NAMESPACE_BEGIN
+
+PrecompileResult alloc_success(size_t buf_size)
```
Baltoli: Dead code now?

category/execution/ethereum/precompiles.cpp
```diff
@@ -188,4 +189,249 @@ check_call_precompile(State &, CallTracerBase &, evmc_message const &msg)
...
+}
+
+PrecompileResult ecrecover_execute(byte_string_view const input)
+{
+    static constexpr auto kSecp256k1n =
+        0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141_u256;
+
+    alignas(8) std::basic_string<uint8_t> d(128, '\0');
```
Baltoli: What's the intent here? This will align the stack variable, not the memory pointed to by the string

Baltoli (changes requested):

cmake/precompile_deps.cmake
dhil: I think they have been moved out because they don't (readily) compile on RISC-V. Though, I agree. These should be pulled into a separate cmake file.

Baltoli: That's what I presumed was the case - perhaps they go into a preceding PR to this one if they need to be split out?

dhil: Yes we can do. I am just in the process of rebasing this PR. Afterwards, I will make a single commit hoisting these bits out. We can cherry-pick that subsequently.

dhil: I've looked further into this now. I think these dependencies can be moved back into the root `CMakeLists.txt` file. I think the whole zkvm build architecture can changed since this patch -- if I am correct then the zkvm builds no longer uses the root `CMakeLists.txt`, suggesting that we can do away with the `precompile_deps.cmake` entirely?

dhil: I've removed `precompile_deps.cmake` now. I've reverted to the original `CMakeLists.txt` file with the addition of linking the precompiles library.

cmake/precompile_deps.cmake
dhil: Stale now, as `precompile_deps.cmake` has been removed.

category/execution/CMakeLists.txt
dhil: I believe it is defined by the silkpre cmake project when it is pulled in.

zkvm/category/execution/ethereum/precompiles_impl.hpp
dhil: I think `alloc_success` is dead code. It cannot find any references of it our code bases. You are right `init_trusted_setup` should be inline.

category/execution/ethereum/precompiles.cpp
dhil: I think it is unintended. I'll remove it.

zkvm/category/execution/ethereum/precompiles_bls12.hpp
dhil: I think each templated definition is instantiated at both `G1` and `G2` in the cpp file.

zkvm/guest/CMakeLists.txt
mkolosick: I believe the `target_compile_definitions` fixes this. But also having cherry-picked this onto the zkvm pr it seems that this might not even matter in the end.

zkvm/guest/CMakeLists.txt
```diff
@@ -70,12 +70,34 @@ add_library(evmc::evmc ALIAS evmc)
...
+target_link_libraries(evmc_cpp INTERFACE evmc::evmc)
+
 # ethash — only keccak, no ethash or global_context (needs std::mutex)
 set(ETHASH_BUILD_ETHASH OFF CACHE BOOL "" FORCE)
 set(ETHASH_TESTING OFF CACHE BOOL "" FORCE)
 set(ETHASH_INSTALL_CMAKE_CONFIG OFF CACHE BOOL "" FORCE)
 add_subdirectory("${THIRD_PARTY_DIR}/ethash" ethash)
 
+# blst (BLS12-381)
+set(BLST_SOURCE_DIR "${THIRD_PARTY_DIR}/blst")
```
Baltoli: It's not clear to me that we even need BLST in the ZK guest. I don't see any usage of it in the shadowing guest code - all the functions go through the accelerators API?

zkvm/category/execution/ethereum/precompiles_bls12.hpp
Baltoli:
```cmake
add_library(monad-zkvm-precompiles OBJECT
    "${MONAD_ROOT}/category/execution/ethereum/precompiles.cpp"
    "${MONAD_ROOT}/category/execution/ethereum/precompiles_gas_cost_impl.cpp"
)
```
right, but in `precompiles_bls12.cpp`, which isn't built for the ZK guest

mkolosick: fixed

Baltoli (dismissed):

goodlyrottenapple (dismissed):

zkvm/category/execution/ethereum/precompiles_impl.hpp
```diff
@@ -0,0 +1,629 @@
...
+
+    zkvm_bytes_32 commitment_versioned_hash{};
+    constexpr uint8_t VERSION_HASH_VERSION_KZG = 1;
+    if (zkvm_sha256(
+            commitment_versioned_hash.data,
+            sizeof(zkvm_kzg_commitment),
+            commitment->data) != ZKVM_EOK) {
+        return PrecompileImplResult::failure();
+    }
+    commitment_versioned_hash.data[0] = VERSION_HASH_VERSION_KZG;
```
dhil, line 369: New code since last approval/review. This is a solution to a previously dangling AI comment:
> kzg_to_version_hashed() ignores the return code of zkvm_sha256(). If SHA-256 fails, h will contain uninitialized bytes and the versioned-hash check can behave nondeterministically. Zero-initialize the hash buffer and treat a non-ZKVM_EOK status as a mismatch by returning the zero hash.

I've inlined the function `kzg_to_version_hashed` here to handle the error case correctly.

I believe this is not a problem on the non-zkvm execution client, because the sha256 implementation there can never fail (at least its interface doesn't communicate failure).

CC @Baltoli @goodlyrottenapple

goodlyrottenapple (approved):

Baltoli (approved):

PR #2212, opened 2026-04-16
[dual-timeline 1/N] Add secondary timeline metadata, root offset ring…

maxkozlovsky:
…, and lifecycle management

Introduce the foundational infrastructure for dual-timeline support:

Types (category/mpt/detail/timeline.hpp):
- timeline_id enum (primary/secondary) and NUM_TIMELINES constant
- timeline_compaction_state struct with per-timeline compaction boundary

Metadata (category/mpt/detail/db_metadata.hpp):
- secondary_timeline_header_t carved from future_variables_unused: version_lower_bound_, next_version_, active_ flag
- Total db_metadata size unchanged (528512 bytes) for backward compat

UpdateAux (trie.hpp, trie.cpp, update_aux.cpp):
- Per-timeline compaction state array (timeline_[NUM_TIMELINES]) with tl(timeline_id) accessor, replacing bare member fields
- Secondary root offset ring mapped from cnv chunk 0's unused space (65536 entries, 512KB per copy)
- root_offsets_delegator parameterized on timeline_id
- timeline_active(), get_root_offset_at_version(v, tid), db_history_{min,max}_version(tid), version_is_valid_ondisk(v, tid)
- Lifecycle: activate_secondary_timeline (initializes compaction from primary), deactivate_secondary_timeline, promote_secondary_to_primary

Tests:
- db_metadata_test: layout, field offsets, zero-init semantics, read/write round-trip, memcpy survival, no overlap with consensus
- timeline_test: enum values, default construction, trivial copyability
- update_aux_test: secondary ring mapping, activate/deactivate lifecycle, ring push/read, promote, per-timeline version queries, version_is_valid_ondisk per-timeline

### DIFF
--- a/category/mpt/db_metadata_context.cpp
+++ b/category/mpt/db_metadata_context.cpp
@@ -23,6 +23,7 @@
 #include <category/mpt/config.hpp>
 #include <category/mpt/db_metadata_context.hpp>
 #include <category/mpt/detail/db_metadata.hpp>
+#include <category/mpt/detail/timeline.hpp>
 #include <category/mpt/trie.hpp>
 #include <category/mpt/util.hpp>
 
@@ -30,6 +31,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cerrno>
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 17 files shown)


maxkozlovsky:
Thanks for the reviews — addressed in push 95309dbf:

- **Clang-tidy errors**: added missing `#include <cstddef>`, `<cstring>`, `<utility>`, and `<category/mpt/detail/timeline.hpp>`; made `cleanup` scope_exits and read-only delegator locals `const`; used `auto const &` in the migration / ring-init range loops.
- **`reinterpret_cast<std::atomic_*>` writes → `start_lifetime_as<std::atomic_*>`** in `activate_secondary_timeline` and `deactivate_secondary_timeline`, matching the reader pattern and avoiding the object-lifetime UB.
- **`promote_secondary_to_primary` std::swap of atomic fields**: replaced with an inline `atomic_swap` lambda that does acquire-load + release-store through `start_lifetime_as<std::atomic_uint64_t>`. Kept the existing TODOs; added a clarifying comment on the `active_` asymmetry (primary has no separate `active_` byte, so post-swap the new-secondary's `active_=1` intentionally leaves the old-primary accessible as the new secondary — that's the promote-as-fork semantics).
- **`db_history_min_valid_version(secondary)`**: now short-circuits on `!timeline_active(tid)` before reading `version_lower_bound_`, closing the activate/deactivate race window.
- **Empty-span OOB in delegator**: added `MONAD_ASSERT(!root_offsets_chunks_.empty())` in `push`, `assign`, and `operator[]`.
- **`%lu` on `uint64_t`**: switched to `LOG_INFO` + `{}` style.

category/mpt/db_metadata_context.cpp
```diff
@@ -542,51 +694,826 @@ uint64_t DbMetadataContext::db_history_max_version() const noexcept
...
+                            : &m->secondary_timeline.version_lower_bound_;
+        auto *pver_nv = (primary_idx == 0)
+                            ? &m->root_offsets.next_version_
+                            : &m->secondary_timeline.next_version_;
+        auto *sver_lb = (primary_idx == 0)
+                            ? &m->secondary_timeline.version_lower_bound_
+                            : &m->root_offsets.version_lower_bound_;
+        auto *sver_nv = (primary_idx == 0)
+                            ? &m->secondary_timeline.next_version_
+                            : &m->root_offsets.next_version_;
```
Chen-Yifan, line 1045: `fork_version` input here and the outer `activate_secondary_timeline()` seems unnecessary, we do set the correct value for `next_version_` and `version_lower_bound_` when first write a version to secondary db, see `fast_forward_next_version()` in `write_new_root_node()`.

maxkozlovsky: removed version from activate

category/mpt/db_metadata_context.cpp
```diff
@@ -104,7 +134,153 @@ DbMetadataContext::DbMetadataContext(AsyncIO &io)
...
+                "writable mapping to migrate to %s. Open with write access.",
+                detail::db_metadata::PREVIOUS_MAGIC,
+                detail::db_metadata::MAGIC);
+        }
+        for (auto const &copy : copies_) {
+            auto *const m = copy.main;
+            if (!is_previous_magic(m)) {
+                continue;
+            }
+            auto const g = m->hold_dirty();
```
0xSqualo, line 196: @maxkozlovsky @Chen-Yifan Should `MONAD007` dirty-copy recovery run before the migration loop, since `hold_dirty()` here can clear a pre-existing dirty bit during `--upgrade` and cause a partially written metadata copy to be migrated instead of healed from the clean backup?

Chen-Yifan: I think you're right. The dirty recovery should happen before the migration for upgrade, and it should copy the corresponding layout based on the whether the current magic string is 007 or 008.

maxkozlovsky: Added copy restore before upgrade if one of them is dirty. Added a test to verify, which fails without the change.

Chen-Yifan (approved):

Chen-Yifan (approved):
Note that to upgrade to the release that first includes this commit, it requires node operators to run `monad-mpt --upgrade` before restart the clients.

kkuehlz (approved):

marcograss (commented):
approved with the follow up #2302 and #2387

PR #2213, opened 2026-04-17
Evm-as: fix PushLabel encoding at offset zero for Pre-Shanghai

anne-laure-s:
Closes https://github.com/category-labs/monad-private/issues/37

Fixes pre-Shanghai `PUSH0` calls in `evm-as`.

### DIFF
--- a/category/vm/utils/evm-as/compiler.hpp
+++ b/category/vm/utils/evm-as/compiler.hpp
@@ -209,6 +209,15 @@ namespace monad::vm::utils::evm_as
                         size_t const offset = it->second;
                         size_t const n =
                             offset == 0 ? offset : byte_width(offset);
+                        if constexpr (traits::evm_rev() < EVMC_SHANGHAI) {
+                            if (n == 0) {
+                                // Special case for zero offset before Shanghai,
+                                // as PUSH0 is not available.
+                                emit_byte(mc::EvmOpCode::PUSH1);
+                                emit_byte(0x00);
+                                return;
+                            }
+                        }
                         emit_byte(
                             mc::EvmOpCode::PUSH0 + static_cast<uint8_t>(n));
                         // Note: assumes we are executing on a
@@ -350,7 +359,10 @@ namespace monad::vm::utils::evm_as
                             }
                             size_t offset = it->second;
                             if (offset == 0) {
-                                std::string const str = std::format("PUSH0");
+                                static constexpr std::string_view str =
+                                    traits::evm_rev() < EVMC_SHANGHAI
+                                        ? "PUSH1 0x00"
+                                        : "PUSH0";
                                 os << str;
                                 return str.size();
                             }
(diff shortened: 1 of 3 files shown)


andreaslyn (dismissed):

category/vm/utils/evm-as/compiler.hpp
```diff
@@ -208,6 +208,13 @@ namespace monad::vm::utils::evm_as
                         size_t const offset = it->second;
                         size_t const n =
                             offset == 0 ? offset : byte_width(offset);
+                        if (offset == 0 && traits::evm_rev() < EVMC_SHANGHAI) {
```
Baltoli, line 211: We'd usually hoist the check on the revision to compile time using `if constexpr`. Something like:
```cpp
if constexpr (traits::evm_rev() < EVMC_SHANGHAI) {
  ...
} else {
  ...
}
```
which also avoids the early return

category/vm/utils/evm-as/compiler.hpp
```diff
@@ -350,7 +357,10 @@ namespace monad::vm::utils::evm_as
                             }
                             size_t offset = it->second;
                             if (offset == 0) {
-                                std::string const str = std::format("PUSH0");
+                                std::string const str =
+                                    traits::evm_rev() < EVMC_SHANGHAI
+                                        ? std::format("PUSH1 0x00")
+                                        : std::format("PUSH0");
```
Baltoli, line 363: You can do this at compile time as well with an IIFE:
```cpp
std::string const str = [] {
  if constexpr (...) {
    return ...;
  } else {
    return ...;
  }
}();
```

I'd also drop the useless `std::format` wrappers

category/vm/utils/evm-as/resolver.hpp
```diff
@@ -97,6 +97,13 @@ namespace monad::vm::utils::evm_as
                                 size_t const n =
                                     offset == 0 ? 0 : byte_width(offset);
 
+                                if (n == 0 &&
```
Baltoli, line 100: as above

Baltoli (changes requested):

category/vm/utils/evm-as/compiler.hpp
Baltoli: Even better:
```cpp
static constexpr std::string_view str = traits::evm_rev() < EVMC_SHANGHAI ? ... : ...;
```

category/vm/utils/evm-as/compiler.hpp
anne-laure-s: Changed all occurrences in the new code ; https://github.com/category-labs/monad/pull/2213/changes/dc3f46f657fd884885d06cb5bfb8d6abf12311df changes one occurrence in the pre-existing code

category/vm/utils/evm-as/compiler.hpp
anne-laure-s: Applied

category/vm/utils/evm-as/resolver.hpp
anne-laure-s: Changed

Baltoli (approved):

goodlyrottenapple (approved):

Baltoli:
Merging as the change is approved; the failing CI is Claude maxing out its usage limit.

PR #2216, opened 2026-04-20
Apply east-const to function parameter definitions

Baltoli:
## Summary

Mechanical pass that adds `const` (east-const) to function parameter **definitions** across the codebase, covering both value parameters (`int x` → `int const x`) and the pointer itself for pointer parameters (`T *p` → `T *const p`, leaving any pointee qualifier alone).

A small number of **declarations** are also touched (~19, all in headers) — in each case a leftover top-level `const` on a by-value parameter is removed so the declaration matches the codebase's existing no-top-level-qualifier convention for declarations. The language ignores top-level qualifiers on parameters in declarations, so there's no behavioral or ABI effect; the normalization came out of the AST tooling re-emitting these headers.

References (`T&`, `T&&`), typedefs that hide pointers (e.g. `va_list`), parameters that are mutated in the body or member-initializer list, and move-only parameters consumed via `std::move(name)` are deliberately left non-const.

### Scope

- `category/`, `cmd/`, `test/`
- 198 files, +1,231 / -1,084

### Tooling

A libclang-based Python tool walked every TU in `compile_commands.json`, collected the candidate edits from the AST, deduplicated across TUs, and emitted a single JSON patch set that an applier script then rewrote on disk. Compile errors from each subsystem surfaced the cases where a parameter was actually mutated (read-only assignment, discarded qualifier, reference-to-non-const binding, std::move silently falling back to copy); those were reverted by hand so the build stays green.

### Notable exceptions

- `category/async/sender_errc.hpp` is left unchanged: adding `const` to `make_status_code(sender_errc)` causes boost::outcome's `quick_status_code_from_enum` machinery to recurse into itself as a constexpr expansion.
- A handful of params that are legitimately mutated (e.g. `event_ring_init_file`'s `error_name`, `stack_top` in `call_runtime.hpp`, several `bytes`/`prefix_index` counters in the MPT code, `int argc` in test `main`s) stay non-const.

## Test plan

- [x] `cmake --build build --parallel` — clean
- [x] `ctest --test-dir build --output-on-failure --timeout 500` — 3520 / 3520 pass
- [x] `scripts/apply-clang-format.sh` applied

### DIFF
--- a/category/async/config.hpp
+++ b/category/async/config.hpp
@@ -98,7 +98,8 @@ struct chunk_offset_t
     }
 
     constexpr chunk_offset_t(
-        uint32_t id_, file_offset_t offset_, file_offset_t spare_ = max_spare)
+        uint32_t const id_, file_offset_t const offset_,
+        file_offset_t const spare_ = max_spare)
         : offset(offset_ & max_offset)
         , id(id_ & max_id)
         , spare{spare_ & max_spare}
@@ -140,7 +141,7 @@ struct chunk_offset_t
                static_cast<file_offset_t>(offset);
     }
 
-    void set_spare(uint16_t value) noexcept
+    void set_spare(uint16_t const value) noexcept
     {
         MONAD_ASSERT(value < max_spare);
         spare = value & max_spare;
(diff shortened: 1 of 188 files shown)


Baltoli:
@claude review

Baltoli:
@claude re-review

jhunsaker (dismissed):

jhunsaker (approved):

PR #2219, opened 2026-04-21
Disas tests: re-enable in CI and refresh stale goldens

Baltoli:
## Summary

Three commits — two preparatory cleanups followed by the actual fix.

### Commit 1 — \`73d5329a0\` Remove unused callgrind instruction-count regression harness

Deletes ~247 lines of Python + six JSON fixtures (\`test_callgrind.py\`, \`callgrind_parser.py\`, \`callgrind/keccak_*.json\`).

- Introduced July 2023 (\`6540a5d79\`, \`afbfe1380\`, \`b65516bf5\`) as an experimental perf-regression tool; seminal commit labelled the three keccak specs as "Example callgrind tests".
- Only ever had those three specs; all pointed at \`ethash-test\`.
- \`ethash-test\` is not built here — \`CMakeLists.txt:137\` hardcodes \`ETHASH_TESTING=NO\`.
- No new specs added in 30 months. No other pytest module imports \`callgrind_parser\`.
- The harness has never actually run in CI (see Commit 2's bug #1).
- Substantive touches since Oct 2023: none (only bulk relocations and formatting).

If instruction-count regression testing becomes wanted again, the 2023 implementation is recoverable from git.

### Commit 2 — \`26c943047\` Disas tests: re-enable in CI and refresh stale goldens

The \`test_disas\` pytest suite has been silently skipped in CI since its introduction. Two bugs conspired:

1. **Wrong pytest invocation.** \`scripts/test.sh\` ran \`pytest-3 --pyargs monad\`, which from the repo root (the Dockerfile's cwd, \`docker/Dockerfile:45\`) resolves the \`monad\` package to the top-level \`monad/__init__.py\` — which has no tests. Collection returned 0 items; \`|| [ $? -eq 5 ]\` tolerated the empty result. Switched to a path-based invocation: \`pytest-3 category/core/monad/tests/\`.
2. **Wrong CC match.** The codegen-sensitive gate checked \`CC == "clang"\`, but the docker matrix sets \`CC=clang-19\` (\`docker.yaml:28\`), so the check always failed. Now matches \`clang*\`.

**Broken specs.** Two specs referenced \`libmonad_core.a\`, which does not exist — \`monad_core\` is an OBJECT library. Point them at \`libmonad_core_disas.a\`, where the symbols are archived via the existing \`PUBLIC monad_core\` link dep (\`category/core/CMakeLists.txt:233\`). Added a one-line comment next to each spec explaining the indirection.

**Stale goldens.** Regenerated the three pre-existing goldens under Clang 19 + RelWithDebInfo:
- \`core-tl_tid.dis\`: trailing whitespace only (objdump version drift).
- \`core-spinlock.dis\`: objdump whitespace plus an LLVM TLS-access improvement (\`add %fs:0x0\` + \`mov\` → \`mov %fs:(...)\`); spinlock logic identical.
- \`fiber/priority_queue.dis\`: was captured against a \`std::priority_queue\` + \`pthread_mutex\` implementation that has since been replaced by \`oneapi::tbb::concurrent_priority_queue\`. New golden matches current source — but the fixture is deleted entirely in commit 3; see below.

### Commit 3 — \`4999b94cd\` Delete fiber/priority_queue disas fixture

\`PriorityQueue::pop()\` and \`push()\` are one-line wrappers over \`oneapi::tbb::concurrent_priority_queue::try_pop\` / \`push\` (see \`category/core/fiber/priority_queue.cpp\`). The disas golden therefore pins:
- ~10 instructions of prologue,
- one \`call\` whose relocation target is a ~900-character demangled TBB-internal template instantiation,
- ~5 instructions of epilogue.

Signal: low (Monad-authored code footprint ≈ 5 lines per method). Noise: high (any TBB version bump changes the relocation target and trips the test with a giant diff that's always a regenerate, never a real regression). Track record bears this out — the previous std::priority_queue-era golden survived the swap to TBB for ~18 months without anyone noticing, so the test was clearly not being read when it drifted. Correctness is covered by unit tests; this fixture was catching nothing useful.

After this commit, \`disas/\` contains only fixtures pinning Monad-authored code where codegen matters (spinlock primitives, TLS init). A follow-up PR will add a uint256 fixture in the same spirit.

## Test plan
- [x] \`pytest-3 category/core/monad/tests/ -v\` from repo root with \`CC=clang-19\` + \`CMAKE_BUILD_TYPE=RelWithDebInfo\` → 2/2 passed (core-spinlock, core-tl_tid).
- [x] Same invocation with \`CC=gcc-15\` → both deselected by \`-k not test_disas\`.
- [x] Shell pattern sanity-checked: \`clang-19\` and \`clang\` run the suite; \`gcc-15\` skips.
- [ ] CI green on the Clang 19 + RelWithDebInfo matrix entry.


Baltoli:
@claude review

jhunsaker (approved):

PR #2220, opened 2026-04-21
Added compiler only VM mode

andreaslyn:


category/vm/vm.cpp
```diff
@@ -151,6 +156,29 @@ namespace monad::vm
...
+    {
+        if (MONAD_UNLIKELY(mode_ == CompilerOnly)) {
+            auto const ncode = compiler_.cached_compile<traits>(
+                code_hash, icode, compiler_config_);
+            auto const entry = ncode->entrypoint();
+            if (MONAD_LIKELY(entry != nullptr)) {
+                return execute_native_entrypoint_raw<traits>(rt_ctx, entry);
+            }
+            LOG_WARNING("WARNING: VM: fallback to interpreter: "
+                        "compilation failed in CompilerOnly mode.");
```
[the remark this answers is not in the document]
andreaslyn: I believe that the `cached_compile` function does not attempt to compile again after a failed compilation, because the `ncode` is always cached regardless of compilation failure.

category/vm/vm.hpp
```diff
@@ -122,13 +122,23 @@ namespace monad::vm
 
     class VM
     {
+    public:
+        enum Mode
+        {
+            Dual,
+            CompilerOnly,
+            InterpreterOnly
+        };
```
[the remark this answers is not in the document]
andreaslyn: There is not such implicit cast in c++

test/vm/unit/monad_vm_interface_tests.cpp
```diff
@@ -584,53 +605,93 @@ TEST(MonadVmInterface, execute_raw)
...
 
     auto max_code_size_offset = vm.compiler_config().max_code_size_offset;
     auto const compile_threshold =
         native::max_code_size(max_code_size_offset, warm_icode->code_size());
 
-    // Execute with interpreter on warm cache until compilation is started.
-    do {
-        execute_raw(EvmTraits<EVMC_SHANGHAI>{}, warm_hash, warm_vcode);
-        vm.compiler().debug_wait_for_empty_queue();
+    auto pre_pre_warm_vcode = vm.find_varcode(warm_hash);
```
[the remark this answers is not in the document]
andreaslyn: Removed it

anne-laure-s (approved):
The logic of the code seems to preserve the previous logic.

This was in the previous code, but might be useful to mention that `async_compile` is called even in InterpreterOnly mode, and its behavior is then decided silently by other variables (ie `mode == Dual`).

I think this is error prone and confusing, we shouldn’t call `async_compile` if we don’t compile. If there is some caching logic involved behind the usage of `async_complie`, it should be made explicit.

Baltoli:
@claude review

category/vm/vm.cpp
Baltoli, line 172: Is a logged warning the best outcome here? My intuition would be that we'd want a harder failure if compilation fails, but open to being convinced otherwise

Baltoli:
Looks reasonable to me other than my minor comment around logging + Claude's comments. No major concerns but will wait for Asymmetric to review

category/vm/vm.cpp
andreaslyn: Compilation failure is expected to happen on certain pathological contracts. If CompilerOnly mode is enabled for some reason, then an attacker will be able to halt the chain if this is changed to a (release) assertion, for example.

andreaslyn: The intention is that native code is inserted into the code cache by `cached_compile` even when compilation fails, so this should not spam the warning on every call to the same contract. It should instead call `execute_intercode_raw` on line 126 when compilation has failed.

category/vm/compiler.hpp
```diff
@@ -53,12 +55,12 @@ namespace monad::vm
         std::atomic<uint64_t> num_unexpected_compilation_errors_{0};
         std::atomic<uint64_t> num_size_out_of_bound_compilation_errors_{0};
 
-        // must be called non-concurrently
         void event_new_compiled_code_cached(
             SharedIntercode const &icode, SharedNativecode const &ncode,
             auto compile_start, auto compile_end) noexcept
         {
             if constexpr (utils::collect_monad_compiler_stats) {
+                std::lock_guard<std::mutex> lock_{mutex_};
```
andreaslyn: I have made this function concurrent after Asymmetric made me aware that this function was not concurrent and called via `cached_compile`.

marcograss (approved):

dhil (approved):

PR #2222, opened 2026-04-21
Reorg rlp `parse*_metadata` functions

goodlyrottenapple:
add unified RLP parse metadata function which returns decoded view and its type

### DIFF
--- a/category/execution/ethereum/rlp/decode.hpp
+++ b/category/execution/ethereum/rlp/decode.hpp
@@ -25,10 +25,14 @@
 
 #include <boost/outcome/try.hpp>
 
+#include <concepts>
+#include <type_traits>
+#include <utility>
+
 MONAD_RLP_NAMESPACE_BEGIN
 
 template <unsigned_integral T>
-constexpr Result<T> decode_raw_num(byte_string_view const enc)
+inline Result<T> decode_raw_num(byte_string_view const enc)
 {
     if (MONAD_UNLIKELY(enc.size() > sizeof(T))) {
         return DecodeError::Overflow;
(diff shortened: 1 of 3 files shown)


category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +58,114 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
-        end = i + length;
+        auto const payload = enc.substr(i, length);
+        enc = enc.substr(end);
+        return std::pair{type, payload};
     }
+}
 
-    if (MONAD_UNLIKELY(end > enc.size() || end < i)) {
+constexpr Result<std::pair<RlpType, byte_string_view>>
+parse_metadata(byte_string_view &enc)
```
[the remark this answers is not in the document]
goodlyrottenapple: swapping fields would lead to a compiler error.

goodlyrottenapple:
@claude review

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +60,136 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
+    };
+
+    template <ParseMode Mode>
+    using parse_metadata_result_t = std::conditional_t<
+        Mode == ParseMode::Any, Result<std::pair<RlpType, byte_string_view>>,
+        Result<byte_string_view>>;
+
+    template <ParseMode Mode>
+    [[gnu::always_inline]] inline constexpr parse_metadata_result_t<Mode>
+    parse_metadata(byte_string_view &enc)
```
dhil: I think this function is morally four functions: `parse_metadata`, `parse_string`, `parse_list`, `parse_payload`, where the first one validates the input and delegates to either of the other three; `parse_string` and `parse_list` both tail call `parse_payload`.

dhil (dismissed):

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +60,145 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
+namespace detail
+{
 
-    if (MONAD_UNLIKELY(enc[0] >= 0xc0)) {
-        return DecodeError::TypeUnexpected;
-    }
+    enum class ParseMetadataOptions
+    {
+        ReturnRlpType,
+    };
```
Baltoli, line 75: Why do we need the single-variant enum here? Will there be more options in the future?

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +60,145 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
+
+    template <ParseMetadataOptions... Options>
+    using parse_metadata_result_t = std::conditional_t<
+        ((Options == ParseMetadataOptions::ReturnRlpType) || ...),
+        Result<std::pair<RlpType, byte_string_view>>, Result<byte_string_view>>;
```
Baltoli, line 92: i.e. I don't see how this type will ever be anything other than `Result<std::pair<RlpType, byte_string_view>>`. But if it's for future extension then that seems fine

category/execution/ethereum/rlp/decode.hpp
goodlyrottenapple: No, but wasn't sure what the idiomatic thing to do is. i don't like using bools, as `foo<true>` doesn't really tell you much. Do you have a better way to structure the code here?

category/execution/ethereum/rlp/decode.hpp
goodlyrottenapple: if you don't provide an option, it will revert to `Result<byte_string_view>`

category/execution/ethereum/rlp/decode.hpp
```diff
@@ -56,90 +60,145 @@ constexpr Result<size_t> decode_length(byte_string_view const enc)
...
 
-    if (enc[0] < 0x80) // [0x00, 0x7f]
+    template <RlpType Ty, ParseMetadataOptions... Options>
+    [[gnu::always_inline]] inline parse_metadata_result_t<Options...>
+    extract_payload(byte_string_view &enc, size_t i, size_t length)
```
Baltoli:
```suggestion
    extract_payload(byte_string_view &enc, size_t const i, size_t const length)
```

category/execution/ethereum/rlp/test/test_decode.cpp
```diff
@@ -60,3 +61,96 @@ TEST(Rlp, DecodeAfterEncodeString)
...
+        ASSERT_FALSE(result.has_error());
+        EXPECT_EQ(result.value().first, RlpType::String);
+        EXPECT_EQ(result.value().second, to_byte_string_view("dog"));
+        EXPECT_EQ(enc.size(), 0);
+    }
+
+    // List payload
+    {
+        auto const inner =
+            monad::byte_string({0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g'});
```
Baltoli: Is `monad::` not already in scope?

category/execution/ethereum/rlp/decode.hpp
Baltoli, line 92: If it is the case that we need / use both branches, maybe use a named structure rather than `std::pair`

category/execution/ethereum/rlp/decode.hpp
Baltoli: Your comment below addressed my confusion - I realised that what I'd missed was the zero-arity case with no options. This seems fine given that

category/execution/ethereum/rlp/decode.hpp
goodlyrottenapple: Explained the purpose of this type and `ParseMetadataOptions...` in the comment above

dhil (dismissed):

goodlyrottenapple:
@claude re-review

Baltoli (approved):

dhil (approved):

es3n1n (commented):
👍

PR #2227, opened 2026-04-23
Port SHA256, RIPEMD160 precompiles away from silkpre

Baltoli:
Add a `category/core/crypto` module with an `init_crypto()` entry point that loads the KZG trusted setup, registers the OpenSSL default and legacy providers, and probes every digest consumed by consensus-critical code (SHA256, RIPEMD160). Idempotency is provided by the thread-safe initialization of a function-local static, so FFI consumers and the CLI can invoke it freely without racing on provider load or leaking handles on repeat attempts. A misconfigured OpenSSL fails at startup.

Migrate the SHA256 and RIPEMD160 precompiles, and the EIP-7685 requests-hash in `process_requests`, from silkpre to OpenSSL's EVP digest API. The fetched `EVP_MD *` is cached per-algorithm at initialization time. Null-input handling (previously a silkpre-specific workaround in `sha256_execute`) is centralized in the shared `digest()` helper.

Introduce `category/core/crypto/digest.hpp`, a small templated helper that owns one `thread_local EVP_MD_CTX` per digest type and drives it with `EVP_DigestInit_ex` / `EVP_DigestUpdate` / `EVP_DigestFinal_ex`. Each instantiation is shared across translation units, so both the precompile path and `compute_requests_hash` reuse the same warm context on the same thread, and switching algorithms doesn't tear down OpenSSL's per-algorithm internal state. The header documents the fiber-atomic invariant the helper relies on (no fiber yield between Init and Final).

Update `cmd/monad/main.cpp`, `category/rpc/monad_executor.cpp`, the blockchain / precompile tests, and the VM fuzzer / execution benchmark binaries to call `init_crypto()` at startup. Wire `libcrypto>=3.0` into the root, core, and execution CMake targets, and add `libssl-dev` to the Ubuntu deps script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,7 @@ find_package(Boost REQUIRED COMPONENTS context filesystem json CONFIG)
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(brotli REQUIRED IMPORTED_TARGET libbrotlienc libbrotlidec)
 pkg_check_modules(crypto++ REQUIRED IMPORTED_TARGET libcrypto++)
+pkg_check_modules(libcrypto REQUIRED IMPORTED_TARGET libcrypto>=3.0)
 pkg_check_modules(zstd REQUIRED IMPORTED_TARGET libzstd)
 
 # ankerl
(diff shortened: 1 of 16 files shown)


Baltoli:
@claude re-review

Baltoli:
@claude re-review; noting previous comments

Baltoli:
Closing as we've decided not to use OpenSSL; the library structure and performance don't properly suit our use case.

PR #2228, opened 2026-04-23
[dual-timeline 2/N] Implement dual-timeline support in MPT database

maxkozlovsky:
Wire the secondary-timeline metadata (from 1/N) into the write/read paths and expose dual-timeline as a Db-level API. A single on-disk DB hosts two timelines (primary, secondary), each with its own root offset ring, per-timeline compaction state, and StateMachine; storage rings and free-list are shared. Use case is zero-downtime state-root-hash-algorithm migration.

API: a Db is bound to one timeline. Existing constructors produce a primary Db. For secondary access:

- activate_secondary_timeline(machine, fork_version) -> Db: activates the on-disk secondary and returns a Db bound to it.
- open_secondary_timeline(machine) -> optional<Db>: attaches to an already-active secondary without modifying metadata; use after reopen/restart.

Primary-only lifecycle: deactivate_secondary_timeline() and promote_secondary_to_primary(). Both assert impl_.use_count()==1 -- the secondary Db must be destroyed first. promote also clears the primary Db's StateMachine binding so a missed close+reopen traps on the next upsert instead of silently writing with the old machine.

Internals: timeline_id threaded through Db::upsert -> do_update ->
write_new_root_node and find -> find_blocking -> read_node_blocking. Per-timeline compact_offsets / curr_upsert_auto_expire_version on UpdateAux and unflushed_version on RWOnDisk. Combined GC boundary: release_unreferenced_chunks takes the component-wise min of compact_offset_pair across all active timelines.

15 tests in dual_timeline_test.cpp cover activate/deactivate, concurrent operation, promote + reopen, reactivation, open-without-activate, GC protection during primary history expiration, and load_root boundary conditions.


category/mpt/update_aux.cpp
```diff
@@ -223,6 +235,10 @@ void UpdateAux::rewind_to_match_offsets()
 void UpdateAux::clear_ondisk_db()
 {
     MONAD_ASSERT(is_on_disk());
+    // No finalized version means no anchor for either timeline; wipe both.
```
Chen-Yifan: suggest removing this, and makes deactivation of secondary timeline explicit. 

I found if I create a db with secondary timeline, close them, reopen with `rewind_to_latest_finalized = true` will silently remove secondary timeline.

category/mpt/update_aux.cpp
```diff
@@ -244,19 +260,64 @@ void UpdateAux::clear_ondisk_db()
...
         auto const g = metadata_ctx_->main_mutable(which)->hold_dirty();
-        metadata_ctx_->root_offsets(which).rewind_to_version(version);
+        if (primary_needs_rewind) {
+            metadata_ctx_->root_offsets(timeline_id::primary, which)
+                .rewind_to_version(version);
+        }
+        if (secondary_needs_rewind) {
+            metadata_ctx_->root_offsets(timeline_id::secondary, which)
+                .rewind_to_version(version);
+        }
```
0xSqualo, line 317: @Chen-Yifan @maxkozlovsky Is it intentional for `rewind_to_latest_finalized` to rewind an active secondary timeline to the primary finalized version even when that version may not exist in the secondary root ring, rather than leaving the secondary unchanged or rewinding to its latest valid version at or below the target?

Chen-Yifan: I believe this is intentional, latest_finalized version applies to both primary and secondary timeline

Chen-Yifan (approved):

category/mpt/db.cpp
```diff
@@ -796,14 +869,20 @@ class Db::RWOnDisk final : public Impl
     virtual Node::SharedPtr
     load_root_for_version(uint64_t const version) override
     {
-        if (!aux().metadata_ctx().version_is_valid_ondisk(version)) {
+        if (!aux().metadata_ctx().timeline_active(tid_)) {
+            return nullptr;
+        }
+        if (tid_ == timeline_id::primary &&
+            !aux().metadata_ctx().version_is_valid_ondisk(version)) {
```
Chen-Yifan:
```suggestion
       // version_is_valid_ondisk(version, tid_) already folds in the
        // timeline_active(tid_) check and bounds, so an inactive secondary
        // or an out-of-range version returns null without a worker round-trip.
        if (!aux().metadata_ctx().version_is_valid_ondisk(version, tid_)) {
```

category/mpt/traverse.hpp
```diff
@@ -103,8 +103,8 @@ namespace detail
                     continue;
                 }
                 MONAD_ASSERT(aux.is_on_disk());
-                auto const next_node_ondisk =
-                    read_node_blocking(aux, node.fnext(idx), version);
+                auto const next_node_ondisk = read_node_blocking(
+                    aux, node.fnext(idx), version, timeline_id::primary);
```
Chen-Yifan, line 107: does traverse only support primary?

category/mpt/db.cpp
```diff
@@ -1214,6 +1308,68 @@ UpdateAux &Db::aux()
...
 }
 
+Db Db::activate_secondary_timeline(
+    std::unique_ptr<StateMachine> secondary_machine)
+{
+    MONAD_ASSERT(impl_);
+    MONAD_ASSERT(secondary_machine);
+    MONAD_ASSERT(is_on_disk() && !is_read_only());
+    auto *const rw = static_cast<RWOnDisk *>(impl_.get());
+    MONAD_ASSERT(rw->worker_thread_use_count() == 1);
```
Chen-Yifan, line 1319: We should assert that current `rw->tid_ == primary` for this and the following 3 functions
activate_secondary_timeline,
open_secondary_timeline,
promote_secondary_to_primary,
deactivate_secondary_timeline.

Only asserting `worker_thread_use_count() == 1` is not sufficient, user could have been created secondary and deactivated primary db.

Chen-Yifan (changes requested):
leaving a few more comments that I overlooked before, please dismiss my approval earlier

category/mpt/traverse.hpp
maxkozlovsky: yes there were no use cases yet for traverse on secondary

kkuehlz (dismissed):

Chen-Yifan (approved):

kkuehlz (approved):

ogechno:
Approved

PR #2229, opened 2026-04-24
docs: document east-const parameter convention in CLAUDE.md

Baltoli:
## Summary
- Record the east-const rule established by the recently-merged sweep (5d78ce23a) under **Code Conventions** in CLAUDE.md.
- Call out the carve-outs already used in that PR — references, unnamed params, mutated params, `std::move`-consumed move-only params, lambdas, and declarations in headers — so future code (and Claude sessions) can apply the rule without re-deriving it from a 189-file diff.

## Test plan
- [ ] Rendering spot-check on the PR web view — just a new subsection, no structural changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -187,6 +187,19 @@ Test code uses `MONAD_TEST_NAMESPACE_BEGIN/END` (from `monad/test/config.hpp`), 
 - Include paths use the full `category/` prefix: `#include <category/core/hex.hpp>`
 - Naming: `snake_case` for functions, `PascalCase` for types (some older code uses `snake_case` types), trailing `_` for private members
 
+### East-const on parameter definitions
+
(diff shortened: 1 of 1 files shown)


Baltoli:
Addressed both carve-out comments in `ea3f21577`:

- **Header-resident definitions (Copilot):** Rewrote the rule around the declaration-vs-definition distinction (has-body) rather than file location, and now explicitly call out that in-header inline, `constexpr`, template, and member function definitions all get east-const. The original sweep did touch 87 header files, so scoping the rule to `.cpp` as Copilot suggested would understate it.
- **`va_list` carve-out (github-actions):** Added typedefs-that-hide-pointers as a bullet with `va_list`/`format_err.c:33` as the concrete example and the reason (array-typedef implementations break `va_arg`/`va_copy` under const).

Also restructured the carve-outs into a bullet list so additions remain scannable.

jhunsaker (approved):

PR #2231, opened 2026-04-24
[dual-timeline 3/N] Add monad-mpt --upgrade and gate MONAD007 migration

maxkozlovsky:
## Summary

- Adds `monad-mpt --upgrade`: the operator-facing step that brings an on-disk MPT pool up to the current metadata version (MONAD008) and flushes it durably to disk before monad services start.
- Closes the migration race that #2228 leaves open: removes automatic ctor-level migration from the normal open path and gates it behind a new `allow_migration` flag on `storage_pool::creation_flags`. Services now abort on `PREVIOUS_MAGIC` with a clear "run `monad-mpt --upgrade`" message; only `monad-mpt --upgrade` sets the flag.
- Operator runbook at `docs/monad_mpt_upgrade.md`.

Stacked on #2228 (`max/dual-timeline`).

## Shape

- `storage_pool::creation_flags::allow_migration` + `is_migration_allowed()` accessor (default false).
- `DbMetadataContext` ctor consults `is_migration_allowed()` before entering the MONAD007→MONAD008 block. Without the opt-in: `MONAD_ABORT_PRINTF` with the runbook message. The preexisting `can_write_to_map_` gate is preserved.
- Public `DbMetadataContext::sync_metadata_to_disk()` forwarder. The tool calls it after the ctor returns — idempotent on already-current pools, also covers the case where a prior run msync'd but the box was power-cycled before the kernel flushed.
- `--upgrade` joins the existing `cli_ops_group` in `cli_tool_impl.cpp` (mutually exclusive with `--create`, `--truncate`, `--create-empty`, `--rewind-to`, `--reset-history-length` via the existing `require_option(0, 1)`).

## Tests

New:
- `cli_tool.upgrade_idempotent_on_current_pool` — `--create` then `--upgrade` → exits 0, stdout contains `"already on version MONAD008"`.
- `cli_tool.upgrade_rejects_combined_mutation` — `--upgrade --create` → CLI11 parse error.
- `cli_tool.upgrade_migrates_monad007_pool` — end-to-end: fabricates a MONAD007 layout over a `--create`'d file via the pool's own chunk API (correct offsets, real chunk count), runs `--upgrade`, reopens read-only, verifies MONAD008 magic and that `history_length` survived at its new offset.
- `cli_tool.upgrade_requires_storage` — `--upgrade` without `--storage` rejected.
- `update_aux_death_test.aborts_on_monad007_without_allow_migration` — service-path abort with the `"monad-mpt --upgrade"` message.

### DIFF
--- a/category/async/storage_pool.cpp
+++ b/category/async/storage_pool.cpp
@@ -657,6 +657,7 @@ storage_pool::storage_pool(
     storage_pool const *const src, clone_as_read_only_tag_)
     : is_read_only_(true)
     , is_read_only_allow_dirty_(false)
+    , is_migration_allowed_(false)
     , is_newly_truncated_(false)
 {
     devices_.reserve(src->devices_.size());
@@ -711,6 +712,7 @@ storage_pool::storage_pool(
     creation_flags const flags)
     : is_read_only_(flags.open_read_only || flags.open_read_only_allow_dirty)
     , is_read_only_allow_dirty_(flags.open_read_only_allow_dirty)
+    , is_migration_allowed_(flags.allow_migration)
     , is_newly_truncated_(mode == mode::truncate)
 {
     devices_.reserve(sources.size());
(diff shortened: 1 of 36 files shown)


maxkozlovsky:
merged into previous prs

PR #2232, opened 2026-04-24
Rust TrieDB reshape

andr-dev:
This change cleans up some of the naming in the triedb rust crate and prepares the crate to use `cxx`.


rust/crates/monad-triedb/src/lib.rs
```diff
@@ -418,15 +418,15 @@ impl TriedbHandle {
...
 
     pub fn latest_verified_block(&self) -> Option<u64> {
-        parse_triedb_block_num(unsafe { ffi::triedb_latest_verified_block(self.db_ptr) })
+        parse_triedb_block_num(unsafe { ffi::triedb_latest_verified_version(self.db_ptr) })
     }
 
     pub fn earliest_finalized_block(&self) -> Option<u64> {
-        parse_triedb_block_num(unsafe { ffi::triedb_earliest_finalized_block(self.db_ptr) })
+        parse_triedb_block_num(unsafe { ffi::triedb_earliest_version(self.db_ptr) })
     }
```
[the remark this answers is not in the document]
andr-dev: This rename exposed that the `earliest_finalized_block` method actually just produces the earliest version in the database. I believe for execution to start up, it's required that this version be finalized?

rust/crates/monad-triedb/include/ffi.h
```diff
@@ -61,37 +62,41 @@ typedef void (*triedb_async_traverse_callback_fn)(
...
+uint64_t triedb_latest_finalized_version(TriedbRoInner *);
 // returns MAX if doesn't exist
-uint64_t triedb_latest_verified_block(triedb *);
+uint64_t triedb_latest_verified_version(TriedbRoInner *);
 
 // returns MAX if doesn't exist
-uint64_t triedb_earliest_finalized_block(triedb *);
+uint64_t triedb_earliest_version(TriedbRoInner *);
+// returns MAX if doesn't exist
+uint64_t triedb_latest_version(TriedbRoInner *);
```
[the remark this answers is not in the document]
andr-dev: This is intended as it will be used in a subsequent PR.

rust/crates/monad-triedb/CMakeLists.txt
```diff
@@ -27,15 +27,15 @@ endif()
...
   IMPORTED_LOCATION "${MONAD_EXECUTION_DIR}/libmonad_execution.so"
 )
 
-add_subdirectory("../../../../" "monad-execution-build"
+add_subdirectory("../../../" "monad-execution-build"
 )# needed for monad-trie
 
-add_library(triedb_driver SHARED "${PROJECT_SOURCE_DIR}/include/triedb.h"
-                                 "${PROJECT_SOURCE_DIR}/src/triedb.cpp")
+add_library(triedb_driver SHARED "${PROJECT_SOURCE_DIR}/include/ffi.h"
```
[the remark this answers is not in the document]
andr-dev: We'll be removing the `CMakeLists.txt` soon.

rust/crates/monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -134,23 +143,19 @@ int triedb_read(
...
+        void *user;
 
         void set_value(
             monad::async::erased_connected_operation *state,
             monad::async::result<monad::byte_string> result)
         {
             uint8_t const *value = nullptr;
             int length = 0;
-            triedb_async_read_callback_fn const callback = callback_;
-            void *const user = user_;
```
es3n1n, line 152: i think this change introduces a use-after-free, since the `delete state` statement prior to calling callback is destroying `this` as well (`monad::async::connect` returns the state struct that owns the receiver by value). 

so all the subsequent accesses to `callback`, `user` (L182) will be using a free'd pointer

andr-dev: I think I see what you mean, do you know why we used to copy the pointers locally and call `delete state` before calling the callback fn? Is this a common C++ idiom or something?

es3n1n: I _think_ it's done to free op-owned resources before user code runs (which can be slow / chain another read, and there's no point holding that memory), and to let the allocator's tcache hand the same block back to the next allocation

rust/crates/monad-triedb/src/ffi.cpp
```diff
@@ -392,36 +402,36 @@ void triedb_async_ranged_get(
                 value.size());
         });
     (new auto(monad::async::connect(
-         monad::mpt::make_get_node_sender(db->ctx_.get(), prefix, block_id),
-         detail::GetNodeReceiver(
+         monad::mpt::make_get_node_sender(&db->async_ctx, prefix, block_id),
+         TraverseRootReceiver(
```
Chen-Yifan: `GetRootForTraverseReceiver` would be a better name because this is the receiver for `make_get_node_sender()`.

rust/crates/monad-triedb/src/ffi.cpp
```diff
@@ -255,11 +265,11 @@ namespace detail
 
         virtual std::unique_ptr<TraverseMachine> clone() const override
         {
-            return std::make_unique<Traverse>(*this);
+            return std::make_unique<CallbackTraverse>(*this);
         }
     };
 
-    struct TraverseReceiver
+    struct TraverseFinishedReceiver
```
Chen-Yifan: I would prefer the original name `TraverseReceiver` because it pairs with `TraverseSender`

rust/crates/monad-triedb/src/ffi.cpp
```diff
@@ -177,26 +182,31 @@ void triedb_async_read(
...
-        receiver_t{callback, user}));
+        AsyncReadReceiver{callback, user}));
     state->initiate();
 }
 
-namespace detail
+namespace
 {
-    class Traverse final : public monad::mpt::TraverseMachine
+    class CallbackTraverse final : public monad::mpt::TraverseMachine
```
Chen-Yifan: suggest name change to `TraverseMachineWithCallback`?

Chen-Yifan (commented):
Suggested some name changes. The use-after-free concern raised by @es3n1n is valid.

marcograss (approved):
LGTM

rust/crates/monad-triedb/src/lib.rs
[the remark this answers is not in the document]
Chen-Yifan: one corner case at the beginning of a chain is that `earliest_version` is set to some valid value, but it hasn't finalized anything so `latest_finalized_version = INVALID_BLOCK_NUM`. In this case, `earliest_finalized_block ` should be `INVALID_BLOCK_NUM` instead of `= earliest_version`. 
This is not introduced by this PR, but would suggest fixing it by exposing a `triedb_earliest_finalized_version`

Chen-Yifan (approved):

rust/crates/monad-triedb/src/lib.rs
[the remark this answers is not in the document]
andr-dev: Agreed, will change in a subsequent PR to keep this one non-behavioral..

Baltoli (approved):

PR #2233, opened 2026-04-24
[execution]: Automatically distribute priority fees to stakers

kkuehlz:
The staking contract already keeps state of the current proposer. This was the larger piece of work, and this PR strings together existing functionality in the staking contract.

At the end of each transaction, instead of transferring fees to the beneficary, is collected in a special staging address. MIP-11 refers to this as the "distribution account", which is the constant `PRIORITY_FEE_DIST_ADDRESS` introduced in this PR.

At the end of each block, the balance in distribution account balance is transferred to the block proposers validator pool using an internal version of the externalRewards() precompile. If the fees are less than the dust, no balance is transferred.

https://github.com/monad-crypto/MIPs/blob/main/MIPs/MIP-11.md

### DIFF
--- a/category/execution/monad/staking/priority_fee.cpp
+++ b/category/execution/monad/staking/priority_fee.cpp
@@ -1,0 +1,54 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/core/int.hpp>
+#include <category/execution/ethereum/state3/state.hpp>
+#include <category/execution/ethereum/trace/call_tracer.hpp>
+#include <category/execution/monad/staking/priority_fee.hpp>
+#include <category/execution/monad/staking/staking_contract.hpp>
+
+MONAD_STAKING_NAMESPACE_BEGIN
+
+void collect_priority_fee(State &state, uint256_t const &fee)
+{
+    state.add_to_balance(PRIORITY_FEE_DIST_ADDRESS, fee);
+}
+
+void distribute_priority_fees(State &state)
+{
+    uint256_t const fees = state.get_balance(PRIORITY_FEE_DIST_ADDRESS);
+    if (fees == 0) {
+        return;
+    }
+
+    // Drain unconditionally: the subtract is outside the push/pop, so a
+    // failed distribution burns the fees rather than leaving them in the
+    // dist address.
+    state.subtract_from_balance(PRIORITY_FEE_DIST_ADDRESS, fees);
+
+    state.push();
+    state.add_to_balance(STAKING_CA, fees);
+    NoopCallTracer call_tracer;
+    StakingContract contract(state, call_tracer);
+    auto const res = contract.distribute_priority_fees(fees);
+    if (res.has_error()) {
+        state.pop_reject();
+    }
+    else {
+        state.pop_accept();
+    }
+}
+
+MONAD_STAKING_NAMESPACE_END
--- a/category/execution/monad/staking/priority_fee.cpp
+++ b/category/execution/monad/staking/priority_fee.cpp
@@ -14,10 +14,13 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #include <category/core/int.hpp>
+#include <category/core/log.hpp>
+#include <category/core/monad_exception.hpp>
 #include <category/execution/ethereum/state3/state.hpp>
 #include <category/execution/ethereum/trace/call_tracer.hpp>
 #include <category/execution/monad/staking/priority_fee.hpp>
 #include <category/execution/monad/staking/staking_contract.hpp>
+#include <category/execution/monad/staking/util/staking_error.hpp>
 
 MONAD_STAKING_NAMESPACE_BEGIN
 
@@ -44,6 +47,16 @@ void distribute_priority_fees(State &state)
     StakingContract contract(state, call_tracer);
     auto const res = contract.distribute_priority_fees(fees);
     if (res.has_error()) {
+        LOG_ERROR(
+            "staking: distribute priority fee reverted: {}",
+            res.error().message());
+        // At the start of block execution, the proposer id is always cleared to
+        // 0 and is set by the reward syscall. If the reward syscall was not
+        // included, `UnknownValidator` will be returned, and that is the only
+        // state we permit.
+        MONAD_ASSERT_THROW(
+            res.error() == StakingError::UnknownValidator,
+            "fee distribution error");
         state.pop_reject();
     }
     else {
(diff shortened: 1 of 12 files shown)


category/execution/monad/staking/staking_contract.cpp
```diff
@@ -1838,6 +1817,51 @@ Result<void> StakingContract::syscall_snapshot(
...
 
+Result<void> StakingContract::apply_external_reward(
+    u64_be const val_id, Address const &sender,
+    uint256_t const &external_reward, bool const apply_commission)
+{
+    // 1. Only validators in the consensus set can invoke this method.
+    auto val_execution = vars.val_execution(val_id);
+    if (MONAD_UNLIKELY(!val_execution.exists())) {
+        return StakingError::UnknownValidator;
+    }
```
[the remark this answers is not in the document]
kkuehlz: fixed wording.

category/execution/monad/staking/priority_fee.cpp
```diff
@@ -0,0 +1,86 @@
...
+        return;
+    }
+
+    // Drain the dist address unconditionally.  Failed distributions burn the
+    // fees. In practice, this only can happen when the fee is below the
+    // staking dust threshold, or if consensus provides an invalid proposer
+    // address.
+    state.subtract_from_balance(
+        PRIORITY_FEE_DIST_ADDRESS, priority_fees_this_block);
+
```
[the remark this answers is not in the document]
kkuehlz: this is by design and there is a test for it.

category/execution/monad/staking/util/constants.hpp
```diff
@@ -36,6 +36,9 @@ using namespace intx::literals;
 // staking contract address
 inline constexpr Address STAKING_CA{0x1000};
 
+inline constexpr Address PRIORITY_FEE_DIST_ADDRESS{
+    0xfee5fee5fee5fee5fee5fee5fee5fee5fee5fee5_address};
```
Baltoli, line 38: This isn't in the MIP - should the intermediate fee collection address be called out there at all?

kkuehlz: @jhb10c

Baltoli (dismissed):
Approving the traits / versioning changes; will let @jhb10c sign off on the math / staking components

jhunsaker:
I thought we agreed this was a generic redistribution account?

category/execution/monad/staking/staking_contract.cpp
```diff
@@ -1837,6 +1842,47 @@ Result<void> StakingContract::syscall_snapshot(
...
+    uint256_t const active_stake = consensus_view.stake().load().native();
+    if (MONAD_UNLIKELY(active_stake == 0)) {
+        return StakingError::NotInValidatorSet;
+    }
+
+    // 2. Sub-dust fees burn rather than distribute.
+    //
+    // NOTE: max external reward is deliberately NOT checked. In-protocol
+    // distribution has no upper-bound.
+    if (MONAD_UNLIKELY(fees < limits::min_external_reward())) {
```
jhb10c: This should check that the balance after commission is removed is below the minimum. Since there could be a scenario where fees - commission < min_external_reward. This would result in an update to the accumulator with sub min_external_reward values.

category/execution/monad/staking/staking_contract.cpp
```diff
@@ -1837,6 +1842,47 @@ Result<void> StakingContract::syscall_snapshot(
...
+        auto const split, split_commission(fees, commission_rate));
+    auto const &[commission, del_reward] = split;
+    auto auth = vars.delegator(val_id, val_execution.auth_address());
+    BOOST_OUTCOME_TRY(
+        auto const auth_reward,
+        checked_add(auth.rewards().load().native(), commission));
+    auth.rewards().store(auth_reward);
+
+    // 4. Update validator accumulator.
+    BOOST_OUTCOME_TRY(apply_reward(
```
jhb10c: Another question related to the above context is what is the intended behaviour if we are below the threshold. Since the commissions by pass the accumulators completely any amount is "safe" to credit. So is the expected behaviour that below threshold commission is still applied but remainder is burned? or is it that below threshold both amounts is burned?

category/execution/monad/staking/staking_contract.hpp
```diff
@@ -620,6 +620,11 @@ class StakingContract
     Result<void> syscall_on_epoch_change(byte_string_view, uint256_t const &);
 
     Result<void> syscall_snapshot(byte_string_view, uint256_t const &);
+
+    /////////////////////////////////
+    //  In-protocol functionality  //
+    /////////////////////////////////
+    Result<void> distribute_priority_fees(uint256_t const &fees);
```
jhb10c, line 627: Note: remove _priority_fees from function naming

category/execution/monad/staking/staking_contract.cpp
kkuehlz: nice. added a test, confirmed this, made fix.

category/execution/monad/staking/staking_contract.cpp
kkuehlz: commission is going to be reverted as well. keeps the code simpler. it's not worth the complexity to preserve dust commission

category/execution/monad/validate_monad_block_test.cpp
0xSqualo: @kkuehlz is this missing setting `txns[x].data = reward_calldata()`?

It seems this test case may be silently succeeding because of the [early shortcircuit here](https://github.com/category-labs/monad/blob/82d8767f909d38f7eebfb6e3beea7d5cf1c2955e/category/execution/monad/validate_monad_block.cpp#L92).

jhb10c (dismissed):
There are a few minor nits. 

- The failed path for the consensus validator check is not specifically tested unlike the missing execution validator. This would only happen if execution validator joins and is not in consensus set yet. But from the code it follows that it would revert so not necessary to add test. 
- the comments in distribute_priority_fees are misnumbered. 

Otherwise the tests cover the overall behaviour of distribute_priority_fees (both of them). This function being the spine of this update. 

The layout of this function is straight forward: 
- If validator exists and has been set this block then allocate commission to auth account and remainder if sufficient to delegator pool. If Validator does not exist or reward not set then all fees are burned.

andreaslyn (dismissed):

category/execution/monad/validate_monad_block_test.cpp
kkuehlz: done

andreaslyn (dismissed):

category/execution/monad/staking/priority_fee.cpp
```diff
@@ -0,0 +1,67 @@
...
+    auto const res = contract.distribute_priority_fees(fees);
+    if (res.has_error()) {
+        LOG_ERROR(
+            "staking: distribute priority fee reverted: {}",
+            res.error().message());
+        // At the start of block execution, the proposer id is always cleared to
+        // 0 and is set by the reward syscall. If the reward syscall was not
+        // included, `UnknownValidator` will be returned, and that is the only
+        // state we permit.
+        MONAD_ASSERT_THROW(
```
marcograss, line 57: just a nit, here the monad_assert_throw only tolerates UnknownValidator. Any other error is unreachable today but it feels a bit fragile if the code gets changed

marcograss (commented):
I left a comment for a nit improvement

marcograss (dismissed):

category/execution/monad/staking/priority_fee.cpp
kkuehlz: this was by design. we don't want to absorb other errors, like a `MathError` from the reward helper.

jhb10c (dismissed):
this lgtm as per the previous review, updated enforcements, and fuzzing

Baltoli (dismissed):
Approving but has a rebase conflict from MIP-8's trait; will reapprove when fixed

Baltoli (dismissed):

andreaslyn (approved):

jhb10c (approved):

marcograss (approved):

PR #2241, opened 2026-04-28
fix(execution): preserve access lists for reverted frames

mmv08:
# Preserve access-list traces for reverted frames

Context: this was discovered while working on https://github.com/category-labs/monad-bft/issues/2684. After fixing the RPC layer so `eth_createAccessList` could return execution failures inside the result object, the test still failed because the returned access list was empty. Further debugging showed that the execution tracer was losing reverted-frame storage accesses before RPC could encode them.

## Problem

`eth_createAccessList` should still return the access list collected during simulation when the simulated call reverts. The revert should be reported inside the JSON-RPC result, not by discarding the access-list result as a top-level RPC error.

This was not happening on Monad testnet. A contract that reads storage and then
reverts returned a top-level JSON-RPC error:

```bash
curl --request POST \
  --url https://testnet-rpc.monad.xyz/ \
  --header 'Content-Type: application/json' \
  --header 'User-Agent: insomnia/12.5.0' \
  --data '{
        "params": [{
                "to": "0x25Db08EdD8ffC4FF026fECa9F38c0f635d563132",
                "data": "0x7300000000000000000000000000000000000000000000000000000000000000"
        }
        ],
        "method": "eth_createAccessList",
        "jsonrpc": "2.0",
        "id": 73
}'
```

Observed Monad response:

```json
{"jsonrpc":"2.0","error":{"code":-32603,"message":"execution reverted"},"id":73}
```

The equivalent behavior on Sepolia returns a normal JSON-RPC result containing both the access list and the execution error:

```bash
curl --request POST \
  --url https://sepolia.drpc.org/ \
  --header 'Content-Type: application/json' \
  --header 'User-Agent: insomnia/12.5.0' \
  --data '{
        "params": [{
                "to": "0xCBBD82B84BE1C5975bd6928858143eE2fDd46122",
                "data": "0x7300000000000000000000000000000000000000000000000000000000000000"
        }
        ],
        "method": "eth_createAccessList",
        "jsonrpc": "2.0",
        "id": 73
}'
```

Observed Sepolia response:

```json
{
  "id": 73,
  "jsonrpc": "2.0",
  "result": {
    "accessList": [
      {
        "address": "0xcbbd82b84be1c5975bd6928858143ee2fdd46122",
        "storageKeys": [
          "0x7300000000000000000000000000000000000000000000000000000000000000"
        ]
      }
    ],
    "error": "execution reverted",
    "gasUsed": "0x63d0"
  }
}
```

The important behavior is that the access list survives the revert. The caller can still see which account and storage key were touched before execution failed.

## Reproduction Contract

The contract used to isolate the problem is intentionally small: it reads a storage slot selected by calldata, then immediately reverts.

Runtime bytecode:

```text
6000355460006000fd
```

Disassembly:

```text
60 00    PUSH1 0x00      ; calldata offset
35       CALLDATALOAD    ; read the 32-byte storage slot from calldata
54       SLOAD           ; access that slot so it appears in the access list
60 00    PUSH1 0x00      ; revert data offset
60 00    PUSH1 0x00      ; revert data length
fd       REVERT          ; fail after the storage read
```

Initcode from the original walkthrough:

```text
6009600c60003960096000f36000355460006000fd
```

Disassembly:

```text
60 09    PUSH1 0x09      ; runtime length
60 0c    PUSH1 0x0c      ; runtime offset in this initcode
60 00    PUSH1 0x00      ; destination memory offset
39       CODECOPY        ; copy runtime into memory
60 09    PUSH1 0x09      ; runtime length
60 00    PUSH1 0x00      ; source memory offset
f3       RETURN          ; return runtime as deployed code
```

The high-signal property is that the contract can only produce the expected access list if the execution layer preserves the `SLOAD` access across the subsequent `REVERT`.

## Root Cause

The state rollback behavior was correct. Reverted frames should be rejected, and their frame-local state should not leak into the accepted parent state.

The bug was in the access-list tracing path. `eth_call_impl` ran `trace::run_tracer` after execution returned. For a reverted call, the execution path had already rejected the frame:

```text
execute_call_message
  -> VM executes contract
     -> SLOAD records accessed storage in the current State frame
     -> REVERT
  -> post_call(state, result)
     -> state.pop_reject()
  -> return to eth_call_impl
  -> trace::run_tracer(state_tracer, state)
     -> AccessListTracer reads state.current()
     -> reverted-frame storage access is already gone
```

So `AccessListTracer` was reading rollback-sensitive access metadata too late. Changing `State::pop_reject()` would have been the wrong fix because consensus rollback semantics should remain independent from RPC observability.

## Solution

The fix keeps rollback semantics unchanged and treats frame rejection as a state-tracer lifecycle event. State tracers that care about rejected-frame metadata can observe the frame immediately before it is popped, while tracers that do not care about this event remain no-ops.

Execution code receives the same tracing dependencies used elsewhere in the client: a `CallTracerBase&` for call tracing and a `trace::StateTracer&` for state tracing. `EvmcHost` stores both references, and the shared `reject_frame` helper calls `trace::on_frame_reject` directly while the failed frame is still visible in `State`.

`trace::on_frame_reject(StateTracer&, State&)` mirrors `trace::run_tracer`: it dispatches on the state-tracer variant. `std::monostate`, `PrestateTracer`, and `StateDiffTracer` do nothing for this event. `AccessListTracer` captures access metadata from the frame that is about to be rejected.

`AccessListTracer` owns an accumulated map of observed accesses. The frame-reject hook copies access-list metadata from the currently pushed frame into tracer-owned storage while that frame is still visible in `State`. The capture is scoped to the current frame's dirty accounts, so it preserves rollback-sensitive accesses without re-recording unrelated accepted state.

Failed `CALL` and `CREATE` paths go through a shared `reject_frame` helper. That helper notifies the state tracer, then calls `State::pop_reject()`, preserving the existing rollback behavior including the RIPEMD touched-account rule.

The final encode step still observes the accepted state through `trace::run_tracer` before producing the JSON result. That gives the tracer both categories of data:

- accepted-frame accesses still visible in `State`
- rejected-frame accesses that were captured before rollback

`trace::reset` clears access-list accumulation before each transaction execution attempt, so speculative retries do not leak observation state from an abandoned attempt.

## Tests

This PR adds focused coverage for the behavior that was missing:

- `EthCallFixture.access_list_trace_reverted_call`
  - runs the tiny `CALLDATALOAD; SLOAD; REVERT` contract through
    `monad_executor_eth_call_submit`
  - asserts the execution status is `EVMC_REVERT`
  - asserts an encoded access-list trace exists
  - asserts the trace contains the touched storage key
- `EthCallFixture.access_list_trace_nested_reverted_call`
  - verifies access-list tracing also preserves storage touched inside a nested
    reverted call
- `TraitsTest.create_revert_preserves_access_list_trace`
  - verifies rejected `CREATE` frames snapshot access-list metadata before the
    created account is rolled back
- `access_list_state_view_excludes_rejected_frame`
  - documents that rejected frame accesses must not leak back through `State`
  - reinforces that this is tracer-specific capture, not a rollback change
- `access_list_records_rejected_frame_storage`
  - verifies the state-tracer lifecycle hook preserves rejected-frame storage
    accesses
- `access_list_records_rejected_frame_regular_account`
  - verifies the same hook preserves rejected-frame account accesses without
    storage keys

Verified locally:

```text
cmake --build cmake-build-debug --target state_tracer_test monad_executor_test
./cmake-build-debug/category/execution/state_tracer_test
./cmake-build-debug/category/rpc/monad_executor_test --gtest_filter='EthCallFixture.access_list_trace:EthCallFixture.access_list_trace_reverted_call:EthCallFixture.access_list_trace_empty:EthCallFixture.access_list_trace_nested:EthCallFixture.access_list_trace_nested_reverted_call'
```

**Note**: this is my first PR in the repo, any tips on how to better test these changes are very welcome!

### DIFF
--- a/category/execution/ethereum/evmc_host.hpp
+++ b/category/execution/ethereum/evmc_host.hpp
@@ -26,6 +26,7 @@
 #include <category/execution/ethereum/reserve_balance.hpp>
 #include <category/execution/ethereum/state3/state.hpp>
 #include <category/execution/ethereum/trace/call_tracer.hpp>
+#include <category/execution/ethereum/trace/state_tracer.hpp>
 #include <category/execution/ethereum/transaction_gas.hpp>
 #include <category/vm/evm/delegation.hpp>
 #include <category/vm/evm/traits.hpp>
@@ -114,9 +115,11 @@ struct EvmcHost final : public EvmcHostBase
     std::optional<uint256_t> base_fee_per_gas_;
     uint64_t i_;
     ChainContext<traits> const &chain_ctx_;
+    trace::StateTracer &state_tracer_;
 
     EvmcHost(
-        CallTracerBase &call_tracer, evmc_tx_context const &tx_context,
+        CallTracerBase &call_tracer, trace::StateTracer &state_tracer,
+        evmc_tx_context const &tx_context,
         BlockHashBuffer const &block_hash_buffer, State &state,
         Transaction const &tx, std::optional<uint256_t> const base_fee_per_gas,
         uint64_t const i, ChainContext<traits> const &chain_ctx,
@@ -126,6 +129,7 @@ struct EvmcHost final : public EvmcHostBase
         , base_fee_per_gas_{base_fee_per_gas}
         , i_{i}
         , chain_ctx_{chain_ctx}
+        , state_tracer_{state_tracer}
     {
     }
 
@@ -240,9 +244,9 @@ struct EvmcHost final : public EvmcHostBase
     }
 };
 
-static_assert(sizeof(EvmcHost<EvmTraits<EVMC_LATEST_STABLE_REVISION>>) == 128);
+static_assert(sizeof(EvmcHost<EvmTraits<EVMC_LATEST_STABLE_REVISION>>) == 136);
 static_assert(alignof(EvmcHost<EvmTraits<EVMC_LATEST_STABLE_REVISION>>) == 8);
-static_assert(sizeof(EvmcHost<MonadTraits<MONAD_NEXT>>) == 128);
+static_assert(sizeof(EvmcHost<MonadTraits<MONAD_NEXT>>) == 136);
 static_assert(alignof(EvmcHost<MonadTraits<MONAD_NEXT>>) == 8);
 
 MONAD_NAMESPACE_END
--- a/category/execution/ethereum/execute_transaction.cpp
+++ b/category/execution/ethereum/execute_transaction.cpp
@@ -330,6 +330,7 @@ Result<evmc::Result> ExecuteTransaction<traits>::execute_impl2(State &state)
         get_tx_context<traits>(tx_, sender_, header_, chain_.get_chain_id());
     EvmcHost<traits> host{
         call_tracer_,
+        state_tracer_,
         tx_context,
         block_hash_buffer_,
         state,
@@ -425,6 +426,7 @@ Result<Receipt> ExecuteTransaction<traits>::operator()()
         state.set_original_nonce(sender_, tx_.nonce);
 
         call_tracer_.reset();
+        trace::reset(state_tracer_);
 
         auto result = execute_impl2(state);
 
@@ -449,6 +451,7 @@ Result<Receipt> ExecuteTransaction<traits>::operator()()
         State state{block_state_, Incarnation{header_.number, i_ + 1}};
 
         call_tracer_.reset();
+        trace::reset(state_tracer_);
 
         auto result = execute_impl2(state);
 
--- a/category/execution/ethereum/state3/state.hpp
+++ b/category/execution/ethereum/state3/state.hpp
@@ -39,7 +39,6 @@
 #include <cstdint>
 #include <deque>
 #include <optional>
-#include <vector>
 
 MONAD_NAMESPACE_BEGIN
 
@@ -112,6 +111,12 @@ public:
     void pop_accept();
 
     void pop_reject();
+
+    // Return addresses marked dirty (including touched/accessed accounts) in
+    // the currently pushed frame. Intended for observers that must inspect
+    // frame-local metadata immediately before pop_accept() or pop_reject();
+    // callers must not retain references beyond the frame pop.
+    Set<Address> const &current_frame_dirty_accounts() const;
 
     ////////////////////////////////////////
 
--- a/category/execution/ethereum/test/test_call_trace.cpp
+++ b/category/execution/ethereum/test/test_call_trace.cpp
@@ -161,8 +161,17 @@ TYPED_TEST(TraitsTest, execute_success)
         ChainContext<typename TestFixture::Trait>::debug_empty();
     constexpr std::span<std::optional<Address> const> authorities_empty{};
     uint256_t base_fee{0};
+    trace::StateTracer noop_state_tracer = std::monostate{};
     EvmcHost<typename TestFixture::Trait> host{
-        call_tracer, tx_context, buffer, s, tx, base_fee, 0, chain_ctx};
+        call_tracer,
+        noop_state_tracer,
+        tx_context,
+        buffer,
+        s,
+        tx,
+        base_fee,
+        0,
+        chain_ctx};
 
     auto const result =
         ExecuteTransactionNoValidation<typename TestFixture::Trait>(
@@ -238,8 +247,17 @@ TYPED_TEST(TraitsTest, execute_reverted_insufficient_balance)
         ChainContext<typename TestFixture::Trait>::debug_empty();
     constexpr std::span<std::optional<Address> const> authorities_empty{};
     uint256_t base_fee{0};
+    trace::StateTracer noop_state_tracer = std::monostate{};
     EvmcHost<typename TestFixture::Trait> host{
-        call_tracer, tx_context, buffer, s, tx, base_fee, 0, chain_ctx};
+        call_tracer,
+        noop_state_tracer,
+        tx_context,
+        buffer,
+        s,
+        tx,
+        base_fee,
+        0,
+        chain_ctx};
 
     auto const result =
         ExecuteTransactionNoValidation<typename TestFixture::Trait>(
@@ -320,8 +338,17 @@ TYPED_TEST(TraitsTest, create_call_trace)
         ChainContext<typename TestFixture::Trait>::debug_empty();
     constexpr std::span<std::optional<Address> const> authorities_empty{};
     uint256_t base_fee{0};
+    trace::StateTracer noop_state_tracer = std::monostate{};
     EvmcHost<typename TestFixture::Trait> host{
-        call_tracer, tx_context, buffer, s, tx, base_fee, 0, chain_ctx};
+        call_tracer,
+        noop_state_tracer,
+        tx_context,
+        buffer,
+        s,
+        tx,
+        base_fee,
+        0,
+        chain_ctx};
 
     auto const result =
         ExecuteTransactionNoValidation<typename TestFixture::Trait>(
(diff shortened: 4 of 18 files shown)


category/rpc/monad_executor_test.cpp
```diff
@@ -3088,6 +3193,121 @@ TEST_F(EthCallFixture, access_list_trace_nested)
...
+        ACCESS_LIST_TRACER,
+        true);
+    f.get();
+
+    EXPECT_TRUE(ctx.result->status_code == EVMC_SUCCESS);
+
+    std::vector<uint8_t> const encoded_trace(
+        ctx.result->encoded_trace,
+        ctx.result->encoded_trace + ctx.result->encoded_trace_len);
+
```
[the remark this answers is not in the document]
mmv08: this was fixed

category/rpc/monad_executor_test.cpp
```diff
@@ -3088,6 +3194,123 @@ TEST_F(EthCallFixture, access_list_trace_nested)
...
+            "address" : "0x00000000000000000000000000000000bbbbbbbb",
+            "storageKeys" : [
+                "0x0000000000000000000000000000000000000000000000000000000000000002"
+            ]
+        }
+    ])";
+
+    EXPECT_EQ(
+        nlohmann::json::parse(expected),
+        nlohmann::json::from_cbor(encoded_trace));
```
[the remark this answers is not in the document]
mmv08: adjust tracer to be deterministic

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -161,8 +161,17 @@ TYPED_TEST(TraitsTest, execute_success)
...
-        call_tracer, tx_context, buffer, s, tx, base_fee, 0, chain_ctx};
+        call_tracer,
+        tx_context,
+        buffer,
+        s,
+        tx,
+        base_fee,
+        0,
+        chain_ctx,
+        noop_state_tracer};
```
Baltoli: I'd reorder this constructor so that the call tracer and state tracer are adjacent - that way you avoid having to explicitly specify the defaulted parameters at the end of the list when you have a state tracer

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -240,6 +295,30 @@ namespace trace
 
     EXPLICIT_TRAITS_MEMBER(AccessListTracer::should_exclude_address);
 
+    void on_frame_reject(StateTracer &tracer, State &state)
+    {
+        std::visit(
+            Cases{
+                [](std::monostate const &) {},
+                [](PrestateTracer &) {},
+                [](StateDiffTracer &) {},
```
Baltoli: You can collapse all the no-op cases into one `auto&&` limb - the access list tracer one is the most constrained so will get selected correctly

mmv08: have just updated the pr!

category/execution/ethereum/execute_transaction.cpp
```diff
@@ -425,6 +426,7 @@ Result<Receipt> ExecuteTransaction<traits>::operator()()
         state.set_original_nonce(sender_, tx_.nonce);
 
         call_tracer_.reset();
+        trace::reset(state_tracer_);
```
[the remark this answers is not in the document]
mmv08: @Baltoli would you say this is something that should be fixed?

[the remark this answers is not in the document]
Baltoli: You can't get reverting system transactions (i.e. `ExecuteSystemTransaction::operator()` can't cause reverted frames to pile up incorrectly), _but_ I think it keeps the intent of the code clearer to reset the tracer there as well - calling `reset` is basically free

category/execution/ethereum/state3/state.hpp
```diff
@@ -113,6 +114,21 @@ class State
...
+    void for_each_current_frame_dirty_account(Func func) const
+    {
+        MONAD_ASSERT(version_);
+        MONAD_ASSERT(dirty_.size() == version_);
+
+        for (auto const &address : dirty_.back()) {
+            func(address);
+        }
+    }
+
```
dhil: Can we provide a `const& const` method similar to `current()` here rather than internalizing a bespoke iteration method?

category/execution/ethereum/state3/state.hpp
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <category/core/address.hpp>
+#include <category/core/assert.h>
```
Baltoli: I don't think the header needs this extra include

category/rpc/monad_executor_test.cpp
```diff
@@ -2886,6 +2886,112 @@ TEST_F(EthCallFixture, access_list_trace)
...
+        commit_sequential(tdb, {}, {}, BlockHeader{.number = i});
+    }
+
+    static constexpr auto sender =
+        0x00000000000000000000000000000000deadbeef_address;
+
+    // CALLDATALOAD(0); SLOAD; REVERT(0, 0)
+    static constexpr auto contract_address =
+        0x00000000000000000000000000000000aaaaaaaa_address;
+    auto const contract_code = 0x6000355460006000fd_bytes;
```
dhil: Nit:
```suggestion
    static constexpr auto contract_address =
        0x00000000000000000000000000000000aaaaaaaa_address;
    // CALLDATALOAD(0); SLOAD; REVERT(0, 0)
    auto const contract_code = 0x6000355460006000fd_bytes;
```

dhil: Same in the other tests.

category/execution/ethereum/trace/state_tracer.hpp
```diff
@@ -98,9 +98,30 @@ namespace trace
...
+
+        // Capture rollback-sensitive accesses from the frame that is about to
+        // be rejected. Must be called while that State frame is still pushed.
+        void capture_rejected_frame_accesses(State const &);
+
+        friend void on_frame_reject(
+            std::variant<
+                std::monostate, PrestateTracer, StateDiffTracer,
+                AccessListTracer> &,
+            State &);
```
Baltoli: I'd just make the record function public to avoid explicitly spelling out the variant type here as a friend

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -200,30 +202,82 @@ namespace trace
...
+
+        std::vector<AccessListEntry> entries;
+        entries.reserve(accesses_.size());
+        for (auto const &[address, storage_keys] : accesses_) {
+            auto &entry = entries.emplace_back();
+            entry.address = address;
+            entry.storage_keys.assign(storage_keys.begin(), storage_keys.end());
+            // Match go-ethereum's access-list tracer output order: storage
+            // keys are sorted within each address, then entries are sorted by
+            // address below.
```
dhil, line 257: Just curious here. Suppose we didn't sort the entries, would it break some downstream tools then?

category/execution/ethereum/evmc_host.hpp
```diff
@@ -126,9 +129,17 @@ struct EvmcHost final : public EvmcHostBase
         , base_fee_per_gas_{base_fee_per_gas}
         , i_{i}
         , chain_ctx_{chain_ctx}
+        , state_tracer_{state_tracer}
     {
     }
 
+    // Notify state tracers while the failed frame is still present in State;
```
dhil:
```suggestion
    // Notify the state tracer while the failed frame is still present in State;
```

category/execution/ethereum/evmc_host.hpp
```diff
@@ -126,9 +129,17 @@ struct EvmcHost final : public EvmcHostBase
...
+    // callers reject or discard the frame immediately after this hook.
+    void on_frame_reject()
```
dhil: As far as I can tell this function is only invoked in `reject_frame` in `evm.cpp`. I am not sure this warrants an abstraction. Instead we can simply make the member `state_tracer` public (following `call_tracer`) such that we can call `trace::on_frame_reject` directly in `evm.cpp`.

category/execution/ethereum/trace/state_tracer.cpp
mmv08: I don't think it would. It's more about having more determinism in RPC results for tests/snapshots/etc. But the specification does not mandate an order.

I can remove it to simplify the PR

dhil (approved):

Baltoli (approved):

PR #2251, opened 2026-05-04
Switch to use monad for vm fuzzer state

andreaslyn:


test/vm/fuzzer/assertions.cpp
```diff
@@ -13,63 +13,71 @@
...
+            MONAD_ASSERT(v.size() == 1);
+            auto const [_, is_new] = a_accs.insert({k, v.recent()});
+            MONAD_ASSERT(is_new);
+        }
+        std::unordered_map<Address, AccountState> b_accs;
+        for (auto const &[k, v] : b.current()) {
+            MONAD_ASSERT(v.size() == 1);
+            auto const [_, is_new] = b_accs.insert({k, v.recent()});
+            MONAD_ASSERT(is_new);
+        }
```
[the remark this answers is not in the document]
andreaslyn: It is a pre-condition that size is 1.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -122,183 +150,208 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
-        [](std::pair<address const, Account> const &p) noexcept {
-            auto const &acc = p.second;
-            return acc.erase_if_empty && acc.is_empty();
-        });
-
-    return result;
+    MONAD_ASSERT(rev == EVMC_OSAKA); // TODO switch to monad revisions
+    using traits = EvmTraits<EVMC_OSAKA>;
+    return message_call<traits>(
+        tstate, block_hash_buffer, tx, msg, block_header);
```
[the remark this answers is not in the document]
andreaslyn: Only osaka is supported at the moment

scripts/vm/par-fuzzer.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "Usage: $0 N FUZZER_ARGS..."
+    echo "  where 1 <= N <= 100"
+    exit 1
+fi
+
+pids=
+for i in `seq 1 $n`; do
+    "$script_dir"/fuzzer.sh --seed $i $@ &
+    pids+=($!)
+done
```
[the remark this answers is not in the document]
andreaslyn: Most if this is redundant or wrong. I will update `pids=()`.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -494,6 +554,86 @@ static bool toss(Engine &engine, double p)
...
+    std::condition_variable cv_;
+    std::thread thread_;
+    bool done_;
+
+    static constexpr auto timeout_ = std::chrono::seconds{20};
+
+public:
+    TimeoutWaitThread()
+        : thread_{[this] { this->timeout_wait(); }}
+        , done_{}
```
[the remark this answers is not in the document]
andreaslyn: Correct.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -122,183 +150,208 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
+    tstate.state.access_account(msg.sender); // tx sender is always warm
 
-    ++sender_acc.nonce;
-    sender_acc.balance -= block_gas_left * effective_gas_price;
+    auto tx = tx_from(tstate, msg);
 
-    Host host{rev, vm, state, block, hashes, tx};
+    MONAD_ASSERT(tx.to.has_value());
+    tstate.state.access_account(*tx.to);
+    tstate.state.add_to_balance(*tx.to, load_be<uint256_t>(msg.value));
```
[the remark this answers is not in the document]
andreaslyn: Yes it should credit the sender

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -494,6 +554,86 @@ static bool toss(Engine &engine, double p)
...
+        .transactions_root =
+            some_good_constant(engine).value.template store_be<bytes32_t>(),
+        .receipts_root =
+            some_good_constant(engine).value.template store_be<bytes32_t>(),
+        .prev_randao =
+            some_good_constant(engine).value.template store_be<bytes32_t>(),
+        .difficulty = load_be<uint256_t>(pre_difficulty),
+        .number = block_number,
+        .gas_limit = random_uint32(engine) % (block_gas_limit + 1),
+        .gas_used = random_uint32(engine) % (block_gas_limit + 1),
```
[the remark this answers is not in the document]
andreaslyn: This is intentionally setting random values.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -122,183 +150,208 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
+    tstate.state.access_account(msg.sender); // tx sender is always warm
 
-    ++sender_acc.nonce;
-    sender_acc.balance -= block_gas_left * effective_gas_price;
+    auto tx = tx_from(tstate, msg);
 
-    Host host{rev, vm, state, block, hashes, tx};
+    MONAD_ASSERT(tx.to.has_value());
+    tstate.state.access_account(*tx.to);
+    tstate.state.add_to_balance(*tx.to, load_be<uint256_t>(msg.value));
```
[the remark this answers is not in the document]
andreaslyn: Yes it should credit the sender

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -577,25 +732,25 @@ static void do_run(
...
                 [&](auto const &address) {
-                    if (auto *found = evmone_state.find(address);
-                        found != nullptr) {
-                        return found->code;
-                    }
-
-                    return evmc::bytes{};
+                    TransitionState tstate{monad_state, monad_vm};
+                    return tstate.state.get_code(address);
                 },
```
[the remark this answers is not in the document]
andreaslyn: This does not affect performance significantly, because the `fuzz_iteration` function is dominant.

scripts/vm/par-fuzzer.sh
```diff
@@ -0,0 +1,24 @@
...
+
+trap 'kill ${pids[@]} 2>/dev/null' EXIT INT TERM
```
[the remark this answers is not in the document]
andreaslyn: This is largely correct, but insignificant

scripts/vm/par-fuzzer.sh
```diff
@@ -0,0 +1,24 @@
...
+
+wait -n
```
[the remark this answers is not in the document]
andreaslyn: The suggested change is not preserving the intended semantics

category/vm/compiler.cpp
```diff
@@ -61,10 +62,21 @@ namespace monad::vm
 
     template <Traits traits>
     SharedNativecode Compiler::compile(
-        SharedIntercode const &icode, CompilerConfig const &config)
+        SharedIntercode const &icode, CompilerConfig const &pre_config,
+        bytes32_t const &fallback_base_name __attribute__((unused)))
```
Baltoli: Really tiny nitpick: we use `[[maybe_unused]]` everywhere else in the repo. It doesn't matter per se, but just for consistency

category/vm/utils/debug.cpp
```diff
@@ -14,14 +14,46 @@
...
         std::getenv("MONAD_COMPILER_FUZZING");
     bool is_fuzzing_monad_vm = is_fuzzing_monad_vm_env &&
                                std::strcmp(is_fuzzing_monad_vm_env, "1") == 0;
+
+    static auto const is_compiler_runtime_debug_trace_enabled_env =
+        std::getenv("MONAD_COMPILER_DEBUG_TRACE");
+    bool is_compiler_runtime_debug_trace_enabled =
+        is_compiler_runtime_debug_trace_enabled_env &&
+        std::strcmp(is_compiler_runtime_debug_trace_enabled_env, "1") == 0;
+
```
Baltoli, line 40: Very minor: we have two of this pattern now, so could save some intermediate variables and clean up slightly via a function:
```cpp
static bool env_flag_enabled(char const *const name)
{
    char const *const value = std::getenv(name);
    return value && std::strcmp(value, "1") == 0;
}
```

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -60,21 +61,50 @@
...
+
+struct InterpreterFuzzerVmVariant
+{
+};
+
+enum class FuzzerVmTag
+{
+    Compiler,
+    Interpreter
+};
```
goodlyrottenapple, line 103: Is this the same as `vm::VM::Mode`, i.e. is there a reason not to just use that enum?

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -578,25 +740,25 @@ static void do_run(
                 contract_addresses,
                 {genesis_address},
                 [&](auto const &address) {
-                    if (auto *found = evmone_state.find(address);
-                        found != nullptr) {
-                        return found->code;
-                    }
-
-                    return evmc::bytes{};
+                    TransitionState tstate{monad_state, monad_vm};
```
goodlyrottenapple: can we perhaps pull `TransitionState tstate{monad_state, monad_vm};` construction out of the lambda? seems a bit heavy to re-build each time it's called?

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -495,6 +563,85 @@ static bool toss(Engine &engine, double p)
...
+template <typename Engine>
+static TestBlockHashBuffer generate_test_block_hash_buffer(Engine &engine)
+{
+    TestBlockHashBuffer b;
+    for (size_t i = 0; i < BlockHashBuffer::N; ++i) {
+        auto pre_hash = some_good_constant(engine).value;
+        if (!pre_hash) {
+            pre_hash = random_constant(engine).value;
+        }
+        b.set_blockhash(i, store_be_as<bytes32_t>(pre_hash));
```
goodlyrottenapple, line 538: Based on the `if (!pre_hash) { ... }` do we want to assert that `pre_hash` is not zero or is this not a hard constraint, only something we ideally want to avoid?

test/vm/fuzzer/fuzzer.cpp
andreaslyn: VM mode has an additional Dual mode. It might make sense to test Dual mode now that we are using monad as the host, but this requires more thought than enabling Dual mode and calling it a day.

goodlyrottenapple: we could leave an assert that dual mode is currently disabled in the fuzzer and not expose it in CLI parsing?

test/vm/fuzzer/fuzzer.cpp
andreaslyn: Sure

test/vm/fuzzer/fuzzer.cpp
andreaslyn: Good point. I have added a comment also.

test/vm/fuzzer/fuzzer.cpp
andreaslyn: We could, but if we keep the types like this we do not need an assert. Indeed that was the original motivation for introducing this enum.

category/vm/vm.cpp
```diff
@@ -226,6 +252,17 @@ namespace monad::vm
 
     EXPLICIT_TRAITS_MEMBER(VM::execute_native_entrypoint_raw);
 
+    void VM::debug_set_execute_override(ExecuteOverride f
+                                        __attribute__((unused)))
```
Baltoli:
```suggestion
                                        [[maybe_unused]])
```

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -123,184 +153,215 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
-        [](std::pair<address const, Account> const &p) noexcept {
-            auto const &acc = p.second;
-            return acc.erase_if_empty && acc.is_empty();
-        });
-
-    return result;
+    MONAD_ASSERT(rev == MONAD_ETH_OSAKA); // TODO switch to monad revisions
+    using traits = EvmTraits<MONAD_ETH_OSAKA>;
+    return message_call<traits>(
+        tstate, block_hash_buffer, tx, msg, block_header);
```
Baltoli, line 256: Are you going to follow up on this? I presume the CLI revision selection will break in this state but not a big problem obviously

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -123,184 +153,215 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
+    }
+
+    TransitionState(FuzzerTestStateRef ts, FuzzerVm &execution_vm)
+        : test_state{ts}
+        , vm{vm_mode(execution_vm)}
+        , block_state{ts->trie_db, vm}
+        , state{block_state, Incarnation{0, 0}}
+    {
+        if (std::holds_alternative<evmc::VM>(execution_vm)) {
+            auto &e = std::get<evmc::VM>(execution_vm);
```
Baltoli: `get_if` is a bit shorter but not important really

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -123,184 +153,215 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
+                        *host,
+                        context,
+                        to_evmc_revision(rev),
+                        *msg,
+                        code,
+                        code_size);
+                });
+        }
+        else if (std::holds_alternative<CompilerConfig>(execution_vm)) {
+            auto const &config = std::get<CompilerConfig>(execution_vm);
```
Baltoli: Same

test/vm/fuzzer/fuzzer.cpp
andreaslyn: Yes here https://github.com/category-labs/monad/pull/2399

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -123,184 +153,215 @@ static constexpr std::string_view to_string(evmc_status_code const sc) noexcept
...
+
+private:
+    void commit(uint64_t block_number)
+    {
+        block_state.merge(state);
+        auto [released_state, released_code, _] =
+            std::move(block_state).release();
+        test::commit_simple(
+            test_state->trie_db,
+            std::move(*released_state),
```
Baltoli: I think this isn't actually moving anything (const ref parameter)

Baltoli (commented):
Only one design question: is it necessary / deliberate that we end up constructing a fresh VM with every `TransitionState`, and therefore end up with several VMs per iteration? If it is necessary I'm happy to approve when nitpicks are addressed

test/vm/fuzzer/fuzzer.cpp
andreaslyn: Good spot

andreaslyn:
> Only one design question: is it necessary / deliberate that we end up constructing a fresh VM with every `TransitionState`, and therefore end up with several VMs per iteration? If it is necessary I'm happy to approve when nitpicks are addressed

Good point. Sharing vm instance between iterations makes sense.

Baltoli (dismissed):

category/vm/utils/debug.hpp
```diff
@@ -34,4 +39,21 @@ namespace monad::vm::utils
...
+    extern bool is_compiler_runtime_debug_trace_enabled;
+#else
+    static constexpr bool is_compiler_runtime_debug_trace_enabled = false;
+#endif
+
+#ifdef MONAD_COMPILER_TESTING
+    std::optional<std::string>
+    make_compiler_asm_log_path(bytes32_t const &base_name);
+#else
+    __attribute__((always_inline)) constexpr std::optional<std::string>
```
goodlyrottenapple:
```suggestion
    [[gnu::always_inline]] constexpr std::optional<std::string>
```

goodlyrottenapple (dismissed):

goodlyrottenapple (approved):

Baltoli (approved):

PR #2253, opened 2026-05-04
eth_createAccessList: deduplicate storage keys by page under MIP-8

brett-monad:
Under page-gas (monad_pricing_version >= 2 / MONAD_NEXT), the EVM warms an entire 128-slot page when any slot on that page is accessed. Including multiple slots from the same page in the access list wastes users' gas (1,900 per entry) with no benefit.

AccessListTracer::encode now keeps only the minimum slot per page when page_gas_active(), reducing the returned access list to the cheapest set that achieves the same warming effect.

Adds monad_pricing_version 2 and page_gas_active()/storage_page_shift() to MonadTraits (concept + defaults + MonadRevision), and five MonadTraitsTest cases covering same-page deduplication (2-slot and 3-slot), different-page retention, one-address two-pages, and pre-MONAD_NEXT unchanged behaviour.

### DIFF
--- a/category/execution/ethereum/trace/state_tracer.cpp
+++ b/category/execution/ethereum/trace/state_tracer.cpp
@@ -29,6 +29,7 @@
 #include <category/execution/ethereum/state3/account_state.hpp>
 #include <category/execution/ethereum/state3/state.hpp>
 #include <category/execution/ethereum/trace/state_tracer.hpp>
+#include <category/execution/monad/db/storage_page.hpp>
 #include <category/vm/evm/explicit_traits.hpp>
 #include <category/vm/evm/traits.hpp>
 
@@ -261,11 +262,21 @@ namespace trace
         for (auto const &[address, storage_keys] : accesses_) {
             auto &entry = entries.emplace_back();
             entry.address = address;
-            entry.storage_keys.assign(storage_keys.begin(), storage_keys.end());
             // Match go-ethereum's access-list tracer output order: storage
             // keys are sorted within each address, then entries are sorted by
             // address below.
+            entry.storage_keys.assign(storage_keys.begin(), storage_keys.end());
             std::ranges::sort(entry.storage_keys);
+            if constexpr (traits::mip_8_active()) {
+                // Under page-gas (MIP-8), warming one slot warms all 128 slots
+                // on its page, so keep only one representative per page. The
+                // keys are already sorted ascending, so each page's slots are
+                // contiguous and unique() retains the first (= minimum) of
+                // each run -- avoiding charging users for redundant entries.
+                auto const dups = std::ranges::unique(
+                    entry.storage_keys, {}, compute_page_key);
+                entry.storage_keys.erase(dups.begin(), dups.end());
+            }
         }
         std::ranges::sort(entries, {}, &AccessListEntry::address);
 
--- a/category/execution/ethereum/trace/state_tracer_test.cpp
+++ b/category/execution/ethereum/trace/state_tracer_test.cpp
@@ -53,6 +53,7 @@
 
 #include <test_resource_data.h>
 
+#include <algorithm>
 #include <bit>
 #include <optional>
 #include <vector>
@@ -77,6 +78,10 @@ namespace
         0x8d8ebb65ec00cb973d4fe086a607728fd1b9de14aa48208381eed9592f0dee9a_bytes32;
     constexpr auto key7 =
         0xff896b09014882056009dedb136458f017fcef9a4729467d0d00b4fd413fb1f1_bytes32;
+    constexpr auto page0_slot1 =
+        0x0000000000000000000000000000000000000000000000000000000000000001_bytes32;
+    constexpr auto page1_slot0 =
+        0x0000000000000000000000000000000000000000000000000000000000000080_bytes32;
     constexpr auto value1 =
         0x0000000000000000000000000000000000000000000000000000000000000003_bytes32;
     constexpr auto value2 =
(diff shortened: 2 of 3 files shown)


category/vm/evm/traits.hpp
```diff
@@ -243,15 +255,30 @@ namespace monad
 
         // Pricing version 1 activates the changes in:
         // Monad specification §4: Opcode Gas Costs and Gas Refunds
+        // Pricing version 2 activates page-level storage gas scheduling
+        // (MIP-8).
         static consteval uint8_t monad_pricing_version() noexcept
         {
+            if constexpr (Rev >= MONAD_NEXT) {
+                return 2;
+            }
```
Baltoli: We don't want this to be a "pricing version" - that language was used for a wholesale repricing of precompiles / opcodes to match Monad resource costs. Instead, we should define a function `mip_8_active()` and predicate on that. That flag will end up being added by several different PRs, so it'll just end up rebasing out of whichever PR lands second.

category/vm/evm/traits.hpp
```diff
@@ -243,15 +255,30 @@ namespace monad
...
 
+        static consteval bool page_gas_active() noexcept
+        {
+            return monad_pricing_version() >= 2;
+        }
+
+        static consteval uint8_t storage_page_shift() noexcept
+        {
+            return 7; // MIP-8: 128-slot pages, fixed for all page-gas revisions
+        }
```
Baltoli: This doesn't need to be a parameter, it's fine to hard-code it for now

category/vm/evm/traits.hpp
```diff
@@ -243,15 +255,30 @@ namespace monad
...
                 return 1;
             }
 
             return 0;
         }
 
+        static consteval bool page_gas_active() noexcept
+        {
+            return monad_pricing_version() >= 2;
+        }
```
Baltoli: See above

category/vm/evm/traits.hpp
```diff
@@ -56,13 +56,15 @@ namespace monad
         { T::eip_7883_active() } -> std::same_as<bool>;
         { T::eip_7951_active() } -> std::same_as<bool>;
         { T::mip_3_active() } -> std::same_as<bool>;
+        { T::page_gas_active() } -> std::same_as<bool>;
```
Baltoli: MIP-8

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -51,6 +51,26 @@ namespace trace
         return std::format("0x{}", to_hex(view));
     }
 
+    namespace
+    {
+        template <uint8_t shift>
```
Baltoli: again no need to be generic over page size; folding the constant 7 in will simplify this code

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -205,11 +221,31 @@ namespace trace
...
+                }
+                for (auto const &[_, slot] : page_to_min_slot) {
+                    keys.push_back(bytes_to_hex(slot.bytes));
+                }
+            }
+            else {
+                for (auto const &key : accessed) {
+                    keys.push_back(bytes_to_hex(key.bytes));
+                }
             }
```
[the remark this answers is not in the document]
Baltoli: > but the execution path still warms exact slots only

yes, this is pending execution changes for the MIP itself

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -51,6 +51,22 @@ namespace trace
         return std::format("0x{}", to_hex(view));
     }
 
+    namespace
+    {
+        constexpr bytes32_t compute_page_key(bytes32_t const &key) noexcept
+        {
+            constexpr uint8_t shift = constants::MIP8_STORAGE_PAGE_SHIFT;
+            bytes32_t result{};
+            result.bytes[0] = static_cast<uint8_t>(key.bytes[0] >> shift);
```
Baltoli: No real need for an intermediate variable here I don't think:
```suggestion
            bytes32_t result{};
            result.bytes[0] = static_cast<uint8_t>(key.bytes[0] >> constants::MIP8_STORAGE_PAGE_SHIFT);
```

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -51,6 +51,22 @@ namespace trace
         return std::format("0x{}", to_hex(view));
     }
 
+    namespace
+    {
+        constexpr bytes32_t compute_page_key(bytes32_t const &key) noexcept
```
Baltoli: Maybe worth a comment here that this function is doing the big-endian version of the shift / offset calculation in the MIP?

category/vm/evm/traits.hpp
```diff
@@ -241,8 +249,8 @@ namespace monad
             return false;
         }
 
-        // Pricing version 1 activates the changes in:
-        // Monad specification §4: Opcode Gas Costs and Gas Refunds
+        // Pricing version 1 activates Monad-specific opcode/precompile costs
+        // (Monad specification §4). Max return value is 1.
```
Baltoli: No need to update this comment

category/execution/ethereum/trace/state_tracer.cpp
brett-monad: shift is used multiple times so I decided to keep as it is more concise

marcograss (commented):
a couple of nits:

- after this PR base_sload_cost is not used anymore, should it be kept?
- compute_page_key is duplicated in several project location, could it be refactored in a single implementation to avoid drift? (the >> 7 shift)

marcograss (dismissed):

Baltoli:
> compute_page_key is duplicated in several project location, could it be refactored in a single implementation to avoid drift? (the >> 7 shift)

Yeah there are several PRs that all introduce that code, I think this PR will be one of the last to merge and so will rebase its implementation out

category/execution/ethereum/trace/state_tracer.cpp
```diff
@@ -261,7 +262,27 @@ namespace trace
...
+                    auto const page_key = compute_page_key(key);
+                    auto const [it, inserted] =
+                        page_to_min_slot.try_emplace(page_key, key);
+                    if (!inserted && key < it->second) {
+                        it->second = key;
+                    }
+                }
+                for (auto const &[_, slot] : page_to_min_slot) {
+                    entry.storage_keys.push_back(slot);
+                }
```
Baltoli: I think there's a better way to get the minimum pages, actually - we already have the storage keys per address, so instead of copying them out into a separately allocated map, we could instead deduplicate the existing vector of storage keys under the projection to their page. Something like this (with the existing comment):
```cpp
  entry.storage_keys.assign(storage_keys.begin(), storage_keys.end());
  std::ranges::sort(entry.storage_keys);
  if constexpr (traits::mip_8_active()) {
      auto const dup = std::ranges::unique(
          entry.storage_keys, [](bytes32_t const &a, bytes32_t const &b) {
              return compute_page_key(a) == compute_page_key(b);
          });
      entry.storage_keys.erase(dup.begin(), dup.end());
  }
```

Baltoli (approved):

dhil (approved):

marcograss:
approved

PR #2254, opened 2026-05-05
core: copy silkpre's sha256 and rmd160 into third_party

Baltoli:
Vendors two silkpre crypto sources into `category/core/crypto/` so monad's call sites no longer go through the silkpre wrappers. Two commits, one per primitive.

### sha256 (7c1cb9820)

Copies `silkpre/lib/silkpre/sha256.{c,h}` verbatim apart from:
  - renaming `silkpre_sha256` to `monad_sha256` (and the matching header guard) so the new symbol can coexist with silkpre's during the migration
  - making the internal helper `cpuid` static — it was non-static in silkpre upstream, which only worked because there was a single copy of the file in the link

`NOTICE` + `SILKPRE-LICENSE.txt` are dropped alongside, mirroring the pattern in `utils/clang-tidy-auto-const/`. Sources are added to `monad_core` with per-file `-Wno-conversion` / `-Wno-sign-conversion` to keep `-Werror` honest without modifying the vendored code, and a `.clang-format` / `.clang-tidy` pair in the directory excludes the vendored files from the project lint/format scripts.

In-tree direct callers are switched to `monad_sha256`:
  - `sha256_execute` (precompiles_impl.cpp) is rewritten to malloc its output buffer and call `monad_sha256` directly, matching the ownership pattern used by `identity_execute` / `point_evaluation_execute`
  - `kzg_to_version_hashed` (precompiles_impl.cpp) and the two call sites in `compute_requests_hash` (process_requests.cpp) just swap the symbol name

### rmd160 (f370b1eee)

Mirror of the sha256 vendor: copies `silkpre/lib/silkpre/rmd160.{c,h}` verbatim apart from renaming `silkpre_rmd160` to `monad_rmd160` (and the matching header guard). The vendored file exports a single symbol, so no static-helper fixups were needed.

The KU Leuven 1996 MIT-style notice already in the file headers is self-contained, so no separate LICENSE drop is required; `NOTICE` is extended to describe the rmd160 provenance, and the per-file `-Wno-conversion` / `-Wno-sign-conversion` treatment is reused.

`ripemd160_execute` (precompiles_impl.cpp) is rewritten to malloc its 32-byte output, zero the top 12 bytes, and write the 20-byte digest at offset 12 — matching `silkpre_rip160_run`'s behaviour exactly.

### Notes

`silkpre_sha256_run` and the ripemd160 silkpre wrapper, used by silkpre internally, still call silkpre's own implementations, but neither is reached from monad code. Dropping the AArch64 path from the vendored sha256.c is a follow-up.

### DIFF
--- a/category/execution/ethereum/precompiles_impl.cpp
+++ b/category/execution/ethereum/precompiles_impl.cpp
@@ -36,13 +36,15 @@
 
 #include <evmc/evmc.h>
 
+#include <silkpre_vendor/rmd160.h>
+#include <silkpre_vendor/sha256.h>
+
 #include <setup/settings.h>
 #include <setup/setup.h>
 
 #include <stdio.h>
 
 #include <silkpre/precompile.h>
-#include <silkpre/sha256.h>
 
 #include <cstdint>
 #include <cstdlib>
@@ -58,8 +60,7 @@ namespace
     {
         constexpr uint8_t VERSION_HASH_VERSION_KZG = 1;
         monad::bytes32_t h;
-        // TODO: remove silkpre
-        silkpre_sha256(
+        monad_sha256(
             h.bytes,
             commitment.bytes,
             sizeof(KZGCommitment),
@@ -114,6 +115,15 @@ static inline PrecompileResult silkpre_execute(byte_string_view const input)
     return {EVMC_SUCCESS, output, output_size};
 }
 
+// Substitute a pointer to the empty string when `input.data()` is null.
+// monad_sha256 / monad_rmd160 invoke undefined behaviour on a null input
+// pointer even when the length is zero.
+static inline uint8_t const *nonnull_input_data(byte_string_view const input)
+{
+    return input.data() != nullptr ? input.data()
+                                   : reinterpret_cast<uint8_t const *>("");
+}
+
 PrecompileResult ecrecover_execute(byte_string_view const input)
 {
     auto const clamped_input = input.substr(0, 128);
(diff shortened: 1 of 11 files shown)


category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -114,20 +115,30 @@ PrecompileResult ecrecover_execute(byte_string_view const input)
...
-    }
-    return silkpre_execute<silkpre_sha256_run>(input);
+    auto *const output = static_cast<uint8_t *>(std::malloc(32));
+    MONAD_ASSERT(output != nullptr);
+
+    // Avoid passing a null pointer to monad_sha256; even with length zero the
+    // buffer-state setup would invoke undefined behaviour.
+    auto const *const data = input.data() != nullptr
+                                 ? input.data()
+                                 : reinterpret_cast<uint8_t const *>("");
```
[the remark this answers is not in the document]
mkolosick: This is wrong and misleading. `memcpy(..., nullptr, 0)` (which is called in `calc_chunk` when a length of zero is passed) is undefined behavior even if implementations *probably* just treat it as a no-op.

mkolosick (dismissed):

third_party/silkpre_vendor/src/silkpre_vendor/rmd160.c
```diff
@@ -0,0 +1,389 @@
...
+    for (size_t remaining = len; remaining >= 64; remaining -= 64) {
+        for (unsigned i = 0; i < 16; ++i) {
+            current[i] = load32(ptr);
+            ptr += 4;
+        }
+        rmd160_compress(buf, current);
+    }
+
+    rmd160_finish(buf, ptr, len);
+
```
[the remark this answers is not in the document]
Baltoli: True of existing code; in the interests of unmodified porting I've left this as-is. Not an issue in practice given gas costing

category/execution/ethereum/precompiles_impl.cpp
```diff
@@ -114,20 +116,30 @@ PrecompileResult ecrecover_execute(byte_string_view const input)
...
 PrecompileResult ripemd160_execute(byte_string_view const input)
 {
-    return silkpre_execute<silkpre_rip160_run>(input);
+    auto *const output = static_cast<uint8_t *>(std::malloc(32));
+    MONAD_ASSERT(output != nullptr);
+
+    // Ethereum's RIPEMD-160 precompile returns the 20-byte digest left-padded
+    // with 12 zero bytes to a 32-byte ABI word.
+    std::memset(output, 0, 12);
+    monad_rmd160(output + 12, input.data(), input.size());
```
0xSqualo: @Baltoli should the `input` have a `nullptr` guard / ternary here to align with `sha256_execute()` here on input sanitization?

Baltoli: I don't _think_ there's a path to nullptr-memcpy as there is for SHA, but it probably doesn't hurt to add one defensively

mkolosick: the ripemd code pretty straightforwardly does nothing when the length is zero so I don't think this is necessary but it's okay to do.

mkolosick (dismissed):

goodlyrottenapple (dismissed):

0xSqualo (dismissed):

goodlyrottenapple (dismissed):

0xSqualo (approved):

goodlyrottenapple (approved):

PR #2261, opened 2026-05-11
[zkVM] CodeTracer for witness generation

goodlyrottenapple:
### Summary
  Adds a `CodeTracer` variant to the `StateTracer` so witness generation can record every code preimage read during block execution.

### Changes
  - **`trace::CodeTracer`** (new `StateTracer` alternative): holds a `Code` map of `code_hash → SharedIntercode`.
  `trace::on_read_code` inserts when the active tracer is `CodeTracer`, no-op otherwise.
  - **EVM host code reads**: `EvmcHostBase::get_code_size` / `copy_code` now resolve the code hash and call `on_read_code` when the `StateTracer` is a `CodeTracer`. `state_tracer_` has been promoted from `EvmcHost<traits>` to `EvmcHostBase` (`sizeof` 64 → 72) so the base class can hook these paths.
  - **Reserve-balance code reads**: threads `trace::StateTracer&` through `revert_transaction`, `init_reserve_balance_context`, `ReserveBalance::init_from_tx`, `dipped_into_reserve`, and `is_delegated`, recording the delegated-code lookups that happen.
  - **System-call tracing**: `process_requests` (EIP-7002 / EIP-7251 system calls) now also takes a `StateTracer&` and records code reads against it. To accommodate this, `state_tracers` is sized `senders.size() + 1`: index `0` is the block-level tracer used by `process_requests`, and transaction `i` now uses `state_tracers[i + 1]`.

### DIFF
--- a/category/execution/ethereum/evmc_host.cpp
+++ b/category/execution/ethereum/evmc_host.cpp
@@ -18,6 +18,7 @@
 #include <category/core/bytes.hpp>
 #include <category/core/config.hpp>
 #include <category/core/int.hpp>
+#include <category/core/likely.h>
 #include <category/execution/ethereum/block_hash_buffer.hpp>
 #include <category/execution/ethereum/block_hash_history.hpp>
 #include <category/execution/ethereum/core/receipt.hpp>
@@ -35,14 +36,15 @@
 MONAD_NAMESPACE_BEGIN
 
 EvmcHostBase::EvmcHostBase(
-    CallTracerBase &call_tracer, evmc_tx_context const &tx_context,
-    BlockHashBuffer const &block_hash_buffer, State &state,
-    bool const log_native_transfers) noexcept
+    CallTracerBase &call_tracer, trace::StateTracer &state_tracer,
+    evmc_tx_context const &tx_context, BlockHashBuffer const &block_hash_buffer,
+    State &state, bool const log_native_transfers) noexcept
     : block_hash_buffer_{block_hash_buffer}
     , tx_context_{tx_context}
     , state_{state}
     , call_tracer_{call_tracer}
     , log_native_transfers_{log_native_transfers}
+    , state_tracer_{state_tracer}
 {
 }
 
@@ -86,6 +88,17 @@ EvmcHostBase::get_balance(evmc::address const &address) const noexcept
 size_t EvmcHostBase::get_code_size(evmc::address const &address) const noexcept
 {
     try {
+        if (MONAD_UNLIKELY(
+                std::holds_alternative<trace::CodeTracer>(state_tracer_))) {
+            bytes32_t const hash = state_.get_code_hash(address);
+            if (hash == NULL_HASH) {
+                return 0;
+            }
+            auto const vcode = state_.read_code(hash);
+            MONAD_ASSERT(vcode);
+            trace::on_read_code(state_tracer_, hash, vcode->intercode());
+            return vcode->intercode()->size();
+        }
         return state_.get_code_size(address);
     }
     catch (...) {
@@ -114,6 +127,17 @@ size_t EvmcHostBase::copy_code(
     size_t const size) const noexcept
 {
     try {
+        if (MONAD_UNLIKELY(
+                std::holds_alternative<trace::CodeTracer>(state_tracer_))) {
+            bytes32_t const hash = state_.get_code_hash(address);
+            if (hash != NULL_HASH) {
+                auto const vcode = state_.read_code(hash);
+                MONAD_ASSERT(vcode);
+                trace::on_read_code(state_tracer_, hash, vcode->intercode());
+                return vcode->intercode()->copy_code(offset, data, size);
+            }
+            return 0;
+        }
         return state_.copy_code(address, offset, data, size);
     }
     catch (...) {
--- a/category/execution/ethereum/evmc_host.hpp
+++ b/category/execution/ethereum/evmc_host.hpp
@@ -57,9 +57,11 @@ protected:
     bool const log_native_transfers_;
 
 public:
+    trace::StateTracer &state_tracer_;
+
     EvmcHostBase(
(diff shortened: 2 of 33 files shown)


category/execution/ethereum/evmc_host.hpp
```diff
@@ -59,9 +59,11 @@ class EvmcHostBase : public vm::Host
     bool const log_native_transfers_;
 
 public:
+    trace::StateTracer &state_tracer_;
```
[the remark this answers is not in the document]
goodlyrottenapple: state_tracer_ was previously public in EvmcHost, so leaving the same visibility.

category/execution/ethereum/evmc_host.cpp
```diff
@@ -113,6 +125,14 @@ size_t EvmcHostBase::copy_code(
...
 {
     try {
+        if (MONAD_UNLIKELY(
+                std::holds_alternative<trace::CodeTracer>(state_tracer_))) {
+            bytes32_t const hash = state_.get_code_hash(address);
+            if (hash != NULL_HASH) {
+                auto const vcode = state_.read_code(hash);
+                trace::on_read_code(state_tracer_, hash, vcode->intercode());
+            }
+        }
```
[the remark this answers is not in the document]
mkolosick: I would like to see reduced duplication here, factoring out the actual copying block into a `SharedVarCode` helper between this and `State::copy_code`.

category/execution/ethereum/execute_block.cpp
```diff
@@ -210,7 +210,9 @@ Result<std::vector<Receipt>> execute_block_transactions(
 {
     MONAD_ASSERT(senders.size() == transactions.size());
     MONAD_ASSERT(senders.size() == call_tracers.size());
-    MONAD_ASSERT(senders.size() == state_tracers.size());
+    // state_tracers[0] is reserved for block-level system calls (see
+    // process_requests); tx i uses state_tracers[i + 1].
+    MONAD_ASSERT(senders.size() + 1 == state_tracers.size());
```
mkolosick: Is there a reason not to pull the block-level system call state_tracer (`state_tracers[0]`) out into its own argument/field?

goodlyrottenapple: replaced with a separate `system_call_state_tracer`

mkolosick (dismissed):

goodlyrottenapple:
@claude re-review

mkolosick (dismissed):

mkolosick:
I was worried about the maintainability of having to make sure to add `on_read_code` at new places that read code, but I think the future witness generation based tests will serve to catch this issue.

mkolosick (dismissed):

category/execution/ethereum/state3/state.hpp
```diff
@@ -190,6 +190,20 @@ class State
 
     size_t get_code_size(Address const &);
 
+    [[gnu::always_inline]]
+    static inline size_t copy_code(
```
dhil: This function appears to be independent of `State`, so I would think it should be defined outside of `State`.

dhil (commented):
This looks reasonable to me within the current state of things wrt intertwining of execution and tracing logic.

category/execution/ethereum/state3/state.hpp
goodlyrottenapple: yeah, not sure where the best place for it would be tho? could move to state3/utils.hpp perhaps?

dhil: We have some utils in `category/core/` (we even have a `category/core/util/`, but there is only one thing there).

category/execution/ethereum/state3/state.hpp
```diff
@@ -190,6 +190,20 @@ class State
...
+        vm::SharedIntercode const icode, size_t const offset,
```
Baltoli: We only read through `icode` here - should take as a const ref rather than by value

category/execution/ethereum/evmc_host.cpp
```diff
@@ -86,6 +88,17 @@ EvmcHostBase::get_balance(evmc::address const &address) const noexcept
...
+            bytes32_t const hash = state_.get_code_hash(address);
+            if (hash == NULL_HASH) {
+                return 0;
+            }
+            auto const vcode = state_.read_code(hash);
+            MONAD_ASSERT(vcode);
+            trace::on_read_code(state_tracer_, hash, vcode->intercode());
+            return vcode->intercode()->size();
+        }
         return state_.get_code_size(address);
```
Baltoli, line 102: @dhil note the short-circuiting here - something worth thinking about in your tracing redesign.

category/execution/ethereum/state3/state.hpp
Baltoli: Maybe it should live on `Intercode`?

Baltoli: Which would have the side effect of also resolving my comment below re. const ref

category/execution/ethereum/trace/state_tracer.hpp
```diff
@@ -121,8 +122,31 @@ namespace trace
...
+    /// Records every code preimage read during execution, keyed by
+    /// code_hash. Used by witness generation to assemble the codes section
+    /// of the post-block witness. Insertions happen from the EVM host's
+    /// code-read entry points; production execution uses `std::monostate`
+    /// instead, so the recording path has zero cost. The map is held by
+    /// `unique_ptr` so it does not bloat `sizeof(StateTracer)` when the
+    /// variant is in the `std::monostate` alternative.
+    struct CodeTracer
+    {
+        std::unique_ptr<Code> codes{std::make_unique<Code>()};
```
Baltoli: Doesn't need concurrent access, so can probably just be a plain `ankerl` map rather than `Code`?

Baltoli (commented):
Seems largely fine to me; a few minor points

Baltoli (approved):

dhil (approved):

PR #2263, opened 2026-05-12
Upgrade alloy rust version to 2.0

eerkaijun:
Block timestamp is an optional field in rpc responses. The version bump introduces this field, and since we have access to block timestamp, we can populate this field

### DIFF
--- a/rust/crates/monad-exec-events/src/block.rs
+++ b/rust/crates/monad-exec-events/src/block.rs
@@ -74,6 +74,8 @@ impl ExecutedBlock {
             excess_blob_gas: Some(0),
             parent_beacon_block_root: Some(alloy_primitives::B256::ZERO),
             requests_hash: Some(alloy_primitives::B256::ZERO),
+            block_access_list_hash: None,
+            slot_number: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


rust/crates/monad-exec-events/src/block.rs
```diff
@@ -74,6 +74,8 @@ impl ExecutedBlock {
             excess_blob_gas: Some(0),
             parent_beacon_block_root: Some(alloy_primitives::B256::ZERO),
             requests_hash: Some(alloy_primitives::B256::ZERO),
+            block_access_list_hash: None,
```
andr-dev, line 77: Shouldn't this be set to `keccak256(rlp_encode(block_access_list))`? Similarly for `slot_number` although that seems like an eth consensus detail we won't be copying over ...

eerkaijun: those are for upcoming eth forks which are not applicable yet

andr-dev (approved):

Baltoli (approved):

dhil (approved):

PR #2264, opened 2026-05-12
[barrett reduction] Clarify proof sketches and tighten generic template constraints

cmcl:
This PR revises proof sketches and tightens generic template constraints based on findings from [the Rocq model and correctness proofs branch](https://github.com/category-labs/monad/tree/cmcl/barrett-rocq-model).   Code changes are either semantics-preserving replacements or along dormant paths that are not exercised by the exported Barrett aliases, i.e. there is no change to exported Barrett behaviour

### DIFF
--- a/category/core/runtime/uint256.hpp
+++ b/category/core/runtime/uint256.hpp
@@ -1931,8 +1931,8 @@ namespace monad::vm::runtime
             /**
              * For Barrett reduction, we compute the quotient approximant
              *     q_hat = floor((x*reciprocal_) / 2^SHIFT)
-             * Under some conditions we can pre-divide x by 2^SHIFT to narrow
-             * the multiplications involved. The quantity computed is
+             * Under some conditions we can pre-divide x by 2^PRE_PRODUCT_SHIFT
+             * to narrow the multiplications involved. The quantity computed is
              *     q_hat = floor(
              *         floor(x/2^PRE_PRODUCT_SHIFT)*reciprocal_
              *         / 2^POST_PRODUCT_SHIFT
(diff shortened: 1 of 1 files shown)


cmcl:
@claude Can you review the changes?

category/core/runtime/uint256.hpp
```diff
@@ -2104,27 +2125,31 @@ namespace monad::vm::runtime
...
-                // therefore
-                //     q - 2 <= q1_hat <= q
+                if constexpr (!need_quotient) {
+                    static_assert(
+                        POST_PRODUCT_BIT_SHIFT == 0,
+                        "Barrett reduce<false> requires a word-aligned "
+                        "post-product shift");
+                }
+
+                // Let I = INPUT_BITS and B = PRE_PRODUCT_SHIFT.
```
m-alvarez: If the symbolic proof isn't unsound, I'd rather keep it than replace it by a rewrite in English

cmcl: The intended proof idea was fine, but the sketch used `S` and `B` ambiguously.  `S` should be the reciprocal scale, `SHIFT == INPUT_BITS`.  `B` should be `PRE_PRODUCT_SHIFT`, the number of low bits dropped before the multiply.  The old comment used `S` correctly in `floor(r * x / 2^S)`, but then split `x` as if `S`, rather than `B`, determined the low discarded part.  There was also a hard-coded 256-bit constant which does not align with the generic `INPUT_BITS` context here, for example when `INPUT_BITS == 257` for `addmod`.  Also, `B` is such that `2^B <= d`; equality is possible when `MIN_DENOMINATOR` is a power of two.

I'm happy to replace this with a revised symbolic proof sketch (as well as resurrect some of the prior wording) instead of the English-only description:

```text
Let S = SHIFT = INPUT_BITS and B = PRE_PRODUCT_SHIFT.

We are approximating the shifted product
    floor(r * x / 2^S)
by the shifted product
    floor(r * floor(x / 2^B) / 2^(S - B)).

As seen in the definition of PRE_PRODUCT_SHIFT, B is chosen so that
    2^B <= d.

To see why this is sound, let
    q = floor(x / d)
    x = x_1 * 2^B + x_0
        where x_1 = floor(x / 2^B) and x_0 = x mod 2^B.

Then x_0 < 2^B <= d.

Let
    q1 = floor((x_1 * 2^B) / d)
    q_hat = floor(r * x_1 / 2^(S - B))
          = floor((r * x_1 * 2^B) / 2^S).

By the correctness proof of the reciprocal above, applied to x_1 * 2^B,
we know
    q1 - 1 <= q_hat <= q1.

Since x and x_1 * 2^B differ by x_0, and x_0 < d, we also know
    q1 <= q <= q1 + 1.

Therefore
    q - 2 <= q_hat <= q.
```

m-alvarez: The clarifications are great. I much prefer the (updated) symbolic version, thanks.

andreaslyn (approved):

PR #2272, opened 2026-05-15
Add MONAD_ASSUME and use it to deduplicate RLP encoder preconditions

goodlyrottenapple:
The RLP encoders in encode.hpp had a repeated `#ifdef NDEBUG / __builtin_unreachable() / abort()` pattern guarding output-buffer preconditions. Factor it into a single MONAD_ASSUME(cond) macro in category/core/assume.h: aborts in debug, emits no code in release while still propagating the assumption to the optimizer so downstream bounds checks can be eliminated.

### DIFF
--- a/category/core/assume.h
+++ b/category/core/assume.h
@@ -1,0 +1,33 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <category/core/assert.h>
+
+/// Assert a precondition in debug builds and propagate it to the optimizer
+/// in release builds. In NDEBUG mode this enables the compiler to assume
+/// `cond` holds and may allow it to eliminate downstream checks, though
+/// evaluating `cond` may still emit instructions. Violating `cond` in
+/// release is undefined behavior; in debug, the failure is reported via
+/// MONAD_ASSERT with the condition string, source location, and backtrace.
+#ifdef NDEBUG
+    #define MONAD_ASSUME(cond)                                                 \
+        if (!(cond)) {                                                         \
+            __builtin_unreachable();                                           \
+        }
+#else
+    #define MONAD_ASSUME(cond) MONAD_ASSERT(cond)
+#endif
--- a/category/core/rlp/encode.hpp
+++ b/category/core/rlp/encode.hpp
@@ -15,6 +15,7 @@
 
 #pragma once
 
+#include <category/core/assume.h>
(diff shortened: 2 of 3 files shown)


category/core/rlp/encode.hpp
```diff
@@ -81,34 +76,23 @@ constexpr size_t string_length(byte_string_view const s)
 constexpr std::span<unsigned char>
 encode_string(std::span<unsigned char> d, byte_string_view const s)
 {
+    MONAD_ASSUME(d.size() > 0);
```
mkolosick, line 79: Was this just UB before?

mkolosick (dismissed):

category/core/rlp/encode.hpp
goodlyrottenapple: My thinking is that you need both conditions to be respected by the caller, as the single `MONAD_ASSUME(d.size() >= s.size());` may become true due to an underflow when e.g. taking a sub-span of an empty span as @maxkozlovsky pointed out.

category/core/assume.h
```diff
@@ -0,0 +1,36 @@
...
+/// Assert a precondition in debug builds and propagate it to the optimizer
+/// in release builds. In NDEBUG mode this enables the compiler to assume
+/// `cond` holds and may allow it to eliminate downstream checks, though
+/// evaluating `cond` may still emit instructions. Violating `cond` in
+/// release is undefined behavior; in debug, the failure is reported via
+/// MONAD_ASSERT with the condition string, source location, and backtrace.
+#ifdef NDEBUG
+    #define MONAD_ASSUME(cond)                                                 \
+        do {                                                                   \
+            if (!(cond)) {                                                     \
```
dhil: Nit:
```suggestion
            if (MONAD_UNLIKELY(!(cond))) {                                                     \
```

category/core/assume.h
```diff
@@ -0,0 +1,36 @@
...
+                __builtin_unreachable();                                       \
+            }                                                                  \
+        }                                                                      \
+        while (0)
+#else
+    #define MONAD_ASSUME(cond) MONAD_ASSERT(cond)
```
dhil, line 32: `MONAD_ASSERT` is defined as an `if-then-else`, whereas `MONAD_ASSUME` is defined as a `do .. while` in `NDEBUG`. The difference is small, but subtle, because in the debug case `MONAD_ASSUME(cond);` is two statements (`MONAD_ASSERT` expansion + the empty statement), whereas in `NDEBUG` mode the same macro expression is one statement (here `;` terminates the `do .. while` statement). For consistency, it might be better to implement `MONAD_ASSUME` as
```c
#define MONAD_ASSUME(cond) \
  if (MONAD_UNLIKELY(!(cond)) { \
    __builtin_unreachable();    \
  }

Comment (goodlyrottenapple) [reply category/core/assume.h:32]: copilot shouted at me earlier when i defined MONAD_ASSUME as an `if ...`. Don't have good intuition whether it was just hand-wringing? Happy to change back...

File: category/core/assume.h (hunk 2 above)
Comment (goodlyrottenapple) [reply category/core/assume.h]: This is redundant afaik. are you thinking more for documentation?

File: category/core/rlp/encode.hpp (hunk 1 above)
Comment (mkolosick) [reply category/core/rlp/encode.hpp:79]: Agreed the check seems right just surprised nothing was there for this before.

File: category/core/assume.h (hunk 3 above)
Comment (dhil) [reply category/core/assume.h:32]: One consequence is that in debug mode you can compile and run the program with
```
MONAD_ASSUME(true)
```
whereas it fails to parse in `NDEBUG` mode.

Baltoli (approved):

mkolosick (approved):

PR #2273, opened 2026-05-15
[execution/ethereum] Generalize EIP-7685 request hashing

cmcl:
Rename the request-processing trait gate from eip_7002_active to eip_7685_active so block-level request hash handling is tied to the EIP that defines requests_hash rather than to a specific request type.

Expose BlockRequest and make compute_requests_hash operate on a typed request list.

Add focused unit tests for EIP-7685 request hash semantics.

### DIFF
--- a/category/execution/ethereum/execute_block.cpp
+++ b/category/execution/ethereum/execute_block.cpp
@@ -342,7 +342,7 @@ Result<std::vector<Receipt>> execute_block(
         process_withdrawal(state, block.withdrawals);
     }
 
-    if constexpr (traits::eip_7002_active()) {
+    if constexpr (traits::eip_7685_active()) {
         BOOST_OUTCOME_TRY(
             auto const computed_requests_hash,
             process_requests<traits>(
--- a/category/execution/ethereum/process_requests.cpp
+++ b/category/execution/ethereum/process_requests.cpp
@@ -32,9 +32,17 @@
 #include <boost/outcome/try.hpp>
 #include <intx/intx.hpp>
 
+#include <array>
+#include <cstdint>
+#include <utility>
+#include <vector>
+
 using BOOST_OUTCOME_V2_NAMESPACE::success;
 
 MONAD_ANONYMOUS_NAMESPACE_BEGIN
+
+constexpr uint8_t WITHDRAWAL_REQUEST_TYPE = 0x01;
+constexpr uint8_t CONSOLIDATION_REQUEST_TYPE = 0x02;
 
 template <Traits traits>
 Result<byte_string> system_call(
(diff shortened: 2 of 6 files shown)


category/execution/ethereum/execute_block.cpp
```diff
@@ -342,7 +342,7 @@ Result<std::vector<Receipt>> execute_block(
         process_withdrawal(state, block.withdrawals);
     }
 
-    if constexpr (traits::eip_7002_active()) {
+    if constexpr (traits::eip_7685_active()) {
```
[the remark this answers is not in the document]
cmcl: Not correct, I believe: requests processing is Ethereum-specific.

goodlyrottenapple (dismissed):

category/execution/ethereum/process_requests.cpp
```diff
@@ -125,75 +132,79 @@ Result<byte_string> system_call(
...
-
-MONAD_NAMESPACE_BEGIN
-
 template <Traits traits>
 Result<bytes32_t> process_requests(
     Chain const &chain, State &state, BlockHashBuffer const &block_hash_buffer,
     BlockHeader const &header, ChainContext<traits> const &chain_ctx)
 {
+    std::vector<BlockRequest> requests;
+    requests.reserve(2);
```
dhil: Would the following work?
```suggestion
    std::array<BlockRequest, 2> requests{};
```

category/execution/ethereum/process_requests.hpp
```diff
@@ -21,12 +21,25 @@
...
 
+struct BlockRequest
+{
+    uint8_t type;
+    byte_string data;
+};
+
+// EIP-7685: Canonical callers must supply requests ordered by ascending request
+// type.  The hash function preserves the supplied order.
+bytes32_t compute_requests_hash(std::span<BlockRequest const> requests);
```
dhil, line 41: Is it always the case that this function is called with `requests.size() == 2` in production (I see that one of the tests exercise different sizes)? If so, then maybe it worth constraining the signature further with static extent, e.g.

```suggestion
bytes32_t compute_requests_hash(std::span<BlockRequest const, 2> requests);
```

cmcl: It will be 3 once [deposit requests](https://github.com/category-labs/monad/pull/2277) are supported.  Including the length in the type might be rather brittle if more request ~objects~ types are added by subsequent EIPs .

category/execution/ethereum/process_requests.cpp
cmcl: It does, provided the `push_back` calls are changed accordingly.  I'm inclined to keep the original vector-based solution though since it matches other instances within the PR, and the heap allocation is small.

Baltoli:
```cpp
return compute_requests_hash(std::array<BlockRequest>{
    {WITHDRAWAL_REQUEST_TYPE, std::move(withdrawal_output)}
    {CONSOLIDATION_REQUEST_TYPE, std::move(consolidation_output)}
});
```

category/execution/ethereum/process_requests.cpp
```diff
@@ -125,43 +133,39 @@ Result<byte_string> system_call(
...
-    auto const hash_request = [&](uint8_t type, byte_string const &data) {
-        // EIP-7685 flat requests encoding: empty request types are excluded
-        // from the hash.
-        if (data.empty()) {
-            return;
+    std::vector<uint8_t> inner_hashes;
+    inner_hashes.reserve(32 * requests.size());
+
+    for (auto const &req : requests) {
+        // EIP-7685 flat requests encoding: empty request data are excluded from
```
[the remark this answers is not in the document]
cmcl: Builds fine.

category/execution/ethereum/process_requests.cpp
```diff
@@ -125,43 +133,39 @@ Result<byte_string> system_call(
...
 
-    // TODO: EIP-6110 deposits
-    hash_request(0x00, {});
-    hash_request(0x01, withdrawal_output);
-    hash_request(0x02, consolidation_output);
+        bytes32_t inner;
+        byte_string buf;
+        buf.reserve(1 + req.data.size());
+        buf.push_back(req.type);
+        buf.insert(buf.end(), req.data.begin(), req.data.end());
```
dhil: I think the following may work:
```suggestion
        buf.append_range(req.data);
```
I think it is conceptually simpler.

category/execution/ethereum/process_requests.cpp
```diff
@@ -125,43 +133,39 @@ Result<byte_string> system_call(
...
+        silkpre_sha256(inner.bytes, buf.data(), buf.size(), true);
+        inner_hashes.insert(inner_hashes.end(), inner.bytes, inner.bytes + 32);
```
dhil: Following the above, I think we can simplify to:
```suggestion
        inner_hashes.append(inner.bytes, inner.bytes + 32);
```

dhil (dismissed):
Looks good to me. I've attached some very minor nits.

category/execution/ethereum/process_requests.cpp
cmcl: I can use `append_range` here too which additionally avoids the arithmetic

dhil (approved):

goodlyrottenapple (approved):

PR #2278, opened 2026-05-19
[execution]: Memory Bound LRU Cache

kkuehlz:
[execution]: Memory Bound LRU Cache
Introduces a memory bound LRU cache for variable length storage slots.
Under page storage, the cache will hold encoded pages, which have different
sizes.

LruWeightCache does exactly what we need.  I experimented with several
slab allocator of fixed sized pool approaches. Rolling one by hand has
marginal improvements: around 10ns on insert time. Perhaps real world
data could inform the slab class sizes. For now, this simpler solution
is preferred.


kkuehlz:
The `storage_page_t` holds a compressed representation so this is no longer necessary.

PR #2286, opened 2026-05-21
Improve block stats

magedm:
Detailed gas reporting:
- 'gase': gas used - growth gas
- 'gasu': gas used
- 'gasc': gas charged
- 'gasl': gas limit

Before this PR only gas charged was reported as 'gas' and used in gps computation. After this PR 'gase' is used in gps computation for more conservative results.

For real time:
- Report finalization time per block.
- Changed execution total from 'tot' to 'etot' (as it does not include finalization, reported separately).

For replay:
- Report finalization latency as 'fin'.
- Removed the useless timestamp 'ts' field (useful for realtime, but useless for replay).

For both real time and replay:
- Added 'blr' field for block read latency.

[UPDATE 2026-06-02]
Keeping defunct 'gas' (in replay and real time) and 'tot' (in real time) temporarily for compatibility with current scripts. Will remove in a separate PR later after scripts are updated.

### DIFF
--- a/category/execution/runloop/runloop_ethereum.cpp
+++ b/category/execution/runloop/runloop_ethereum.cpp
@@ -93,12 +93,11 @@ Result<void> process_ethereum_block(
     fiber::PriorityPool &priority_pool, Block &block,
     BlockHeader const &parent_header, bytes32_t const &block_id,
     bytes32_t const &parent_block_id, bool const enable_tracing,
-    ExecutionEventRecorder *const exec_recorder)
+    ExecutionEventRecorder *const exec_recorder,
+    std::chrono::steady_clock::time_point const block_begin,
+    std::chrono::microseconds const blr_time)
 {
     static_assert(traits::evm_rev() >= MONAD_ETH_CONSTANTINOPLE);
-
-    [[maybe_unused]] auto const block_start = std::chrono::system_clock::now();
-    auto const block_begin = std::chrono::steady_clock::now();
 
     record_block_start(
         exec_recorder,
@@ -222,8 +221,12 @@ Result<void> process_ethereum_block(
 
     // Commit prologue: database finalization, computation of the Ethereum
     // block hash to append to the circular hash buffer
+    auto const fin_begin = std::chrono::steady_clock::now();
     db.finalize(block.header.number, block_id);
     db.update_verified_block(block.header.number);
+    [[maybe_unused]] auto const fin_time =
+        std::chrono::duration_cast<std::chrono::microseconds>(
+            std::chrono::steady_clock::now() - fin_begin);
     exec_output.eth_block_hash =
         to_bytes(keccak256(rlp::encode_block_header(exec_output.eth_header)));
     block_hash_buffer.set(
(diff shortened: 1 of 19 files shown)


magedm:
@claude review once

Fix for the prior P2: runtime/call.cpp now captures the predicate in bool grew, performs the 25,000 raw deduction inside the existing if, then bumps add_growth_gas(25000) only after the combined-gas OOG check at line 141 passes.

Re P3: growth_gas is just stats. The debug-assert `state.growth_gas() <= tx_gas_used` in execute_transaction.cpp catches mis-accounting in debug-build CI; in release the worst case is an incorrect stats number, and `Context::add_growth_gas` is null-safe. Relying on existing CI rather than adding stats-specific tests.

category/vm/runtime/call.cpp
```diff
@@ -131,6 +131,7 @@ namespace monad::vm::runtime
             if (has_value &&
                 !ctx->host->account_exists(ctx->context, &dest_address)) {
                 ctx->gas_remaining -= 25000;
+                ctx->add_growth_gas(25000);
             }
```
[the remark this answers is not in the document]
magedm: Good comment. Fixed.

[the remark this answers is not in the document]
magedm: Fixed.

category/execution/ethereum/state3/state.hpp
```diff
@@ -240,6 +241,10 @@ class State
      */
     bool record_balance_constraint_for_debit(
         Address const &, uint256_t const &debit);
+
+    void add_growth_gas(int64_t g) noexcept;
```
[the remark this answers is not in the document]
magedm: growth_gas is just stats. The debug-assert `state.growth_gas() <= tx_gas_used` in execute_transaction.cpp catches mis-accounting in debug-build CI; in release the worst case is an incorrect stats number, and `Context::add_growth_gas` is null-safe. Relying on existing CI rather than adding stats-specific tests.

category/execution/ethereum/state3/state.cpp
```diff
@@ -717,4 +717,20 @@ bool State::record_balance_constraint_for_debit(
     return false;
 }
 
+void State::add_growth_gas(int64_t const g) noexcept
+{
+    MONAD_DEBUG_ASSERT(g >= 0);
+    growth_gas_ += static_cast<uint64_t>(g);
+}
```
Baltoli: Why not have this just take a `uint64_t` directly? afaict the only calls are:
- literals
- passed from trait functions we control

Doing so will avoid the assert + cast. Also maybe worth checking overflow?

magedm: Thanks! Updated. Originally I wanted to match the existing signed traits gas cost functions. But those can be subtractive, whereas growth gas is purely additive.

magedm:
Moved growth_gas into a struct Stats under State.

magedm:
Updated PR to handle MIP-8 changes.

category/vm/runtime/types.hpp
```diff
@@ -392,6 +395,17 @@ namespace monad::vm::runtime
         template <Traits traits>
         evmc::Result copy_to_evmc_result();
 
+        [[gnu::always_inline]]
+        constexpr void add_growth_gas(uint64_t const g) noexcept
+        {
+            if (growth_gas_ptr != nullptr) {
```
andreaslyn, line 403: Is it possible to have `growth_pas_ptr != nullptr` an invariant of Context?

category/vm/vm.cpp
```diff
@@ -65,6 +65,7 @@ namespace monad::vm
         auto const &icode = vcode->intercode();
         auto rt_ctx =
             runtime::Context::from(host_itf, host_ctx, msg, icode->code_span());
+        rt_ctx.growth_gas_ptr = host.growth_gas_ptr();
```
andreaslyn: Can this be initialized in `Context::from` using the given host interface? and `Context::empty` will probably want the `growth_gas_ptr` as an argument.

category/execution/runloop/runloop_ethereum.cpp
```diff
@@ -227,32 +230,40 @@ Result<void> process_ethereum_block(
...
         block_time,
         block.transactions.size() * 1'000'000 /
             (uint64_t)std::max(1L, block_metrics.tx_exec_time.count()),
         block.transactions.size() * 1'000'000 /
             (uint64_t)std::max(1L, block_time.count()),
+        block_metrics.gas_exec,
+        block_metrics.gas_used,
         exec_output.eth_header.gas_used,
-        exec_output.eth_header.gas_used /
+        exec_output.eth_header.gas_used,
```
goodlyrottenapple, line 265: is this outputting the same `exec_output.eth_header.gas_used` twice, and if so, is this intentional?

goodlyrottenapple: oh, nm just re-read the PR message

category/vm/runtime/types.hpp
magedm: - growth_gas_ptr != nullptr is intentionally not an invariant. Enforcing it would require every test that constructs Context directly to own a sink, so we keep the null-check to let those tests stay unchanged.
- The vm::Host mocks (HostMock in monad_vm_interface_tests.cpp and MockedHost in test/vm/unit/runtime/mocked_host.hpp) return nullptr, and several test sites construct Context::from(...) directly with evmc::MockedHost (which isn't a vm::Host).
- Production is always non-null, as EvmcHostBase::growth_gas_ptr() returns &state_.stats().growth_gas.

category/vm/vm.cpp
magedm: - Doing this inside Context::from would need either (a) recovering a vm::Host* from the evmc context or (b) changing Context::from's signature to take a vm::Host& instead of the C interface.
- Several tests construct Context::from with evmc::MockedHost (from the evmc library, not vm::Host-derived), so the recovery in option (a) would be UB and the vm::Host&-signature in option (b) couldn't be called at all.
- The current pattern keeps Context::from at the C interface layer and lets VM::execute do the vm::Host-level wiring separately.
- Context::empty's only caller (TestContext) doesn't observe growth_gas, so the struct default is sufficient there.

category/vm/runtime/types.hpp
andreaslyn: I think it is not particularly difficult to change all the tests that construct context. I have done this before some times.

category/vm/vm.cpp
andreaslyn: Passing the pointer as argument to both `Context::from` and `Context::empty` is fine.

magedm:
Decoupled growth-gas accounting from State and Host. Introduced a top-level TransactionStats struct as the per-tx accounting object. Threaded it as its own parameter through the tx-execution chain.

Since evmc::Host::call has a fixed signature, growth accounting can't be threaded through it as a parameter across nested-opcode dispatch. Instead, the code routes the pointer through runtime::Context and derives it inside the EvmcHost::call override.

@copilot review

category/vm/runtime/types.hpp
magedm: The downside is not just changing tests but also having to add an unused local variable to be the target of growth gas pointer. It adds more clutter to the tests.

category/vm/vm.cpp
magedm: Done.

category/vm/runtime/types.hpp
andreaslyn: I think that it would be better to return vm growth gas as part of the evmc result instead of passing around the pointer. Some benefits with this design are:

* context can always initialize growth gas to 0
* avoids all the null checks
* less memory traffic when increasing growth gas

cmcl:
FYI: My push just fixes the copyright notice in the new file (see #2429).

PR #2290, opened 2026-05-22
Replace bindgen with cxx in monad-triedb

andr-dev:
This change swaps out `bindgen` for `cxx`, a framework for safely calling C++ code from Rust and vice-versa. This allows for the following improvements:
1. No need to build+link special `triedb-driver` library, C++ build is handled by `cxx`
2. No more `nullptr` checks, native support for C++ references
3. Significantly better type sharing
  a. Byte slice pointer+len arg pattern replaced by `::rust::Slice<uint8_t const>`, reduces borrow safety bug surface area since `cxx` enforces lifetimes on Rust side
  b. `std::unique_ptr<T>` becomes `cxx::UniquePtr<T>`, no more `*_free` fns + manual `Drop` impls
  c. `std::vector<T>` enables no-memcpy valset return
4. No more hand-written signature match assertions
5. Less hand-written `unsafe` (hidden behind `cxx` machinery)

On top of the above, this change enables the following performance improvements:
1. All callback functions can now be resolved at link time removing dynamic dispatch
2. `NodeCursor` can be passed across the FFI boundary, enabling zero-copy reads (future PR)


rust/crates/monad-triedb/include/ffi.h
```diff
@@ -15,112 +15,106 @@
...
+        ::rust::Str dbdirpath, uint64_t node_lru_max_mem,
+        bool disable_mismatching_storage_pool_check);
+
+    inline size_t
+    triedb_poll(TriedbRoInner &inner, bool const blocking, size_t const count)
+    {
+        return inner.db.poll(blocking, count);
+    }
+
+#define MONAD_RUST_METADATA_GETTER(name, ret, call)                            \
```
andr-dev: This will be used in the next PR when we introduce `TriedbRwHandle`.

andr-dev:
@claude review.

rust/crates/monad-triedb/src/read.rs
```diff
@@ -0,0 +1,247 @@
...
+
+use std::sync::{atomic::AtomicUsize, Arc};
+
+use cxx::UniquePtr;
+
+use crate::{ffi, NibblesView, TriedbRoHandle};
+
+/// A value read from the triedb, holding a cursor that pins the underlying node in the cache. The
+/// byte view is valid as long as the `ffi::NodeCursor` is alive.
+pub struct TriedbNodeCursor {
```
Chen-Yifan: mpt::NodeCursor is already owning, not need to wrap it with unique ptr. The lifetime around this is awkward.

rust/crates/monad-triedb/src/ffi.cpp
```diff
@@ -13,207 +13,163 @@
...
+    TriedbRoInner::TriedbRoInner(
         std::vector<std::filesystem::path> dbname_paths,
-        uint64_t const node_lru_max_mem)
-        : io_ctx{monad::mpt::ReadOnlyOnDiskDbConfig{
-              .disable_mismatching_storage_pool_check = true,
-              .dbname_paths = std::move(dbname_paths)}}
-        , db{io_ctx}
-        , async_ctx{db, node_lru_max_mem}
+        uint64_t const node_lru_max_mem,
+        bool const disable_mismatching_storage_pool_check)
```
Chen-Yifan: no need for the new flags.

rust/crates/monad-triedb/src/read.rs
andr-dev: Returning `mpt::NodeCursor` enables zerocopy reads of triedb which is useful since consensus RLP decodes every value read, as far as I can tell. With this, we no longer need to materialize a copied byte string. The simplest implementation of this is returning a pointer to the `mpt::NodeCursor`.

I think we could also get rid of this indirection by creating a rust version of NodeCursor so that the rust compiler understands size+align and then in C++ define a move constructor from C++ version to rust. This however is a bit more messy and probably deserves a followup.

Chen-Yifan: you are already introducing a TriedbNodeCursor def on rust side, I would recommend doing it the correct way in the very first. However, I think all the code reorganization should be moved to the next PR to make review easier.

Chen-Yifan (commented):
Overall, I'm also not convinced that the API changes, code reorganization + zerocopy thing should be included in this PR. The bindgen -> cxx deserves its own PR, and let's move the impl to the next PR. That will make the purpose more clear, and easier to review.

andr-dev:
Agreed, removed the node cursor stuff and kept the simple copied bytes approach. I'll add the node cursor correctly in the next change.

rust/crates/monad-triedb/build.rs
```diff
@@ -13,20 +13,38 @@
...
-            .build("triedb_driver");
-    }
+    monad_build::MonadCxx::new("src/ffi.rs")
+        .file("src/ffi.cpp")
+        .defines([
+            ("MONAD_CXX_CTYPES_USE_EVMC_HPP", "1"),
+            ("QUILL_ROOT_LOGGER_ONLY", "1"),
+        ])
+        .includes([
+            "third_party/ankerl",
```
Chen-Yifan, line 30: since it no longer uses cmake that uses the whole execution lib, instead it scopes down to only compile triedb related stuff, the following include paths are no longer needed:

     "third_party/ethash/include",
      "third_party/evmc/include",
      "third_party/fiber/include",
      "third_party/quill/quill/include",
      "third_party/unordered_dense/include"

rust/crates/monad-build/src/lib.rs
Chen-Yifan, line 62: Now this part is confusing, this PR narrowed down the triedb build dependency from the whole execution tree to only necessary ones for triedb, and renamed the lib from `triedb_driver` to`monad_triedb_ffi`. However, the old name `triedb_driver` is still used in Dockerfile, but now no longer refers to anything. 

The `ENV TRIEDB_TARGET` in Dockerfile has been not accurate for a while, what it really means is to build the whole execution lib instead of just triedb (because of event ring, statesync and more other stuff are needed by the rust side)
Since you're already editing this env var, can we rename it to something like `MONAD_BUILD_EXECUTION=1`?

rust/crates/monad-triedb/include/ffi.h
```diff
@@ -15,112 +15,158 @@
...
+            std::vector<std::filesystem::path> dbname_paths,
+            uint64_t node_lru_max_mem);
+    };
+
+    std::unique_ptr<TriedbRoInner>
+    triedb_open(::rust::Str dbdirpath, uint64_t node_lru_max_mem);
+
+    // pumps async reads, processing no more than count maximum, returning how
+    // many were processed.
+    inline size_t
```
Chen-Yifan, line 88: why changing these to inline?

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -13,25 +13,232 @@
...
+        sender,
+
+        completed_counter,
+        concurrency_tracker,
+    } = unsafe { *Box::from_raw(ctx as *mut AsyncReadContext) };
+
+    let _ = sender.send(found.then(|| value.into()));
+
+    completed_counter.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+    drop(concurrency_tracker);
```
kkuehlz: why?

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -13,25 +13,242 @@
...
+    }
+
+    // Generates VectorElement for Validator so CxxVector<Validator> is usable.
+    impl CxxVector<Validator> {}
+
+    #[namespace = "monad::rust::ffi"]
+    extern "Rust" {
+        type CallbackContext;
+
+        unsafe fn callback_async_read(ctx: *mut CallbackContext, value: &[u8], found: bool);
```
kkuehlz, line 147: comment the ownership on the API

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -13,25 +13,242 @@
...
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub struct Bytes32(pub [u8; 32]);
+
+unsafe impl cxx::ExternType for Bytes32 {
+    type Id = cxx::type_id!("monad::bytes32_t");
+    type Kind = cxx::kind::Trivial;
+}
+
+#[repr(C)]
+pub struct OpaqueCallbackContext(());
```
kkuehlz, line 40: Switch to this

```
  extern "Rust" {
      type AsyncReadContext;
      type AsyncTraverseContext;

      fn callback_async_read(ctx: Box<AsyncReadContext>, value: &[u8], found: bool);
      fn callback_traverse_value(ctx: &AsyncTraverseContext, key: &[u8], value: &[u8]);
      fn callback_traverse_finished(ctx: Box<AsyncTraverseContext>, completed: bool);
  }

```

rust/crates/monad-triedb/src/read.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    #[inline]
+    fn latest_voted_block_id(&self) -> Option<[u8; 32]> {
+        parse_block_id(crate::ffi::triedb_latest_voted_block_id(&self.inner))
+    }
+
+    #[inline]
+    fn traverse(&mut self, key: NibblesView<'_>, block_id: u64, sender: ffi::AsyncTraverseSender) {
+        let ctx = Box::into_raw(Box::new(ffi::AsyncTraverseContext::new(
```
kkuehlz, line 132: we won't need `into_raw()` or `from_raw()` once we get rid of the opaque type.

rust/crates/monad-triedb/include/ffi.h
```diff
@@ -15,112 +15,158 @@
...
-// value. Call triedb_finalize when
-// done with the value.
-void triedb_async_read(
-    TriedbRoInner *, uint8_t const *key, uint8_t key_len_nibbles,
-    uint64_t block_id, triedb_async_read_callback_fn callback, void *user);
-
-// traverse the trie.
-enum triedb_async_traverse_callback
+    inline ::rust::Slice<uint8_t const>
+    node_cursor_as_bytes(NodeCursor const &cursor)
```
kkuehlz: this seems like the wrong namespace. should this be ni `monad::rust`?

rust/crates/monad-triedb/src/ffi.rs
andr-dev: Decided offline to do this in a separate PR.

0xSqualo (approved):

PR #2291, opened 2026-05-24
feat: unified clap-style --help output for C++ binaries

Arjun-Somvanshi:
## Summary

  `monad-ctl` (in [monad-bft](https://github.com/category-labs/monad-bft)) dispatches `--help` across a mix of Rust binaries (`monad-keystore`,
  `monad-ledger-tail`, `monad-sign-name-record`) and the three C++ binaries shipped from this repo (`monad`, `monad-cli`, `monad-mpt`). The two languages use different argparsers — `clap` (Rust) vs `CLI11` (C++) — and their default `--help` layouts diverge in section headers, flag separators, value placeholders, and defaults rendering.

  This PR adds a shared `monad::cli::HelpFormatter` (subclass of `CLI::Formatter`) and installs it in all three C++ binaries so they emit `clap`-style help that matches the Rust binaries' output.

  ## What changed

  - **New** `category/core/cli/help_formatter.{hpp,cpp}` — clap-style overrides for `make_help`, `make_description`, `make_usage`, `make_subcommands`, `make_option*`.
  - **New** `category/core/cli/help_formatter_test.cpp` — 13 gtest cases covering header, usage line, separator (`-h, --help` with space), placeholder rendering, `[default: …]` / `[required]` suffixes, boolean-flag default suppression, snake_case alias collapsing, long-only indent, no `--version` flag, and option-group flattening.
  - Install the formatter in `cmd/monad/main.cpp`, `cmd/monad_cli.cpp`, `category/mpt/cli_tool_impl.cpp`. Each call site is a one-liner: `monad::cli::HelpFormatter{GIT_COMMIT_HASH}.install(cli);`.
  - Hoist `GIT_COMMIT_HASH` computation from `cmd/CMakeLists.txt` to root `CMakeLists.txt` so `monad-mpt` and `cli_tool_test` can use the same value without duplicating the `git rev-parse` block. `target_compile_definitions` for the new consumers added in `category/mpt/CMakeLists.txt` and `category/mpt/test/CMakeLists.txt`.

  ## Layout details

  Matches clap's "long" help style:
  - `<bin> <git_commit>` header line; binary name in bold.
  - `Usage:`, `Options:`, `Commands:` — bold + underlined.
  - Option names in bold; first canonical long name only (snake_case aliases like `--block-db,--block_db` collapse to `--block-db`).
  - `<UPPER_SNAKE>` value placeholders; descriptions on the next line indented to column 10.
  - `[default: X]` and `[required]` suffixes. Boolean flag defaults (`false`) suppressed.
  - ANSI escapes emitted only when stdout is a TTY and `NO_COLOR` is unset/empty.
  - No `--version` printer (this PR does not add one; left as a follow-up — the existing `--version` in `monad-cli` is an unrelated business option).

### DIFF
--- a/category/core/cli/help_formatter.cpp
+++ b/category/core/cli/help_formatter.cpp
@@ -1,0 +1,320 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/core/cli/help_formatter.hpp>
+
+#include <category/core/config.hpp>
+
+#include <algorithm>
+#include <cctype>
+#include <cstddef>
+#include <cstdlib>
+#include <format>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include <unistd.h>
+
+MONAD_NAMESPACE_BEGIN
+
+namespace cli
+{
+    namespace
+    {
+        // ANSI escape sequences kept short; \x1b[0m fully resets attributes.
+        constexpr std::string_view ANSI_BOLD = "\x1b[1m";
+        constexpr std::string_view ANSI_BOLD_UNDERLINE = "\x1b[1;4m";
+        constexpr std::string_view ANSI_RESET = "\x1b[0m";
+
+        bool detect_color()
+        {
+            char const *const no_color = std::getenv("NO_COLOR");
+            if (no_color != nullptr && no_color[0] != '\0') {
+                return false;
+            }
+            return ::isatty(STDOUT_FILENO) != 0;
+        }
+
+        std::string to_placeholder(std::string s)
+        {
+            for (char &c : s) {
+                if (c == '-') {
+                    c = '_';
+                }
+                else {
+                    c = static_cast<char>(
+                        std::toupper(static_cast<unsigned char>(c)));
+                }
+            }
+            return s;
+        }
+
+        bool is_option_group(CLI::App const *const sub)
+        {
+            return dynamic_cast<CLI::Option_group const *>(sub) != nullptr;
+        }
+
+        void collect_options(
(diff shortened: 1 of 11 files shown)


category/core/cli/help_formatter.cpp
```diff
@@ -0,0 +1,354 @@
...
+    {
+        if (name.empty()) {
+            name = app->get_name();
+        }
+
+        std::string out = "\n";
+        out += wrap(ANSI_BOLD_UNDERLINE, use_color_, "Usage:");
+        out += ' ';
+        out += wrap(ANSI_BOLD, use_color_, name);
+        out += " [OPTIONS]";
```
Baltoli: Rather than using `.append` or `+=` to build interpolated strings, we should use `std::format` instead.

category/core/cli/help_formatter.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+        auto const real_subs = app->get_subcommands(
+            [](CLI::App const *const s) { return !is_option_group(s); });
+        if (!real_subs.empty()) {
+            out += " [COMMAND]";
```
Baltoli, line 145: Note that `+=` is fine here because there's no interpolation; the `std::format` version would be something like `std::format_to(std::back_inserter(out), " [COMMAND]")` which isn't as clear and boils down to the same thing in the end

Baltoli:
@claude review

category/core/cli/help_formatter.cpp
Arjun-Somvanshi: fixed in [a474cfb](https://github.com/category-labs/monad/commit/a474cfb4b9b997ebdd8e66f4d3a761c45ffd24e4)

category/core/cli/help_formatter.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+        std::string wrap(
+            std::string_view const escape, bool const enabled,
+            std::string_view const inner)
+        {
+            if (!enabled) {
+                return std::string{inner};
+            }
+            return std::format("{}{}{}", escape, inner, ANSI_RESET);
+        }
```
Baltoli: This could just be folded into a private member I think, to save the repetition of `use_color_` at call sites?:
```cpp
  std::string styled(std::string_view escape, std::string_view inner) const
  {
      return use_color_ ? std::format("{}{}{}", escape, inner, ANSI_RESET)
                        : std::string{inner};
  }
```

category/core/cli/help_formatter.cpp
```diff
@@ -0,0 +1,325 @@
...
+        CLI::App const *const app, std::string name,
+        CLI::AppFormatMode const mode) const
+    {
+        if (name.empty()) {
+            name = app->get_name();
+        }
+
+        std::string out;
+        out += make_description(app);
+        out += make_usage(app, name);
```
Baltoli: Could be a `std::format`

dhil (dismissed):

dhil (approved):

goodlyrottenapple (approved):

PR #2294, opened 2026-05-26
[cmake] build libhugetlbfs, don't take it from the system

kjcamann:
libhugetlbfs does not meet our requirements for what dependencies can come from the system, namely it is not common enough: some popular Linux distributions do not contain it in their package manager.

In this commit we consume it as a third_party submodule, and kick off the local autotools-based build using ExternalProject_Add.

### DIFF
--- a/third_party/libhugetlbfs/CMakeLists.txt
+++ b/third_party/libhugetlbfs/CMakeLists.txt
@@ -1,0 +1,57 @@
+# Copyright (C) 2025-26 Category Labs, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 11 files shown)


third_party/libhugetlbfs/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# This cmake project can be used to build just the execution event SDK library
```
cmcl, line 16: Stale comment regarding event SDK

kjcamann:
Closed in favor of #2299

PR #2295, opened 2026-05-26
Move library linking to MonadCMake

andr-dev:


rust/crates/monad-event-ring/build.rs
```diff
@@ -18,14 +18,16 @@ fn main() {
         monad_build::repository_root().join("category/event"),
         monad_build::MonadCMakeLinkage::Static,
     )
+    .link_libraries([
+        "zstd",
+        if build_rs::input::cargo_cfg_target_os() == "linux" {
```
andr-dev, line 23: Switched to `build-rs` since `target_os` evaluates to the `build.rs` host target, not the cargo build target.

https://doc.rust-lang.org/cargo/reference/build-script-examples.html#reading-target-configuration

kjcamann (approved):

Baltoli (approved):

PR #2299, opened 2026-05-27
[event] remove libhugetlbfs

kjcamann:
We only needed a few functions from hugeutils.c and maintaining the build system around this was becoming a pain. Here we just do the same thing libhugetlbfs does, but do it ourselves.

As a ninja change, various functions (some unrelated to this exact change) returning ENAMETOOLONG when a user-provided path buffer was too small now return ERANGE. I realized ERANGE was the better option (matching getcwd(2)) for a small buffer vs. ENAMETOOLONG which is for PATH_MAX overflow. A function added in the commit uses ERANGE that way, and to keep the other C APIs in sync they were all changed.

### DIFF
--- a/category/core/mem/hugetlb_path.c
+++ b/category/core/mem/hugetlb_path.c
@@ -19,14 +19,20 @@
 #include <category/core/path_util.h>
 #include <category/core/srcloc.h>
 
-#include <hugetlbfs.h>
-
+#include <ctype.h>
 #include <errno.h>
 #include <stddef.h>
+#include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 
 #include <fcntl.h>
 #include <linux/limits.h>
+#include <linux/magic.h>
+#include <mntent.h>
+#include <sys/stat.h>
+#include <sys/vfs.h>
+#include <unistd.h>
 
 thread_local char g_error_buf[PATH_MAX];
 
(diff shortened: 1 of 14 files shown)


category/core/mem/hugetlb_path.c
```diff
@@ -60,33 +204,25 @@ int monad_hugetlbfs_open_dir_fd(
         pathbuf_size = sizeof local_pathbuf;
     }
     if (params->page_size == 0) {
-        long default_size = gethugepagesize();
-        if (default_size == -1) {
-            return FORMAT_ERRC(errno, "no default huge page size configured");
+        rc = monad_get_default_hugepage_size(&resolve_size) != 0;
```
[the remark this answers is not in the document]
kjcamann: Fixed

jhunsaker:
@claude review

jhunsaker:
@claude review again

category/event/CMakeLists.txt
```diff
@@ -65,6 +55,7 @@ target_sources(monad_event PUBLIC
   "../core/event/event_ring_util.h"
   "../core/event/test_event_ctypes.h"
   "../core/mem/align.h"
+  "../core/mem/hugetlb_path.h"
```
cmcl: Nit: On non-Linux targets, this header is installed but `hugetlb_path.c` is not compiled, so these declarations have no definitions.

category/core/mem/hugetlb_path.c
```diff
@@ -37,12 +44,150 @@ thread_local char g_error_buf[PATH_MAX];
...
+        if (strcmp(mount.mnt_type, "hugetlbfs") != 0) {
+            continue; // Not hugetlbfs, skip
+        }
+        // Parse the mount options, looking for pagesize=X so we can check this
+        // mount against `required_pagesize`
+        strlcpy(mntopts, mount.mnt_opts, sizeof mntopts);
+        for (char const *tok = strtok_r(mntopts, ",", &saveptr); tok != nullptr;
+             tok = strtok_r(nullptr, ",", &saveptr)) {
+            uint64_t pagesize;
+            constexpr char PAGESIZE_OPT[] = "pagesize=";
```
cmcl: I think this changes behaviour compared with libhugetlbfs. The kernel docs describe the hugetlbfs `pagesize` mount option as optional: [if `pagesize` is not specified, the platform default huge page size will used](https://www.kernel.org/doc/html/v6.7/admin-guide/mm/hugetlbpage.html#using-huge-pages). That means a valid default-size hugetlbfs mount may not have a `pagesize=` option to parse from `/proc/mounts`.

Would it be more robust to treat `/proc/mounts` only as candidate mount enumeration, then use `statfs(mount.mnt_dir, ...)` to compare `f_bsize == required_pagesize`? If we are already calling `statfs`, we could also check `f_type == HUGETLBFS_MAGIC`, matching `monad_hugetlbfs_check_path` and avoiding the `"hugetlbfs"` string comparison.

Co-Authored-By: Codex GPT 5.5

category/core/mem/hugetlb_path.c
```diff
@@ -37,12 +44,150 @@ thread_local char g_error_buf[PATH_MAX];
         &MONAD_SOURCE_LOCATION_CURRENT(),                                      \
         __VA_ARGS__)
 
+// Simple function for parsing human-readable sizes like "256 kB", "8M" or "1G"
+static int expand_byte_size(char const *const text, uint64_t *const size)
```
cmcl:
```suggestion
static int expand_byte_size(char const *const text, size_t *const size)
```

The only callers effectively compare to `size_t` anyway.  The call in `find_hugetlbfs_path_for_size` might be removable (if adopting `statfs.f_bsize`) and the call in `monad_get_default_hugepage_size` would no longer require type punning.  If this takes `size_t *`, the `sscanf` should use `%zu`.

category/core/mem/hugetlb_path.c
kjcamann: Yes this is much better, will fix

category/core/mem/hugetlb_path.c
kjcamann: Done

category/event/CMakeLists.txt
kjcamann: Fixed

cmcl (dismissed):

cmcl (approved):

Baltoli (approved):

dhil (approved):

PR #2300, opened 2026-05-27
[execution/ethereum] Add EIP-7691 blob parameters

cmcl:
This PR adds the minimal EIP-7691 blob-parameter support needed to unblock Prague Ethereum replay.  The PR introduces `eip_7691_active()` to gate the blob gas helpers rather than gating on the EVMC revision.

It makes the existing blob fee and blob gas limit calculations fork-aware:

- Cancun keeps the EIP-4844 values:
  - max blobs per block: 6
  - blob base fee update fraction: 3,338,477
- Prague and later Ethereum revisions use the EIP-7691 values:
  - max blobs per block: 9
  - blob base fee update fraction: 5,007,716

The change intentionally stays smaller than a general blob schedule implementation.  It deliberately does not implement:

- parent-aware header validation of excess_blob_gas
- target_blobs_per_block (required only by parent header validation)
- EIP-7840-style chain-config blob schedules

A separate draft PR will be opened for parent-aware header validation to workshop the best approach.  That PR will be able to remove the remaining eip4844_blobs excess-blob-gas exclusions.

Co-Authored-By: Codex GPT 5.5

### DIFF
--- a/category/execution/ethereum/validate_block.cpp
+++ b/category/execution/ethereum/validate_block.cpp
@@ -212,8 +212,7 @@ constexpr Result<void> static_validate_4844(Block const &block)
                 blob_gas_used += get_total_blob_gas(tx);
             }
         }
-        constexpr uint64_t MAX_BLOB_GAS_PER_BLOCK = 786432;
-        if (MONAD_UNLIKELY(blob_gas_used > MAX_BLOB_GAS_PER_BLOCK)) {
+        if (MONAD_UNLIKELY(blob_gas_used > max_blob_gas_per_block<traits>())) {
             return BlockError::GasAboveLimit;
         }
         if (MONAD_UNLIKELY(
--- a/category/execution/ethereum/validate_transaction.cpp
+++ b/category/execution/ethereum/validate_transaction.cpp
@@ -212,7 +212,8 @@ Result<void> static_validate_transaction(
 
             if (MONAD_UNLIKELY(
                     tx.max_fee_per_blob_gas <
-                    get_base_fee_per_blob_gas(excess_blob_gas.value()))) {
+                    get_base_fee_per_blob_gas<traits>(
+                        excess_blob_gas.value_or(0)))) {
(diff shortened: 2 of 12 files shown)


category/execution/ethereum/validate_block.cpp
```diff
@@ -212,7 +212,8 @@ constexpr Result<void> static_validate_4844(Block const &block)
                 blob_gas_used += get_total_blob_gas(tx);
             }
         }
-        constexpr uint64_t MAX_BLOB_GAS_PER_BLOCK = 786432;
+        constexpr uint64_t MAX_BLOB_GAS_PER_BLOCK =
+            max_blob_gas_per_block<traits>();
```
Baltoli: I'd just inline this into the conditional now that it's no longer a magic number

category/execution/ethereum/validate_transaction.cpp
```diff
@@ -212,7 +212,8 @@ Result<void> static_validate_transaction(
 
             if (MONAD_UNLIKELY(
                     tx.max_fee_per_blob_gas <
-                    get_base_fee_per_blob_gas(excess_blob_gas.value()))) {
+                    get_base_fee_per_blob_gas<traits>(
+                        excess_blob_gas.value_or(0)))) {
```
Baltoli, line 216: Why the change from `.value()` to `.value_or(0)`?

cmcl: Copilot highlighted there are some tests which pass `std::nullopt`.  My previous fix to use a `has_value()` guard meant that calls with `excess_blob_gas` set to `std::nullopt` would allow `max_fee_per_blob_gas` to be `0` whereas the `MIN_BASE_FEE_PER_BLOB_GAS` specified by [EIP 4844](https://eips.ethereum.org/EIPS/eip-4844) is `1`

Baltoli (approved):

goodlyrottenapple (approved):

PR #1628, opened 2025-09-09
Division and modular arithmetic via Barrett reduction

m-alvarez:
# Benchmarks
### Udivrem

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 115.35 | 99.05  | 88.83   | 56.13   | 134.66 |
| Division                 | 232.92 | 382.35 | 298.32  | 206.30  | 276.28 |

### Addmod

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 142.47 | 142.04 | 111.14  | 112.57  | 187.12 |
| Division                 | 283.85 | 436.81 | 362.67  | 256.83  | 318.99 |

### Mulmod

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 318.10 | 318.86 | 290.94  | 278.39  | 415.61 |
| Division                 | 511.98 | 852.60 | 824.30  | 777.86  | 740.81 |

### Mulmod by a constant

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 107.44 | 143.00 | 176.75  | 218.53  | 238.22 |
| Division                 | 501.78 | 841.55 | 822.95  | 774.18  | 736.42 |

### Udivrem

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 107.80 | 125.03 | 92.13   | 61.73   | 144.25 |
| Division                 | 296.85 | 389.71 | 321.43  | 237.60  | 308.19 |

### Addmod

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 149.81 | 135.43 | 109.00  | 105.94  | 195.59 |
| Division                 | 337.28 | 496.55 | 415.57  | 304.16  | 370.96 |

### Mulmod

| Divisor significant bits | 1-64   | 65-128 | 129-192 | 193-256 | 1-256  |
|--|--|--|--|--|--|
| Barrett                  | 303.96 | 300.92 | 291.90  | 278.89  | 410.74 |
[...]

PR #2303, opened 2026-05-29
Inline DIV/MOD when divisor fits in 64 bits

jhunsaker:


andreaslyn (commented):
We are currently working on a general approach to optimize div/mod by a constant via Barrett reduction https://github.com/category-labs/monad/pull/1628. Our guess is that Barrett reduction will be more efficient than inlining division. But it would be good to have fair benchmarks of both approaches for comparison.

The fuzzer is failing on this PR at the moment.

PR #2304, opened 2026-06-01
[execution/vm] Introduce in-tree monad_evm_revision, replacing evmc_revision

Baltoli:
## Summary

First step in replacing deprecated evmc leaf types with our own: introduce an in-tree `monad_evm_revision` enum and migrate the codebase off evmc's `evmc_revision`. This also decouples us from evmc's release cadence — future Ethereum forks can be added without waiting on an evmc bump.

Two commits, as discussed:

1. **Add `monad_evm_revision`** — additive, zero behavior change.
   - `category/vm/evm/revision.h`: the new C enum, free of any evmc include, mirroring `evmc_revision` 1:1 (modeled on the sibling `monad/revision.h`).
   - `category/vm/evm/revision.{hpp,cpp}`: `to_evmc_revision()` / `from_evmc_revision()` shims, with `static_assert`s enforcing the 1:1 value correspondence (the only place that includes evmc for this purpose).

2. **Migrate call sites** (~89 files) — behavior-preserving rename.
   - `EvmTraits` template parameter, the `Traits` concept's `evm_rev()`, `Chain::get_revision()`, `SWITCH_EVM_TRAITS` dispatch labels, precompiles, runtime, opcodes, fuzzing, and every `>= EVMC_<fork>` feature gate now use `monad_evm_revision` / `MONAD_EVM_*`. Identical underlying values make this a pure rename.

## evmc/evmone boundary

`evmc_revision` survives **only** at the genuine evmc/evmone boundaries, all of which are on test and benchmark paths: the `evmc_vm` ABI execute callbacks and the evmone `baseline::execute` / test-`Host` calls. Each converts explicitly via the shims rather than relying on implicit enum decay, so the remaining reify points are visible and removable once evmone is gone.

## Future forks

Adding a post-Osaka fork is now a localized edit (enum value + `SWITCH_EVM_TRAITS` case + traits flags) with no evmc dependency. Caveat: a future-only fork can't be exercised through the evmone-based differential tests until evmone is removed, since `to_evmc_revision` would have no evmc value to map it to.

## Out of scope

The existing `#include <evmc/evmc.h>` lines in migrated files are left in place; stripping transitive evmc includes is a natural follow-up but would broaden the blast radius well beyond this rename.

## Testing

- Full project build: green.
- `check-trait-instantiations.py`: no overlaps.
- 2656 VM / EVM / traits / precompile / execution tests: 100% pass.
- clang-format clean on all changed files.

### DIFF
--- a/category/vm/evm/revision.h
+++ b/category/vm/evm/revision.h
@@ -56,7 +56,7 @@ enum monad_eth_revision
     MONAD_ETH_MAX_REVISION = MONAD_ETH_EXPERIMENTAL,
 
     // The latest known EVM revision with finalized specification.
-    MONAD_ETH_LATEST_STABLE_REVISION = MONAD_ETH_CANCUN
+    MONAD_ETH_LATEST_STABLE_REVISION = MONAD_ETH_PRAGUE
 };
 
 char const *monad_eth_revision_to_string(enum monad_eth_revision rev);
--- a/category/vm/evm/revision.h
+++ b/category/vm/evm/revision.h
@@ -1,0 +1,71 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <evmc/evmc.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+// Monad's in-tree EVM fork revision enum. This is a drop-in replacement for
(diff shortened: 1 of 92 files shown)


category/vm/evm/revision.h
```diff
@@ -0,0 +1,72 @@
...
+
+    // The latest known EVM revision with finalized specification.
+    MONAD_EVM_LATEST_STABLE_REVISION = MONAD_EVM_CANCUN
+};
+
+char const *monad_evm_revision_to_string(enum monad_evm_revision rev);
+
+// Value-preserving conversions to/from evmc's evmc_revision. The casts are safe
+// because monad_evm_revision mirrors evmc_revision 1:1 (asserted in
+// revision.cpp). Needed only at the evmc/evmone boundaries; see the note above.
```
dhil: Maybe this comment should be in the `.cpp` file rather than on the interface declaration as it concerns some implementation details.

dhil (dismissed):

dhil (dismissed):

category/vm/evm/revision.h
Baltoli: Good call — moved the implementation rationale (value-preserving cast, 1:1 correspondence enforced by the static_asserts) down to revision.cpp, and left only an interface-level description on the declaration. Done in 46de78aec.

dhil (dismissed):

dhil (dismissed):

goodlyrottenapple (dismissed):

dhil (approved):

mkolosick (approved):

PR #2305, opened 2026-06-02
[execution/ethereum] Restructure intrinsic_gas as additive feature gates

Baltoli:
## Summary

Restructures `intrinsic_gas` in `transaction_gas.cpp` from a descending `evm_rev() < EVMC_X` if-ladder into an additive accumulator that gates each cost term on the feature that introduced it.

The old chain re-listed the base terms in every branch and encoded the cumulative intent as *descending* thresholds, so you had to diff adjacent branches to see which EIP added which cost. The new form reads top-to-bottom: base cost, plus each feature's term when active.

## Bug fixed

The old `else` branch covered **Cancun and later** and added `g_authorization(tx)` — but EIP-7702 authorization lists are a **Prague** feature. So Cancun went through the EIP-7702 pricing path. The authorization term is now gated at `>= EVMC_PRAGUE`.

This was **latent**: a decoded Cancun transaction can never carry a non-empty authorization list (the RLP decoder only populates it for type-4 txns, and validation rejects type-4 before Prague), so `g_authorization` returned 0 in practice. Incorrect but unreachable from real input — hence no test, as discussed.

## Behavior

Equivalent to the *corrected* ladder for every supported revision. Uses the existing `eip_2929_active()` flag; the two remaining `evm_rev()` comparisons are positioned to become `eip_3860_active()` / `eip_7702_active()` once the feature-flag migration lands.

## Testing

- Builds cleanly (`monad_execution`, GCC-15 RelWithDebInfo / AVX2 toolchain).
- `clang-format-19` clean.

### DIFF
--- a/category/execution/ethereum/transaction_gas.cpp
+++ b/category/execution/ethereum/transaction_gas.cpp
@@ -112,24 +112,22 @@ uint64_t intrinsic_gas(Transaction const &tx) noexcept
 {
     static_assert(traits::evm_rev() > EVMC_HOMESTEAD);
 
-    if constexpr (traits::evm_rev() < EVMC_BERLIN) {
-        return 21'000 + g_data<traits>(tx) + g_txn_create(tx);
+    uint64_t gas = 21'000 + g_data<traits>(tx) + g_txn_create(tx);
+
+    // EIP-2930: access-list and storage-key cost (Berlin)
+    if constexpr (traits::evm_rev() >= EVMC_BERLIN) {
+        gas += g_access_and_storage(tx);
     }
-    else if constexpr (traits::evm_rev() < EVMC_SHANGHAI) {
-        return 21'000 + g_data<traits>(tx) + g_txn_create(tx) +
-               g_access_and_storage(tx);
+    // EIP-3860: per-word initcode cost (Shanghai)
+    if constexpr (traits::evm_rev() >= EVMC_SHANGHAI) {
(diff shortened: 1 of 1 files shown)


category/execution/ethereum/transaction_gas.cpp
```diff
@@ -112,24 +112,22 @@ uint64_t intrinsic_gas(Transaction const &tx) noexcept
...
-    if constexpr (traits::evm_rev() < EVMC_BERLIN) {
-        return 21'000 + g_data<traits>(tx) + g_txn_create(tx);
-    }
-    else if constexpr (traits::evm_rev() < EVMC_SHANGHAI) {
-        return 21'000 + g_data<traits>(tx) + g_txn_create(tx) +
-               g_access_and_storage(tx);
+    uint64_t gas = 21'000 + g_data<traits>(tx) + g_txn_create(tx);
+
+    // EIP-2930: access-list and storage-key cost (Berlin)
+    if constexpr (traits::eip_2929_active()) {
```
m-alvarez: The comment says we're testing for EIP-2930 but the actual check is for 2929. Is there a reason to not check `traits::evm_rev() >= EVMC_BERLIN` instead?

dhil (dismissed):

m-alvarez (dismissed):

dhil (approved):

goodlyrottenapple (approved):

PR #2319, opened 2026-06-05
[execution/ethereum] Replace constexpr validators with inline

cmcl:
~Replace~ Remove misleading `constexpr` annotations on the static block validation helpers ~with explicit `inline`~. These helpers are not used in constant-evaluation contexts. ~but `constexpr` previously implied `inline`; switching to `inline` preserves previous code generation.~

### DIFF
--- a/category/execution/ethereum/validate_block.cpp
+++ b/category/execution/ethereum/validate_block.cpp
@@ -163,8 +163,7 @@ Result<void> static_validate_header(BlockHeader const &header)
 EXPLICIT_TRAITS(static_validate_header);
 
 template <Traits traits>
-constexpr Result<void>
-static_validate_ommers(Chain const &chain, Block const &block)
+Result<void> static_validate_ommers(Chain const &chain, Block const &block)
 {
     // YP eq. 33
     if (compute_ommers_hash(block.ommers) != block.header.ommers_hash) {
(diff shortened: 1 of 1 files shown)


Baltoli:
I don't think these should even be marked `inline` given that they're in a TU rather than a header.

dhil:
I would argue that the `constexpr` annotations should remain, because a feature of `constexpr` expressions/statements is that they do not permit undefined behaviour.

Baltoli:
> I would argue that the constexpr annotations should remain, because a feature of constexpr expressions/statements is that they do not permit undefined behaviour.

Only true if the function is actually constant-evaluated, right? That's a property of an evaluation rather than the body of the function itself.

Baltoli (approved):

dhil:
> > I would argue that the constexpr annotations should remain, because a feature of constexpr expressions/statements is that they do not permit undefined behaviour.
> 
> Only true if the function is actually constant-evaluated, right? That's a property of an evaluation rather than the body of the function itself.

Yes I think that's correct.

dhil (approved):

Baltoli:
These functions are never constant-evaluated, so the specifier is meaningless - you don't get any UB protection because every evaluation is at runtime

dhil:
> These functions are never constant-evaluated, so the specifier is meaningless - you don't get any UB protection because every evaluation is at runtime

Yes, we discussed it offline yesterday and reached the same conclusion.

PR #2324, opened 2026-06-08
Thread Monad consensus round number into evmc_tx_context (EIP-7843)

brett-monad:
## Summary
  
  Carries the Monad consensus round into the EVM tx context (for a future EIP-7843
  opcode): adds `std::optional<uint64_t> slot_number` to the EL `BlockHeader`, populated
  from `MonadConsensusBlockHeader::block_round` on the consensus path; `get_tx_context`
  reads it into `evmc_tx_context.block_round` (0 for EVM/eth). Bumps `evmc`/`evmone`
  submodules to the merged commits adding `block_round`. **Type + plumbing only.**

  `slot_number` is in-memory only (not RLP-encoded) — wire format/hashing unchanged; the
  round still travels the wire via the consensus header's own `block_round`.

  **Deferred to the opcode/EIP-7843-fork PR:** `BlockHeader` RLP encoding + revision
  gating, and the SLOTNUM opcode.

  ## Testing

  Builds clean (GCC + Clang `build-fuzz`); full `ctest` passes; unit test verifies round propagation.

### DIFF
--- a/category/execution/ethereum/core/block.hpp
+++ b/category/execution/ethereum/core/block.hpp
@@ -56,6 +56,19 @@ struct BlockHeader
     std::optional<uint64_t> excess_blob_gas{std::nullopt}; // EIP-4844
     std::optional<bytes32_t> parent_beacon_block_root{std::nullopt}; // EIP-4788
     std::optional<bytes32_t> requests_hash{std::nullopt}; // EIP-7685
+    // EIP-7843. In-memory only: deliberately NOT RLP-encoded. Encoding is
+    // deferred to the 7843 activation fork. Adding it to encode_block_header
+    // now would change every block hash with no fork gate, and would lock an
+    // on-wire position that EIP-7843 leaves unpinned relative to a future
+    // EIP-7928 block_access_list_hash. The consensus round is durably carried
+    // in MonadConsensusBlockHeader::block_round, so replay paths can
+    // repopulate this from there without any on-wire encoding.
+    // Coverage: only the consensus execution path (propose_block in
+    // runloop_monad) populates this. Any entrypoint that does not repopulate
+    // it -- RPC/trace re-execution in monad_executor, historical-block replay
+    // runloops -- leaves it nullopt, so the round read via value_or(0) is 0
+    // there.
+    std::optional<uint64_t> slot_number{std::nullopt};
 
     friend bool operator==(BlockHeader const &, BlockHeader const &) = default;
 };
@@ -70,10 +83,10 @@ struct Block
     friend bool operator==(Block const &, Block const &) = default;
 };
 
-static_assert(sizeof(BlockHeader) == 760);
+static_assert(sizeof(BlockHeader) == 776);
 static_assert(alignof(BlockHeader) == 8);
 
-static_assert(sizeof(Block) == 840);
+static_assert(sizeof(Block) == 856);
 static_assert(alignof(Block) == 8);
 
 MONAD_NAMESPACE_END
--- a/category/execution/ethereum/evmc_host_test.cpp
+++ b/category/execution/ethereum/evmc_host_test.cpp
@@ -71,7 +71,8 @@ bool operator==(evmc_tx_context const &lhs, evmc_tx_context const &rhs)
            !std::memcmp(
                lhs.block_base_fee.bytes,
                rhs.block_base_fee.bytes,
-               sizeof(evmc_bytes32));
+               sizeof(evmc_bytes32)) &&
+           lhs.block_round == rhs.block_round;
 }
 
 TYPED_TEST(TraitsTest, get_tx_context)
@@ -120,6 +121,14 @@ TYPED_TEST(TraitsTest, get_tx_context)
         hdr.prev_randao.bytes,
         sizeof(hdr.prev_randao));
     EXPECT_EQ(pos_result, ctx);
+
+    // slot_number (EIP-7843) is surfaced as block_round (unset -> 0 is
+    // covered by the comparisons above).
+    hdr.slot_number = 9'000'000'000;
+    EXPECT_EQ(
+        get_tx_context<typename TestFixture::Trait>(tx, from, hdr, 1)
+            .block_round,
+        uint64_t{9'000'000'000});
 }
 
 TYPED_TEST(TraitsTest, emit_log)
--- a/cmd/monad/runloop_monad.cpp
+++ b/cmd/monad/runloop_monad.cpp
@@ -283,6 +283,13 @@ Result<BlockExecOutput> propose_block(
     block.header.parent_hash =
         to_bytes(keccak256(rlp::encode_block_header(db.read_eth_header())));
 
+    // EIP-7843: surface the Monad consensus round to execution via the EL
+    // header's slot_number (in-memory only for now; not RLP-encoded). It is
+    // read into evmc_tx_context.block_round by get_tx_context and system_call.
+    // NOTE: only this path populates slot_number; RPC/trace re-execution does
+    // not (see EXE-60).
+    block.header.slot_number = consensus_header.block_round;
+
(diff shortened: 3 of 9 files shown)


Baltoli:
_**_Summary of a conversation between me and @claude_**_

## Design note: where `block_round` should live

The PR is correct and builds as-is — this is a scoping/design suggestion for the threading, not a correctness issue. Summary of a discussion on the `round` plumbing, in case it's useful before this lands.

### The concern

Threading `round` as a trailing `uint64_t` param through `execute_block` → `execute_block_transactions` → `dispatch_transaction` (×2) → `ExecuteTransaction` → `get_tx_context`, with `= 0` defaults at each layer, has a few smells: the call site now has to spell out `false, // trace_transfers` just to reach `round`; the defaults are inconsistent (eth `dispatch_transaction` defaults, monad one doesn't); and four layers of `= 0` is a silent-zero footgun for a consensus-relevant value.

### Recommendation: put it on `BlockHeader` as the EIP-7843 `slot_number`

Since `block_round` is meant to be 1:1 with the EIP-7843 slot number, the spec-aligned home is the EL header, exactly as Geth does (`types.Header.SlotNumber`, `rlp:"optional"`). Our `BlockHeader` already carries the same trailing-optional EIP pattern (1559/4895/4844/4788/7685), and `MonadConsensusBlockHeader` already embeds the EL header as `execution_inputs` right next to `block_round` — so this mirrors Ethereum's CL→EL flow.

```cpp
// BlockHeader, after requests_hash
std::optional<uint64_t> slot_number{std::nullopt}; // EIP-7843
```
```cpp
// get_tx_context — reads what it already receives; zero new params
.block_round = hdr.slot_number.value_or(0),
```

Populate it from `consensus_header.block_round` where `execution_inputs` is lifted into `block.header`. `get_tx_context` already takes `header_`, so **the entire threading diff disappears** — no trailing param on the five functions, no `round_` member. (A `ChainContext::get_block_round()` accessor was considered as an alternative carrier — branchless, avoids the `is_monad_trait_v` guard — but the header is the better home given replay below.)

### Agreed staging: 7843 now, no BAL interposition

- **Now:** append `slot_number` after `requests_hash`, revision-gated, fed from `block_round`. No `block_access_list_hash` (7928), **no placeholder**. EIP-7843's text only adds `slotNumber: uint64` — it does not mention 7928 or pin slot's position after BAL, so we're not forced to carry BAL to emit slot, and a placeholder BAL hash wouldn't buy replay compat anyway (wrong value for any non-empty BAL block).
- **When G\* / 7928 lands:** implement real BAL and reconcile the field order on the `EvmTraits` replay path to match the fork's canonical `[…, requests_hash, BAL, slot]`.

### Two notes so the deferral stays safe

1. **SLOTNUM spec tests are state tests, not header-hash tests** — they drive the slot via the env and assert the opcode returns it. Passing them needs only the opcode + the value reaching `tx_context` (which the header-fed `get_tx_context` provides). The header *wire* order is exercised only by blockchain/replay tests for the BAL fork, which we're deferring. So this plan passes SLOTNUM the moment the opcode is wired.
2. **The later wire-format revisit is free on the replay path, but a fork boundary for Monad's own chain.** Nothing emits slot on the `EvmTraits` path yet, so reordering to canonical later is free there. But once `slot_number` is live on Monad mainnet in the after-`requests_hash` position, interposing BAL before it changes emitted field order — a Monad hard-fork-level re-encode, not a silent edit. Normal (EIP activations are forks), just worth planning BAL adoption as a fork.

### Net

If 7843/replay is in scope, the header field is both neater and the only option that gives EL encoding parity eventually — and it obviates this PR's threading. Suggest: keep `slot_number` as the home, wire `SLOTNUM` → `tx_context.block_round` and drop the threaded `round` when the opcode lands. Before locking the on-wire position, confirm the target fork's full header order against the execution-spec/engine-API header definition (authoritative for ordering) rather than 7843 alone.

Baltoli (changes requested):
See note from @claude

brett-monad:
@claude review this PR

category/execution/ethereum/evmc_host_test.cpp
```diff
@@ -120,6 +120,16 @@ TYPED_TEST(TraitsTest, get_tx_context)
         hdr.prev_randao.bytes,
         sizeof(hdr.prev_randao));
     EXPECT_EQ(pos_result, ctx);
+
+    // block_round comes from BlockHeader::slot_number (EIP-7843); operator==
+    // above does not compare it, so assert it explicitly.
```
Baltoli: This isn't true, is it?
```cpp
    std::optional<uint64_t> slot_number{std::nullopt};

    friend bool operator==(BlockHeader const &, BlockHeader const &) = default;
```

m-alvarez: I think the comment is talking about the `operator==` on `evmc_tx_context` defined in this file on l. 48 but maybe that operator should be updated to account for the new field?

Baltoli: Ah, right, I misread the reference - but agreed, no reason not to update it

cmd/monad/runloop_monad.cpp
```diff
@@ -283,6 +283,11 @@ Result<BlockExecOutput> propose_block(
     block.header.parent_hash =
         to_bytes(keccak256(rlp::encode_block_header(db.read_eth_header())));
 
+    // EIP-7843: surface the Monad consensus round to execution via the EL
+    // header's slot_number (in-memory only for now; not RLP-encoded). It is
+    // read into evmc_tx_context.block_round by get_tx_context.
+    block.header.slot_number = consensus_header.block_round;
```
Baltoli, line 291: > 🤖 Generated by Claude

Could you add a `// TODO(EXE-60):` to the analogous place in RPC? This is _currently_ the only path that populates `slot_number`, and since it's in-memory only (not RLP-encoded), it can't be recovered from the EL header. The RPC/trace re-execution paths in `monad_executor.cpp` build blocks where `slot_number` is `nullopt`, so once `SLOTNUM` (EIP-7843) is wired, `debug_traceBlock`/historical re-execution would observe round 0 — diverging from how the block actually executed.

Consensus is unaffected (this chokepoint covers proposed + finalized/sync-replayed blocks); it's RPC/trace fidelity that needs the same repopulation from `MonadConsensusBlockHeader::block_round` before the opcode lands. Tracked in EXE-60.

cmcl (approved):

category/execution/ethereum/core/block.hpp
```diff
@@ -56,6 +56,19 @@ struct BlockHeader
...
+    // on-wire position that EIP-7843 leaves unpinned relative to a future
+    // EIP-7928 block_access_list_hash. The consensus round is durably carried
+    // in MonadConsensusBlockHeader::block_round, so replay paths can
+    // repopulate this from there without any on-wire encoding.
+    // Coverage: only the consensus execution path (propose_block in
+    // runloop_monad) populates this. Any entrypoint that does not repopulate
+    // it -- RPC/trace re-execution in monad_executor, historical-block replay
+    // runloops -- leaves it nullopt, so the round read via value_or(0) is 0
+    // there.
+    std::optional<uint64_t> slot_number{std::nullopt};
```
dhil, line 71: We never seem to use the fact that this thing can be `std::nullopt`. Is this intended to be used in a future patch?

dhil: Answer: it will be used in a later patch.

dhil (approved):

category/execution/ethereum/core/block.hpp
brett-monad: not exercised today, but intended for EXE-60 and 7843 serialization

mkolosick (approved):

Baltoli (approved):

PR #2326, opened 2026-06-09
Page-store dual-db: 4 integration fixes + MonadDevnetFork test chain

maxkozlovsky:
## Summary

Surfaced and fixed while building a full-cluster end-to-end test of the slot → page DB migration on `vicky/page-store-release1-dual-db`. Four small, scoped commits:

- `cmd/monad: allow booting single-db on a slot primary with no secondary` — the live runloop required a slot primary to have an active page secondary, making the pre-migration state unrepresentable. With this change: slot + active secondary → dual-write; slot + no secondary → single-db on the previous revision; page primary → single-db.
- `monad-statesync/ffi: pass chain_config to client_context_create` — the C API gained a leading `enum monad_chain_config` but the Rust FFI shim had not been updated. Paired with the monad-bft side, which threads `chain_config` from `monad-node`'s `chain_id`.
- `cmd/monad: register monad state machine before opening the on-disk Db` — `Db(OnDiskDbConfig)` resolves the `StateMachine` factory from the DB's persisted `state_machine_kind` at open time, so the monad (page) factory must be registered before the open call. Previously it was registered later in the per-chain runloop dispatch, which made a page-encoded DB unopenable.
- `Add MonadDevnetFork test-only chain (chain_id 20144)` — env-var-driven `MONAD_NINE → MONAD_NEXT` fork via `MONAD_DEVNETFORK_NEXT_AT`. Test-only chain, not registered for production. Needed by the integration test to exercise the slot→page migration across a fork crossing.

The monad-execution docker fix (skip C++ tests in the `runner` stage) is being submitted as a separate PR.

Verified by the monad-integration test `test_offline_dual_db_lifecycle` (PASS 134 s, single-node). Full status in `docs/dual_timeline_migration_poc.md` of the integration test PR.

## Test plan

- [x] Local single-node integration test passes end-to-end (full migration lifecycle, balances survive promote)
- [ ] CI build + test

### DIFF
--- a/category/execution/CMakeLists.txt
+++ b/category/execution/CMakeLists.txt
@@ -105,6 +105,7 @@ add_library(
   "ethereum/db/partial_trie_db.cpp"
   "ethereum/db/state_machine_init.cpp"
   "ethereum/db/state_machine_init.hpp"
+  "ethereum/db/storage_key.hpp"
   "ethereum/db/trie_db.cpp"
   "ethereum/db/trie_db.hpp"
   "ethereum/db/trie_rodb.hpp"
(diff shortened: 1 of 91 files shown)


maxkozlovsky:
all merged

PR #2335, opened 2026-06-11
[rpc] Graceful state override failure

dhil:
This patch turns the hard assertion failure on state override application failure into a graceful exit, that returns a message to the user.


dhil:
@claude please review this PR.

jhunsaker (approved):

PR #2339, opened 2026-06-15
[rpc] Improve the expressiveness of `eth_simulateV1`

dhil:
This patch increases the expressive power of `eth_simulateV1` when used with state overrides. This is achieved by inserting a "storage prefetch" loop in `apply_state_overrides`, which ensures that `set_storage` sees the correct state after resetting the account incarnation.

### DIFF
--- a/category/rpc/monad_executor.cpp
+++ b/category/rpc/monad_executor.cpp
@@ -185,6 +185,19 @@ namespace
 
             // Remove all override
             if (!state_delta.state.empty()) {
+                // A possible side-effect of `set_to_state_incarnation` is that
+                // it can alter what `set_storage` sees as the original storage
+                // (c.f. `update_state`). Therefore we prime the original
+                // storage slots against the existing block state incarnation
+                // before resetting the account incarnation. Consequently, merge
+                // validation remains consistent when a previous simulated
+                // block already mutated these slots.
+
+                // As with `state_diff`, access the account first.
+                (void)state.get_nonce(address);
+                for (auto const &[key, _] : state_delta.state) {
+                    (void)state.get_storage(address, key);
+                }
                 state.set_to_state_incarnation(address);
                 update_state(state_delta.state);
             }
--- a/category/rpc/monad_executor_test.cpp
+++ b/category/rpc/monad_executor_test.cpp
@@ -8002,22 +8002,19 @@ TEST_F(EthCallFixture, eth_simulate_v1_simple_transfer_withdrawals_monad)
     monad_executor_destroy(executor);
 }
 
-// Tests that failure to apply state overrides fails gracefully.
-TEST_F(EthCallFixture, eth_simulate_v1_state_override_graceful_failure)
+// Tests that state overrides can be applied after prior simulated mutations
+// in the same eth_simulate request.
+TEST_F(EthCallFixture, eth_simulate_v1_state_override)
 {
     static constexpr Address contract =
         0x00000000000000000000000000000000deadbeef_address;
     static constexpr Address sender =
         0x00000000000000000000000000000000feedface_address;
 
-    // Idea: We deploy a simple contract that does an SSTORE, then we call it
-    // such that it mutates the `BlockState` object held by the eth_simulate_v1
-    // context. Subsequently, we call the same contract again with a state
-    // override for the mutated field, which causes `block_state.can_merge` to
-    // fail because the override is incompatible with the mutated state.
-    // NOTE(dhil): This simulation pattern should probably be allowed, but for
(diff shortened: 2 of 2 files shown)


dhil:
@claude critically review this PR please.

category/rpc/monad_executor_test.cpp
```diff
@@ -7345,9 +7342,443 @@ TEST_F(EthCallFixture, eth_simulate_v1_state_override_graceful_failure)
...
+
+    nlohmann::json output = nlohmann::json::from_cbor(
+        ctx.result->encoded_trace,
+        ctx.result->encoded_trace + ctx.result->encoded_trace_len);
+
+    ASSERT_EQ(output.size(), 2);
+    ASSERT_EQ(output[0]["calls"].size(), 1);
+    ASSERT_EQ(output[1]["calls"].size(), 1);
+    EXPECT_EQ(output[0]["calls"][0]["status"], "0x1");
+    EXPECT_EQ(output[1]["calls"][0]["status"], "0x1");
```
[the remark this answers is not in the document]
dhil: I've implemented a variation of this suggestion.

category/rpc/monad_executor.cpp
```diff
@@ -189,6 +189,19 @@ namespace
 
             // Remove all override
             if (!state_delta.state.empty()) {
+                // A possible side-effect of `set_to_state_incarnation` is that
```
aa755, line 188: I think we should just get rid of incarnations. I dont see why we really need them post cancun. IIUC, post cancun, accounts cannot really be recreated except within a single transaction.

dhil: This is an interesting point, thanks!

dhil: I investigated this idea further. In principle I think we can dispense of incarnations, however, right now `set_to_state_incarnation` has the side effect of zeroing the storage of the account, which is required to correctly implement the override "state". Thus in the spirit of minimal changes, I think we should defer the removal of incarnations to a subsequent patch, because it will require us to extend the state object with some form of storage zeroing mechanism.

category/rpc/monad_executor.cpp
```diff
@@ -185,6 +185,19 @@ namespace
...
+                // storage slots against the existing block state incarnation
+                // before resetting the account incarnation. Consequently, merge
+                // validation remains consistent when a previous simulated
+                // block already mutated these slots.
+
+                // As with `state_diff`, access the account first.
+                (void)state.get_nonce(address);
+                for (auto const &[key, _] : state_delta.state) {
+                    (void)state.get_storage(address, key);
+                }
```
aa755, line 200: Could full account-override `state` still leave stale cached storage for omitted keys?

  In [this commit](https://github.com/category-labs/monad/commit/f0f806e3363592082f77d9f5f41b43a26cac1897), with Codex I added a failing test case which has block 1 write slot `0 = 1`, then block 2 applies a full
  `state` override for the same account containing only slot `1 = 0xff`. Reading slot `0` then returns `1`, but I think it should return `0`.

  Reason: `state` appears to mean whole-storage replacement, unlike `stateDiff`, which is partial. Codex says Geth handles this distinction directly: `state` calls `statedb.SetStorage(...)`, while `stateDiff` loops
  over keys and calls `SetState(...)`. `SetStorage` is documented as replacing the entire storage and creates a fresh object containing only the supplied map entries.

  Geth refs found by Codex (I have not checked these as I am not familiar with Geth):
  - override handling: https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/override/override.go#L849-L864
  - `SetStorage`: https://github.com/ethereum/go-ethereum/blob/master/core/state/statedb.go#L3053-L3092

  [This commit](https://github.com/category-labs/monad/commit/a91698d265b226cdaae87a31f8dbfe5a57e40d41) fixes the issue by clearing cached `BlockState` storage when the merged account incarnation changes.

  I also added [one more defensive fix](https://github.com/category-labs/monad/commit/71759038ecf14ba4de7d3766a487776fde7047c4): `State::set_to_state_incarnation` now clears the transaction-local `State` storage
  cache too. Without this, [a contrived unit test](https://github.com/category-labs/monad/commit/6c4236c87f8354e709c8248229ef45976e4d4a69) can keep an old cached key alive after the incarnation reset. I am not sure
  whether a normal user request can trigger that path, but clearing the local cache matches the same storage-epoch reset invariant as the `BlockState` fix.

  Locally: the tests fail before their respective fixes, pass after the fixes.

dhil: Thanks! This is insightful. These are some of the details I wasn't confident about at all. I will incorporate this idea into the PR.

PR #2340, opened 2026-06-15
Caching of riscv cross compiler and bump github workflow ubuntu version

mkolosick:


jhunsaker (dismissed):

eclark0426 (dismissed):

scripts/vm/benchmark-analysis/pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "benchmark-analysis"
 version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
-requires-python = ">=3.12"
```
dhil, line 6: Is there a reason for this change? Or is it some automatic tool update?

dhil (dismissed):

scripts/vm/benchmark-analysis/pyproject.toml
mkolosick: Ubuntu 26.04 default set of installs for python starts at 3.13 so we need to either change the install method or just bump the version.

.github/workflows/code-quality.yml
```diff
@@ -32,16 +32,14 @@ jobs:
             submodules: false
       - name: Install clang-format-${{ env.LLVM_VERSION }}
         run: |
-          wget https://apt.llvm.org/llvm.sh
-          sudo bash llvm.sh ${{ env.LLVM_VERSION }}
```
dhil, line 36: Just flagging that this caused problems in the past, c.f. https://github.com/category-labs/monad/pull/2072.

scripts/vm/benchmark-analysis/pyproject.toml
dhil: I read the constraint as meaning "at least version 3.12", if that reading is correct then 3.13 also ought to work.

.github/workflows/code-quality.yml
mkolosick: Do you remember what the problem was? Was it related to how we installed `clang-format` or just the disappearance of the upstream action?

dhil: I think the problem was that some third party repo broke, which the entire world seemed to depend on. Probably safe and robust enough to move to the package offered by `apt` provided it meets our requirements.

scripts/vm/benchmark-analysis/pyproject.toml
mkolosick: Fair. I'll revert this part.

scripts/vm/benchmark-analysis/.python-version
```diff
@@ -1 +1 @@
-3.12
```
dhil, line 1: @mkolosick should this change be reverted too then? Or does this imply a strict equality on the version?

mkolosick: This is only used by `.github/workflows/test-vm.yml` to signal what version to install.

mkolosick: Which needed to be bumped with the move to ubuntu 26.04

dhil (approved):

goodlyrottenapple (approved):

Baltoli (approved):

PR #2344, opened 2026-06-16
[execution] Fiber synchronization abstraction.

dhil:
This patch adds a zero-cost wrapper over Boost fibers promise synchronization. The purpose of this wrapper is to abstract the details of `boost::fibers::promise` such that we can shadow it on platforms that do not support Boost fibers. Behavior is unchanged: fibers still wait on predecessor completion in the same baton-passing flow.


dhil:
@claude please critically review this PR.

Baltoli (approved):

goodlyrottenapple (approved):

PR #2345, opened 2026-06-16
Replace deprecation assertions with >=

eclark0426:


eclark0426:
@claude please review

eclark0426:
I intentionally left out the one line that claude mentioned, since it was not introduced by the deprecation commits, and doesn't seem to serve the same purpose of "this logic was changed and is no longer valid for earlier forks"

jhunsaker (approved):

PR #2359, opened 2026-06-18
mpt: enable Db::traverse and traverse_blocking on the secondary timeline

Chen-Yifan:
Both traverse entry points asserted the bound timeline was primary, and the traversal internals hardcoded timeline_id::primary when validating versions and reading nodes off disk. read_node_blocking and version_is_valid_ondisk already take a timeline_id, so the assertion was the only thing blocking secondary-timeline traversal.

Thread the timeline through the traverse path: preorder_traverse_blocking, preorder_traverse_ondisk, and TraverseSender now carry a timeline_id (defaulting to primary), and the Db impls pass their bound tid. Drop the primary-only assertions in Db::traverse, Db::traverse_blocking, and the templated children_of overload.

All new parameters default to timeline_id::primary, so existing primary callers are unaffected.

### DIFF
--- a/category/mpt/db.cpp
+++ b/category/mpt/db.cpp
@@ -239,7 +239,7 @@ public:
         size_t const concurrency_limit) override
     {
         return preorder_traverse_ondisk(
-            aux(), std::move(node), machine, version, concurrency_limit);
+            aux(), std::move(node), machine, version, tid_, concurrency_limit);
     }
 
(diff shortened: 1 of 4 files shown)


kkuehlz (approved):

maxkozlovsky (approved):

marcograss (commented):
approved

PR #2363, opened 2026-06-22
[docs] Add slot → page storage migration operator runbook

maxkozlovsky:
## Summary

Adds `docs/page_store_migration_runbook.md` — an operator/devops runbook for the online slot → page (page-store) database migration.

Covers, self-contained for an operator audience:

- The dual-timeline model (slot primary + page secondary, dual-write window, fork flips canonical state root).
- The single hard invariant: every node must activate **and** populate its page secondary before the chain crosses the fork, or it fail-stops on `read_storage_page is only valid on a page-encoded TrieDb`.
- The per-node command sequence with expected success/error log lines: `--activate-secondary --state-machine monad`, `monad-cli` snapshot dump/load `--secondary`, `--promote-secondary`, `--deactivate-secondary`.
- Cluster rolling orchestration (Phases A–D, one node offline at a time, promote staggered freely post-fork).
- Verification criteria, rollback / guard-rails table, and production-scale cautions.

Derived from the cluster integration tests in category-labs/monad-integration#432 and the associated design notes; all test- and source-internal references were stripped so the doc stands alone for operators.

## Test plan

Docs only — no code change.

### DIFF
--- a/docs/page_store_migration_runbook.md
+++ b/docs/page_store_migration_runbook.md
@@ -1,0 +1,405 @@
+# Slot → Page Storage Migration — Operator Runbook
+
+Operational procedure for migrating a Monad node's on-disk MPT database from the
+**slot-encoded** layout to the **page-encoded** (page-store) layout, online,
+without resyncing from genesis.
+
+> **Audience:** node operators / devops running validators and full nodes.
+> Assumes familiarity with starting/stopping a node and locating its triedb.
+
+---
+
+## 1. What this migration is
+
+The page-store change replaces the per-storage-slot trie encoding with a
+page-packed encoding (a bitmap plus packed values, bucketed by page key). Rather
+than rewrite the live database in place, migration runs a **dual-timeline**
+window:
+
+- The node keeps its existing **slot** database as the *primary* timeline.
+- It stamps and populates a second **page** database as the *secondary* timeline.
+- For a window it **dual-writes** every committed block to both.
+- At a scheduled chain fork the **canonical** state root flips from slot to page.
+- After the fork the operator **promotes** the page secondary to primary and
+  **deactivates** the now-stale slot timeline, leaving a single page-encoded
+  database.
+
+State-machine kinds, as named on the CLI:
+
+| Kind on CLI | Encoding | Role during migration |
+|-------------|----------|-----------------------|
+| `ethereum`  | slot     | the pre-migration primary |
+| `monad`     | page     | the secondary you activate, then promote to primary |
+
+### The fork is the hinge
+
+The page-store fork is a scheduled protocol upgrade defined by the chain config —
+a single timestamp, identical across the whole cluster. It is not set by the
+operator: it ships in a later release, once every node has migrated to
+dual-timeline and verification (§4) passes. Canonical state selection is driven
+entirely by whether a block is before or after it:
+
+- **Before the fork** — the slot primary owns the canonical `state_root` stamped
+  into the block header; the page secondary shadows it.
+- **At/after the fork** — the page secondary owns the canonical `state_root`; the
+  slot primary shadows it.
+
+Which database is "on disk primary" (the thing `--promote-secondary` changes) is
+**independent of which timeline is canonical**: pre-fork the slot is canonical no
+matter which ring is on-disk primary, post-fork the page is. That independence is
+what lets promote be staggered across a cluster (§3.2).
+
+---
+
+## 2. The one invariant you must not violate
+
+**Every node must activate AND populate its page secondary before the chain
+crosses the fork.**
+
+A node still slot-only when the chain crosses the fork cannot follow the
+now-canonical page timeline: it fail-stops — the process crashes cleanly, with no
+data corruption. Recovery is then forward only: the node must activate its
+secondary and rejoin via statesync. **Promote and deactivate, by contrast, may
+happen any time after the fork and may be staggered freely.**
+
+Because the fork point ships in a separate release (§1, §3.2), this is not a race
+against a fixed deadline: migrate and verify the whole fleet first, and only then
+does the fork-setting release go out. The constraint is ordering — that release
+must not ship while any node is still slot-only.
+
+---
+
+## 3. Procedure
+
+Every node needs the page-store-enabled `monad-mpt` and `monad-cli` staged first —
+the `--state-machine monad` kind and the snapshot `--secondary` flag do not exist
+on an older slot-only build.
+
+The per-node command sequence (§3.1) is the same whether you migrate a single
+node or roll it across a cluster (§3.2). For a cluster you run §3.1's two offline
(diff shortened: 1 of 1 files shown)


docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,291 @@
...
+**Step 1 — Stop the node.**
+
+**Step 2 — Activate the page secondary (stamps an empty page timeline).**
+
+```bash
+monad-mpt --activate-secondary --state-machine monad --storage "$TRIEDB"
+```
+- Success line: `Activated secondary timeline`
+- The command also prints the database's latest finalized version:
+  `Latest finalized is <N>`. **Use that `<N>`** for the snapshot below — do not
```
Chen-Yifan: I'm thinking we should probably make the default version as latest finalized, so it doesn't need to rely on the operator grepping for `<N>`

docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,291 @@
...
+- This re-encodes all state at version `<N>` from slot form into page form; the
+  empty secondary now holds the full state.
+
+**Step 5 — Restart the node → dual-write mode.**
+The node now commits every block to both the slot primary and the page secondary.
+Wait for it to rejoin and catch up, and confirm block production resumes.
+
+> Optional safety: boot with `--dual-db-migration-mode` to cross-check every read
+> against both timelines during the dual-write window. It costs read latency — use
+> it to build confidence on the first nodes rather than fleet-wide.
```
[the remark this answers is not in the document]
Chen-Yifan: This part of the doc is outdated, I removed the `--dual-db-migration-mode` from the cli option for `monad` binary. It automatically commits to all active timelines without the need of an additional flag.

docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,291 @@
...
+
+**Step 6 — Cross the fork.** No operator action: block timestamps reach the
+configured fork time and the canonical state root flips to the page secondary.
+From here the page timeline is the source of truth.
+
+**Step 7 — Stop the node** (now post-fork).
```
Chen-Yifan, line 142: worth mentioning that Step 7 and onward can happen any time after the fork, and doesn't have to be immediately after.

maxkozlovsky: ok added

docs/page_store_migration_runbook.md
maxkozlovsky: I'll make the snapshot change

docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,300 @@
...
+
+**Step 10 — Restart the node → single-database, page-encoded, post-fork.**
+Confirm it resumes and commits new blocks without the `read_storage_page` abort.
+
+### 4.2 Cluster rolling migration
+
+Drive §4.1 across the cluster in phases, one node offline at a time, keeping
+quorum throughout.
+
+- **Phase A — Rolling activate + populate (Steps 1–5), node by node.**
```
pdobacz, line 166: It could be useful to know how long is it estimated the Steps 1-5 would take (on testnet, on mainnet, or as function of the state size?), assuming that `snapshot dump -> snapshot load` take up the majority of time needed to cycle a node.

This gives an idea of how much time to allocate to the entire process before the fork cross timestamp.

EDIT: I see this in §7 now, so a reference from here would suffice. More precise estimates may also help ('allow x minutes per y gb of state on machine z' or similar?)

docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,300 @@
...
+  For each node: stop → activate → snapshot dump → snapshot load → restart →
+  wait for it to rejoin and the cluster's latest block to advance, *then* move to
+  the next node. **This phase must finish on every node before the fork time**
+  (§2). Keep load flowing; the cluster keeps producing blocks throughout.
+
+- **Phase B — Cross the fork.** Once Phase A is complete fleet-wide, let the
+  chain cross the configured fork time. Every node is now dual-writing and flips
+  canonical to page together at the fork.
+
+- **Phase C — Rolling promote + deactivate (Steps 7–10), node by node.** Safe to
```
pdobacz, line 179: Similarly to above comment - some gauge on at which point / at what rate the dual-db operation is expected to lead to degradation of node. Similarly, a reference to §7 works best

docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,300 @@
...
+  including contracts with dense/multi-page storage, returns identical values
+  pre- and post-migration. This exercises the page encoding directly and is where
+  a silent encoding bug would surface.
+- **State-root continuity** — the canonical `state_root` at the last pre-promote
+  block, read back from the promoted page primary, matches what was committed.
+- **Historical queries still resolve** — queries at pre-fork block heights still
+  answer after the page primary takes over.
+- **First post-fork commit succeeds** — the node commits a post-fork block
+  without `read_storage_page is only valid on a page-encoded TrieDb`.
+
```
pdobacz, line 294: Is the node logging/warning/monitoring the fact it is running in dual mode? Assuming it does, maybe it could be documented in here as well, what to expect when a node is ready for the fork-cross and when it is past the cross-fork and should promote and deactivate.

pdobacz (commented):
Great doc! Left some comments with optional suggestions

docs/page_store_migration_runbook.md
```diff
@@ -0,0 +1,300 @@
...
+---
+
+## 5. Verification / success criteria
+
+The migration is correct only if state is byte-identical across the promote.
+Check, per node and at the cluster level:
+
+- **Block production never stalls** beyond your interruption budget; transactions
+  submitted during the migration confirm (allow a small in-flight tolerance).
+- **Account state matches** across the promote — spot-check `eth_getBalance`,
```
Chen-Yifan: I don't know how realistic and representative it is to check specific account balances and storage slots. If all nodes that have been configured dual timeline are able to reach consensus and produce blocks, that means the new state root match and is conanical. 
We can add this: To gain confidence before crossing the fork, node operators can grep for the secondary state root of specific block numbers and cross check with other nodes, here is the log line:
```
        LOG_INFO(
            "block={}, block_id={} state_root primary={} secondary={}",
            block.header.number,
            block_id,
            db.state_root(),
            secondary_db->state_root());
```

docs/page_store_migration_runbook.md
maxkozlovsky: added timing estimates

docs/page_store_migration_runbook.md
maxkozlovsky: added mention of specific log line

docs/page_store_migration_runbook.md
maxkozlovsky: added specific log line to doc

PR #2375, opened 2026-06-23
Simplify monad-triedb callback logic

andr-dev:
This change removes the dynamic dispatch of callback functions in the triedb wrapper C library replacing it with calls to function declarations which are defined in rust.

The singular traverse callback was also split into two different callback functions, simplifying the interface and making it easier to reason about ownership (callback_traverse_value called multiple times with referential ownership, callback_callback_finished called once at the end to destroy values and return result).

The net result is a reduction in lines of code with a simpler ownership model.


rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -35,3 +36,147 @@ pub use self::bindings::{validator_data, validator_set};
...
+    let value = if value_ptr.is_null() {
+        None
+    } else {
+        Some(unsafe { std::slice::from_raw_parts(value_ptr, value_len) }.to_vec())
+    };
+
+    let _ = sender.send(value);
+
+    completed_counter.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+}
```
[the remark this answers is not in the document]
andr-dev: We can add an explicit `drop(concurrency_tracker)` after the `completed_counter.fetch_add(1, ...)` but I don't think there's any defined semantics as to how these values get updated with respect to each other.

rust/crates/monad-triedb/include/ffi.h
```diff
@@ -38,42 +38,30 @@ int triedb_read(
...
     uint8_t const **value, uint64_t block_id);
 
-typedef void (*triedb_async_read_callback_fn)(
-    uint8_t const *value, int length, void *user);
-// calls (*completed) when read is
-// complete. length is -1 if key not
-// found. If >=0, returns length of
-// value. Call triedb_finalize when
-// done with the value.
+typedef struct CallbackContext CallbackContext;
```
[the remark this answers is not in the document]
andr-dev: That's why it's an opaque struct that is only forward declared.

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -35,3 +36,147 @@ pub use self::bindings::{validator_data, validator_set};
...
+    ctx: *const OpaqueCallbackContext,
+    key_ptr: *const u8,
+    key_len: usize,
+    value_ptr: *const u8,
+    value_len: usize,
+) {
+    let AsyncTraverseContext {
+        data,
+        sender: _,
+
```
dhil, line 138: Stray newline?
```suggestion
```

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -35,3 +36,147 @@ pub use self::bindings::{validator_data, validator_set};
...
+        concurrency_tracker: _,
+    } = unsafe { &*(ctx as *const AsyncTraverseContext) };
+
+    let key = unsafe { std::slice::from_raw_parts(key_ptr, key_len) };
+    let value = unsafe { std::slice::from_raw_parts(value_ptr, value_len) };
```
dhil, line 143: I think we should check whether `key_ptr` and `value_ptr` are non-null, because it is UB to pass a null pointer to `from_raw_parts`.

rust/crates/monad-triedb/src/lib.rs
```diff
@@ -337,22 +199,15 @@ impl TriedbHandle {
...
-            data: std::sync::Mutex::new(Default::default()),
-            sender,
-            concurrency_tracker: Arc::new(()),
-        });
+        let ctx = Box::new(AsyncTraverseContext::new(sender, Arc::new(())));
 
         unsafe {
-            let context = Box::into_raw(traverse_context) as *mut std::ffi::c_void;
-            // sync result is already handled by traverse_callback
             let _result = ffi::triedb_traverse(
```
dhil, line 205: Little bit suspicious that we ignore the return value. Though, I suppose the callback is always invoked by `triedb_traverse`.

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -35,3 +36,147 @@ pub use self::bindings::{validator_data, validator_set};
...
+        concurrency_tracker: _,
+    } = ctx;
+
+    let value = if value_ptr.is_null() {
+        None
+    } else {
+        Some(unsafe { std::slice::from_raw_parts(value_ptr, value_len) }.to_vec())
+    };
+
+    let _ = sender.send(value);
```
dhil, line 87: Can `send` ever fail here? If not then I think it is worthwhile to slap a comment at the invocation site here to explain why. If it can, then we ought to handle the error here.

rust/crates/monad-triedb/src/ffi.rs
andr-dev: This is an intentional line to separate the core data/sender and the "extra" fields (concurrency tracker).

rust/crates/monad-triedb/src/ffi.rs
andr-dev: The old code doesn't check whether these ptrs are null because I think they're guaranteed to not be null on the C++ side. This is however another reason we should switch to cxx which lets us pass a rust slice / std::shared_ptr across the FFI boundary cleanly.

rust/crates/monad-triedb/src/lib.rs
andr-dev: It returns a bool which indicates whether the traverse complete. This can also be derived from whether the sender produces a None. We probably should return bool here but this should be changed in a separate PR.

rust/crates/monad-triedb/src/ffi.rs
andr-dev: Send can fail if the other side of the oneshot channel (the Receiver) is dropped before `send` is called. This error however shouldn't be handled because it's not the responsibility of this FFI function + there's nothing to be done anyways, it just means the request was canceled.

rust/crates/monad-triedb/src/ffi.rs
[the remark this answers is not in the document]
dhil: Are we certain that no logic in the codebase depends on the old ordering of effects?

rust/crates/monad-triedb/src/ffi.rs
```diff
@@ -35,3 +36,147 @@ pub use self::bindings::{validator_data, validator_set};
...
+    let _: unsafe extern "C" fn(*mut OpaqueCallbackContext, *const u8, usize) =
+        bindings::monad_rust_triedb_callback_async_read;
+    let _: unsafe extern "C" fn(*mut OpaqueCallbackContext, *const u8, usize) =
+        monad_rust_triedb_callback_async_read;
+};
+
+#[derive(Debug, Clone)]
+pub struct TraverseEntry {
+    pub key: Vec<u8>,
+    pub value: Vec<u8>,
```
dhil, line 102: Should these morally be `Box<[u8]>`? I think you previously suggested to me that we ought to use `Box<[u8]>` for fixed byte strings.

rust/crates/monad-triedb/include/ffi.h
```diff
@@ -38,42 +38,30 @@ int triedb_read(
...
+
+void monad_rust_triedb_callback_async_read(
+    CallbackContext *ctx, uint8_t const *value, size_t value_len);
+void monad_rust_triedb_callback_traverse_value(
+    CallbackContext const *ctx, uint8_t const *key, size_t key_len,
+    uint8_t const *value, size_t value_len);
+void monad_rust_triedb_callback_traverse_finished(
+    CallbackContext *ctx, bool completed);
```
dhil, line 49: I think these context pointers can be made `const`, i.e. `CallContext *const ctx`.

PR #2382, opened 2026-06-24
mip8: drop unused base_sload_cost and dedup compute_page_key

brett-monad:
Two cleanups to the MIP-8 gas-scheduling code (both introduced in ee641451c):

- traits: remove MonadTraits::base_sload_cost(), which has no callers anywhere in the tree, unlike its siblings base_sstore_cost() and page_write_cost().

- page_tracker: now that the storage_page PR has merged, drop the temporary detail::compute_page_key (with its own local PAGE_KEY_SHIFT) and call the canonical monad::compute_page_key from storage_page.hpp, removing the duplicated big-endian >> 7 page-key logic. Also drops the now-unused core/int.hpp include.

### DIFF
--- a/category/execution/ethereum/state3/page_tracker.hpp
+++ b/category/execution/ethereum/state3/page_tracker.hpp
@@ -17,7 +17,7 @@
 
 #include <category/core/bytes.hpp>
 #include <category/core/config.hpp>
-#include <category/core/int.hpp>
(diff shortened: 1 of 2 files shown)


brett-monad:
@claude review this PR

Baltoli (approved):

Chen-Yifan (approved):

PR #2386, opened 2026-06-25
[page_store] Make TrieRODb storage reads encoding-aware

Chen-Yifan:
## What

`monad_executor` reads state through `TrieRODb`, which always decoded
slot-format storage leaves. Against a page-encoded (post-mip8) db this returns
wrong/empty values for any `SLOAD`. This PR makes `TrieRODb::read_storage` follow
the db's actual encoding.

## Changes

- **`TrieRODb` encoding-aware**: `read_storage` now looks up the `page_key`
  leaf and extracts the slot at its offset when the db is page-encoded,
  mirroring the writable `TrieDb`. Slot-encoded path is unchanged.
- **`mpt::RODb::state_machine_type()`**: reads the (primary) timeline's
  `state_machine_kind` from `db_metadata`, so `TrieRODb` can detect its
  encoding at construction.
- **`Db::is_page_encoded()` is now pure virtual**, forcing every `Db` to
  declare its encoding. `TrieRODb` overrides it; `PartialTrieDb` returns
  `false` (page support left as a TODO).

## Testing

`monad_executor_test`: the storage-reading tests (`prestate_override_state`,
`eth_simulate_v1_time_travel`) now run as `TYPED_TEST` over both slot- and
page-encoded dbs via `EthCallEncodingFixture`, exercising the page read path
through both the eth_call and eth_simulate executor paths. Storage-agnostic
tests stay slot-only.

### DIFF
--- a/category/execution/ethereum/db/db.hpp
+++ b/category/execution/ethereum/db/db.hpp
@@ -26,6 +26,7 @@
 #include <category/execution/ethereum/core/withdrawal.hpp>
 #include <category/execution/ethereum/state2/state_deltas.hpp>
 #include <category/execution/ethereum/trace/call_frame.hpp>
+#include <category/execution/monad/db/storage_page.hpp>
 #include <category/vm/vm.hpp>
 
 #include <cstdint>
@@ -36,15 +37,10 @@
 MONAD_NAMESPACE_BEGIN
 
 class CommitBuilder;
-struct storage_page_t;
 
 struct Db
 {
-    virtual bool is_page_encoded() const
-    {
-        return false;
-    }
-
+    virtual bool is_page_encoded() const = 0;
     virtual std::optional<Account> read_account(Address const &) = 0;
 
     virtual bytes32_t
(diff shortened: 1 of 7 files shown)


maxkozlovsky (dismissed):

marcograss:
on the stateless/witness path, PartialTrieDb still reports its storage as slot-encoded behind a [TODO](https://github.com/category-labs/monad/blob/bcc96ab5e6c5/category/execution/ethereum/db/partial_trie_db.hpp#L161-L164); is leaving it page-unaware intentional for now?
Once mip-8 turns page encoding on, won't a witness SLOAD through there decode page bytes as a slot and quietly hand back a wrong value (or abort)

Chen-Yifan:
> on the stateless/witness path, PartialTrieDb still reports its storage as slot-encoded behind a [TODO](https://github.com/category-labs/monad/blob/bcc96ab5e6c5/category/execution/ethereum/db/partial_trie_db.hpp#L161-L164); is leaving it page-unaware intentional for now? Once mip-8 turns page encoding on, won't a witness SLOAD through there decode page bytes as a slot and quietly hand back a wrong value (or abort)

This is intentional. The zkvm part is currently in prototype stage targeting eth blocks, and it doesn't have a production impact. Also partialTrieDb doesn't decode data from a on disk db, it is generated from a witness.

maxkozlovsky (approved):

kkuehlz (approved):

marcograss:
approved

PR #2389, opened 2026-06-25
updating to quill 12.0.0

mkolosick:
This moves towards being able to use clang-21.

### DIFF
--- a/category/core/log.cpp
+++ b/category/core/log.cpp
@@ -16,47 +16,56 @@
 #include <category/core/config.hpp>
 #include <category/core/log.hpp>
 
+#include <quill/Backend.h>
+#include <quill/Frontend.h>
+#include <quill/sinks/ConsoleSink.h>
+#include <quill/sinks/FileSink.h>
+#include <quill/sinks/NullSink.h>
+
 #include <filesystem>
+#include <utility>
 
-MONAD_ANONYMOUS_NAMESPACE_BEGIN
-
-constexpr char default_pattern[] =
-    "%(time) [%(thread_id)] %(file_name):%(line_number) "
-    "LOG_%(log_level)\t%(message)";
-constexpr char default_time_format[] = "%Y-%m-%d %H:%M:%S.%Qns";
-
-MONAD_ANONYMOUS_NAMESPACE_END
+quill::Logger *monad_root_logger;
 
 MONAD_NAMESPACE_BEGIN
 
 void init_root_logger(quill::LogLevel const level)
 {
-    auto const stdout_handler = quill::stdout_handler();
-    stdout_handler->set_pattern(
-        default_pattern, default_time_format, quill::Timezone::GmtTime);
-    quill::Config cfg;
-    cfg.default_handlers.emplace_back(stdout_handler);
-    quill::configure(cfg);
-    quill::start(true);
-    quill::get_root_logger()->set_log_level(level);
+    monad_root_logger = quill::Frontend::create_or_get_logger(
+        "root",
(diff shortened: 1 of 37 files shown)


category/core/log_ffi.cpp
```diff
@@ -119,19 +126,29 @@ class LogCallbackHandler : public quill::Handler
...
         monad_log const log = {
-            .syslog_level = to_syslog_level(log_event.log_level()),
-            .message = log_message.data(),
-            .message_len = log_message.size(),
+            .syslog_level = to_syslog_level(log_level),
+            .message = log_statement.data(),
+            .message_len = log_statement.size(),
         };
         write_fn_(&log, user_);
     }
```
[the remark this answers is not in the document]
mkolosick: @kjcamann

kjcamann (dismissed):
LGTM

category/core/log.hpp
```diff
@@ -15,51 +15,56 @@
...
-//
-// Today this re-exports `<quill/Quill.h>`, which transitively provides
-// every quill name our code uses (LogLevel, Logger, Handler, FileHandler,
-// the bundled `fmtquill` formatter, etc.). The `fmt` alias points at
-// `fmtquill` rather than `fmtquill::v10` so it survives the inline-namespace
-// tag bump (v10 → v12) in the upcoming quill v11 upgrade — call sites that
-// say `fmt::format(...)` keep working without source changes.
-//
-// The upcoming quill v11 upgrade splits the umbrella across
-// Backend/Frontend/LogMacros and removes the QUILL_ROOT_LOGGER_ONLY mode;
```
Baltoli, line 32: https://github.com/category-labs/monad/blob/68cb02c00b9aca5f1975c3fc1ee531cb2150e860/cmake/compile_options.cmake#L30

Can we remove this compile option now?

category/core/log.cpp
```diff
@@ -16,47 +16,54 @@
...
+    monad_root_logger = quill::Frontend::create_or_get_logger(
+        "root",
+        quill::Frontend::create_or_get_sink<quill::NullSink>("null_sink"));
+    quill::Backend::start();
 }
 
 void flush_logger()
 {
-    quill::flush();
+    monad_root_logger->flush_log();
```
Baltoli: Is it possible to get here from a signal shutdown or error path before `monad_root_logger` is non-null? Is it worth the hit of a null check here?

category/core/log.cpp
```diff
@@ -16,47 +16,54 @@
...
+#include <filesystem>
+#include <utility>
 
-MONAD_ANONYMOUS_NAMESPACE_END
+quill::Logger *monad_root_logger;
 
 MONAD_NAMESPACE_BEGIN
 
 void init_root_logger(quill::LogLevel const level)
 {
```
Baltoli, line 33: Is it worth an assertion here to check we don't re-initialise? Will anything bad happen if we do re-initialise?

category/core/log.hpp
mkolosick: Good catch yes we can.

category/core/log.cpp
mkolosick: `create_or_get_logger` returns the existing logger if it's already created and `quill::Backend::start()` is guarded to only run once so this function is safe even if called multiple times.

category/core/log.cpp
mkolosick: It does not appear possible currently, but this is never called on a hot path so we should just do the null-check.

Baltoli (approved):

kjcamann (approved):

PR #2390, opened 2026-06-25
[statesync] Re-export monad_chain_config enum constants from the FFI

maxkozlovsky:
Re-exports the bindgen-generated `monad_chain_config_CHAIN_CONFIG_*` constants (the C `enum monad_chain_config` from `category/execution/ethereum/chain/chain_config.h`) from the `monad-statesync` `ffi` module.

These values were already generated by bindgen but lived in the private `bindings` module, so Rust consumers could only see the `monad_chain_config` *type*, not its discriminants. Exposing them lets the bft side map a chain to the statesync client (`monad_statesync_client_context_create`) using named constants sourced directly from the C header, instead of duplicating raw `0..4` discriminants.

Execution half of a 2-PR change; the bft consumer is category-labs/monad-bft#3152.

### DIFF
--- a/rust/crates/monad-statesync/src/ffi.rs
+++ b/rust/crates/monad-statesync/src/ffi.rs
@@ -14,10 +14,12 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 pub use self::bindings::{
-    monad_chain_config, monad_statesync_client, monad_statesync_client_context,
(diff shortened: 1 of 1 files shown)


Chen-Yifan (approved):
Approved. Though CHAIN_CONFIG_ETHEREUM_MAINNET and CHAIN_CONFIG_HIVE_NET are not necessary, statesync only works for monad chains and it will abort on them

kkuehlz (approved):

PR #2393, opened 2026-06-26
triedb: add page-aware storage FFI (is_page_encoded + decode_storage_page_slot)

maxkozlovsky:
## What

Adds two read-only FFI entry points to the `monad-triedb` crate so a Rust caller can serve `eth_getStorageAt` against a **page-encoded** timeline:

- `triedb_is_page_encoded(db)` — true when the primary timeline is the Monad (page-encoded) state machine: storage is keyed by `keccak(page_key)` (`page_key = slot >> 7`) and leaves are encoded pages.
- `triedb_decode_storage_page_slot(leaf, offset)` — decode an encoded page leaf and return the 32-byte value at the slot's low-7-bit `offset`. The page wire-format decode stays in C++ (single source of truth, reuses `decode_storage_page`).

Plus the thin Rust wrappers (`ffi.rs`, `lib.rs`).

## Why

After the dual-db migration promotes the page timeline, the primary is page-encoded. A slot-only reader looks up `keccak(slot)` and finds nothing → `eth_getStorageAt` returns `0` for all real storage (a silent, customer-facing correctness hole). This FFI lets the bft RPC read page-encoded storage correctly. Consumer: companion PR in `monad-bft` (`triedb_env.rs` page-aware `get_storage_at`).

## Discussion — is this the right seam?

The split here is: **the Rust caller computes `page_key = slot >> 7` and the offset**, looks up `keccak(page_key)`, then calls this FFI to **decode only**. The main alternative is a single C++ entry point `triedb_read_storage(addr, slot)` that owns the whole page-aware read.

Trade-off as I see it:
- *Current split:* keeps the bft async/LRU cache keyed by `page_key`, so the 128 slots in a page share one disk read + cache entry; only the format-specific decode is delegated to C++ (format lives in one place). Downside: page-key derivation is duplicated outside `storage_page.hpp`.
- *C++ owns the whole read:* zero duplication, but the bft async poller + cache (keyed by nibble path) would need to either lose page-level cache sharing or be taught about pages.

Flagging for review before this lands more permanently. Happy to flip to the C++-owned approach if that's preferred.

### DIFF
--- a/rust/crates/monad-triedb/include/ffi.h
+++ b/rust/crates/monad-triedb/include/ffi.h
@@ -38,6 +38,28 @@ int triedb_read(
     TriedbRoInner *, uint8_t const *key, uint8_t key_len_nibbles,
     uint8_t const **value, uint64_t block_id);
 
+// true if the primary timeline is page-encoded (Monad state machine), in which
+// case storage is keyed by keccak(page_key) (page_key = slot >> 7) and the leaf
+// is an encoded page; otherwise storage is slot-encoded.
+bool triedb_is_page_encoded(TriedbRoInner *);
+
+// Compute the storage page key for a 32-byte slot key on a page-encoded db:
+// page_key = slot >> 7. Writes the 32-byte big-endian page key (the key the
+// storage trie is looked up by) to out_page_key.
+void triedb_compute_page_key(uint8_t const *slot_key, uint8_t *out_page_key);
+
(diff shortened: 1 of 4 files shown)


maxkozlovsky:
Consumer / companion PR (the eth_getStorageAt caller): category-labs/monad-bft#3153

Chen-Yifan (approved):

kkuehlz (approved):

PR #2399, opened 2026-07-02
Switch to use ocaml spec for vm fuzzer

andreaslyn:


test/vm/fuzzer/generator.hpp
```diff
@@ -21,6 +21,10 @@
 #include <category/core/cases.hpp>
 #include <category/core/int.hpp>
 #include <category/core/runtime/uint256.hpp>
+
+#include <category/execution/monad/reserve_balance/reserve_balance_contract.hpp>
+#include <category/execution/monad/staking/util/constants.hpp>
```
Baltoli, line 26: It doesn't really matter here (and I won't block the PR on it), but this is a circular dependency `vm -> execution -> vm`. We should maybe pull the fuzzer out into its own target that depends on both `vm` and `execution`?

category/vm/fuzzing/generator/generator.hpp
```diff
@@ -658,90 +666,43 @@ namespace monad::vm::fuzzing
...
-            else {
-                MONAD_ABORT();
-            }
-        }();
+        static_assert(traits::monad_rev() >= MONAD_EIGHT);
+        std::vector<Address> pre_compiles;
+        for (uint64_t i = 1; i <= 17; ++i) {
+            pre_compiles.emplace_back(i);
+        }
+        pre_compiles.emplace_back(256);
```
Baltoli: Tiny nitpick: this is quoted everywhere in writing as `0x1001` rather than `256`.

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -437,42 +426,45 @@ static arguments parse_args(int const argc, char **const argv)
...
 
     TransitionState monad_tstate{monad_state};
     auto const monad_result =
-        transition(monad_tstate, msg, rev, block_hash_buffer, block_header);
+        transition<traits>(monad_tstate, msg, block_hash_buffer, block_header);
 
-    assert_equal(evmone_result, monad_result, strict_out_of_gas);
+    // Ideally we will verify out-of-gas error strictly when testing the
+    // interpreter. However the monad-ml evm does not agree on specific error
+    // codes at present.
```
Baltoli, line 449: The spec doesn't return OOG error codes, because that code isn't in the YP / Monad spec (it's an RPC detail) - correct?

test/vm/monadml_evm/monadml_evm.hpp
```diff
@@ -0,0 +1,21 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <evmc/evmc.h>
+
+extern "C" struct evmc_vm *evmc_create_monadml_evm_default() noexcept;
```
Baltoli: https://github.com/category-labs/monad-execution-specs/blob/9cc25e7be005f18da57a6316129ed024806e0091/bindings/vm/libmonadml_evm_template.ml#L205 wrong binding name?

test/vm/fuzzer/fuzzer.cpp
```diff
@@ -407,22 +404,14 @@ static arguments parse_args(int const argc, char **const argv)
...
-        {"LONDON", MONAD_ETH_LONDON},
-        {"PARIS", MONAD_ETH_PARIS},
-        {"SHANGHAI", MONAD_ETH_SHANGHAI},
-        {"CANCUN", MONAD_ETH_CANCUN},
-        {"PRAGUE", MONAD_ETH_PRAGUE},
-        {"OSAKA", MONAD_ETH_OSAKA},
-        {"AMSTERDAM", MONAD_ETH_AMSTERDAM},
-        {"LATEST", MONAD_ETH_LATEST_STABLE_REVISION}};
+    auto const rev_map =
+        std::map<std::string, monad_revision>{{"MONAD_EIGHT", MONAD_EIGHT}};
```
Baltoli, line 408: Perhaps out of scope here, but we should maybe add an assertion / FFI method to keep this in sync with the version used by the spec. If (say) we accidentally bumped the spec version but not the version here, we'd silently swallow the divergence

test/vm/monadml_evm/CMakeLists.txt
Baltoli, line 1: Broad comment that it's worth going through and asking Claude to update its documentation to reflect this new build info / fuzzing config

Baltoli (changes requested):
Some stale references?
```
     test/vm/fuzzer/assertions.cpp:85:        evmc::Result const &evmone_result, evmc::Result const &compiler_result,
     test/vm/fuzzer/assertions.cpp:89:            evmone_result.create_address.bytes,
     test/vm/fuzzer/assertions.cpp:92:        MONAD_ASSERT(evmone_result.gas_left == compiler_result.gas_left);
     test/vm/fuzzer/assertions.cpp:93:        MONAD_ASSERT(evmone_result.gas_refund == compiler_result.gas_refund);
     test/vm/fuzzer/assertions.cpp:96:            std::span(evmone_result.output_data, evmone_result.output_size),
     test/vm/fuzzer/assertions.cpp:100:        switch (evmone_result.status_code) {
     test/vm/fuzzer/assertions.cpp:104:                evmone_result.status_code == compiler_result.status_code);
     test/vm/fuzzer/assertions.hpp:33:        evmc::Result const &evmone_result, evmc::Result const &compiler_result,
```

test/vm/fuzzer/generator.hpp
andreaslyn: Good point. I was thinking about it as the generator was located in the test/vm module, which is only true for the fuzzer driver files. I am included to move the fuzzer library to test/vm before merging this to avoid the cycle. What do you think?

test/vm/fuzzer/fuzzer.cpp
andreaslyn: The spec does return out of gas error some times, but does not agree exactly with the interpreter on the choice of error codes. I did start work on modifying the ocaml spec to agree with interpreter here, but realized that it would require a relatively large amount of work, so i decided to proceed with this PR without matching exact out of gas error code

test/vm/monadml_evm/monadml_evm.hpp
andreaslyn: Yes this has been renamed at some point i think

test/vm/fuzzer/fuzzer.cpp
andreaslyn: If the version is not the same, i would expect the fuzzer to catch the difference between spec and monad.

test/vm/monadml_evm/CMakeLists.txt
andreaslyn: Let us do that in a separate PR

test/vm/fuzzer/generator.hpp
Baltoli: Yes, I think we should move all the fuzzing code to `test/vm`

test/vm/fuzzer/fuzzer.cpp
Baltoli: Sounds reasonable, thanks - we can open an issue to align them if possible (cc @m-alvarez)

test/vm/monadml_evm/CMakeLists.txt
Baltoli: sure

test/vm/fuzzer/fuzzer.cpp
Baltoli: Seems better to fail fast, but it's low priority so no need to fix it here

Baltoli (dismissed):

docker/Dockerfile
```diff
@@ -32,6 +32,24 @@ RUN rustup toolchain install 1.91.1 --profile minimal --component clippy
...
+    apt-get install -y opam && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    rm -rf /opt
+
+ARG OCAML_VERSION=5.4.1
+
+ENV OPAMROOT=/root/.opam
+ENV OPAMSWITCH="$OCAML_VERSION"
+ENV PATH="/root/.opam/$OCAML_VERSION/bin:$PATH"
```
dhil: The correctness of this path depends on an opam implementation detail. There is a more robust way to code this by invoking `opam exec`. See my other comment below.

test/vm/monadml_evm/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+find_program(DUNE dune REQUIRED)
```
dhil: Nit: I suppose it would be more robust to find opam here and then do `opam exec -- dune build --profile release` down below. This way we don't have to depend on the file structure of opam.

dhil (dismissed):
LGTM. One nit comment about using `opam exec` rather than coding the opam path into `$PATH`.

goodlyrottenapple (dismissed):

test/vm/monadml_evm/CMakeLists.txt
andreaslyn: Fixed, thanks

Baltoli (dismissed):

goodlyrottenapple (dismissed):

dhil (dismissed):

Baltoli (dismissed):

dhil (approved):

goodlyrottenapple (approved):

PR #2400, opened 2026-07-02
mpt: max version should return INVALID_BLOCK_NUM when rewind below its valid range

Chen-Yifan:
## Problem

Rewinding to a version before a ring's first valid root (e.g. a secondary
timeline activated after the rewind target) left `next_version` at
`target + 1`. The ring then reported a `max_version()` whose slot held no
root: `get_latest_version()` named a version that could not be loaded, and
readers of the latest root got null.

## Change

* `rewind_to_version` on the root offsets ring now resets the ring to the
  empty state when the target predates its entire valid range, so
  `max_version()` reports `INVALID_BLOCK_NUM`, identical to a freshly
  activated secondary. The next commit reseeds it through the existing
  fast forward path.
* Rewinding into a gap between valid roots is refused with an assert.
  Together these restore the invariant that a non-empty ring always has a
  valid root at `max_version()`, and the `sec_root != INVALID_OFFSET`
  assert in `UpdateAux::rewind_to_version` stays as a corruption tripwire.
* Refactor: `reset_all(version)` is split into `reset_all()` (empty state)
  and `advance_next_version(version)` (write cursor positioning used by
  `fast_forward_next_version`). No behavior change.

### DIFF
--- a/category/mpt/db_metadata_context.cpp
+++ b/category/mpt/db_metadata_context.cpp
@@ -707,7 +707,7 @@ void DbMetadataContext::update_root_offset(
         ro.assign(i, root_offset);
         if (root_offset == INVALID_OFFSET && i == db_history_max_version(tid) &&
             i == db_history_min_valid_version(tid)) {
-            ro.reset_all(0);
+            ro.reset_all();
             MONAD_ASSERT(ro.max_version() == INVALID_BLOCK_NUM);
         }
     };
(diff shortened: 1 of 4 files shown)


maxkozlovsky (dismissed):

Chen-Yifan:
@claude review the updated PR please

PR #2407, opened 2026-07-07
EIP-8024 (stack manipulation opcodes)

brett-monad:
Implement EIP-8024: backward-compatible DUPN/SWAPN/EXCHANGE

EIP-8024 adds three stack-manipulation opcodes that take an immediate operand selecting which stack item(s) to act on, lifting the fixed-depth limit of the legacy DUP1–16 / SWAP1–16 family:

- DUPN (0xE6) — duplicate a stack item selected by the immediate
- SWAPN (0xE7) — swap the top item with one selected by the immediate
- EXCHANGE (0xE8) — swap two stack items selected by the immediate

It is backward-compatible because the immediate byte is encoded so it can never be JUMPDEST (0x5B) or a PUSH byte (0x60–0x7F): legacy jumpdest analysis is unaffected, and any disallowed immediate behaves as INVALID.

This PR implements all three opcodes across both execution engines (interpreter and x86 JIT) and the evm-as assembler, with unit tests. EIP-8024 is still a draft under peer review, so it is gated behind an `eip_8024_active()` predicate.

### Spec tests

No change to the Amsterdam exclusion list, and the Amsterdam suite is disabled
at this commit. Nothing needs excluding: the pinned bundle is generated from a
spec in which `0xE6`/`0xE7`/`0xE8` are active, so its
`frontier/opcodes/all_opcodes` fixture expects them to be defined.

Unit tests are the gate on this PR.


brett-monad:
@claude review this PR

brett-monad:
@claude review this PR

category/vm/evm/explicit_traits.hpp
```diff
@@ -145,7 +148,8 @@
     template void id<&f<::monad::EvmTraits<MONAD_ETH_CANCUN>>>();              \
     template void id<&f<::monad::EvmTraits<MONAD_ETH_PRAGUE>>>();              \
     template void id<&f<::monad::EvmTraits<MONAD_ETH_OSAKA>>>();               \
-    template void id<&f<::monad::EvmTraits<MONAD_ETH_AMSTERDAM>>>();
+    template void id<&f<::monad::EvmTraits<MONAD_ETH_AMSTERDAM>>>();           \
+    template void id<&f<::monad::EvmTraits<MONAD_ETH_EXPERIMENTAL>>>();
```
Baltoli: Why is this dragging in `EXPERIMENTAL`?

category/vm/evm/opcodes.hpp
```diff
@@ -241,6 +242,11 @@ namespace monad::vm::compiler
         LOG2 = 0xA2,
         LOG3 = 0xA3,
         LOG4 = 0xA4,
+        // EIP-8024 backward-compatible stack manipulation (gated on
+        // eip_8024_active()). Same opcode bytes as EOF EIP-663.
```
Baltoli: EOF isn't going to get implemented, I don't think we need a comment re. 663

category/vm/evm/opcodes.hpp
```diff
@@ -290,6 +296,21 @@ namespace monad::vm::compiler
...
+    // the static min_stack/stack_increase below exist ONLY to produce the
+    // correct net stack delta (stack_increase - min_stack): +1 for DUPN, 0 for
+    // SWAPN/EXCHANGE. Nothing may rely on their static min_stack.
+    consteval inline void
+    add_eip8024_opcodes(std::array<OpCodeInfo, 256> &table)
+    {
+        add_opcode(DUPN, table, {"DUPN", 1, 0, 1, false, 3, 0});
+        add_opcode(SWAPN, table, {"SWAPN", 1, 0, 0, false, 3, 0});
+        add_opcode(EXCHANGE, table, {"EXCHANGE", 1, 0, 0, false, 3, 0});
+    }
```
Baltoli: Stylistic nitpick - as above, I don't think we need to pull in `EXPERIMENTAL` so this function should just get folded into the `AMSTERDAM` table generator

category/vm/interpreter/instruction_table.hpp
```diff
@@ -99,6 +122,12 @@ namespace monad::vm::interpreter
             return (traits::evm_rev() >= first) ? impl : invalid;
         };
 
+        // Gate on a feature predicate rather than a bare revision (used by
+        // EIP-8024, which is staged behind eip_8024_active()).
+        constexpr auto when = [](bool const active, InstrEval impl) {
+            return active ? impl : invalid;
+        };
```
Baltoli: Nice

Baltoli: Indeed, probably a refactoring that's worth doing more generally outwith the scope of this PR

category/vm/compiler/ir/x86/virtual_stack.cpp
```diff
@@ -485,6 +485,35 @@ namespace monad::vm::compiler::native
         e = std::move(t);
     }
 
+    void Stack::exchange(int32_t const index_a, int32_t const index_b)
```
andreaslyn, line 472: I think that swap can now be implemented simply by `exchange(top_index_, swap_index)`

category/vm/evm/opcodes.hpp
```diff
@@ -797,6 +830,88 @@ namespace monad::vm::compiler
...
+
+    constexpr bool eip8024_pair_disallowed(uint8_t const x)
+    {
+        return x > 81 && x < 128;
+    }
+
+    constexpr uint32_t eip8024_decode_single(uint8_t const x)
+    {
+        MONAD_DEBUG_ASSERT(!eip8024_single_disallowed(x));
+        return (static_cast<uint32_t>(x) + 145u) % 256u;
```
andreaslyn: I would not rely on compiler optimizations here

```suggestion
        return (static_cast<uint32_t>(x) + 145u) & 255u;
```

category/vm/evm/opcodes.hpp
```diff
@@ -797,6 +830,88 @@ namespace monad::vm::compiler
...
+    }
+
+    constexpr std::pair<uint8_t, uint8_t> eip8024_decode_pair(uint8_t const x)
+    {
+        MONAD_DEBUG_ASSERT(!eip8024_pair_disallowed(x));
+        auto const k = static_cast<uint8_t>(x ^ 143);
+        auto const q = static_cast<uint8_t>(k / 16);
+        auto const r = static_cast<uint8_t>(k % 16);
```
andreaslyn:
```suggestion
        auto const q = static_cast<uint8_t>(k >> 4);
        auto const r = static_cast<uint8_t>(k & 15);
```

category/vm/evm/opcodes.hpp
```diff
@@ -797,6 +830,88 @@ namespace monad::vm::compiler
...
+
+    constexpr bool eip8024_pair_operand_valid(uint8_t const n, uint8_t const m)
+    {
+        return n >= 1 && n < m && (n + m) <= 30;
+    }
+
+    constexpr uint8_t eip8024_encode_single(uint32_t const n)
+    {
+        MONAD_DEBUG_ASSERT(eip8024_single_operand_valid(n));
+        return static_cast<uint8_t>((n + 111u) % 256u);
```
andreaslyn:
```suggestion
        return static_cast<uint8_t>((n + 111u) & 255u);
```

category/vm/interpreter/instruction_table.hpp
```diff
@@ -1488,6 +1517,115 @@ namespace monad::vm::interpreter
...
+        if (MONAD_UNLIKELY(gas_remaining < 0)) {
+            ctx.exit(OutOfGas);
+        }
+
+        uint8_t const imm = instr_ptr[1];
+        if (MONAD_UNLIKELY(compiler::eip8024_pair_disallowed(imm))) {
+            ctx.exit(Error);
+        }
+
+        auto const [n_val, m_val] = compiler::eip8024_decode_pair(imm);
```
andreaslyn: The original architecture was to have compiler functionality depend on runtime/interpreter and not the other way around.

andreaslyn (commented):
I suppose we will want some emitter unit tests for the new instructions.

category/vm/interpreter/instruction_table.hpp
andreaslyn: Let us leave this as is. I was confused. Having these in evm is fine.

category/vm/compiler/ir/basic_blocks.hpp
```diff
@@ -389,6 +389,60 @@ namespace monad::vm::compiler::basic_blocks
...
+                else if (opcode == SWAPN) {
+                    auto const n = eip8024_decode_single(imm);
+                    min_stack = static_cast<uint8_t>(n + 1);
+                    stack_increase = static_cast<uint8_t>(n + 1);
+                }
+                else {
+                    auto const m = eip8024_decode_pair(imm).second;
+                    min_stack = static_cast<uint8_t>(m + 1);
+                    stack_increase = static_cast<uint8_t>(m + 1);
+                }
```
Baltoli: Let's pull this out into a helper function rather than having it inline here - saves mutating local variables etc

category/vm/compiler/ir/basic_blocks.hpp
```diff
@@ -389,6 +389,60 @@ namespace monad::vm::compiler::basic_blocks
...
+        // decoding gives the operand-dependent stack effect. Unlike the generic
+        // path below, compute min_stack/stack_increase from the decoded operand
+        // and stash the raw immediate in the index field (decoded again in
+        // codegen). A disallowed immediate behaves as INVALID; crucially we
+        // return WITHOUT consuming that byte, so the outer scan can still
+        // discover a 0x5B there as a JUMPDEST -- matching the interpreter and
+        // EIP-8024's backward-compatibility guarantee.
+        if constexpr (traits::eip_8024_active()) {
+            if (is_eip8024_opcode(opcode)) {
+                uint8_t imm = 0;
```
Baltoli: In fact, the entire body of this `if (is_eip8024` block should maybe just get pulled out into a `decode_eip8024` function? Keeps the scanning logic simple

category/vm/compiler/ir/instruction.hpp
```diff
@@ -102,6 +102,10 @@ namespace monad::vm::compiler
         Dup = 0x80,
         Swap = 0x90,
         Log = 0xA0,
+        // EIP-8024 (gated on eip_8024_active()); same bytes as EOF EIP-663.
```
Baltoli: No need for the comment; not consistent with the rest of the table here

category/vm/compiler/ir/x86.cpp
```diff
@@ -262,6 +263,25 @@ namespace
         case Swap:
             emit.swap(instr.index());
             break;
+        // For DupN/SwapN/Exchange, index() holds the raw EIP-8024 immediate
+        // byte, decoded here to the operand. The decoders assert (in debug) the
+        // byte is not disallowed; that precondition is guaranteed upstream by
+        // BasicBlocksIR::scan_from, which returns InvalidInstruction for a
+        // disallowed immediate and therefore never emits one of these
+        // instructions with a disallowed index.
```
Baltoli: I don't think this comment is useful. The same pattern of decoder-enforced preconditions isn't *new* here, we do the same in regular `DUP` etc.

category/vm/evm/opcodes.hpp
```diff
@@ -241,6 +242,11 @@ namespace monad::vm::compiler
         LOG2 = 0xA2,
         LOG3 = 0xA3,
         LOG4 = 0xA4,
+        // EIP-8024 backward-compatible stack manipulation (gated on
+        // eip_8024_active()).
```
Baltoli: Again not a useful comment - it's out of sync with the rest of the table, and the information in the enum isn't actually gated on that feature flag - only consumers of it are.

category/vm/evm/opcodes.hpp
Baltoli: Will need rebasing on top of Ryan's refactor of this table.

category/vm/evm/traits.hpp
```diff
@@ -150,6 +152,20 @@ namespace monad
...
 
+        // EIP-8024: backward-compatible DUPN/SWAPN/EXCHANGE. Draft (under peer
+        // review), so it is staged on AMSTERDAM, the next fork above
+        // LATEST_SUPPORTED_EVM_FORK. AMSTERDAM is already wired into the
+        // dispatch machinery (explicit instantiations, the runtime switch, and
+        // the opcode/storage tables), so the draft rides that existing
+        // scaffolding instead of needing a revision added just to carry it. No
+        // MonadTraits maps to AMSTERDAM (evm_rev() caps at OSAKA) and it is
+        // excluded from evmone/fuzzing, so the feature is inert outside
+        // explicit EvmTraits<AMSTERDAM> unit tests.
```
Baltoli: Another comment with no real value - there's a lot of code in this file with well-established norms for reading and writing it, we don't need max-verbosity comments for every new trait function

category/vm/evm/traits.hpp
```diff
@@ -294,6 +310,15 @@ namespace monad
             return Rev >= MONAD_FOUR;
         }
 
+        // EIP-8024: backward-compatible DUPN/SWAPN/EXCHANGE. Off on-chain for
+        // now (draft, under peer review). SHIP on the next Monad fork by
+        // flipping this to `return Rev >= MONAD_NEXT;` AND adding
+        // add_eip8024_opcodes(table) to make_opcode_table<MonadTraits<...>>.
```
Baltoli: Again

category/vm/interpreter/instruction_table.hpp
```diff
@@ -85,6 +85,29 @@
     }                                                                          \
     while (false);
 
+// EIP-8024 opcodes (DUPN/SWAPN/EXCHANGE) carry a single immediate byte, so
+// advance the instruction pointer by two. Delta is the net stack change from
+// the opcode table (+1 for DUPN, 0 for SWAPN/EXCHANGE).
```
Baltoli: This is actually a very useful comment

category/vm/interpreter/instruction_table.hpp
Baltoli: Again will get rebased out on top of Ryan's changes

category/vm/interpreter/instruction_table.hpp
```diff
@@ -339,22 +368,22 @@ namespace monad::vm::interpreter
...
+            when(traits::eip_8024_active(), dupn<traits>), // 0xE6,
+            when(traits::eip_8024_active(), swapn<traits>), // 0xE7,
+            when(traits::eip_8024_active(), exchange<traits>), // 0xE8,
+            invalid, // 0xE9,
+            invalid, // 0xEA,
+            invalid, // 0xEB,
+            invalid, // 0xEC,
+            invalid, // 0xED,
+            invalid, // 0xEE,
+            invalid, // 0xEF,
```
Baltoli, line 348: Nice

category/vm/interpreter/instruction_table.hpp
```diff
@@ -1488,6 +1517,115 @@ namespace monad::vm::interpreter
         MONAD_VM_NEXT(SWAP1 + (N - 1));
     }
 
+    // EIP-8024: DUPN/SWAPN/EXCHANGE. Unlike DUP*/SWAP*, the operand is a
```
Baltoli: The comment here is useful, but let's pull the dynamic stack checking logic out into a function rather than repeating it

category/vm/interpreter/instructions_fwd.hpp
```diff
@@ -378,6 +378,22 @@ namespace monad::vm::interpreter
         runtime::Context &, Intercode const &, uint256_t const *, uint256_t *,
         int64_t, uint8_t const *);
 
+    // EIP-8024 (gated on traits::eip_8024_active()).
```
Baltoli: Again not super useful

category/vm/interpreter/intercode.cpp
```diff
@@ -57,6 +57,15 @@ namespace monad::vm::interpreter
...
 
+        // Only PUSH data is skipped. The EIP-8024 opcodes (DUPN/SWAPN/EXCHANGE)
+        // are intentionally NOT special-cased here: their immediate byte is
+        // encoded so it can never be 0x5B (JUMPDEST) or a PUSH byte, so an
+        // unmodified legacy scan that treats them as 1-byte opcodes and
+        // inspects the following byte as an opcode yields exactly the correct
+        // valid-JUMPDEST set. This is EIP-8024's backward-compatibility
+        // guarantee; teaching this scan to skip their immediate would instead
+        // hide a legitimate JUMPDEST that follows an instruction with a
+        // disallowed immediate.
```
Baltoli: I think the substance of this comment is really useful to have here, but it needs to be less verbose - the key bit of information is that valid 8024 opcodes will never have a `0x5B` immediate, so the code as written pre-8024 is correct

category/vm/utils/evm-as/validator.hpp
```diff
@@ -167,6 +167,54 @@ namespace monad::vm::utils::evm_as::internal
...
+            // EIP-8024's static opcode-table min_stack is a placeholder; the
+            // real requirement is dynamic and derives from the decoded operand
+            // (mirrors basic_blocks.hpp).
+            size_t min_stack = 0;
+            size_t stack_increase = 0;
+            if (eip8024.opcode == compiler::EvmOpCode::DUPN) {
+                auto const n = compiler::eip8024_decode_single(eip8024.imm);
+                min_stack = n;
+                stack_increase = n + 1;
+            }
```
Baltoli: I've been sort of staring at this code for a while and I realise that what I've been mulling over is: why do we need to repeat this pattern of computing `min_stack` and `stack_increase` dynamically? The opcode and immediate byte are both known at construction time, so why not set the stack effects then and treat these like regular instructions? Am I missing something?

Baltoli (changes requested):
I haven't reviewed the tests in detail yet, but some high level comments:

* Will need rebasing post @ryankeleti refactoring the instruction table - your `when` predicate is now upstreamed, for example.
* Claude's comments are broadly very noisy / uninformative here. It's got a habit of inserting explanatory prose into places with strong conventions (e.g. the traits file - 5 lines of comment here is noisy and distracts from the easy-to-read, strongly conventioned code).
* I have a design question around instruction construction / decoding - why do we need the stack effects to be computed dynamically if the opcode and immediate are known at construction time?

category/vm/interpreter/instruction_table.hpp
```diff
@@ -65,25 +69,9 @@
...
-            analysis,                                                          \
-            stack_bottom,                                                      \
-            stack_top + delta,                                                 \
-            gas_remaining,                                                     \
-            instr_ptr);                                                        \
-    }                                                                          \
-    while (false);
+#define MONAD_VM_NEXT(OP) MONAD_VM_NEXT_N(OP, 1)
+#define MONAD_VM_NEXT_PUSH(OP) MONAD_VM_NEXT_N(OP, ((OP) - PUSH0) + 1)
+#define MONAD_VM_NEXT_IMM(OP) MONAD_VM_NEXT_N(OP, 2)
```
Baltoli, line 75: Nice

category/vm/interpreter/instruction_table.hpp
```diff
@@ -1503,6 +1491,56 @@ namespace monad::vm::interpreter
...
+        runtime::Context &ctx, Intercode const &analysis,
+        uint256_t const *stack_bottom, uint256_t *stack_top,
+        int64_t gas_remaining, uint8_t const *instr_ptr)
+    {
+        auto const ops = check_requirements_eip8024<SWAPN, traits>(
+            ctx, stack_bottom, stack_top, gas_remaining, instr_ptr[1]);
+
+        auto const top = stack_top->to_avx();
+        *stack_top = *(stack_top - ops.n);
+        *(stack_top - ops.n) = uint256_t{top};
```
Baltoli: Are there uses of `to_avx()` elsewhere in the instruction table? I think we should keep this code portable, as `to_avx()` won't work on the ZK build. If the codegen from writing a C++ swap of the two slots is worse, then we can try to shadow the code, but let's avoid that if at all possible. It's _likely_ that the compiler will emit an AVX swap here anyway.

test/ethereum_test/exclude/MONAD_NEXT_amsterdam.cmake
```diff
@@ -17,5 +17,7 @@
 # as support is added.
 # This is intended to keep test suite clean within stacked
 # Amsterdam PRs.
-set(MONAD_NEXT_amsterdam_included_tests "*/amsterdam/eip7843_slotnum/*")
+set(MONAD_NEXT_amsterdam_included_tests
+    "*/amsterdam/eip7843_slotnum/*"
+    "*/amsterdam/eip8024_dupn_swapn_exchange/*")
```
Baltoli: See comment on the logging PR about this logic; convenient for development but needs cleaning up for merge

test/vm/unit/emitter_tests.cpp
```diff
@@ -4438,3 +4438,167 @@ TEST(Emitter, ReleaseSrcAndDestRegression)
...
+            EXPECT_EQ(stack.get(1).get(), third)
+                << "3rd@" << Emitter::location_type_to_string(loc_third)
+                << " 2nd@" << Emitter::location_type_to_string(loc_second);
+
+            (void)emit.finish_contract(rt);
+        }
+    }
+}
+
+TEST(Emitter, exchange_aliased_duplicated_element)
```
Baltoli, line 4561: Good tests - nice one

category/vm/evm/opcodes.hpp
```diff
@@ -42,7 +43,12 @@ namespace monad::vm::compiler
          * program.
          *
          * This value is 0 for all instructions other than the `PUSHN` family,
-         * each of which expects N bytes to follow.
+         * each of which expects N bytes to follow, and the EIP-8024 opcodes
+         * (`DUPN`/`SWAPN`/`EXCHANGE`), each of which expects a single immediate
+         * operand byte. Note that the basic-block scanner does not read the
+         * EIP-8024 immediate through this field -- `decode_eip8024` consumes it
+         * directly, because unlike `PUSHN` data it also determines the
+         * instruction's stack effect and validity.
```
Baltoli: I don't love the wording here - `decode_eip8024` is called _from_ the basic block scanner, it's not an independent unit

category/vm/evm/opcodes.hpp
```diff
@@ -529,9 +539,12 @@ namespace monad::vm::compiler
             unknown_opcode_info,
             unknown_opcode_info,
             unknown_opcode_info,
-            unknown_opcode_info,
-            unknown_opcode_info,
-            unknown_opcode_info,
+            // EIP-8024: only the net delta (stack_increase - min_stack) is
+            // meaningful here; min_stack is a placeholder -- the real
+            // operand-dependent requirement comes from eip8024_stack_effect.
```
Baltoli, line 543: This seems fine - does any code break from assuming the opcode table is faithful here? I presume not

category/vm/evm/opcodes.hpp
```diff
@@ -611,8 +624,117 @@ namespace monad::vm::compiler
...
+            return {static_cast<uint8_t>(n), static_cast<uint8_t>(n + 1)};
+        }
+        if (opcode == SWAPN) {
+            auto const n = eip8024_decode_single(imm);
+            return {static_cast<uint8_t>(n + 1), static_cast<uint8_t>(n + 1)};
+        }
+        // Pin the fall-through: this helper is shared by the basic-block
+        // scanner, the interpreter and the evm-as validator, and any other
+        // opcode reaching here would silently get a pair-decoded stack effect.
+        MONAD_DEBUG_ASSERT(opcode == EXCHANGE);
```
Baltoli: Seems worth having this be a full assert - and more generally, I think we want to move away from debug asserts as a rule, so you might consider just swapping all usages where the assertion is checking a precondition of a function being called (e.g. here, in release mode, if you call `eip8024_stack_effect` with a bad argument, you'd want a failure)

category/vm/interpreter/stack.hpp
```diff
@@ -73,6 +75,84 @@ namespace monad::vm::interpreter
...
+        if (MONAD_UNLIKELY(gas_remaining < 0)) {
+            ctx.exit(OutOfGas);
+        }
+
+        auto const ops = [imm]() -> Eip8024Operands {
+            if constexpr (is_exchange) {
+                auto const [n_val, m_val] = compiler::eip8024_decode_pair(imm);
+                return {
+                    static_cast<ptrdiff_t>(n_val),
+                    static_cast<ptrdiff_t>(m_val)};
```
Baltoli: I haven't verified, so this might actually be the best way to have it, but if you find yourself `static_cast`-ing between types like this, you might want to reconsider what `Eip8024Operands` actually carries. Is `ptrdiff_t` the best type there?

category/vm/evm/opcodes.hpp
brett-monad: updated the comment a bit for clarity

brett-monad:
@claude review this PR

category/vm/interpreter/instruction_table.hpp
brett-monad: to_avx appears to be misnamed and wasn't introduced by this branch.  Maybe a small follow on commit to fix?

brett-monad: So it turns out that to_avx will compile with the ZK build but the code generation is poor.  The latest code now compiles to the fewest instructions on all three targets

category/vm/compiler/ir/x86/virtual_stack.cpp
```diff
@@ -392,7 +392,7 @@ namespace monad::vm::compiler::native
         auto &e = at(top_index_);
 
         auto const rem = e->stack_indices_.erase(top_index_);
-        MONAD_DEBUG_ASSERT(rem == 1);
+        MONAD_ASSERT(rem == 1);
```
dhil, line 395: Why are these being changed from debug to non-debug?

brett-monad: Just a general preference for actual asserts

dhil: These changes seem unrelated to the topic of this PR. I think, for general PR hygiene and for discussion of the downstream PR, it would be better to carve these out into a separate PR against `main`.

category/vm/utils/evm-as/instruction.hpp
```diff
@@ -58,6 +58,22 @@ namespace monad::vm::utils::evm_as
...
+        constexpr explicit Eip8024I(
+            compiler::EvmOpCode const opcode, uint8_t const imm)
+            : opcode(opcode)
+            , imm(imm)
+        {
+        }
+
+        compiler::EvmOpCode opcode;
+        uint8_t imm;
+    };
```
dhil, line 75: This not a blocker, but I am wondering whether it would better to just generalize `PlainI` to carry an immediate -- we could collapse `PushI` into it too. I might do this in a refactoring after this PR has landed.

dhil (commented):
One thing that seems to be missing from this patch is an extension to the microbenchmark suite with the new opcodes (or well atleast confirming that it isn't broken in some way).

category/vm/compiler/ir/x86/virtual_stack.cpp
dhil: (did this PR get retargeted to `main`? Because my comment above appears to be partially stale.)

brett-monad: yes, it is now retargeted to main.  something to note is that the spec tests for these amsterdam PRs will be addressed in a follow up PR

category/vm/utils/evm-as/instruction.hpp
brett-monad: Worth preserving that Eip8024I's immediate is already-encoded and validity-constrained, unlike PushI's semantic uint256_t

category/vm/compiler/ir/x86/virtual_stack.cpp
brett-monad: As it turns out this has not been retargeted to main; it is still stacked. Either way, once all the Amsterdam PRs are approved they will be committed together with the Amsterdam spec-test PR.

category/vm/compiler/ir/x86/emitter.cpp
```diff
@@ -2334,6 +2334,16 @@ namespace monad::vm::compiler::native
...
     }
 
+    // No discharge
+    void Emitter::exchange(uint8_t const n, uint8_t const m)
+    {
+        MONAD_ASSERT(n > 0);
+        MONAD_ASSERT(m > n);
+        stack_.exchange(
+            stack_.top_index() - static_cast<int32_t>(n),
+            stack_.top_index() - static_cast<int32_t>(m));
```
Baltoli, line 2344: Redundant casts?

Baltoli (changes requested):
I think we can probably improve the opcode encoding / decoding scheme here. Currently, the raw encoded byte is getting stored for every family of opcode - but we could actually just decode the _logical_ index once at bytecode analysis time and reuse that everywhere it's needed.

There's a special case for `EXCHANGE` - in that case I think we'd want to add an `exchange_indices` method that returns a pair and decodes the nibbles.

Baltoli:
One more thing to consider is the following (out of defensiveness):
1. Add `has_operand_dependent_stack_effect` as a new predicate alongside the opcode helpers
2. `static_assert` that predicate isn't true when using `check_requirements`

category/vm/compiler/ir/x86/emitter.cpp
brett-monad: Removed.

brett-monad:
Implemented as requested: the immediate is decoded once at analysis time, `Instruction::index()` holds the logical operand, EXCHANGE has `exchange_indices()`, and `check_requirements` static_asserts on the new `has_operand_dependent_stack_effect`. For consistency the evm-as side got the same treatment, which we had earlier declined: `Eip8024I` carries the operands and the byte is encoded only at emission.

category/vm/evm/opcodes.hpp
```diff
@@ -611,8 +623,155 @@ namespace monad::vm::compiler
         return opcode >= LOG0 && opcode <= LOG4;
     }
 
+    // The EIP-8024 opcodes (DUPN/SWAPN/EXCHANGE), each carrying a single
+    // immediate byte (unlike the DUP1-16/SWAP1-16 families).
+    constexpr bool is_eip8024_opcode(uint8_t const opcode)
+    {
+        return opcode == DUPN || opcode == SWAPN || opcode == EXCHANGE;
```
dhil, line 630: We can tighten this code to guarantee that its image is branchless by exploiting the fact that the opcodes form a contiguous sequence (0xe6-0xe8)
```suggestion
        return static_cast<uint8_t>(EXCHANGE - opcode) < 0x3;
```

I believe this rewrite is safe due to the **defined** behaviour of unsigned integer wrapping. E.g. reasoning equationally

```
Case 1, opcode \in {DUPN, SWAPN, EXCHANGE}:

   (EXCHANGE - opcode) < 0x3
= { by unsigned arithmetic }
   (EXCHANGE - opcode) mod 256 < 0x3
= { by case analysis on opcode }
    (EXCHANGE - 0xe6) mod 256 < 0x3 = 0x2 mod 256 < 0x3 = 0x2 < 0x3
    (EXCHANGE - 0xe7) mod 256 < 0x3 = 0x1 mod 256 < 0x3 = 0x1 < 0x3
    (EXCHANGE - 0xe8) mod 256 < 0x3 = 0x0 mod 256 < 0x3 = 0x0 < 0x3

Thus the relevant cases all evaluate to true.

Case 2, opcode > EXCHANGE

Now, fix opcode = EXCHANGE + k, for 0x1 <= k <= 0x17, then

    EXCHANGE - opcode > 0x2
= { by unsigned arithmetic }
    (EXCHANGE - opcode) mod 256 > 0x2
= { by definition }
    (EXCHANGE - (EXCHANGE + k)) mod 256 > 0x2
= { by distributivity }
    (EXCHANGE - EXCHANGE - k) mod 256 > 0x2
= { by basic arithmetic }
    -k mod 256 >= 0x3
= { since 0x1 <= k <= 0x17, -k mod 256 = 0x100 - k }
    0x100 - k > 0x2
~= { taking k = 0x17 establishes the lower bound }
    0x100 - 0x17 > 0x3
= { arithmetic }
    0xe9 > 0x3
= true

Case 3, opcode < DUPN

Fix opcode = DUPN - k, for 0x1 <= k <= 0xe6

    (EXCHANGE - opcode) mod 256 > 0x2
= { by definition }
    (EXCHANGE - (DUPN - k)) mod 256 > 0x2
= { by distributivity }
    (EXCHANGE - DUPN + k) mod 256 > 0x2
= { EXCHANGE - DUPN = 0x2 }
    (0x2 + k) mod 256 > 0x2
= { 0x1 <= k <= 0xe6, hence 0x3 <= 0x2+k <= 0xe8 < 0x100 }
    0x2 + k > 0x2
= { subtract 0x2 }
    k > 0
= true
```

category/vm/evm/opcodes.hpp
```diff
@@ -611,8 +623,155 @@ namespace monad::vm::compiler
...
+     * EIP-8024 immediate encode/decode. The immediate is encoded so it can
+     * never equal JUMPDEST (0x5B) or a PUSH byte (0x60-0x7F); the disallowed
+     * raw ranges are 91..127 (single form: DUPN/SWAPN) and 82..127 (pair form:
+     * EXCHANGE). An instruction with a disallowed immediate behaves as INVALID.
+     *   decode_single(x) -> n in [17, 235]
+     *   decode_pair(x)   -> (n, m) with 1 <= n < m and n + m <= 30
+     */
+    constexpr bool eip8024_single_disallowed(uint8_t const x)
+    {
+        return x > 90 && x < 128;
```
dhil, line 649: Similarly here, I think the following rewrite is correct using a similar reasoning method as above.

```suggestion
        return static_cast<uint8_t>(x - 91) < 37;
```

category/vm/evm/opcodes.hpp
```diff
@@ -611,8 +623,155 @@ namespace monad::vm::compiler
...
+    }
+
+    constexpr bool eip8024_pair_disallowed(uint8_t const x)
+    {
+        return x > 81 && x < 128;
```
dhil, line 654: Here too
```suggestion
        return static_cast<uint8_t>(x - 82) < 46;
```

dhil (commented):
Some initial thoughts about guaranteeing the quality of the C++ codegen. Given that this code is used by the performance-critical compiler code, I think it is we shouldn't rely on the C++ compiler to optimise the code for us. Moreover, since this patch touches multiple parts of the compiler (and interpreter) I think we should verify that the compilation speed (of the monad compiler) has not regressed.

category/vm/evm/opcodes.hpp
andreaslyn: Here are my two cents: This is not on the critical path, so I am personally indifferent between the two solutions here. Both seem easy to understand and insignificant performance difference.

category/vm/utils/parser.cpp
```diff
@@ -188,11 +188,17 @@ namespace monad::vm::utils
...
     std::string show_opcodes(std::vector<uint8_t> const &opcodes)
     {
         std::stringstream ss;
-        auto const &tbl = monad::vm::compiler::make_opcode_table<
-            EvmTraits<MONAD_ETH_LATEST_STABLE_REVISION>>();
+        auto const &tbl =
+            monad::vm::compiler::make_opcode_table<disasm_traits>();
         for (size_t i = 0; i < opcodes.size(); ++i) {
             auto c = opcodes[i];
             ss << std::format("[{:#x}] {:#x} {}\n", i, c, tbl[opcodes[i]].name);
```
0xSqualo, line 204: Should `show_opcodes` label forbidden EIP-8024 immediates as `INVALID_*` while preserving the following byte as a separate instruction?

category/vm/interpreter/stack.hpp
```diff
@@ -73,6 +76,59 @@ namespace monad::vm::interpreter
...
+        // Validate the immediate before charging gas: a disallowed encoding is
+        // invalid regardless of the gas available, so charging first would
+        // report OutOfGas for a malformed instruction. Matches `invalid` and
+        // decode_eip8024, which turns a disallowed immediate into
+        // Terminator::InvalidInstruction at analysis time.
+        if (MONAD_UNLIKELY(!compiler::eip8024_immediate_valid(Instr, imm))) {
+            ctx.exit(Error);
+        }
+
+        gas_remaining -= info.min_gas;
```
0xSqualo, line 104: Should the EIP-8024 handlers charge their required three gas before checking a forbidden immediate, consistent with EIP-8024 and Geth’s interpreter?

[Geth implementation](https://github.com/ethereum/go-ethereum/blob/89def4a53b19dbb796d7ff81b438f50d270aaca0/core/vm/interpreter.go#L187-L201)

category/vm/evm/opcodes.hpp
goodlyrottenapple: compilers optimise both into identical assembly and the current code is more easily understandable imo

category/vm/interpreter/stack.hpp
Baltoli: I think perhaps for consistency, but I don't believe that this would be consensus-visible

category/vm/evm/opcodes.hpp
Baltoli: Yeah I tend to agree here, I think the simple versions of these lines are fine

PR #2414, opened 2026-07-08
Make DbCache a read-through cache 

Chen-Yifan:
## Problem

DbCache is populated only by committed state: the LRU caches are filled at finalization (`on_finalize` -> `insert_in_lru_caches`) and the proposal map at commit (update_proposal_state). A read that misses both fetches the value from the trie and discards it. Nothing that is only read is ever cached.

This has performance impact for the page-encoded (MIP-8) commit path. `PageCommitBuilder` performs a read-modify-write: for every touched page it reads the current page from the db to merge the changed slots before writing. Because those reads aren't cached, the commit-time RMW re-reads pages from disk that execution already read in the same block, the commit phase re-does the read execution just did. On a fresh-storage workload this dominates block time.

## Change

Add read-through: on a storage/account read-miss, insert the value fetched from disk into the LRU so a subsequent read hits.

- `try_read_account` / `try_read_storage` / `try_read_storage_page` now return a `CacheReadStatus` (`Hit` / `MissResolved` / `MissTruncated`) instead of `bool`.
- New `insert_account` / `insert_storage_page` populate the LRU, guarded by the status.
- `TrieDb::read_account` / `read_storage` / `read_storage_page` insert the disk value on `MissResolved`, including empty/absent entries (negative caching).

🤖  Generated with Claude Opus 4.8

### DIFF
--- a/category/execution/ethereum/db/db_cache.hpp
+++ b/category/execution/ethereum/db/db_cache.hpp
@@ -36,6 +36,16 @@
 
 MONAD_NAMESPACE_BEGIN
 
+// Outcome of a cache read.
+enum class CacheReadStatus
+{
+    Hit, // served from proposal map or LRU
+    MissResolved, // both missed, proposal chain walked to the finalized base
+                  // -> disk value == finalized value -> safe to cache this miss
+    MissTruncated, // proposal chain truncated before the finalized base
+                   // -> can't prove finalized-consistent -> don't cache miss
+};
+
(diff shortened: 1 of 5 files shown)


maxkozlovsky (dismissed):

Chen-Yifan:
@claude review please

magedm (approved):

kkuehlz (approved):

maxkozlovsky (approved):

PR #2415, opened 2026-07-09
Throw on maximum code size being exceeded

Baltoli:
This can happen _in theory_ via RPC state overrides, but not in practice because of transport layer restrictions limiting the maximum code size override that can be relayed. This change is therefore defence in depth against future transport layer relaxations; it's consistent with existing fixes to hard assertions reachable from user input.

### DIFF
--- a/category/core/monad_exception.hpp
+++ b/category/core/monad_exception.hpp
@@ -18,6 +18,7 @@
 #include <category/core/backtrace.hpp>
 #include <category/core/config.hpp>
 #include <category/core/likely.h>
+#include <category/core/throw.hpp>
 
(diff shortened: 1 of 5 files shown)


0xSqualo (dismissed):

dhil (dismissed):
I would add a regression test to monad executor tests, that submits an eth call request with a large code size override. I think it should be sufficient to generate one consisting entirely of zeroes. We can generate it programmatically using evm-as. I think the max code size is small enough that this is feasible.

dhil (approved):

andreaslyn (approved):

PR #2417, opened 2026-07-09
Replace new_var's char buffer loop with std::format

ryankeleti:
Instead of holding a char buffer and writing to a stringstream in
reverse, let `std::format` handle converting the next subscript value
to a string. This avoids any potential for invalid buffer accesses.

### DIFF
--- a/category/vm/utils/evm-as/compiler.cpp
+++ b/category/vm/utils/evm-as/compiler.cpp
@@ -17,9 +17,8 @@
 
 #include <array>
 #include <cstddef>
-#include <limits>
+#include <format>
 #include <ostream>
-#include <sstream>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


Baltoli (dismissed):

category/vm/utils/evm-as/compiler.cpp
m-alvarez: Maybe we could bound this iteration by `i < sz`, which means that we'd silently truncate the subscript instead of UB. Both options are bad, but the silent truncation will come up in unit tests.

andreaslyn (dismissed):

cmcl:
Any reason we cannot rewrite the problematic loop to use `operator<<` (on `ctx.next_subscript`) for `std::stringstream`?

Edit: to clarify this will get rid of the buffer entirely.

ryankeleti:
> Any reason we cannot rewrite the problematic loop to use `operator<<` (on `ctx.next_subscript`) for `std::stringstream`?
> 
> Edit: to clarify this will get rid of the buffer entirely.

That's probably the best choice, there might be some performance implications but this code doesn't seem to be very hot. We could also use `std::format`.

cmcl:
> We could also use `std::format`.

There were some murmurs from Claude about this being better (in that the conversion is performed on two-digits at a time)

Baltoli (approved):

mkolosick (approved):

cmcl (approved):

PR #2419, opened 2026-07-09
snapshot: raise fd limit, fail fast, and report open errors on dump

maxkozlovsky:
The filesystem snapshot dumper held one std::ofstream per shard/file open for the whole dump (256 shards x 4 files), overrunning the common 1024 soft RLIMIT_NOFILE part-way through and aborting on a failed ofstream::open with only "failed to open <path>".

- io.cpp: AsyncIO_rlimit_raiser now raises the soft RLIMIT_NOFILE to the hard limit (up to 16384) instead of only warning, and warns with errno on getrlimit/setrlimit failure or when the hard limit is itself too low.
- db_snapshot_filesystem.cpp: open the .blake3 checksum streams lazily at destroy time (halving peak fds); flush and check the data and checksum streams so a truncated dump fails at dump time instead of as a later load-time checksum mismatch; and append strerror(errno) to every open-failure assert (including a previously-unchecked ifstream).
- monad_cli.cpp: pre-flight the open-file limit before opening the db and exit non-zero with a clear message when it cannot fit the dump, deriving the requirement per invocation from --total-shards/--shard-number.

Reported by operators dumping snapshots for the v0.16.0 migration.

### DIFF
--- a/category/async/io.cpp
+++ b/category/async/io.cpp
@@ -132,17 +132,44 @@ namespace detail
     {
         AsyncIO_rlimit_raiser_impl()
         {
+            constexpr rlim_t target = 16384;
             struct rlimit r = {0, 0};
-            getrlimit(RLIMIT_NOFILE, &r);
-            if (r.rlim_cur < 4096) {
-                std::cerr << "WARNING: maximum file descriptor limit is "
-                          << r.rlim_cur
-                          << " which is less than 4096. 'Too many open files' "
-                             "errors may result. You can increase the hard "
-                             "file descriptor limit for a given user by adding "
-                             "to '/etc/security/limits.conf' '<username> hard "
-                             "nofile 16384'."
+            if (getrlimit(RLIMIT_NOFILE, &r) != 0) {
+                std::cerr << "WARNING: getrlimit(RLIMIT_NOFILE) failed: "
+                          << std::strerror(errno)
+                          << "; cannot check or raise the open file limit."
                           << std::endl;
+                return;
+            }
+            // Raise the soft limit toward target, capped at the hard limit
+            // (RLIM_INFINITY means the hard limit imposes no cap). The guard
+            // skips the call when the soft limit already meets the target.
+            rlim_t desired = target;
+            if (r.rlim_max != RLIM_INFINITY && r.rlim_max < desired) {
+                desired = r.rlim_max;
+            }
+            if (desired > r.rlim_cur) {
+                struct rlimit const want = {desired, r.rlim_max};
+                if (setrlimit(RLIMIT_NOFILE, &want) == 0) {
+                    r.rlim_cur = desired;
+                }
+                else {
+                    std::cerr << "WARNING: setrlimit(RLIMIT_NOFILE) to "
+                              << desired << " failed: " << std::strerror(errno)
+                              << "; proceeding with a soft limit of "
+                              << r.rlim_cur << "." << std::endl;
+                }
+            }
+            if (r.rlim_cur < 4096) {
+                std::cerr
+                    << "WARNING: maximum file descriptor limit is "
+                    << r.rlim_cur
+                    << " which is less than 4096. 'Too many open files' "
+                       "errors may result. Raise the hard file descriptor "
+                       "limit for this user (e.g. in "
+                       "'/etc/security/limits.conf': '<username> hard "
+                       "nofile 16384')."
+                    << std::endl;
             }
         }
     } AsyncIO_rlimit_raiser;
--- a/cmd/monad_cli.cpp
+++ b/cmd/monad_cli.cpp
@@ -48,6 +48,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cerrno>
 #include <charconv>
 #include <chrono>
 #include <cmath>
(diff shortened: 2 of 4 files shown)


category/async/io.cpp
```diff
@@ -132,17 +132,44 @@ namespace detail
...
+                return;
+            }
+            // Raise the soft limit toward target, capped at the hard limit
+            // (RLIM_INFINITY means the hard limit imposes no cap). The guard
+            // skips the call when the soft limit already meets the target.
+            rlim_t desired = target;
+            if (r.rlim_max != RLIM_INFINITY && r.rlim_max < desired) {
+                desired = r.rlim_max;
+            }
+            if (desired > r.rlim_cur) {
```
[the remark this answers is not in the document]
maxkozlovsky: unnecessary noise

cmd/monad_cli.cpp
```diff
@@ -905,6 +907,61 @@ int main(int const argc, char *argv[])
...
+            LOG_ERROR(
+                "shard_number ({}) must be < total_shards ({})",
+                shard_number,
+                total_shards);
+            return 1;
+        }
+        // The filesystem dumper holds one fd open per (shard, data file) for
+        // the whole run. The soft limit was already raised toward the hard
+        // limit at startup (AsyncIO_rlimit_raiser in category/async/io.cpp), so
+        // reaching this error means the hard limit itself is too low and must
```
cmcl: Nit: I think the comment about "this error" should be moved further down to the relevant branch.

cmd/monad_cli.cpp
```diff
@@ -905,6 +907,61 @@ int main(int const argc, char *argv[])
...
+            // rather than block a dump that may well succeed. The dumper's
+            // open() asserts remain the backstop.
+            LOG_WARNING(
+                "could not read the open file descriptor limit (getrlimit "
+                "failed: {}); proceeding without the snapshot fd pre-flight "
+                "check",
+                std::strerror(errno));
+        }
+        else if (limit.rlim_cur < required_fds) {
+            LOG_ERROR(
```
cmcl, line 959: Can we include the current hard limit in the error diagnostic?

cmd/monad_cli.cpp
```diff
@@ -905,6 +907,61 @@ int main(int const argc, char *argv[])
...
+        // reaching this error means the hard limit itself is too low and must
+        // be raised (limits.conf / systemd LimitNOFILE).
+        uint64_t const shards_this_run =
+            shard_number < MONAD_SNAPSHOT_SHARDS
+                ? (MONAD_SNAPSHOT_SHARDS - shard_number + total_shards - 1) /
+                      total_shards
+                : 0;
+        // Data-file fds held open simultaneously, plus headroom for the triedb
+        // device(s), io_uring rings, logger, and std streams.
+        constexpr uint64_t fd_headroom = 256;
```
cmcl, line 936: How is this calculated?

cmd/monad_cli.cpp
```diff
@@ -905,6 +907,61 @@ int main(int const argc, char *argv[])
...
+                "could not read the open file descriptor limit (getrlimit "
+                "failed: {}); proceeding without the snapshot fd pre-flight "
+                "check",
+                std::strerror(errno));
+        }
+        else if (limit.rlim_cur < required_fds) {
+            LOG_ERROR(
+                "open file descriptor limit ({}) is too low to dump this "
+                "snapshot, which needs about {} descriptors ({} shards x {} "
+                "files plus overhead). Raise the hard limit for this user "
```
cmcl:
```suggestion
                "files plus headroom). Raise the hard limit for this user "
```

cmd/monad_cli.cpp
cmcl: Just flagging that checksum files were previously opened during `monad_db_dump_snapshot()` before this `LOG_INFO`. They are now opened in `monad_db_snapshot_filesystem_write_user_context_destroy()`, which runs after the log, so a checksum-open failure can trigger an assertion after `success=true` has already been queued for logging.

cmd/monad_cli.cpp
```diff
@@ -55,6 +55,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdlib>
+#include <cstring>
```
cmcl, line 59: To avoid the transitive include:

```suggestion
#include <cerrno>
#include <cstring>
```

cmcl (dismissed):

maxkozlovsky:
Updated to resolve the comments.

cmd/monad_cli.cpp
cmcl: My question here was about `fd_headroom`, not `shards_this_run`.  Is there's anything more concrete that can be said about the choice of `256`?

cmcl (approved):

Chen-Yifan (approved):

cmd/monad_cli.cpp
maxkozlovsky: not really, c++ library would open few files that are hard to quantify, input/output/error, logging, iouring, etc. 256 is large enough number to cover these random uses. It is not possible to describe in a comment without going into unnecessary detail.

PR #2439, opened 2026-07-16
EIP-7843 (`SLOTNUM`)

ryankeleti:
Add opcode `0x4B` (`SLOTNUM`) feature gated under `eip_7843_active()`.

`SLOTNUM` returns the slot number for the current block via
`evmc_tx_context.block_round` in both the interpreter and the x86 compiler.

- Add all-zero `block_access_list_hash` block header field; BAL (EIP-7928)
  is not planned for Monad but proceeds `slot_number` in RLP encoding. There's a corresponding `eip_7928_active()` feature flag set to false for Monad revisions, to stub out any future implementation.
- RLP-encode/decode `BlockHeader::slot_number` as a trailing optional field
  after `requests_hash` and `block_access_list_hash`.
- `propose_block` only populates it when `eip_7843_active()` is true.

This change will be followed shortly by a PR in `monad-bft` adding
`slot_number` and the all-zero `block_access_list_hash` to consensus headers:     https://github.com/category-labs/monad-bft/pull/3185


test/ethereum_test/CMakeLists.txt
```diff
@@ -123,10 +123,6 @@ add_test(NAME MONAD_NEXT_amsterdam_monad_ethereum_test
          COMMAND monad-ethereum-test --fork MONAD_NEXT
                  --blockchain-tests ${CMAKE_BINARY_DIR}/src/MonadAmsterdamSpecTestFixtures/blockchain_tests
                  --gtest_filter=-:${TESTS_EXCLUDED})
```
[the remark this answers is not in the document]
ryankeleti: Fixed

category/vm/interpreter/instruction_table.hpp
```diff
@@ -1245,6 +1249,21 @@ namespace monad::vm::interpreter
...
+        int64_t gas_remaining, uint8_t const *instr_ptr)
+    {
+        if constexpr (traits::eip_7843_active()) {
+            check_requirements<SLOTNUM, traits>(
+                ctx, analysis, stack_bottom, stack_top, gas_remaining);
+            push(stack_top, ctx.env.tx_context->block_round);
+
+            MONAD_VM_NEXT(SLOTNUM);
+        }
+    }
```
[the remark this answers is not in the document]
ryankeleti: Removed traits::eip_7843_active gate, on second inspection is redundant with the guard in the table

category/execution/ethereum/core/rlp/block_rlp.cpp
```diff
@@ -89,6 +89,11 @@ byte_string encode_block_header(BlockHeader const &block_header)
             encode_bytes32(block_header.requests_hash.value());
     }
 
+    if (block_header.slot_number.has_value()) {
+        encoded_block_header +=
+            encode_unsigned(block_header.slot_number.value());
+    }
```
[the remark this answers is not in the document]
ryankeleti: Added assert

ryankeleti:
@claude review this PR

category/vm/evm/opcodes.hpp
```diff
@@ -667,8 +668,15 @@ namespace monad::vm::compiler
...
     make_opcode_table<EvmTraits<MONAD_ETH_AMSTERDAM>>()
     {
-        return make_opcode_table<
+        auto table = make_opcode_table<
             EvmTraits<previous_evm_revision(MONAD_ETH_AMSTERDAM)>>();
+
+        if constexpr (EvmTraits<MONAD_ETH_AMSTERDAM>::eip_7843_active()) {
+            // https://eips.ethereum.org/EIPS/eip-7843
+            add_opcode(0x4B, table, {"SLOTNUM", 0, 0, 1, false, 2, 0});
+        }
```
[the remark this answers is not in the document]
ryankeleti: Outdated - cbd989e63979d192f24190c1827e82a60bddc385

category/rpc/monad_executor_test.cpp
```diff
@@ -2158,6 +2263,125 @@ TEST_F(EthCallFixture, trace_block_with_prestate)
...
+TEST_F(EthCallFixture, trace_block_slot_number_from_persisted_header)
+{
+    // EXE-60 regression test: the historical-trace path must read
+    // slot_number from the target block's persisted header instead
+    // of the slot_number in the header passed over FFI.
+    static constexpr Address ADDR_A =
+        0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_address;
+
+    // SLOTNUM PUSH1 0x00 SSTORE
+    auto const contract_code = 0x4b600055_bytes;
```
Baltoli: This is pretty small so not a big issue, but I'd rather not commit _new_ code that encodes programs as byte literals. There's a nice tool @dhil wrote for embedding EVM code as a DSL ([EvmBuilder](https://github.com/category-labs/monad/blob/add15bad804eacbe4245f0c4fbd6e948690c5fef/category/vm/utils/evm-as/builder.hpp#L44)), which is what we should use for new code.

(being somewhat hypocritical here, if you look at the git blame you can see I'm responsible for a lot of these examples)

category/vm/evm/traits.hpp
```diff
@@ -327,6 +333,11 @@ namespace monad
             return evm_rev() >= MONAD_ETH_OSAKA;
         }
 
+        static consteval bool eip_7843_active() noexcept
+        {
+            return Rev >= MONAD_NEXT;
+        }
```
Baltoli, line 363: This is a slightly subtle one. I think we would actually want to gate on `evm_base` being `>= AMSTERDAM`, and update the base EVM revision returned for `MONAD_NEXT` here. From memory that's what we did for Osaka

category/execution/ethereum/core/block.hpp
```diff
@@ -56,18 +56,14 @@ struct BlockHeader
...
-    // runloops -- leaves it nullopt, so the round read via value_or(0) is 0
-    // there.
+    // EIP-7843. RLP-encoded as a trailing optional field (see
+    // encode_block_header/decode_block_header) once requests_hash is also
+    // present. The consensus round is durably carried in
+    // MonadConsensusBlockHeader::block_round; only the consensus execution
+    // path (propose_block in runloop_monad) populates this field from it.
+    // Any entrypoint that does not repopulate it -- RPC/trace re-execution in
+    // monad_executor, historical-block replay runloops -- leaves it nullopt,
+    // so the round read via value_or(0) is 0 there (see EXE-60).
```
Baltoli: I would probably strip this comment out for the sake of brevity now that this PR fixes the handling of slot numbers. The treatment of this field isn't meaningfully different to the others here (e.g. `requests_hash`).

category/execution/ethereum/core/rlp/block_rlp.cpp
```diff
@@ -184,6 +190,12 @@ Result<BlockHeader> decode_block_header(byte_string_view &enc)
                 if (payload.size() > 0) {
                     BOOST_OUTCOME_TRY(
                         block_header.requests_hash, decode_bytes32(payload));
+
+                    if (payload.size() > 0) {
+                        BOOST_OUTCOME_TRY(
+                            block_header.slot_number,
+                            decode_unsigned<uint64_t>(payload));
```
Baltoli: It's out of scope for this PR, but the deeply nested code here is kind of unpleasant - not your fault, just noticed it while reviewing.

ryankeleti: I agree - I can look into it as a follow-up

category/rpc/monad_executor.cpp
```diff
@@ -1391,6 +1382,16 @@ struct monad_executor
...
 
+                    // EXE-37/EXE-60: overwrite block_header.slot_number with
+                    // the target block's persisted slot_number.
+                    auto const persisted_header =
+                        monad::get_block_header(db, block_number, block_id);
+                    MONAD_ASSERT_THROW(
+                        persisted_header.has_value(), BLOCK_HEADER_ERR_MSG);
+                    BlockHeader call_header = block_header;
+                    call_header.slot_number =
+                        persisted_header.assume_value().slot_number;
```
Baltoli: To make sure I understand your logic here - I presume that you've noticed that the `block_header` in scope doesn't have a slot number because it's been RLP encoded by an FFI consumer that doesn't yet know about slot numbers, right? Am I missing something?

ryankeleti: That's right, arguably I think the full header with slot_number should be passed via RLP over the FFI.

I looked into it a bit more, I see `to_alloy_header` in `rust/crates/monad-exec-events/src/block.rs` defaults `slot_number: None` and it's never set afterwards - would be worthwhile to see if it can be populated on the rust side

test/utils/from_json.cpp
```diff
@@ -122,6 +122,12 @@ monad::BlockHeader read_genesis_blockheader(nlohmann::json const &genesis_json)
             write_to.bytes);
     }
 
+    // Amsterdam fork (EIP-7843)
+    if (genesis_json.contains("slotNumber")) {
+        block_header.slot_number = std::stoull(
+            genesis_json["slotNumber"].get<std::string>(), nullptr, 0);
+    }
```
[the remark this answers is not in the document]
Baltoli: This is consistent with the Prague feature above; also just for test code where the excess robustness doesn't matter

ryankeleti:
Status:
  - PR on consensus side: https://github.com/category-labs/monad-bft/pull/3185
  - Test fixtures need to be updated to include both block_access_list_hash and slot_number

ryankeleti:
Repinned test fixtures:
    Bump MonadSpecTestFixtures 1.1.1 -> 1.2.0
    Bump MonadAmsterdamSpecTestFixtures 0.2.0 -> 0.4.0
    
`MONAD_NEXT_amsterdam_monad_ethereum_test` fails due to missing EIP-7708 implementation

category/execution/ethereum/core/rlp/block_rlp.cpp
```diff
@@ -89,6 +89,18 @@ byte_string encode_block_header(BlockHeader const &block_header)
...
+        encoded_block_header +=
+            encode_bytes32(block_header.block_access_list_hash.value());
+    }
+
+    if (block_header.slot_number.has_value()) {
+        MONAD_ASSERT(block_header.requests_hash.has_value());
+        MONAD_ASSERT(block_header.block_access_list_hash.has_value());
+        encoded_block_header +=
+            encode_unsigned(block_header.slot_number.value());
+    }
```
Baltoli, line 100: There's no case in Monad where we have a BAL hash and no slot number - might be worth asserting that here? Minor point

category/execution/ethereum/core/block.hpp
```diff
@@ -56,19 +56,11 @@ struct BlockHeader
...
-    // runloop_monad) populates this. Any entrypoint that does not repopulate
-    // it -- RPC/trace re-execution in monad_executor, historical-block replay
-    // runloops -- leaves it nullopt, so the round read via value_or(0) is 0
-    // there.
-    std::optional<uint64_t> slot_number{std::nullopt};
+
+    // Monad does not implement BALs. If EIP-7843 is active, we carry
+    // an all-zero BAL hash to maintain parity with Ethereum's RLP format.
+    std::optional<bytes32_t> block_access_list_hash{std::nullopt}; // EIP-7928
+    std::optional<uint64_t> slot_number{std::nullopt}; // EIP-7843
```
Baltoli, line 63: Nice

category/execution/monad/chain/monad_testnet.cpp
```diff
@@ -76,6 +76,7 @@ GenesisState MonadTestnet::get_genesis_state() const
         .excess_blob_gas = 0,
         .parent_beacon_block_root = NULL_ROOT,
         .requests_hash = NULL_HASH,
+        // When Amsterdam is enabled, add block_access_list_hash/slot_number
```
Baltoli: I think it's actually the case that the testnet genesis header has these as `nullopt`, right? The genesis state is anchored at a particular revision, it doesn't change depending on the live revision

category/execution/runloop/runloop_monad.cpp
```diff
@@ -293,11 +293,15 @@ Result<BlockExecOutput> propose_block(
...
-    // read into evmc_tx_context.block_round by get_tx_context and system_call.
+    // header's slot_number. It is read into evmc_tx_context.block_round by
+    // get_tx_context and system_call.
     // NOTE: only this path populates slot_number; RPC/trace re-execution does
     // not (see EXE-60).
-    block.header.slot_number = consensus_header.block_round;
+    if constexpr (traits::eip_7843_active()) {
+        block.header.requests_hash = bytes32_t{};
+        block.header.block_access_list_hash = bytes32_t{};
+        block.header.slot_number = consensus_header.block_round;
```
Baltoli: Will the slot number actually get supplied by the execution inputs once your changes to `monad-bft` arrive? We might want to change this to assert that the execution and consensus views are consistent

test/ethereum_test/exclude/MONAD_NEXT_amsterdam.cmake
```diff
@@ -13,9 +13,9 @@
...
-# does not yet emit; every fixture therefore fails the genesis block-hash
-# check. Exclude the whole suite for now and drop entries here to re-enable
-# individual tests as support lands.
-set(MONAD_NEXT_amsterdam_excluded_tests "BlockchainTests.*")
+# Only include slotnum tests, include other Amsterdam fixtures
+# as support is added.
+# This is intended to keep test suite clean within stacked
+# Amsterdam PRs.
+set(MONAD_NEXT_amsterdam_included_tests "*/amsterdam/eip7843_slotnum/*")
+set(MONAD_NEXT_amsterdam_excluded_tests "")
```
Baltoli: See comments on Brett's PRs: this is fine and convenient to develop with, but we should clean it up for merging.

ryankeleti: This is essentially to keep the PRs green, Brett's last change should  delete this and enable all Amsterdam tests

category/execution/monad/chain/monad_testnet.cpp
ryankeleti: Removed comment

category/execution/runloop/runloop_monad.cpp
ryankeleti: Changed to assert

test/vm/unit/evm_fixture.hpp
```diff
@@ -15,6 +15,7 @@
 
 #pragma once
 
+#include "category/vm/evm/traits.hpp"
```
Baltoli: Tiny nitpick: `<>` rather than `""` for includes. Mixing confuses clang-format

Baltoli (commented):
Will need some minor reworking when #2399 + followups land (re. differential / reference testing); looking generally in good shape though

test/vm/unit/evm_fixture.hpp
ryankeleti: Applied

category/execution/ethereum/validate_block.cpp
```diff
@@ -157,6 +157,22 @@ Result<void> static_validate_header(BlockHeader const &header)
...
+    if constexpr (!traits::eip_7843_active()) {
+        if (MONAD_UNLIKELY(
+                header.block_access_list_hash.has_value() ||
+                header.slot_number.has_value())) {
+            return BlockError::FieldBeforeFork;
+        }
+    }
+    else {
+        if (MONAD_UNLIKELY(
+                !header.block_access_list_hash.has_value() ||
```
marcograss, line 170: Since Monad doesn't implement EIP-7928, `block_access_list_hash` is only ever meant to be the all-zero placeholder — but `validate_block` only checks the field is present, never that it's zero, so a proposer can put any 32-byte value there and it still commits into the header hash.
Is skipping that value check intentional, or should Amsterdam validation reject a nonzero `block_access_list_hash` for Monad?

ryankeleti: I discussed this with @Baltoli and we decided on not validating zero

Baltoli: Prior art for `requests_hash`:
```cpp
    if constexpr (traits::evm_rev() < MONAD_ETH_PRAGUE) {
        if (MONAD_UNLIKELY(header.requests_hash.has_value())) {
            return BlockError::FieldBeforeFork;
        }
    }
    else if (MONAD_UNLIKELY(!header.requests_hash.has_value())) {
        return BlockError::MissingField;
    }
```

category/vm/fuzzing/instruction_data.hpp
```diff
@@ -45,6 +45,9 @@ namespace monad::vm::fuzzing
     //   RETURNDATACOPY  (generate_returndatacopy)
     //   PUSH0 - PUSH32  (generate_push)
     //   CREATE, CREATE2 (generate_create)
+    //
+    // TODO(amsterdam): SLOTNUM is Amsterdam-only and unimplemented by
+    // by pinned evmone.
```
Baltoli: Stale comment now that evmone is gone - just needs updated to refer to the OCaml spec

Baltoli (dismissed):
One minor comment; will reapprove when fixed

Baltoli (dismissed):

goodlyrottenapple (dismissed):

Baltoli (dismissed):

category/rpc/monad_executor.cpp
```diff
@@ -848,6 +838,16 @@ namespace
         std::vector<std::vector<CallFrame>> const empty_call_frames{};
         uint64_t gas_consumed_so_far = 0;
 
+        auto const next_synthetic_round =
```
dhil, line 841: Maybe just name it `next_round`, as it used by the synthetic blocks as well as the user-defined blocks. That way we do not conflate the two different notions of synthetic here.
```suggestion
        auto const next_round =
```

category/rpc/monad_executor.cpp
```diff
@@ -958,6 +965,12 @@ namespace
                 .parent_beacon_block_root = bytes32_t{},
             };
 
+            if (auto const current_round = next_synthetic_round(header)) {
+                current_header.requests_hash = bytes32_t{};
+                current_header.block_access_list_hash = bytes32_t{};
+                current_header.slot_number = current_round;
+            }
```
dhil, line 972: Why not inline these three into the structure construction just above?

rust/crates/monad-exec-events/src/block.rs
```diff
@@ -74,8 +74,8 @@ impl ExecutedBlock {
             excess_blob_gas: Some(0),
             parent_beacon_block_root: Some(alloy_primitives::B256::ZERO),
             requests_hash: Some(alloy_primitives::B256::ZERO),
-            block_access_list_hash: None,
-            slot_number: None,
+            block_access_list_hash: Some(alloy_primitives::B256::ZERO),
```
0xSqualo, line 77: Should the execution-event SDK’s Alloy/RPC header conversion be made fork-aware so that a pre-Amsterdam event does not return Amsterdam-only header fields whose encoding disagrees with its recorded hash?

rust/crates/monad-exec-events/src/block.rs
```diff
@@ -74,8 +74,8 @@ impl ExecutedBlock {
...
+            slot_number: Some(self.start.round),
```
0xSqualo, line 78: Can Ethereum-format replay use the persisted `slot_number` when reconstructing post-Amsterdam execution events, rather than substituting the execution block number?

category/rpc/monad_executor.cpp
```diff
@@ -848,6 +838,16 @@ namespace
         std::vector<std::vector<CallFrame>> const empty_call_frames{};
         uint64_t gas_consumed_so_far = 0;
```
0xSqualo, line 841: Should `eth_simulateV1` mirror Geth by selecting fork rules from each simulated child header when a simulation crosses the Amsterdam activation time?

[Geth implementation](https://github.com/ethereum/go-ethereum/blob/89def4a53b19dbb796d7ff81b438f50d270aaca0/internal/ethapi/simulate.go#L241-L316)

category/rpc/monad_executor.cpp
ryankeleti: I felt it might be a bit noisy to have ternary conditions inside the construction, but I'll apply this change

category/rpc/monad_executor.cpp
Baltoli: Yes, that seems reasonable to do (cc @ryankeleti)

rust/crates/monad-exec-events/src/block.rs
Baltoli: Hmmm, I think you've got a point here. This code as written will end up with extra data being sent to WS subscribers, which we don't want. There's some thought needed here to figure out how to migrate this particular path (this change is the first time we're doing so).

Baltoli: @brett-monad is looking into this; the gist is that we'll pipe in a way for the Rust code that uses this EE data to check the appropriate revision when reifying the Alloy header back to JSON

PR #2440, opened 2026-07-16
[execution/ethereum] Add EIP-4844/7918 blob gas validation

cmcl:
Validates each block’s excess blob gas against its parent header, using the [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844) update rule and [EIP-7918](https://eips.ethereum.org/EIPS/eip-7918) reserve-price adjustment.

Enable the corresponding EIP-4844 and EIP-7918 tests.

## Checklist

- [x] Update internal tracker for EIP-7918 (EXE-45)

### DIFF
--- a/category/execution/ethereum/transaction_gas.cpp
+++ b/category/execution/ethereum/transaction_gas.cpp
@@ -16,6 +16,7 @@
 #include <category/core/assert.h>
 #include <category/core/config.hpp>
 #include <category/core/int.hpp>
+#include <category/execution/ethereum/core/block.hpp>
 #include <category/execution/ethereum/core/transaction.hpp>
 #include <category/execution/ethereum/transaction_gas.hpp>
 #include <category/vm/evm/explicit_traits.hpp>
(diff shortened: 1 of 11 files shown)


category/execution/ethereum/transaction_gas.cpp
```diff
@@ -225,6 +226,46 @@ uint256_t get_base_fee_per_blob_gas(
...
+        auto const threshold = checked_mul(
+            PRICE_RATIO,
+            get_base_fee_per_blob_gas(
+                parent_header.excess_blob_gas.value_or(0),
+                current_blob_schedule));
+
+        uint256_t const execution_base_fee =
+            parent_header.base_fee_per_gas.value_or(uint256_t{});
+
+        if (threshold && execution_base_fee > threshold.assume_value()) {
```
Baltoli, line 253: Is this the morally correct way to assert the checked multiplication error here? I think this code is only called in Ethereum replay anyway, so it's not critical, but just checking that there's no way the hard assumption can bite us

cmcl: Yes, the case of overflow correctly maps to a false reserve-price condition.

Baltoli (dismissed):

category/execution/ethereum/transaction_gas.hpp
```diff
@@ -64,6 +64,9 @@ max_gas_cost(uint64_t const gas_limit, uint256_t const max_fee_per_gas) noexcept
 // EIP-4844
 inline constexpr uint64_t GAS_PER_BLOB = 131'072;
 
+// EIP-7918
+inline constexpr uint64_t BLOB_BASE_COST = 1u << 13;
```
cmcl: Would it be preferable if these aligned in style?  I just noticed I forgot to do this.

I have a preference for the shift computation but I think the constant number is ubiquitous in the codebase.

Baltoli: Oh yeah, good catch - I think the non-shift style is how these things are generally encoded here and in EIPs?

Baltoli (approved):

mkolosick (approved):

PR #2444, opened 2026-07-20
Add MONAD_TRY / MONAD_CATCH_ALL macros

goodlyrottenapple:
`EvmcHost` wraps each callback in `try/catch (...)`, but the zkVM builds with `-fno-exceptions`, so the keywords are a hard compile error. `MONAD_TRY` / `MONAD_CATCH_ALL` hide them behind macros.

### DIFF
--- a/category/core/throw.hpp
+++ b/category/core/throw.hpp
@@ -17,3 +17,9 @@
 
 // Throw a C++ exception
 #define MONAD_THROW(exc, ...) throw exc(__VA_ARGS__)
+
+// Wrap a try/catch pair. On platforms with exceptions disabled, these are
+// redefined to erase the `try` keyword and reduce the `catch` to a dead
+// `if constexpr (false)`, so the catch body still type-checks but never runs.
+#define MONAD_TRY try
+#define MONAD_CATCH(...) catch (__VA_ARGS__)
--- a/zkvm/category/core/throw.hpp
+++ b/zkvm/category/core/throw.hpp
@@ -21,3 +21,5 @@
 #include <zkvm/core/zkvm_halt.h>
 
 #define MONAD_THROW(exc, ...) zkvm_halt(1)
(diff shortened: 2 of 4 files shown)


category/core/throw.hpp
```diff
@@ -17,3 +17,9 @@
 
 // Throw a C++ exception
 #define MONAD_THROW(exc, ...) throw exc(__VA_ARGS__)
+
+// Wrap a try/catch(...) pair. The zkVM mirror erases the `try` keyword and
+// reduces `catch (...)` to a dead `if (false)` so the catch body still
+// type-checks but never runs in the no-exceptions build.
```
Baltoli: Extremely minor nitpick: I'd rephrase this slightly to be more generic - avoid mentioning the zkVM and just specify that it's to allow compilation on platforms with exceptions disabled.

zkvm/category/core/throw.hpp
```diff
@@ -21,3 +21,5 @@
 #include <zkvm/core/zkvm_halt.h>
 
 #define MONAD_THROW(exc, ...) zkvm_halt(1)
+#define MONAD_TRY
+#define MONAD_CATCH_ALL if (false)
```
Baltoli: I'm not sure it makes any difference in practice but for explicitness:
```suggestion
#define MONAD_CATCH_ALL if constexpr (false)
```

category/core/throw.hpp
```diff
@@ -17,3 +17,9 @@
...
+#define MONAD_TRY try
+#define MONAD_CATCH_ALL catch (...)
```
Baltoli: maybe `MONAD_CATCH(...)` expanding to `catch (__VA_ARGS__)` is a bit more general / clearer?

mkolosick (approved):

Baltoli (approved):

PR #2446, opened 2026-07-20
Misc header import fixes

goodlyrottenapple:


cmcl:
> * `db.hpp` / `partial_trie_db.hpp`: `vm/vm.hpp` → `vm/code.hpp`. Both headers only use `vm::SharedIntercode`, which `code.hpp` provides. Downstream files that need the heavier `vm.hpp` (e.g. `runloop_monad.cpp` for `vm::VM`) already get it via `block_state.hpp`.

Just to note that `vm/vm.hpp` is provided by `runloop_monad.hpp` so there's no reliance on a transitive include in this case.

cmcl (dismissed):

cmcl (approved):

Baltoli (approved):

PR #2371, opened 2026-06-23
Remove support for Istanbul fork

eclark0426:

PR #2450, opened 2026-07-20
Use availability predicates in opcode/instruction tables

ryankeleti:
Instead of using revision-chaining with `make_opcode_table` and successively building up newer tables with `add_opcode`, have one table that checks feature-flag availability predicates (e.g. `when(traits::eip_7939_active(), {"CLZ", ...})` for including the opcode `CLZ` if EIP-7939 is in the current fork).

Additionally, `instruction_table` now derives opcode availabilityfrom `opcode_table`, using `!is_unknown_opcode_info<traits>(op)` (through the `avail` helper lambda), keeping the tables in sync.

### DIFF
--- a/category/vm/evm/opcodes.hpp
+++ b/category/vm/evm/opcodes.hpp
@@ -21,9 +21,7 @@
 #include <evmc/evmc.hpp>
 
 #include <array>
-#include <cstddef>
 #include <cstdint>
-#include <exception>
 #include <string_view>
 #include <tuple>
 
@@ -252,13 +250,6 @@ namespace monad::vm::compiler
         SELFDESTRUCT = 0xFF
     };
 
-    consteval monad_eth_revision
-    previous_evm_revision(monad_eth_revision const rev)
-    {
-        MONAD_DEBUG_ASSERT(std::to_underlying(rev) > 0);
-        return monad_eth_revision(std::to_underlying(rev) - 1);
-    }
-
     /**
      * Placeholder value representing an opcode value not currently used by the
      * EVM specification. The value of `unknown_opcode_info` is significant, so
@@ -266,6 +257,17 @@ namespace monad::vm::compiler
      */
     constexpr auto unknown_opcode_info =
         OpCodeInfo{"UNKNOWN", 0, 0, false, false, 0, 0};
+
+    template <Traits traits>
+    consteval std::string_view eip_4399_name()
+    {
+        return traits::eip_4399_active() ? "PREVRANDAO" : "DIFFICULTY";
+    }
+
+    consteval OpCodeInfo when(bool const available, OpCodeInfo const info)
+    {
+        return available ? info : unknown_opcode_info;
+    }
 
     /**
      * Lookup table of opcode info for each possible 1-byte opcode value.
@@ -276,24 +278,11 @@ namespace monad::vm::compiler
      * revisions and become valid in later ones).
      */
     template <Traits traits>
-    consteval std::array<OpCodeInfo, 256> make_opcode_table() = delete;
-
-    template <Traits traits>
-    constexpr std::array<OpCodeInfo, 256> opcode_table =
-        make_opcode_table<traits>();
-
-    consteval inline void add_opcode(
-        uint8_t const opcode, std::array<OpCodeInfo, 256> &table,
-        OpCodeInfo const info)
+    consteval std::array<OpCodeInfo, 256> make_opcode_table()
     {
-        MONAD_DEBUG_ASSERT(table[opcode] == unknown_opcode_info);
-        table[opcode] = info;
-    }
+        static_assert(traits::evm_rev() >= MONAD_ETH_BERLIN);
 
-    template <>
-    consteval std::array<OpCodeInfo, 256>
-    make_opcode_table<EvmTraits<MONAD_ETH_BERLIN>>()
-    {
+        // clang-format off
         return {
             OpCodeInfo{"STOP", 0, 0, 0, false, 0, 0}, // 0x00
             OpCodeInfo{"ADD", 0, 2, 1, false, 3, 0}, // 0x01
@@ -326,7 +315,7 @@ namespace monad::vm::compiler
             OpCodeInfo{"SHL", 0, 2, 1, false, 3, 0}, // 0x1B,
             OpCodeInfo{"SHR", 0, 2, 1, false, 3, 0}, // 0x1C,
             OpCodeInfo{"SAR", 0, 2, 1, false, 3, 0}, // 0x1D,
-            unknown_opcode_info,
+            when(traits::eip_7939_active(), {"CLZ", 0, 1, 1, false, 5, 0}), // 0x1E,
             unknown_opcode_info,
 
             OpCodeInfo{"SHA3", 0, 2, 1, true, 30, 0}, // 0x20,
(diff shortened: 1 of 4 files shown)


category/vm/interpreter/instruction_table.hpp
```diff
@@ -131,7 +131,9 @@ namespace monad::vm::interpreter
             shl<traits>, // 0x1B,
             shr<traits>, // 0x1C,
             sar<traits>, // 0x1D,
-            since(MONAD_ETH_OSAKA, clz<traits>), // 0x1E,
+            avail(
+                !compiler::is_unknown_opcode_info<traits>(CLZ),
+                clz<traits>), // 0x1E,
```
Baltoli: Let's do two things here:
- I think the `is_unknown_opcode_info` helper should just live in the `vm::` namespace rather than in `vm::compiler`. This change _might_ have a blast radius (but I think it's probably fine), so worth doing it in a separate preceding commit.
- Given that every use of the `avail` lambda is doing the same pattern via the opcode table lookup, let's just fold the `!unknown` part into the lambda body. This will make the first part less of an issue, but worth doing regardless.

category/vm/evm/opcodes.hpp
```diff
@@ -267,6 +258,73 @@ namespace monad::vm::compiler
...
+
+    template <Traits traits>
+    consteval bool has_blob()
+    {
+        // EIP-4844: BLOBHASH
+        // https://eips.ethereum.org/EIPS/eip-4844
+        //
+        // EIP-7516: BLOBBASEFEE
+        // https://eips.ethereum.org/EIPS/eip-7516
+        return traits::evm_rev() >= MONAD_ETH_CANCUN;
```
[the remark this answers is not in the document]
ryankeleti: eip_4844_active is false but the opcodes BLOBHASH and BLOBBASEFEE were previously available in the same way before my change

category/vm/evm/opcodes.hpp
```diff
@@ -347,7 +391,14 @@ namespace monad::vm::compiler
...
 
             OpCodeInfo{"ADDRESS", 0, 0, 1, false, 2, 0}, // 0x30,
-            OpCodeInfo{"BALANCE", 0, 1, 1, true, 700, 0}, // 0x31,
+            OpCodeInfo{
+                "BALANCE",
+                0,
+                1,
+                1,
+                true,
+                traits::eip_2929_active() ? 100 : 700,
```
Baltoli: I think we should aim to merge https://github.com/category-labs/monad/pull/2371 ahead of this PR - while doing so will get you an annoying rebase, it'll clean the changes here up substantially

category/vm/evm/opcodes.hpp
```diff
@@ -357,23 +408,58 @@ namespace monad::vm::compiler
...
             OpCodeInfo{"RETURNDATACOPY", 0, 3, 0, true, 3, 0}, // 0x3E,
-            OpCodeInfo{"EXTCODEHASH", 0, 1, 1, true, 700, 0}, // 0x3F,
+            OpCodeInfo{
+                "EXTCODEHASH",
+                0,
+                1,
+                1,
+                true,
+                traits::eip_2929_active() ? 100 : 700,
+                0}, // 0x3F,
```
Baltoli: Separately, once the Istanbul-removal PR goes through, let's disable clang-format for these tables and keep definitions on one line for readability (auto-formatting is good except when it's not, and big tables like this are one place it's not)

category/vm/evm/opcodes.hpp
```diff
@@ -357,23 +408,58 @@ namespace monad::vm::compiler
...
 
             OpCodeInfo{"BLOCKHASH", 0, 1, 1, false, 20, 0}, // 0x40,
             OpCodeInfo{"COINBASE", 0, 0, 1, false, 2, 0}, // 0x41,
             OpCodeInfo{"TIMESTAMP", 0, 0, 1, false, 2, 0}, // 0x42,
             OpCodeInfo{"NUMBER", 0, 0, 1, false, 2, 0}, // 0x43,
-            OpCodeInfo{"DIFFICULTY", 0, 0, 1, false, 2, 0}, // 0x44,
+            OpCodeInfo{// Name change: https://eips.ethereum.org/EIPS/eip-4399
+                       (traits::evm_rev() >= MONAD_ETH_PARIS) ? "PREVRANDAO"
+                                                              : "DIFFICULTY",
```
Baltoli: Let's make this a feature flag + helper at the top of the table, just for readability

category/vm/evm/opcodes.hpp
```diff
@@ -267,6 +258,73 @@ namespace monad::vm::compiler
...
+    template <Traits traits>
+    consteval bool has_blob()
+    {
+        // EIP-4844: BLOBHASH
+        // https://eips.ethereum.org/EIPS/eip-4844
+        //
+        // EIP-7516: BLOBBASEFEE
+        // https://eips.ethereum.org/EIPS/eip-7516
+        return traits::evm_rev() >= MONAD_ETH_CANCUN;
+    }
```
Baltoli: Let's refactor this slightly - I think for everything except the blob opcodes (which are a slightly odd carveout), we should turn these functions into trait feature flags directly (e.g. `traits::eip_3198_active()` for one). Doing so keeps features in one place (the traits definitions).

Then, for the blob opcodes, we can add `has_blob_opcodes` as a separate feature flag (gated on `evm_rev() >= MONAD_ETH_CANCUN`). This means that Monad keeps blobs disabled but the stubbed opcodes available.

ryankeleti: Refactored into feature flags

category/vm/interpreter/instruction_table.hpp
ryankeleti: There are other similar functions (e.g. `is_push_opcode`) in the `vm::compiler` namespace - should we migrate all of them?

Baltoli: I think it _should_ be but maybe if we pull everything in it'll make this PR too noisy. Let's perhaps do this in a followup?

ryankeleti: Sounds good - I'll add the lambda wrapper but defer the namespace change

category/vm/evm/traits.hpp
```diff
@@ -70,17 +70,44 @@ namespace monad
         { T::evm_rev() } -> std::same_as<monad_eth_revision>;
 
         // Feature flags
+
+        // EIP-1153: TLOAD/TSTORE
+        // https://eips.ethereum.org/EIPS/eip-1153
```
Baltoli: Being very picky I don't think we need these comments - the trait function names are self-documenting

category/vm/evm/traits.hpp
```diff
@@ -70,17 +70,44 @@ namespace monad
...
         { T::mip_12_active() } -> std::same_as<bool>;
         { T::can_create_inside_delegated() } -> std::same_as<bool>;
+        // EIP-4844: BLOBHASH
+        // https://eips.ethereum.org/EIPS/eip-4844
+        //
+        // EIP-7516: BLOBBASEFEE
+        // https://eips.ethereum.org/EIPS/eip-7516
+        //
+        // Opcodes are stubs without eip_4844_active
+        { T::has_blob_opcodes() } -> std::same_as<bool>;
```
Baltoli: This is the only place with inconsistent behaviour - I'd call out that the opcodes exist but return stub data in Monad, so have been separated from 4844

category/vm/interpreter/instruction_table.hpp
```diff
@@ -95,8 +95,10 @@ namespace monad::vm::interpreter
     {
         static_assert(traits::evm_rev() >= MONAD_ETH_ISTANBUL);
```
ryankeleti, line 96: Should I leave this or change to `MONAD_ETH_BERLIN`

category/vm/evm/opcodes.hpp
```diff
@@ -267,6 +258,28 @@ namespace monad::vm::compiler
...
+        return 100;
+    }
+
+    template <Traits traits>
+    consteval std::string_view eip_4399_name()
+    {
+        return traits::eip_4399_active() ? "PREVRANDAO" : "DIFFICULTY";
+    }
+
+    consteval OpCodeInfo avail(bool const available, OpCodeInfo const info)
```
Baltoli: While I'm bikeshedding minor issues, let's call this `when`

category/vm/evm/opcodes.hpp
```diff
@@ -267,6 +258,28 @@ namespace monad::vm::compiler
...
+    template <Traits traits>
+    consteval uint32_t sstore_gas_cost()
+    {
+        if constexpr (is_monad_trait_v<traits>) {
+            if constexpr (traits::monad_rev() >= MONAD_NEXT) {
+                return traits::base_sstore_cost();
+            }
+        }
+        return 100;
+    }
```
Baltoli: Maybe just add `base_sstore_cost` to the Ethereum traits? Then you can delete this extra helper

Baltoli (changes requested):

category/vm/evm/opcodes.hpp
ryankeleti: `base_sstore_cost` returns `100` - so `100` in every branch now. Would it be preferable to
(1) remove `base_sstore_cost` entirely, inline constant 100
(2) keep `base_sstore_cost`, have it return 100 for Ethereum traits

category/vm/interpreter/instruction_table.hpp
goodlyrottenapple: Might as well change it I think? There are still a few pre-Berlin asserts left in the code base but we should probably clean those up

goodlyrottenapple (approved):
LGTM

category/vm/interpreter/instruction_table.hpp
Baltoli: Yeah, change it as a drive by

category/vm/evm/opcodes.hpp
Baltoli: (2) - `base_sstore_cost` gets used somewhere else, I think?

Baltoli (approved):

PR #2454, opened 2026-07-22
mpt/async: use std::atomic_ref instead of start_lifetime_as<std::atomic<T>>

Baltoli:
## Problem

libstdc++-16 made `std::start_lifetime_as<T>` hard-require `is_implicit_lifetime_v<T>`. `std::atomic<T>` is not an implicit-lifetime type, so the long-standing pattern of viewing raw bytes in the memory-mapped TrieDB metadata as a `std::atomic` — `start_lifetime_as<std::atomic<T>>(ptr)->load()/->store()` — no longer compiles under gcc-16 / libstdc++-16:

```
error: static assertion failed
  static_assert(is_implicit_lifetime_v<_Tp>);
  • 'std::atomic<unsigned char>' is not an implicit-lifetime type
```

## Fix

Those sites (and their `reinterpret_cast<std::atomic<T>*>` siblings) perform atomic loads / stores / `fetch_add` on plain, non-atomic fields that live in shared, memory-mapped metadata. There was never a real `std::atomic` object at those addresses, so the pattern was always technically UB — libstdc++-16 just began enforcing it. `std::atomic_ref<T>` is the C++20 facility for exactly this: atomic operations on a pre-existing non-atomic object. This PR migrates every such site to `std::atomic_ref<T>`, **preserving each memory order verbatim**.

## Commits (kept split for review)

1. **use std::atomic_ref instead of start_lifetime_as<std::atomic<T>>** — the core fix; converts the 38 `start_lifetime_as<std::atomic<…>>` sites. `detail::db_metadata::is_dirty()` now returns `std::atomic_ref<uint8_t>` by value (callers use `.load()`/`.store()`/the implicit conversion, unchanged). `std::atomic_ref<chunk_offset_t>` resolves to the primary template rather than the custom `std::atomic<chunk_offset_t>` specialisation; a `static_assert(std::atomic_ref<chunk_offset_t>::is_always_lock_free)` pins that both lower to the same lock-free 8-byte atomics — required for the cross-process mmap.

2. **hand out atomic_ref via chunk_bytes_used_at accessor** — replaces `device_t::metadata_t::chunk_bytes_used()`, which returned a `std::span<uint32_t>` that each caller wrapped in `atomic_ref` per access, with `chunk_bytes_used_at(offset, index)` returning the `std::atomic_ref<uint32_t>` directly. Removes the footgun of a plain span whose elements could be accessed non-atomically by accident.

3. **drop the now-unused std::atomic<chunk_offset_t> specialisation** — with all access going through `atomic_ref`, the custom specialisation has no users anywhere in the tree; remove it, its layout static_asserts, and the `<bit>` include it needed for `bit_cast`.

4. **replace reinterpret_cast<std::atomic<T>*> with std::atomic_ref** — the sibling cast pattern that still compiled but expressed the same fake-atomic-object UB: the five `uint64_t` version setters, the `chunk_info_t` load/store, and the word-by-word `atomic_memcpy` loop. Two of these read through a const-qualified path (a const method / a `void const*` source) over genuinely-mutable shared memory; since `std::atomic_ref` gains a const-qualified form only in **C++26 (P3323R1)**, those two keep a documented `const_cast` + `TODO`. `chunk_info_t` has no custom specialisation, so `atomic_ref` uses the primary template exactly as the cast did; a matching lock-free `static_assert` is added.

## What is intentionally *not* changed

Behaviour. Every conversion is a 1:1 swap at the same address, width, and memory order; an independent detailed review of the diff (byte representation of the `chunk_offset_t` primary-vs-custom-atomic path, `atomic_ref::required_alignment` vs the `db_metadata` field layout, and site-by-site memory orders) concluded it is behaviour-preserving.

## Validation

- **Full build green under clang-19 + gcc-16 libstdc++** (the config this gcc-16 port is validated on). No `start_lifetime_as<atomic>` or `reinterpret_cast<atomic>` remain.
- **g++-16 directly:** the previously-failing TUs now compile. A full g++-16 `-O2` build still can't complete, but only because of pre-existing, unrelated `-Werror` false positives in Boost.Outcome headers — nothing in this diff.
- **Tests (clang + libstdc++-16), all green:** `storage_pool_test`, `db_metadata_test`, `db_state_machine_kind_open_test`, `update_aux_test`, and the `db_test` copy / dirty / secondary / auto_expire / version_history subset — covering every changed path, including the two `const` getters, the secondary-timeline migrate/promote logic, `db_copy`/`atomic_memcpy`, and chunk-info add/remove.

### DIFF
--- a/category/async/config.hpp
+++ b/category/async/config.hpp
@@ -16,7 +16,6 @@
 #pragma once
 
 #include <atomic>
-#include <bit>
 #include <cassert>
 #include <compare>
 #include <cstdint>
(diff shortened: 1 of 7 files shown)


Baltoli:
@claude re-review in detail

Baltoli:
A Codex adversarial review flagged one issue, which is that the `atomic_memcpy` function is technically UB - this isn't new to this PR; the old version had the same issue (doing atomic operations over underlying objects that aren't actually `uint64_t`, but rather chunks of a larger object). It was otherwise happy with the changes here.

kkuehlz (approved):

maxkozlovsky (approved):

PR #2455, opened 2026-07-23
statesync: record finalized version only after state root validation

kkuehlz:
monad_statesync_client_finalize previously marked the target finalized before comparing the canonical trie's state root against the trusted header. A peer supplying syntactically valid but corrupted state thus became the client's restart baseline: reopening rewinds to the recorded finalized version and initializes sync progress from it, so a normal retry never re-requested the corrupted range. Validate the root first and only then update the finalized version, and warn when a target is rejected.

### DIFF
--- a/category/statesync/statesync_client.cpp
+++ b/category/statesync/statesync_client.cpp
@@ -18,7 +18,9 @@
 #include <category/core/config.hpp>
 #include <category/core/keccak.hpp>
 #include <category/core/likely.h>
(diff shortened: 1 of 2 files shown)


Chen-Yifan (dismissed):
approved but need to address copilot comments.

Chen-Yifan (approved):

maxkozlovsky (approved):

PR #2460, opened 2026-07-27
Change Result to a basic_outcome type

andreaslyn:


andreaslyn:
Close if favor of https://github.com/category-labs/monad/pull/2467

PR #2461, opened 2026-07-27
Propagate exceptions as Result

andreaslyn:
Change base branch to main when #2460 is merged


andreaslyn:
Close in favor of https://github.com/category-labs/monad/pull/2467

PR #2465, opened 2026-07-28
Build the zkVM RISC-V cross toolchain with nix

goodlyrottenapple:
riscv-gnu-toolchain builds by cloning binutils, gdb and newlib from sourceware.org, which rate limits; cold CI builds now fail outright with HTTP 429. nixpkgs fetches hash-pinned source tarballs from its own CDN, removing the dependency on a host that throttles us.

nixpkgs is pinned to nixos-26.05 for GCC 15.2.0, matching the Ubuntu host gcc-15. The cross system is declared rv64ima/lp64 so newlib is built soft-float; the stock riscv64-embedded target is lp64d, which the zkVM cannot use.


docker/Dockerfile
```diff
@@ -120,31 +120,45 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
+
+FROM base AS base_riscv_toolchain
+
+# The toolchain sits at a hash-named store path that moves whenever the pin
+# does, so resolve it into a stable /riscv symlink here rather than baking
+# the path into this file. The absolute symlink targets inside the store
+# resolve because the store is copied to the same absolute location.
+COPY --from=riscv_toolchain_nix /nix/store /nix/store
+COPY --from=riscv_toolchain_nix /toolchain-path /tmp/toolchain-path
+RUN ln -s "$(cat /tmp/toolchain-path)" /riscv && rm /tmp/toolchain-path
```
Baltoli, line 150: > 🤖 Generated by Claude on Bruce's behalf

The deleted `build-riscv-toolchain.sh` hard-failed the build when the cross GCC version diverged from the host `gcc-15`; that check is now only a manual step in `nix/README.md`. Since `base` already ships `gcc-15`, the invariant can be enforced here for one line:

```suggestion
RUN ln -s "$(cat /tmp/toolchain-path)" /riscv && rm /tmp/toolchain-path

# Keep the cross compiler in lockstep with the host compiler; fail the image
# build on divergence (the check build-riscv-toolchain.sh used to enforce).
RUN test "$(/riscv/bin/riscv64-none-elf-gcc -dumpfullversion)" = "$(gcc-15 -dumpfullversion)"
```

.github/workflows/docker-base-riscv.yaml
```diff
@@ -24,13 +24,13 @@ jobs:
...
 
-      - name: Build and cache base_riscv_gnu_toolchain stage
+      - name: Build and cache base_riscv_toolchain stage
         uses: docker/build-push-action@v5
         with:
           context: .
           file: docker/Dockerfile
           ulimit: "memlock=-1:-1"
-          target: base_riscv_gnu_toolchain
+          target: base_riscv_toolchain
```
Baltoli, line 33: > 🤖 Generated by Claude on Bruce's behalf

Not on a changed line, so no suggestion block — but the `cache-to` just below likely needs `mode=max` now:

```yaml
          cache-to: type=gha,mode=max,scope=riscv-toolchain
```

`cache-to` defaults to `mode=min`, which only exports layers of the final target image. The expensive nix build now lives in the intermediate `riscv_toolchain_nix` stage — only its *COPY result* is a layer of `base_riscv_toolchain` — so this warm-up workflow probably isn't caching the ~8-minute layer it exists to cache. (Currently masked because `test-zk.yml` writes `mode=max` to `scope=zk` on main pushes, but this workflow should be self-sufficient.) Worth verifying with a cold-cache `workflow_dispatch` run.

docker/Dockerfile
```diff
@@ -120,31 +120,45 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
+    readlink -f /toolchain > /toolchain-path && \
+    nix-collect-garbage > /dev/null
+
+FROM base AS base_riscv_toolchain
+
+# The toolchain sits at a hash-named store path that moves whenever the pin
+# does, so resolve it into a stable /riscv symlink here rather than baking
+# the path into this file. The absolute symlink targets inside the store
+# resolve because the store is copied to the same absolute location.
+COPY --from=riscv_toolchain_nix /nix/store /nix/store
```
Baltoli, line 148: > 🤖 Generated by Claude on Bruce's behalf

Optional, take it or leave it: `nix-collect-garbage` can't reclaim the nix image's own profile closure (nix, bash, openssl, …), so this copies a few hundred MB of non-toolchain store paths into `base_riscv_toolchain` and `build_zk`. If image size ever matters, the nix stage can export just the toolchain closure:

```dockerfile
RUN nix --extra-experimental-features "nix-command flakes" \
        copy --no-check-sigs --to /export "$(readlink -f /toolchain)"
```

and then `COPY --from=riscv_toolchain_nix /export/nix/store /nix/store` here. These are CI-only images, so fine to skip.

Baltoli (commented):
> 🤖 Generated by Claude on Bruce's behalf

Inline suggestions from an offline review pass: restore the host/cross GCC lockstep check as a build-time assertion, use `mode=max` for the toolchain cache scope, plus two optional nits (store slimming, digest tag comment) and a question about `aarch64-darwin`.

docker/Dockerfile
```diff
@@ -120,31 +120,45 @@ RUN cd src/rust && cargo check --all-targets --all-features
...
-    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
-# Clone, build, install, strip, and remove the build tree in a single layer.
-# Only /root/riscv is consumed downstream; the build tree must be removed in
-# the same RUN it was created or it persists in the layer.
-RUN /opt/build-riscv-toolchain.sh && \
-    rm -rf /opt
-
-ENV PATH="/root/riscv/bin:${PATH}"
+FROM nixos/nix@sha256:377d4887aca98f0dfa12971c1ea6d6a625a435d8b610d4c95a436843da6fbfd1 AS riscv_toolchain_nix
```
Baltoli, line 124: > 🤖 Generated by Claude on Bruce's behalf

Digest-only pins are the right call, but unreadable when bumping. This digest is `nixos/nix:2.35.1`:

```suggestion
# nixos/nix:2.35.1 — pinned by digest; keep the tag in this comment in sync
FROM nixos/nix@sha256:377d4887aca98f0dfa12971c1ea6d6a625a435d8b610d4c95a436843da6fbfd1 AS riscv_toolchain_nix
```

nix/flake.nix
```diff
@@ -0,0 +1,25 @@
...
+
+  # Pinned to a release branch: nixos-26.05 ships GCC 15.2.0, matching both
+  # the Ubuntu host gcc-15 and the GCC that riscv-gnu-toolchain 2026.05.06
+  # used to build. flake.lock pins the exact revision; bumping it rebuilds
+  # the toolchain, so check the GCC version still matches (see README.md).
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-26.05";
+
+  outputs = { self, nixpkgs }:
+    let
+      systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" ];
```
Baltoli: I think we only want `x86_64-linux`

.github/workflows/docker-base-riscv.yaml
Baltoli: ^ The issue is Buildkit getting cache misses on the intermediate layer that actually builds the compiler

docker/Dockerfile
Baltoli: Seems fine for now and we can revisit

mkolosick (approved):

dhil (approved):

Baltoli (approved):

PR #2469, opened 2026-07-31
EIP-7708 (ETH transfer logs)

brett-monad:
Emit a standardised log for every native ETH movement, so indexers can track native value the way they already track ERC-20 transfers.

`Transfer(address,address,uint256)` (topic0 `0xddf252ad…`) on every non-zero value transfer to a different account: the top-level tx transfer, value-carrying `CALL`, `SELFDESTRUCT`, and `CREATE`/`CREATE2`. This reuses the existing `eth_simulate` native-transfer machinery, re-gated to fire on the consensus path. Two emitter addresses are in play and 7708 does not replace one with the other: the consensus log from `ETH_SYSTEM_ADDRESS` fires whenever 7708 is active, and `eth_simulate`'s synthetic from the ERC-7528 native-token address `0xeeee…` is emitted alongside it when `traceTransfers` is set — geth returns both. Note that `logIndex` counts both, so a consumer dropping the `0xeeee…` entries sees only odd indices.

The address was already a function-local constant in `process_requests.cpp`, the sender the EIP-7002/7251 system calls execute as. 7708 needs the same address, so it is promoted to a shared `ETH_SYSTEM_ADDRESS` rather than leaving two spellings of one consensus constant that can silently diverge.

### There is deliberately no Burn log

An earlier draft of EIP-7708 also specified `Burn(address,uint256)` for ETH destroyed by `SELFDESTRUCT`. That draft then took EIP-8246 as a prerequisite and dropped the log, because after 8246 there is no burn left to report. geth and erigon are in the same configuration.

### Active on MONAD_NEXT

`eip_7708_active()` is `>= MONAD_ETH_AMSTERDAM` in both trait families, and `MonadTraits<MONAD_NEXT>::evm_rev()` is Amsterdam, so the rule is live on the configuration that ships. The tests run over an explicit two-element type list — `MonadRevisionConstant<MONAD_NEXT>` and `EvmRevisionConstant<MONAD_ETH_AMSTERDAM>` — naming both revisions where 7708 is active. They are named rather than derived from `LATEST_SUPPORTED_EVM_FORK` so the suite does not depend on that constant having been advanced.

### Spec tests

No change to the Amsterdam exclusion list. The Amsterdam suite is disabled at
this commit and cannot open here: the fixtures are generated with the whole
fork active, so any Amsterdam EIP not yet in the tree fails its own fixtures —
at this commit the `eip8024_dupn_swapn_exchange` and
`frontier/opcodes/all_opcodes` cases, 47 of 710. It opens once every Amsterdam
EIP is in the tree.

Unit tests are unaffected and are the real gate here: `test_state` 379 and
`test_call_trace` 234 passing, both builds clean.

### Prerequisite

EIP-7708 lists `requires: 8246`, which must be live before this is. Without it
two `SELFDESTRUCT` paths still destroy ETH with no recipient — a `Transfer` log
cannot express that — so 7708 would claim complete native-value logging while
those burns went unlogged.

https://claude.ai/code/session_01T2YcdwU1mj3Lqunp9Y7pqa

### DIFF
--- a/category/execution/ethereum/evmc_host.hpp
+++ b/category/execution/ethereum/evmc_host.hpp
@@ -22,6 +22,7 @@
 #include <category/execution/ethereum/core/contract/abi_encode.hpp>
 #include <category/execution/ethereum/core/contract/abi_signatures.hpp>
 #include <category/execution/ethereum/core/contract/events.hpp>
+#include <category/execution/ethereum/core/receipt.hpp>
 #include <category/execution/ethereum/execute_message.hpp>
 #include <category/execution/ethereum/precompiles.hpp>
 #include <category/execution/ethereum/reserve_balance.hpp>
@@ -46,6 +47,39 @@ static_assert(sizeof(vm::Host) == 24);
 static_assert(alignof(vm::Host) == 8);
 
 class BlockHashBuffer;
+
+// The system sender: the caller the EIP-7002/EIP-7251 system calls execute as,
+// and the emitter EIP-7708 attributes its synthetic logs to. Not owned by any
+// one of them.
+inline constexpr Address SYSTEM_ADDRESS =
+    0xfffffffffffffffffffffffffffffffffffffffe_address;
+
+// EIP-7708 emits its Transfer logs from SYSTEM_ADDRESS on the consensus path
+// (from the activation revision); eth_simulate on pre-activation blocks keeps
+// the ERC-7528 native-token pseudo-address instead. Burn logs have no
+// pre-activation form, so they always come from SYSTEM_ADDRESS.
+inline constexpr Address SIMULATE_NATIVE_TOKEN_LOG_ADDRESS =
+    0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address;
+
+// EIP-7708 Burn log, emitted at finalization for the residual balance of an
+// account created and destroyed in the same transaction (ETH received after it
+// selfdestructed, which is then destroyed). LOG2 from SYSTEM_ADDRESS:
+// topic1 = the burned account, data = the burned amount.
+inline Receipt::Log
+make_eip7708_burn_log(Address const &burned, uint256_t const &amount)
+{
+    static constexpr bytes32_t signature =
+        abi_encode_event_signature("Burn(address,uint256)");
+    static_assert(
+        signature ==
+        bytes32_from_hex("cc16f5dbb4873280815c1ee09dbd06736cffcc184412cf7a71a0"
+                         "fdb75d397ca5"));
+
+    return EventBuilder(SYSTEM_ADDRESS, signature)
+        .add_topic(abi_encode_address(burned))
+        .add_data(abi_encode_uint(u256_be{amount}))
+        .build();
+}
 
 class EvmcHostBase : public vm::Host
 {
@@ -155,6 +189,21 @@ struct EvmcHost final : public EvmcHostBase
     {
         MONAD_TRY
         {
+            // EIP-7708: an account created and destroyed in the same
+            // transaction that sends its balance to itself destroys that
+            // balance, so it emits a Burn log inline (a Transfer does not
+            // apply: from == to) before selfdestruct zeroes it. Residual
+            // balance arriving after this point is burned at finalization
+            // instead.
+            //
+            // Read before the call rather than after only to avoid depending on
+            // State::selfdestruct's internals; destruct() sets a flag without
+            // touching the account, so either order gives the same answer.
+            bool created_this_tx = false;
+            if constexpr (traits::eip_7708_active()) {
+                created_this_tx = state_.is_current_incarnation(address);
+            }
+
             auto const [result, transferred_balance] =
                 state_.selfdestruct<traits>(address, beneficiary);
 
@@ -163,6 +212,19 @@ struct EvmcHost final : public EvmcHostBase
 
             emit_native_transfer_event(
                 address, beneficiary, transferred_balance);
+
+            if constexpr (traits::eip_7708_active()) {
+                if (created_this_tx && beneficiary == address &&
+                    transferred_balance > 0) {
+                    // A burn is value destruction, not a transfer between
+                    // accounts, so it gets no call frame: unlike every other
+                    // log, it is stored on the receipt without a matching
+                    // call_tracer_.on_log. Consistent with the finalization
+                    // burns, which have no frame to attach to at all.
+                    state_.store_log(
+                        make_eip7708_burn_log(address, transferred_balance));
+                }
+            }
 
             return result;
         }
@@ -244,9 +306,14 @@ struct EvmcHost final : public EvmcHostBase
     {
         // Skip emitting native transfer events when no value is transferred or
         // `from` and `to` are the same account (i.e. no net transfer of funds).
-        if (log_native_transfers_ && value > 0 && from != to) {
-            static constexpr Address native_token_address =
-                0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address;
+        //
+        // EIP-7708 makes this a consensus rule from the activation revision;
+        // before then it fires only for eth_simulate's traceTransfers
+        // (log_native_transfers_).
+        constexpr bool consensus = traits::eip_7708_active();
+        if ((consensus || log_native_transfers_) && value > 0 && from != to) {
+            constexpr Address log_address =
+                consensus ? SYSTEM_ADDRESS : SIMULATE_NATIVE_TOKEN_LOG_ADDRESS;
             static constexpr bytes32_t signature =
                 abi_encode_event_signature("Transfer(address,address,uint256)");
             static_assert(
@@ -254,7 +321,7 @@ struct EvmcHost final : public EvmcHostBase
                 bytes32_from_hex("ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a"
                                  "11628f55a4df523b3ef"));
 
-            auto event = EventBuilder(native_token_address, signature)
+            auto event = EventBuilder(log_address, signature)
                              .add_topic(abi_encode_address(from))
                              .add_topic(abi_encode_address(to))
                              .add_data(abi_encode_uint(u256_be{value}))
--- a/category/execution/ethereum/state3/state.cpp
+++ b/category/execution/ethereum/state3/state.cpp
@@ -466,9 +466,22 @@ EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
 
 // YP (87)
 template <Traits traits>
-void State::destruct_suicides()
+std::vector<std::pair<Address, uint256_t>> State::destruct_suicides()
 {
     MONAD_ASSERT(!version_);
+
+    // The EIP-7708 capture below sits in the CANCUN+ branch of the loop, since
+    // only EIP-6780 leaves same-tx-created-and-destroyed accounts to burn.
+    // Enforce that coupling rather than trusting it: activating 7708 on an
+    // earlier revision would silently skip every burn instead of failing.
+    static_assert(
+        !traits::eip_7708_active() || traits::evm_rev() >= MONAD_ETH_CANCUN);
+
+    // EIP-7708: the residual balance of an account created and destroyed in the
+    // same transaction is burned here. Capture (address, balance) before the
+    // reset so the caller can emit a Burn log; returned sorted by address
+    // (empty when the rule is inactive).
+    std::vector<std::pair<Address, uint256_t>> burned;
 
     for (auto &it : current_) {
         auto &stack = it.second;
@@ -482,11 +495,27 @@ void State::destruct_suicides()
             }
             else {
                 if (account->incarnation == incarnation_) {
+                    if constexpr (traits::eip_7708_active()) {
+                        if (account->balance > 0) {
+                            burned.emplace_back(it.first, account->balance);
+                        }
+                    }
                     account.reset();
                 }
             }
         }
     }
+
+    // current_ iterates in an unspecified order, but the Burn logs go into the
+    // consensus receipt, so sort by address for a deterministic emission order.
+    if constexpr (traits::eip_7708_active()) {
+        std::sort(
+            burned.begin(), burned.end(), [](auto const &a, auto const &b) {
+                return a.first < b.first;
+            });
+    }
+
+    return burned;
 }
 
 EXPLICIT_TRAITS_MEMBER(State::destruct_suicides);
--- a/category/execution/ethereum/test/test_call_trace.cpp
+++ b/category/execution/ethereum/test/test_call_trace.cpp
@@ -25,6 +25,7 @@
 #include <category/execution/ethereum/core/account.hpp>
 #include <category/execution/ethereum/core/contract/abi_encode.hpp>
 #include <category/execution/ethereum/core/contract/abi_signatures.hpp>
+#include <category/execution/ethereum/core/receipt.hpp>
 #include <category/execution/ethereum/db/trie_db.hpp>
 #include <category/execution/ethereum/db/util.hpp>
 #include <category/execution/ethereum/evmc_host.hpp>
@@ -206,6 +207,24 @@ TYPED_TEST(TraitsTest, execute_success)
         .depth = 0,
         .logs = std::vector<CallFrame::Log>{},
     };
+
+    if constexpr (TestFixture::Trait::eip_7708_active()) {
+        // EIP-7708: the top-level value transfer emits a Transfer log on the
+        // consensus path (no trace_transfers needed), from SYSTEM_ADDRESS.
+        expected.logs->push_back(
+            {{
+                 .data =
+                     byte_string{store_be_as<bytes32_t, uint256_t>(0x10000)},
+                 .topics =
+                     std::vector{
+                         0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef_bytes32,
+                         abi_encode_address(sender),
+                         abi_encode_address(ADDR_B),
+                     },
+                 .address = SYSTEM_ADDRESS,
+             },
+             0});
+    }
 
     EXPECT_EQ(call_frames[0], expected);
 }
@@ -785,7 +804,9 @@ TYPED_TEST(TraitsTest, simulate_v1_trace)
                         0x0000000000000000000000000000000000000000000000000000000000000100_bytes32,
                         0x0000000000000000000000000000000000000000000000000000000000000101_bytes32,
                     },
-                .address = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address,
+                .address = TestFixture::Trait::eip_7708_active()
+                               ? SYSTEM_ADDRESS
+                               : SIMULATE_NATIVE_TOKEN_LOG_ADDRESS,
             },
             0,
         }},
@@ -891,7 +912,9 @@ TYPED_TEST(TraitsTest, simulate_v1_trace_selfdestruct)
                     0x0000000000000000000000000000000000000000000000000000000000000101_bytes32,
                     0x0000000000000000000000000000000000000000000000000000000000000102_bytes32,
                 },
-            .address = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address,
+            .address = TestFixture::Trait::eip_7708_active()
+                           ? SYSTEM_ADDRESS
+                           : SIMULATE_NATIVE_TOKEN_LOG_ADDRESS,
         },
         1, // position: after the selfdestruct sub-frame
     };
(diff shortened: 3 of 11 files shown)


category/execution/ethereum/evmc_host.hpp
```diff
@@ -244,17 +306,22 @@ struct EvmcHost final : public EvmcHostBase
...
-            static constexpr Address native_token_address =
-                0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address;
+        //
+        // EIP-7708 makes this a consensus rule from the activation revision;
+        // before then it fires only for eth_simulate's traceTransfers
+        // (log_native_transfers_).
+        constexpr bool consensus = traits::eip_7708_active();
+        if ((consensus || log_native_transfers_) && value > 0 && from != to) {
+            constexpr Address log_address =
+                consensus ? SYSTEM_ADDRESS : SIMULATE_NATIVE_TOKEN_LOG_ADDRESS;
```
Baltoli: This behaviour should be confirmed against other client implementations: my previous mental model had been that the logs would appear from _both_ addresses when run with `eth_simulateV1` and Amsterdam activated, but let's make sure that we have this correct in either direction. The spec tests won't cover `eth_simulateV1` so there's the potential for a silent bug here.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -47,6 +48,39 @@ static_assert(alignof(vm::Host) == 8);
 
 class BlockHashBuffer;
 
+// The system sender: the caller the EIP-7002/EIP-7251 system calls execute as,
+// and the emitter EIP-7708 attributes its synthetic logs to. Not owned by any
+// one of them.
+inline constexpr Address SYSTEM_ADDRESS =
+    0xfffffffffffffffffffffffffffffffffffffffe_address;
```
Baltoli: This constant is defined elsewhere (requests processing; we should hoist it to a shared location rather than redefining it).

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -891,7 +912,9 @@ TYPED_TEST(TraitsTest, simulate_v1_trace_selfdestruct)
                     0x0000000000000000000000000000000000000000000000000000000000000101_bytes32,
                     0x0000000000000000000000000000000000000000000000000000000000000102_bytes32,
                 },
-            .address = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_address,
+            .address = TestFixture::Trait::eip_7708_active()
+                           ? SYSTEM_ADDRESS
+                           : SIMULATE_NATIVE_TOKEN_LOG_ADDRESS,
```
Baltoli: Same comment re. `eth_simulateV1` / spec tests

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -1552,3 +1575,240 @@ TYPED_TEST(TraitsTest, simulate_v1_trace_transfers)
...
     }
 }
+
+namespace
+{
+    Receipt::Log eip7708_transfer_log(
+        Address const &from, Address const &to, uint256_t const &value)
+    {
+        return {
+            .data = byte_string{store_be_as<bytes32_t, uint256_t>(value)},
```
Baltoli: Minor nitpick: I think the second template argument can be inferred here from the type of `value`

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -1552,3 +1575,240 @@ TYPED_TEST(TraitsTest, simulate_v1_trace_transfers)
...
+    }
+}
+
+// EIP-7708 is active on MonadTraits from MONAD_NEXT and on EvmTraits from
+// MONAD_ETH_AMSTERDAM. The shared TraitsTest matrix covers the MonadTraits
+// half, so the consensus-path (eip_7708_active()) branches in the cases above
+// do run. AMSTERDAM sits above LATEST_SUPPORTED_EVM_FORK and is excluded from
+// the EvmTraits half of that matrix, so pin a fixture to EvmTraits<AMSTERDAM>
+// to cover it too. Removable once LATEST_SUPPORTED_EVM_FORK includes
+// AMSTERDAM.
```
Baltoli: As part of this PR stack, we should just bump the latest supported EVM fork to Amsterdam rather than do this extra fixture

category/vm/evm/traits.hpp
```diff
@@ -194,6 +195,21 @@ namespace monad
...
+        // EIP-7708: ETH transfers emit a Transfer log. A consensus state-
+        // transition rule (not an opcode); read in the execution layer's
+        // emit_native_transfer_event. Active on AMSTERDAM for unit tests.
+        //
+        // Scope is the EIP's four sites -- tx-level transfer, CALL,
+        // SELFDESTRUCT, CREATE/CREATE2 -- each nonzero-value and to a different
+        // account. Withdrawals (not attached to a transaction, so no natural
+        // emission point) and coinbase/base fees (a distinct category) were
+        // considered by the EIP and excluded: those paths stay silent by
+        // design.
```
Baltoli: Same comment as previous PRs: this comment is too much detail. The trait enablement flag isn't the place to be flagging details about particular EIPs

category/vm/evm/traits.hpp
```diff
@@ -378,6 +394,14 @@ namespace monad
             return evm_rev() >= MONAD_ETH_AMSTERDAM;
         }
 
+        // EIP-7708: value transfers emit a Transfer log, plus monad's Burn log
+        // at finalization for an EIP-6780 same-tx-created-and-destroyed
+        // account. See the EvmTraits overload for the scope the EIP defines.
```
Baltoli: As above

test/ethereum_test/exclude/MONAD_NEXT_amsterdam.cmake
```diff
@@ -13,11 +13,13 @@
...
+# EIP-7708 is the last of the stacked Amsterdam EIPs, so the allowlist opens to
+# the whole suite. It could not open any sooner: the bundle is generated with
+# EIP-7708 active, so every fixture that moves ETH carries a Transfer log in its
+# expected logs bloom, and the 109 fixtures outside amsterdam/ (reserve_balance,
+# mip4_checkreservebalance, eip6780_selfdestruct, eip7702_set_code_tx, ...) fail
+# with "wrong logs bloom" until the rule is live -- however complete the rest of
+# the fork is. Keep using the exclusion list below for individual fixtures that
+# are broken for an unrelated reason.
+set(MONAD_NEXT_amsterdam_included_tests "*")
 set(MONAD_NEXT_amsterdam_excluded_tests "")
```
Baltoli: The merged version of this PR stack shouldn't change the CMake behaviour - when cleaning up, this should be reverted back to just excluded tests

Baltoli:
> Burn(address,uint256) (topic0 0xcc16f5db..), a monad-specific extension, for ETH destroyed when an EIP-6780 same-tx-created-and-destroyed account is deleted

why is there a Monad-specific extension needed?

category/execution/ethereum/test/test_call_trace.cpp
ryankeleti: When we bump `LATEST_SUPPORTED_EVM_FORK` it seems we need to also bump deps like `evmone`, maybe another PR on this stack?

Baltoli: We're removing evmone soon anyway, and I don't think it supports Osaka (we just don't have those revisions in the differential fuzzer)

ryankeleti: Looking at [this comment on line 47](https://github.com/category-labs/monad/blob/c143f5cab38556afc2a348527797424280ecb0e9/category/vm/evm/revision.cpp#L47), we'll need to bump here (required for `vm-unit-tests`). evmone is being removed, but evmc might need updating?

brett-monad:
@claude review this PR

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -695,8 +714,50 @@ TYPED_TEST(TraitsTest, selfdestruct_depth)
...
+    {
+        return Traits::eip_7708_active() ? 2 * n + 1 : n;
+    }
+
+    // Number of traced log entries produced by n value transfers.
+    template <class Traits>
+    constexpr size_t transfer_log_count(size_t const n)
+    {
+        return Traits::eip_7708_active() ? 2 * n : n;
+    }
```
Baltoli, line 752:
```suggestion
    template <Traits traits>
    constexpr size_t synthetic_log_index(size_t const n)
    {
        return traits::eip_7708_active() ? 2 * n + 1 : n;
    }

    // Number of traced log entries produced by n value transfers.
    template <Traits traits>
    constexpr size_t transfer_log_count(size_t const n)
    {
        return traits::eip_7708_active() ? 2 * n : n;
    }
```

category/execution/ethereum/state3/state.cpp
```diff
@@ -466,10 +466,20 @@ EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
 
 // YP (87)
 template <Traits traits>
-void State::destruct_suicides()
+std::vector<std::pair<Address, uint256_t>> State::destruct_suicides()
```
Baltoli: Maybe there's a case to be made for implementing 8246 after all if it simplifies this code?

category/execution/ethereum/state3/state.cpp
```diff
@@ -466,10 +466,20 @@ EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
...
 
+    // The capture below sits in the CANCUN+ branch, since only EIP-6780 leaves
+    // same-tx-created-and-destroyed accounts to burn. Assert the coupling: on
+    // an earlier revision 7708 would silently skip every burn rather than fail.
+    static_assert(
+        !traits::eip_7708_active() || traits::evm_rev() >= MONAD_ETH_CANCUN);
+
+    // EIP-7708: capture (address, balance) before the reset below so the caller
+    // can emit a Burn log for the residual balance destroyed here.
+    std::vector<std::pair<Address, uint256_t>> burned;
```
Baltoli: i.e. if we had 8246 then we wouldn't need this - I hadn't appreciated the interaction previously

category/execution/ethereum/evmc_host.hpp
```diff
@@ -239,29 +284,59 @@ struct EvmcHost final : public EvmcHostBase
...
-            static_assert(
-                signature ==
-                bytes32_from_hex("ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a"
-                                 "11628f55a4df523b3ef"));
-
-            auto event = EventBuilder(native_token_address, signature)
+        // Pre-activation nothing is emitted unless eth_simulate asked for it,
+        // so short-circuit on the flag before comparing value and addresses.
+        if constexpr (!traits::eip_7708_active()) {
+            if (!log_native_transfers_) {
```
dhil, line 261: Maybe a branch hint here. We expect this branch to be taken most of the time.
```suggestion
            if (MONAD_LIKELY(!log_native_transfers_)) {
```

category/rpc/monad_executor_test.cpp
```diff
@@ -293,7 +295,23 @@ namespace
                 gas_specified ? tx.gas_limit : MONAD_ETH_CALL_LOW_GAS_LIMIT,
             .status = EVMC_SUCCESS,
             .depth = 0,
-            .logs = std::vector<CallFrame::Log>{},
+            // EIP-7708 is active on this chain, so the value transfer emits a
+            // consensus Transfer log from SYSTEM_ADDRESS. It reaches the call
+            // trace too -- emit_native_transfer_event pairs store_log with
```
dhil, line 301: This is confusing. What is the `emit_native_transfer_event`? Is this being confused with `emit_native_transfer_logs` from `eth_simulateV1`? This is a test case of `eth_call`. It does not emit transfer logs without the EIP being active.

category/execution/ethereum/state3/state.cpp
brett-monad: Verified that we will be implementing 8246

brett-monad:
Note for the rebase onto the EIP-8246 branch, which sits below this one: **the Burn log becomes unreachable the moment 8246 lands.**

`eip_7708_active()` and `eip_8246_active()` are the same predicate for both trait families — both are `>= MONAD_ETH_AMSTERDAM`. So with 8246 underneath:

- `destruct_suicides`' `if constexpr (eip_7708_active())` capture into `burned` can only fire when 8246 is off, which is never when 7708 is on.
- The inline selfdestruct-to-self Burn emit in `evmc_host.hpp` is likewise unreachable, along with `make_eip7708_burn_log` and the `Burn(address,uint256)` topic.

Since 8246 lands below rather than above, the clean outcome is that the rebased version of this PR never introduces the Burn log at all, rather than adding it and deleting it a commit later. That also means the `std::ignore` call sites in `test_state.cpp` and the `Eip7708StateTest` / `Eip7708FinalizationTest` burn cases never get written.

Two things to fix while rebasing: `make_eip7708_burn_log`'s comment currently says EIP-8246 "would remove both burns; monad has declined it", which the 8246 branch contradicts; and the exclusions for the fixtures 8246 invalidates belong in this PR's `MONAD_NEXT_amsterdam.cmake` rather than the 8246 PR's, because this is the PR that removes the allowlist. Test names there are `fs::relative` paths, so they begin `for_monad_next/` — the `osaka.cmake` form omits that level and matches nothing.

Consistent with geth, whose Amsterdam path emits only the transfer log, and only when the beneficiary differs from the destructing account.

brett-monad:
Done — recording it here so nobody actions the comment above twice.

EIP-8246 now sits directly beneath this PR as #2512, and this PR no longer introduces the `Burn` log at all rather than adding it and deleting it a commit later. `make_eip7708_burn_log`, both emit sites, the `destruct_suicides` vector return (back to `void`), the 23 `std::ignore` call sites and the three burn test suites are all gone; there are no `cc16f5db` references left in the tree.

The stale comment claiming monad had declined 8246 went with it, as did the PR title and description, which said the same thing.

On the fixtures: the exclusions landed here rather than in #2512, for the reason given above — this is the PR that removes the allowlist, so this is where they become live. Measured rather than predicted: 601 of 617 pass, and the 15 that fail are exactly the set carrying the pre-8246 SELFDESTRUCT expectations. Both `burn_logs` fixtures that assert the *absence* of a burn pass and are deliberately not excluded. Names use the `for_monad_next/` prefix, since they are `fs::relative` paths — the `osaka.cmake` form omits that level and would have matched nothing.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -47,6 +47,14 @@ static_assert(alignof(vm::Host) == 8);
 
 class BlockHashBuffer;
 
+// Sender for the EIP-7002/EIP-7251 system calls; emitter for EIP-7708's logs.
+inline constexpr Address SYSTEM_ADDRESS =
+    0xfffffffffffffffffffffffffffffffffffffffe_address;
```
dhil, line 54: This constant is defined in `process_requests.cpp`. I think it makes sense to either move it to `core/address.hpp` or define a new header file with bespoke addresses. I am **not** suggesting we do that in this PR, but it is something we should clean up subsequently.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -239,29 +247,57 @@ struct EvmcHost final : public EvmcHostBase
...
                              .add_data(abi_encode_uint(u256_be{value}))
                              .build();
 
             state_.store_log(event);
             call_tracer_.on_log(std::move(event));
+        };
+
+        // Consensus artifact, so not conditioned on log_native_transfers_.
+        if constexpr (traits::eip_7708_active()) {
+            emit(SYSTEM_ADDRESS);
```
dhil, line 319: I think Geth implements the other emit order, i.e. synthetic ERC-7528 logs first, and then EIP-7708 logs.

Note, it reads like they are moving towards deprecating synthetic logs when EIP-7708 is active, c.f. https://github.com/ethereum/execution-apis/pull/881. Though, I suggest we hold off implementing this until the change has landed in Geth.

category/rpc/monad_executor_test.cpp
```diff
@@ -294,7 +296,23 @@ namespace
...
+            // call_tracer_.on_log, as the LOG opcodes do.
+            .logs =
+                std::vector<CallFrame::Log>{
+                    {Receipt::Log{
+                         .data = byte_string{store_be_as<bytes32_t, uint256_t>(
+                             0x10000)},
+                         .topics =
+                             std::vector{
+                                 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef_bytes32,
```
goodlyrottenapple: can we make this a constant, perhaps with a comment that this is the keccaked `Transfer(address,address,uint256)` event?

goodlyrottenapple: ah, actually we already have `static constexpr bytes32_t transfer_sig` for this further down

category/execution/ethereum/evmc_host.hpp
brett-monad: Agreed, as a follow-up once this lands.

category/execution/ethereum/evmc_host.hpp
brett-monad: Agreed. Went ahead and flipped them, synthetic first; may need an update once 881 lands in geth.

category/rpc/monad_executor_test.cpp
brett-monad: Fixed.

dhil (dismissed):

category/execution/ethereum/test/test_call_trace.cpp
```diff
@@ -70,6 +73,22 @@ namespace
...
+        ASSERT_EQ(state.logs().size(), frame_logs.size());
+        for (size_t i = 0; i < frame_logs.size(); ++i) {
+            EXPECT_EQ(state.logs()[i], frame_logs[i].log);
+        }
+    }
+
+    // The eip_7708_active() branches below only test anything if the matrix
+    // has revisions where the rule is on.
+    static_assert(EvmTraits<MONAD_ETH_AMSTERDAM>::eip_7708_active());
+    static_assert(MonadTraits<MONAD_NEXT>::eip_7708_active());
```
Baltoli, line 91: We don't generally include this kind of static assertion; the feature flags in the test are enough.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -48,6 +49,14 @@ static_assert(alignof(vm::Host) == 8);
 
 class BlockHashBuffer;
 
+// Sender for the EIP-7002/EIP-7251 system calls; emitter for EIP-7708's logs.
+inline constexpr Address SYSTEM_ADDRESS =
+    0xfffffffffffffffffffffffffffffffffffffffe_address;
```
Baltoli, line 54: I might call this `ETH_SYSTEM_ADDRESS` just to be explicit. There's also a Monad system contract / address (the sender of system calls).

category/execution/ethereum/evmc_host.hpp
```diff
@@ -48,6 +49,14 @@ static_assert(alignof(vm::Host) == 8);
...
+
+// ERC-7528 pseudo-address: eth_simulate's traceTransfers emitter, any revision.
```
Baltoli, line 56: This comment is overstated; the address is _just_ from ERC-7528 and this definition isn't 1:1 linked to eth_simulate

category/execution/ethereum/evmc_host.hpp
```diff
@@ -256,29 +265,58 @@ struct EvmcHost final : public EvmcHostBase
         return call_tracer_;
     }
 
+    // Not every ETH movement emits. The EIP excludes withdrawals (not attached
+    // to a transaction, so no natural emission point), priority fees and the
+    // base-fee burn (derivable from the header). Beyond the EIP, monad's
+    // staking contract moves, mints and burns ETH by direct balance mutation
+    // rather than a value transfer, so none of that emits either.
```
Baltoli, line 272: Did we explicitly settle the staking contract choice when building `eth_simulateV1` @dhil? It seems analogous to withdrawals etc. from Ethereum, but I wanted to check that explicitly.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -256,29 +265,58 @@ struct EvmcHost final : public EvmcHostBase
...
+        if constexpr (!traits::eip_7708_active()) {
+            if (MONAD_LIKELY(!log_native_transfers_)) {
+                return;
+            }
+        }
+
+        // Skip when no value moves, or from == to so there is no net transfer.
+        // EIP-7708 words its cases the same way: nonzero-value-transferring, to
+        // a different account. One predicate serves both log kinds below, and
+        // most messages carry no value, so this exit is the common path.
```
Baltoli, line 287: Comment is kind of stating the obvious; I'd delete

category/execution/ethereum/evmc_host.hpp
```diff
@@ -256,29 +265,58 @@ struct EvmcHost final : public EvmcHostBase
...
+        if (MONAD_LIKELY(value == 0 || from == to)) {
+            return;
+        }
+
+        static constexpr bytes32_t signature =
+            abi_encode_event_signature("Transfer(address,address,uint256)");
+        static_assert(
+            signature ==
+            bytes32_from_hex("ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a"
+                             "11628f55a4df523b3ef"));
```
Baltoli, line 297: I'd probably hoist this out of the function and call it `ERC20_TRANSFER_SIGNATURE` or something - that definition could probably live in a different place as well, rather than in the host

category/execution/ethereum/evmc_host.hpp
```diff
@@ -256,29 +265,58 @@ struct EvmcHost final : public EvmcHostBase
...
+        };
+
+        // eth_simulate's ERC-7528 synthetic, kept alongside the consensus log
+        // and emitted first: geth's observed order for calls and creates (its
+        // SELFDESTRUCT differs), not specified anywhere.
+        if (log_native_transfers_) {
+            emit(SIMULATE_NATIVE_TOKEN_LOG_ADDRESS);
+        }
+
+        // Consensus artifact, so not conditioned on log_native_transfers_.
```
Baltoli, line 317: Comment isn't useful

category/execution/ethereum/evmc_host.hpp
```diff
@@ -256,29 +265,58 @@ struct EvmcHost final : public EvmcHostBase
...
                              .add_data(abi_encode_uint(u256_be{value}))
                              .build();
 
             state_.store_log(event);
             call_tracer_.on_log(std::move(event));
+        };
+
+        // eth_simulate's ERC-7528 synthetic, kept alongside the consensus log
+        // and emitted first: geth's observed order for calls and creates (its
+        // SELFDESTRUCT differs), not specified anywhere.
```
Baltoli, line 312:
```suggestion
        // Geth emits eth_simulate transfers before EIP-7708 (implementation detail; unspecified)
```

category/rpc/monad_executor_test.cpp
```diff
@@ -298,7 +303,25 @@ namespace
                 gas_specified ? tx.gas_limit : MONAD_ETH_CALL_LOW_GAS_LIMIT,
             .status = MONAD_STATUS_SUCCESS,
             .depth = 0,
-            .logs = std::vector<CallFrame::Log>{},
+            // CHAIN_CONFIG_MONAD_DEVNET runs MONAD_NEXT, where EIP-7708 is
+            // active, so this eth_call's value transfer emits the consensus
+            // Transfer log from SYSTEM_ADDRESS. It is not eth_simulateV1's
+            // synthetic: that is the emit_native_transfer_logs flag, off here.
+            // EvmcHost::emit_native_transfer_event pairs store_log with
+            // call_tracer_.on_log, so the log reaches the call trace too.
```
Baltoli, line 311: Comment isn't very useful - would delete or reduce this heavily as well

category/rpc/monad_executor_test.cpp
```diff
@@ -6963,10 +6986,43 @@ TEST_F(EthCallFixture, eth_simulate_v1_deploy_and_call)
...
     EXPECT_EQ(output[1]["calls"][0]["returnData"], expected_return_data);
 
-    ASSERT_EQ(output[1]["calls"][0]["logs"].size(), 1);
-    EXPECT_EQ(output[1]["calls"][0]["logs"][0]["data"], expected_return_data);
-    EXPECT_EQ(output[1]["calls"][0]["logs"][0]["topics"].size(), 0);
-    EXPECT_EQ(output[1]["calls"][0]["logs"][0]["blockNumber"], "0x3");
+    // EIP-7708: both value transfers now emit consensus Transfer logs from
+    // SYSTEM_ADDRESS, so the contract's own LOG0 is bracketed by one for the
+    // top-level transfer (sender -> deployed, 10 MON) and one for the internal
+    // call the contract makes (deployed -> beneficiary, 5 MON).
```
Baltoli, line 6992: This is a better comment

category/execution/ethereum/evmc_host.hpp
dhil: No we did not. What semantics do we want here?

category/execution/ethereum/evmc_host.hpp
dhil: There is something nice about it being here which is that this is the only place it is used, so in that sense it is local. I tend to agree with abstracting, naming, etc, however, since it is used only once I am not convinced it is worth it (and this particular instance will disappear with the revamped tracing stuff).

category/execution/ethereum/evmc_host.hpp
Baltoli: Well - the argument here for Ethereum is that there is no transaction associated with withdrawals etc., so there's nowhere to hang logs from (and the transfers are derivable from the headers there anyway). I _think_ this is different in Monad, because staking operations happen via system calls? If there is a transaction "in hand" when we do staking-related balance transfers, then it would make sense to emit the logs for them, even if it's a system transaction.

dhil: I agree. I am inclined to believe that right now balance transfers done via system calls are not observable in the `eth_simulateV1` transfer log. I will test it out, and draft a patch to fix the semantics if needed.

Baltoli: I guess that property is sort of inert at present though, because we don't allow simulation of system calls? It becomes relevant here because they're consensus artifacts

category/execution/ethereum/evmc_host.hpp
Baltoli: Sure, fine by me if this will disappear later

category/execution/ethereum/evmc_host.hpp
dhil: Protocol-wise, I am inclined to believe that we should also emit a transfer log. Then I agree with your other comment about hoisting out the emission logic, because it should be used to emit transfer events inside staking too.

Baltoli: Yep, agreed - thanks for the sense check

category/execution/ethereum/test/test_call_trace.cpp
brett-monad: Removed.

category/execution/ethereum/evmc_host.hpp
brett-monad: Renamed.

category/execution/ethereum/evmc_host.hpp
brett-monad: Reworded.

category/execution/ethereum/evmc_host.hpp
brett-monad: Removed.

category/execution/ethereum/evmc_host.hpp
brett-monad: Removed.

category/execution/ethereum/evmc_host.hpp
brett-monad: Fixed.

category/rpc/monad_executor_test.cpp
brett-monad: Reduced to one line.

PR #2472, opened 2026-07-31
build: emit rerun-if-changed for cmake sources in cargo build scripts

eclark0426:


maxkozlovsky (dismissed):
Approving, 
non-blocking suggestion - exclude test files from the watch list, third_party alone has >50K test files which aren't needed for dependency checks

andr-dev (approved):

maxkozlovsky (approved):

Baltoli (approved):

PR #2473, opened 2026-08-03
New OffsetTrie backend for PartialTrieDb - zero-copy loading from the witness binary

goodlyrottenapple:
**Goal:** replace the flat, hash-keyed RLP node list in the witness with a **pre-resolved, offset-linked 
node graph** the ZK guest reads **in place** — no in-guest hash indexing, no per-node RLP decode, 
no per-node allocation, except for writes on commit, which use a copy on write hash map overlay. When
calculating the state root and walking the trie, we check the overlay first.

The nodes are indexed via a `uint32_t` `NodeId` offset. The `NodeId` space is partitioned by the high bit: 
blob offsets live below it, fresh overlay ids at/above it. Real witnesses are ~MBs, far under 2 GiB imposed by
the partition.

## Conventions

- **Little-endian** for all *structural* integers (offsets, lengths, counts, tags).
- **Offsets** are **absolute, from byte 0 of the node region**, stored as
  `uint_32t`. Witnesses are ~10 MB, so `uint_32t` (4 GiB) is ample. `0` = **null / empty** (byte 0
  is `magic`, never a node, so `0` is a safe sentinel).
- **Account values** (`nonce`, `balance`) are stored **big-endian, full width** so the RLP
  encoder (`encode_to`) can `zeroless_view` them directly — no `bswap`. Slot values
  likewise. **Hashes** (`code_hash`, digest) are raw 32-byte arrays.
- **Paths** (nibbles) are packed 2/byte, **first nibble in the high 4 bits** — the
  `get_nibble`/`set_nibble` convention (`category/core/nibble.h`: nibble `n` is the high
  half of byte `n/2` when `n` is even). The guest builds a zero-copy view straight over the
  buffer with `mpt::NibblesView{/*begin=*/0, /*end=*/path_nlen, path_ptr}`
- **Statically sized fields come first**. Fields such as `child`/`value`/`storage` are fixed size
  and therefore come first, as decoding them does not require reading the value under the pointer

## Header (node region, bytes `[0, 8)`)

```
off  size  field        meaning
0    4     magic        'M','Z','W', 0x01   (0x01 = format version)
4    4     root_off     u32  offset of the ACCOUNT-TRIE root node (0 = empty trie)
```
---

## Node encodings

Every node begins with a 1-byte `tag`. Nodes are **tightly packed** (no padding /
alignment). There are six tags:

### Empty — `tag = 'M'`

pointer at the header, which starts with an 'M'. This keeps the invariant that a `NULL_ID` 
points to a "null"/empty node, so we can have a total `get_current` function and pattern 
match on the result, where a possible option is a null/empty value.

### Branch — `tag = 0x00`
```
u8      tag = 0x00
u32[16] children_offs        16 child offsets; 0 = absent slot
```
Branches carry **no value** — valid for the account and storage
tries (fixed 64-nibble hashed keys never terminate at a branch).

### Extension — `tag = 0x01`
```
u8    tag = 0x01
u32   child_off         child node offset (MUST be non-null)
u8    path_len         
...   path              ceil(path_len/2) bytes
```

### Leaf, account value — `tag = 0x02`
```
u8    tag = 0x02
u32   storage_off       storage-trie root id; 0 = no storage subtree present
b264  code_hash_rlp
u8    acct_rlp_len      length of the account-RLP blob
...   acct_rlp          rlp(nonce) || rlp(balance)
u8    path_len
...   path              ceil(path_len/2) bytes

```

The account code hash, nonce and balance are RLP encoded, Since the code hash is always a fixed 33B
RLP string, it is stored before the variable length nonce and balance. These are prepended with a single byte indicating their length.

### Leaf, storage value — `tag = 0x03`
```
u8    tag = 0x03
u256  value             big-endian (32 bytes, the slot value)
u8    path_len
...   path              ceil(path_len/2) bytes
```

### Digest (unresolved subtree) — `tag = 0xa0`
```
u8    tag = 0xa0
b256  hash              32 raw bytes — keccak of the omitted subtree's RLP
```

We use the tag `0xa0` because it is the RLP string length encoding for the hash `0x80 + 32 (KECCAK256_SIZE)`, hence digest is a valid RLP string we can memcpy when comping the state root hash without re-encoding.

### DIFF
--- a/category/execution/ethereum/db/offset_trie.cpp
+++ b/category/execution/ethereum/db/offset_trie.cpp
@@ -1,0 +1,823 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include <category/execution/ethereum/db/offset_trie.hpp>
+
+#include <category/core/assert.h>
+#include <category/core/byte_string.hpp>
+#include <category/core/bytes.hpp>
+#include <category/core/cases.hpp>
+#include <category/core/keccak.hpp>
+#include <category/core/nibble.h>
+#include <category/core/rlp/encode.hpp>
+#include <category/execution/ethereum/core/account.hpp>
+#include <category/execution/ethereum/core/rlp/bytes_rlp.hpp>
+#include <category/execution/ethereum/core/rlp/int_rlp.hpp>
+#include <category/execution/ethereum/rlp/encode2.hpp>
+#include <category/mpt/config.hpp>
+#include <category/mpt/merkle/compact_encode.hpp>
+#include <category/mpt/nibbles_view.hpp>
+
+#include <algorithm>
+#include <array>
+#include <bit>
+#include <cstdint>
+#include <cstring>
+#include <limits>
+#include <span>
+#include <utility>
+#include <vector>
+
+MONAD_MPT_NAMESPACE_BEGIN
+
+NodeId OffsetTrie::read_root(byte_string_view const blob)
+{
+    unsigned char const *const base = blob.data();
+    size_t const len = blob.size();
+    MONAD_ASSERT(len >= HEADER_LEN);
+    MONAD_ASSERT(
+        base[0] == 'M' && base[1] == 'Z' && base[2] == 'W' && base[3] == 0x01);
+    // Keep blob offsets and overlay ids in disjoint halves of the NodeId
+    // space (blob < OVERLAY_BASE, fresh ids >= OVERLAY_BASE). Bounding the
+    // blob size bounds every offset. Real witnesses are ~MBs; this only
+    // rejects a pathological >=2 GiB blob.
+    MONAD_ASSERT(len < OVERLAY_BASE);
+    NodeId const root = read_node_id(base + 4);
+    MONAD_ASSERT(
+        root == NULL_ID || (static_cast<uint64_t>(root) >= HEADER_LEN &&
+                            static_cast<uint64_t>(root) < len));
+    return root;
+}
+
+OffsetTrie::OffsetTrie(byte_string_view const blob)
+    : blob_(blob)
+    , root{read_root(blob_)}
+{
+    unsigned char const *const base = blob_.data();
+
+    // Prime hashes bottom-up over the blob's nodes (children precede
+    // parents), rejecting any node whose extent leaves the region.
+    unsigned char const *const region_end = blob_.end();
+    uint64_t node_offset = HEADER_LEN;
+    NodeViewBase node{base + node_offset};
+    // The only node carrying the EMPTY tag is the magic header at the NULL_ID
+    // offset, which this walk starts past. checked_end has no EMPTY arm, so
+    // encountering that tag in the blob data aborts as an invalid tag. However,
+    // we have to check the very first node isn't EMPTY (provided it exists).
+    MONAD_ASSERT(node.bytes() == region_end || node.tag() != EMPTY);
+    std::vector<uint64_t> node_offsets((blob_.size() + 63) / 64, 0);
+
+    auto const is_valid_offset = [&](NodeId c) {
+        uint64_t child_offset = static_cast<uint64_t>(c);
+        MONAD_ASSERT(
+            c == NULL_ID ||
+            (child_offset < node_offset && child_offset < blob_.size() &&
+             ((node_offsets[child_offset >> 6] >> (child_offset & 63)) & 1)));
+    };
+    unsigned char rlp_buf[MAX_NODE_RLP];
+
+    while (node.bytes() < region_end) {
+        // checked_end asserts that the current node does not reach past the end
+        // of the region
+        auto next_offset =
+            static_cast<uint64_t>(node.checked_end(region_end) - base);
+
+        match(
+            node,
+            Cases{
+                [&](BranchView b) {
+                    // Validate the 16 children two at a time: a uint64_t load
+                    // spans exactly one pair, low word first.
+                    static_assert(std::endian::native == std::endian::little);
+                    static_assert(
+                        sizeof(uint64_t) == 2 * sizeof(node_id_wire_t));
+                    unsigned char const *const p = b.payload();
+                    uint64_t pair;
+                    for (unsigned i = 0; i < 8; ++i) {
+                        std::memcpy(&pair, p + i * sizeof(pair), sizeof(pair));
+                        is_valid_offset(
+                            NodeId{static_cast<node_id_wire_t>(pair)});
+                        is_valid_offset(
+                            NodeId{static_cast<node_id_wire_t>(pair >> 32)});
+                    }
+                },
+                [&](ExtView e) { is_valid_offset(e.child()); },
+                [&](AccountLeafView a) { is_valid_offset(a.storage()); },
+                [](auto) {}});
+
+        match(
+            node,
+            Cases{
+                [](NullView) {},
+                [](DigestView) {},
+                [&](auto) {
+                    node_rlp_span const rem = encode_rlp<true>(
+                        node, node_rlp_span{rlp_buf}); // priming pass
+                    // Only hash-referenced nodes (canonical RLP >= 32 B)
+                    // are cached; smaller nodes are inlined by their
+                    // parent, so caching their hash would make child_ref
+                    // emit a 32-byte ref where the trie inlines it.
+                    if (rem.rlp_size() >= 32) {
+                        bytes32_t h;
+                        keccak256(rem.rlp_data(), rem.rlp_size(), h.bytes);
+                        hashes_.emplace(NodeId{node_offset}, h);
+                    }
+                }});
+
+        node_offsets[node_offset >> 6] |= 1ull << (node_offset & 63);
+        node = NodeViewBase{base + next_offset};
+        node_offset = next_offset;
+    }
+    MONAD_ASSERT(node.bytes() == region_end); // nodes tile exactly
+    is_valid_offset(root);
+}
+
+NodeViewBase OffsetTrie::find_original(NodeId id, NibblesView key) const
+{
+    NodeViewBase found = empty();
+    while (id != NULL_ID) {
+        id = match(
+            get_original(id),
+            Cases{
+                [&](BranchView b) -> NodeId {
+                    if (key.nibble_size() == 0) { // no value at a branch
+                        return NULL_ID;
+                    }
+                    NodeId const next = b.child(key.get(0));
+                    key = key.substr(1);
+                    return next;
+                },
+                [&](ExtView e) -> NodeId {
+                    NibblesView const ep = e.path();
+                    if (!key.starts_with(ep)) {
+                        return NULL_ID;
+                    }
+                    key = key.substr(ep.nibble_size());
+                    return e.child();
+                },
+                [&](AccountLeafView l) -> NodeId {
+                    if (l.path() == key) {
+                        found = l;
+                    }
+                    return NULL_ID;
+                },
+                [&](StorageLeafView l) -> NodeId {
+                    if (l.path() == key) {
+                        found = l;
+                    }
+                    return NULL_ID;
+                },
+                [](DigestView) -> NodeId {
+                    MONAD_ABORT("incomplete witness: lookup hit a Digest");
+                },
+                [](NullView) -> NodeId {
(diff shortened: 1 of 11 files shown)


category/execution/ethereum/db/offset_trie_db.cpp
```diff
@@ -0,0 +1,793 @@
...
+                                     // is the sub-root
+                }
+                store_.put_storage(sleaf, sleaf_path, sdelta.second);
+            }
+        }
+        for (auto const &[slot, sdelta] : delta.storage) {
+            if (sdelta.second == bytes32_t{} && sdelta.first != bytes32_t{}) {
+                auto const slot_key = keccak256(slot.bytes);
+                if (store_.erase_node(storage, mpt::NibblesView{slot_key}) ==
+                    TrieStore::EraseResult::Erased) {
```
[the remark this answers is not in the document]
goodlyrottenapple: no, this analysis doesn't make sense to me. we want to set the storage root to null if the erase_node deleted the whole tree...

category/mpt/merkle/compact_encode.hpp
```diff
@@ -37,7 +37,7 @@ compact_encode_len(unsigned const si, unsigned const ei)
 
 // Transform the nibbles to its compact encoding
 // https://ethereum.org/en/developers/docs/data-structures-and-encoding/patricia-merkle-trie/
-[[nodiscard]] constexpr byte_string_view compact_encode(
+constexpr void compact_encode_raw(
```
goodlyrottenapple, line 40: probably not the best name.... not sure what a better alternative might be tho

goodlyrottenapple:
@claude re-review please

category/execution/ethereum/db/offset_trie.hpp
```diff
@@ -0,0 +1,569 @@
...
+// + value slot + list header. 700 leaves margin.
+inline constexpr size_t MAX_NODE_RLP = 700;
+
+// A node's stable id. 0 = null; `n` in the [HEADER_LEN, blob_len) range is
+// a blob offset (unless shadowed by an overlay); n >= OVERLAY_BASE is a
+// fresh overlay node.
+//
+// The id is register-width, independent of the 32-bit child field the wire
+// format stores it in (node_id_wire below)
+enum class NodeId : uint64_t
```
goodlyrottenapple, line 71: We get better performance in 64bit RISC-V zkVMs when `NodeId` is 64-bits, even though the wire format is 32-bit wide.

category/execution/ethereum/db/test/test_offset_trie.cpp
```diff
@@ -0,0 +1,412 @@
...
+#include <vector>
+
+using namespace monad;
+using namespace monad::mpt;
+
+namespace
+{
+    mpt::Nibbles nibs_to_path(std::vector<uint8_t> const &nibs)
+    {
+        mpt::Nibbles p{(size_t)nibs.size()};
```
dhil:
```suggestion
        mpt::Nibbles p{static_cast<size_t>(nibs.size())};
```
Although, I'm inclined to believe the cast should be a no-op. I'd expect `.size()` to return a `size_t`, unless we are using a custom allocator somewhere.

category/execution/ethereum/db/test/test_offset_trie.cpp
```diff
@@ -0,0 +1,412 @@
...
+        for (size_t i = 0; i < nibs.size(); ++i) {
+            p.set((unsigned)i, nibs[i]);
```
dhil:
```suggestion
            p.set(static_cast<unsigned>(i), nibs[i]);
```

category/execution/ethereum/db/test/test_offset_trie.cpp
```diff
@@ -0,0 +1,412 @@
...
+    // and the offset bookkeeping live here.
+    struct BlobBuilder
+    {
+        byte_string buf;
+
+        BlobBuilder()
+        {
+            unsigned char const magic[4] = {'M', 'Z', 'W', 0x01};
+            buf.append(magic, 4);
+            buf.append(4, (unsigned char)0); // root_off placeholder
```
dhil:
```suggestion
            buf.append(4, static_cast<unsigned char>(0)); // root_off placeholder
```

dhil (commented):
Initial pass. Some comments about C style casts.

category/execution/ethereum/db/test/test_offset_trie.cpp
```diff
@@ -0,0 +1,415 @@
...
+    {
+        if (node_rlp.size() < 32) {
+            return node_rlp; // inline
+        }
+        return rlp::encode_string2(
+            byte_string_view{oracle_hash(node_rlp).bytes, 32});
+    }
+
+    // canonical RLP of an account leaf: list[ compact(path,leaf), string(acct)
+    // ]
```
dhil: Nit:
```suggestion
    // canonical RLP of an account leaf: 
    // list[ compact(path,leaf), string(acct) ]
```

category/execution/ethereum/db/test/test_offset_trie.cpp
```diff
@@ -0,0 +1,415 @@
...
+    {
+        byte_string buf;
+
+        BlobBuilder()
+        {
+            unsigned char const magic[4] = {'M', 'Z', 'W', 0x01};
+            buf.append(magic, 4);
+            // root_off placeholder
+            buf.append(4, static_cast<unsigned char>(0));
+        }
```
dhil, line 61: Nit: I think this can be simplified to
```suggestion
        BlobBuilder() : buf{'M', 'Z', 'W', 0x01, 0, 0, 0, 0} {}
```

category/execution/ethereum/db/partial_trie_db.cpp
```diff
@@ -13,871 +13,182 @@
...
+            if (is_changed(sdelta) && sdelta.second == bytes32_t{}) {
                 auto const slot_key = keccak256(slot.bytes);
-                trie_delete(storage, mpt::NibblesView{slot_key});
+                if (trie_.erase_node(storage, mpt::NibblesView{slot_key}) ==
+                    OffsetTrie::EraseResult::Erased) {
+                    storage = NULL_ID;
+                };
             }
         }
+
```
dhil: These two loops look fusible to me. Is there a reason to keep them separate other than that's allegedly what the reference does?

goodlyrottenapple: All upserts must happen first before the deletes, otherwise you could end up trying to do branch compression and hit a digest I think.

dhil: OK, lets keep them separate for now then.

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -0,0 +1,722 @@
...
+        root == NULL_ID || (static_cast<uint64_t>(root) >= HEADER_LEN &&
+                            static_cast<uint64_t>(root) < len));
+    return root;
+}
+
+OffsetTrie::OffsetTrie(byte_string_view const blob)
+    : blob_(blob)
+    , root{read_root(blob_)}
+{
+    unsigned char const *const base = blob_.data();
```
dhil, line 68: I am inclined to believe that we should check whether the blob is structurally sound before we decode. Otherwise I think this routine may accept malformed witnesses, that can invoke undefined behaviour by doing an out-of-bounds memory access in the raw memcpy in `StorageLeafView::value()`. For example, I think the witness `4d 5a 57 01 00 00 00 00 03` will cause UB, because:

* The header will check.
* `00 00 00 00` is the root offset `NULL_ID`.
* `03` is `LEAF_STORAGE` which causes `encode_rlp` to invoke. `StorageLeafView::value()` which performs a memcpy past the end.

I think this validation logic can be fused into the loop, e.g.

0. Maintain a set called, say, `seen_node_starts`.

1. At the start of each iteration, perform a bounded parse of the current node bytes. This must compute the node end and validate all tag-specific minimum lengths before any view/hash logic.

2. Record the current node start offset in `seen_node_starts`.

3. Validate every referenced child/storage offsets. Require each non-null reference to already be in `seen_node_starts`. Because the format is post-order, this enforces no forward references, no self-references, no interior stale references.

4. Run the existing priming/hash logic for that node.

5. Set `cursor` to point to the end of the parsed node.

6. After the loop assert `cursor == region_end`. Then validate root once `root == NULL_ID || seen_node_starts.find(root) != seen_node_starts.end()`

category/execution/ethereum/db/partial_trie_db.cpp
goodlyrottenapple: Added an explanatory comment

dhil (dismissed):

category/execution/ethereum/db/offset_trie.hpp
```diff
@@ -0,0 +1,614 @@
...
+{
+};
+inline constexpr NodeId NULL_ID{0};
+
+// Width of a child/root field in the node encoding
+using node_id_wire = uint32_t;
```
mkolosick: nit: `node_id_wire_t`

dhil (dismissed):

category/execution/ethereum/db/offset_trie.hpp
```diff
@@ -0,0 +1,615 @@
...
+    case LEAF_ACCT:
+        return v(AccountLeafView{n});
+    case LEAF_STORAGE:
+        return v(StorageLeafView{n});
+    case DIGEST:
+        return v(DigestView{n});
+    case EMPTY:
+        return v(NullView{n});
+    }
+    MONAD_ABORT("bad node tag");
```
mkolosick: change to 
```
default:
    MONAD_ABORT("bad node tag")
```
to please clang-tidy when I finish getting it working for the execution directory.

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -0,0 +1,759 @@
...
+    unsigned char const *const base = blob.data();
+    size_t const len = blob.size();
+    MONAD_ASSERT(len >= HEADER_LEN);
+    MONAD_ASSERT(
+        base[0] == 'M' && base[1] == 'Z' && base[2] == 'W' && base[3] == 0x01);
+    // Keep blob offsets and overlay ids in disjoint halves of the NodeId
+    // space (blob < OVERLAY_BASE, fresh ids >= OVERLAY_BASE). Bounding the
+    // blob size bounds every offset. Real witnesses are ~MBs; this only
+    // rejects a pathological >=2 GiB blob.
+    MONAD_ASSERT(len <= OVERLAY_BASE);
```
mkolosick: Shouldn't this be `len < OVERLAY_BASE`?

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -0,0 +1,759 @@
...
+                [&](AccountLeafView a) { is_valid_offset(a.storage()); },
+                [](auto) {}});
+
+        match(
+            node,
+            Cases{
+                [](NullView) {},
+                [](DigestView) {},
+                [&](auto) {
+                    unsigned char buf[MAX_NODE_RLP];
```
mkolosick: Should this get hoisted out of the closure/loop?

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -0,0 +1,759 @@
...
+                    node_rlp_span const rem = encode_rlp<true>(
+                        node, node_rlp_span{buf}); // priming pass
+                    // Only hash-referenced nodes (canonical RLP >= 32 B)
+                    // are cached; smaller nodes are inlined by their
+                    // parent, so caching their hash would make child_ref
+                    // emit a 32-byte ref where the trie inlines it.
+                    if (rem.rlp_size() >= 32) {
```
mkolosick, line 132: Is it feasible/worth calculating `rlp_size` without doing the full `rlp` encoding and moving `encode_rlp` inside this check?

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -0,0 +1,759 @@
...
+                }
+                dest = encode_path(dest, l.path(), /*terminating=*/true);
+                return wrap(dest);
+            },
+            [](DigestView) -> std::span<unsigned char> { std::unreachable(); },
+            [](NullView) -> std::span<unsigned char> { std::unreachable(); },
+        })};
+}
+
+// ── mutation — typed byte builders (§4 layout) ───────────────────────────────
```
mkolosick: This comment feels like a dangling reference to something.

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -0,0 +1,759 @@
...
+            // Callers only fold a surviving child into its parent's path.
+            [](NullView) { std::unreachable(); },
+        });
+}
+
+std::pair<NodeId, Nibbles>
+OffsetTrie::upsert_node(NodeId const id, NibblesView const key)
+{
+    hashes_.erase(id); // dirtied along the descent
+    // Leaf split/overwrite, shared by both leaf types. Only re-emitting the
```
mkolosick: typo: "Only re-emitting *if* the"

mkolosick (changes requested):

category/execution/ethereum/db/offset_trie.cpp
goodlyrottenapple: I don't think it's worth the effort. Most of the witness are digest hashes

dhil (approved):

mkolosick (approved):

PR #2477, opened 2026-08-05
Pin CI base image by digest and rebuild weekly

mhumeSF:
Pins `ubuntu:26.04` by digest so upstream tag pushes don't silently invalidate the shared build cache, adds a dependabot config to open a PR when a new image is published, and rebuilds the cached base image weekly so apt packages don't go stale between Dockerfile changes.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1-labs
 
-FROM ubuntu:26.04 AS base
+FROM ubuntu:26.04@sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03 AS base
 
 COPY scripts/ubuntu-build/ /opt
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: docker
+    directory: /docker
+    schedule:
+      interval: weekly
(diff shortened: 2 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1-labs
 
-FROM ubuntu:26.04 AS base
+# ubuntu:26.04 — pinned by digest; keep the tag in this comment in sync
+FROM ubuntu@sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03 AS base
```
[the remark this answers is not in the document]
mhumeSF: Good catch — switched to the combined `ubuntu:26.04@sha256:...` form and dropped the comment.

Baltoli:
> adds a dependabot config to open a PR when a new image is published

any idea how often this is?

docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1-labs
 
-FROM ubuntu:26.04 AS base
+FROM ubuntu:26.04@sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03 AS base
```
Baltoli, line 3: Should we do the same thing in other places we use bare 26.04 (e.g. `monad-bft`)?

mhumeSF:
[https://github.com/docker-library/official-images/commits/master/library/ubuntu](<https://github.com/docker-library/official-images/commits/master/library/ubuntu>)

Looks like about every 2 weeks

docker/Dockerfile
mhumeSF: I'll get this on monad-bft

Yb2411 (approved):

PR #2490, opened 2026-08-18
Implement EIP-7981 to increase access list cost

ryankeleti:
Additional cost is conditional on feature flag `eip_7981_active` for intrinsic/floor gas calculations.

Feature is active for EVM rev `>= MONAD_ETH_AMSTERDAM`.

Note: the 64 per-byte cost specified by EIP-7981 is actually 40 per-byte since we haven't/aren't going to implement EIP-7976

### DIFF
--- a/category/execution/ethereum/transaction_gas.cpp
+++ b/category/execution/ethereum/transaction_gas.cpp
@@ -68,6 +68,17 @@ inline constexpr auto g_access_and_storage(Transaction const &tx) noexcept
     return g;
 }
 
+// EIP-7981
+// Without EIP-7976, the 64 per-byte cost specified by EIP-7981 is 40 per-byte
+inline constexpr uint64_t g_access_list_data(Transaction const &tx) noexcept
+{
+    uint64_t g = tx.access_list.size() * 20u;
+    for (auto const &i : tx.access_list) {
+        g += i.keys.size() * 32u;
+    }
+    return g * 40u;
+}
+
 // EIP-7702
 inline constexpr auto g_authorization(Transaction const &tx) noexcept
 {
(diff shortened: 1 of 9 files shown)


category/execution/ethereum/transaction_gas.cpp
```diff
@@ -124,18 +134,29 @@ uint64_t intrinsic_gas(Transaction const &tx) noexcept
     if constexpr (traits::evm_rev() >= MONAD_ETH_PRAGUE) {
         gas += g_authorization(tx);
     }
+    // EIP-7981: increase access-list cost (Glamsterdam)
```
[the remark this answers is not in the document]
ryankeleti: I guess using Amsterdam is better for execution-side changes vs the general Glamsterdam change?

I didn't feel this comment was necessary to add, but the rest of the branches in this function have a similar style.

ryankeleti:
Do we want `evm_rev() >= MONAD_ETH_AMSTERDAM` + update `evm_rev` instead of `Rev >= MONAD_NEXT`?

category/vm/evm/traits.hpp
```diff
@@ -353,6 +359,11 @@ namespace monad
             return Rev >= MONAD_FOUR;
         }
 
+        static consteval bool eip_7981_active() noexcept
+        {
+            return Rev >= MONAD_NEXT;
```
Baltoli: More consistent semantically to say "this EIP is active when the Eth. revision underlying this Monad one is >= Amsterdam"
```suggestion
            return evm_rev() >= MONAD_ETH_AMSTERDAM;
```

category/execution/ethereum/transaction_gas.cpp
```diff
@@ -68,6 +68,16 @@ inline constexpr auto g_access_and_storage(Transaction const &tx) noexcept
     return g;
 }
 
+// EIP-7981
```
Baltoli, line 71: Normally I favour these comments being terse, but we should call out the divergence from Ethereum's spec for this EIP.

category/execution/ethereum/transaction_gas.cpp
```diff
@@ -68,6 +68,16 @@ inline constexpr auto g_access_and_storage(Transaction const &tx) noexcept
...
+inline constexpr uint64_t g_access_list_data(Transaction const &tx) noexcept
```
Baltoli, line 73: I wouldn't fix it in this PR because it's locally consistent, but `inline constexpr` is redundant here (`inline` implies `constexpr` on a function definition). We should clean that up at some point.

category/execution/ethereum/transaction_gas.cpp
```diff
@@ -124,18 +134,29 @@ uint64_t intrinsic_gas(Transaction const &tx) noexcept
     if constexpr (traits::evm_rev() >= MONAD_ETH_PRAGUE) {
```
Baltoli, line 135: Again out of scope, but a nice cleanup for the future would be to convert legacy checks like this into concrete EIP feature flags.

test/ethereum_test/exclude/MONAD_NEXT_amsterdam.cmake
```diff
@@ -18,4 +18,10 @@
 # does not yet emit; every fixture therefore fails the genesis block-hash
```
Baltoli: Comment / XFAIL mechanism here is now stale

Baltoli (dismissed):
Couple of small comments but looks great otherwise!

category/execution/ethereum/transaction_gas.cpp
ryankeleti: I'll do a follow-up to change to just `constexpr`

Baltoli (approved):

brett-monad (approved):

khordadi:
Looks great!

khordadi (commented):
LGTM

khordadi (commented):
LGTM

PR #2492, opened 2026-08-18
Add always-on cache hit/miss counters, and export the node cache over FFI

maxkozlovsky:
Adds always-on hit/miss/eviction counters to the four caches on the state read
path, and exports the trie-node cache over the triedb FFI so monad-node can
publish it. Cache sizing could not be evaluated on a running node before this:
`LruCache` carried a stats facility behind `MONAD_LRU_CACHE_STATS`, a macro
defined nowhere in the tree or the build, and `static_lru_cache`,
`LruWeightCache` and `NodeCache` had none at all.

### Design

A shared `CacheStats` (three relaxed atomics) embedded in the three cache
templates, exposed two ways:

- `stats()` — cumulative totals, for a metrics scraper
- `block_stats()` — activity since `begin_block_stats()`, for the block metrics
  log line

The old `print_stats()` cleared the counters on read. That gave per-block
deltas at the cost of making the totals unscrapeable, and two readers would
steal counts from each other. A baseline snapshot provides both views without
either reader disturbing the other.

`Db::begin_block_stats()` is pure rather than a defaulted no-op: an
implementation that reports stats but forgets to open the window would print
since-construction totals on a per-block line, which looks more plausible than
the real thing. The three implementations with no stats define it empty
deliberately.

`NodeCache` additionally reports `used_bytes()`. The cache is bounded twice —
by `node_lru_max_mem` and by the slot count derived from it via
`AVERAGE_NODE_SIZE` — and only `used_bytes()` against `size()` shows which of
the two is binding.

The FFI returns `bool` / `Option`, deliberately unlike the
`triedb_storage_stats_read` it otherwise mirrors: all-zero is a legitimate
reading for an idle cache, so it cannot double as an error the way a zero disk
capacity can. Its scope is also narrower than the name suggests and is
documented as such — only the async read and traverse paths consult the cache;
`triedb_read` is blocking and uncached.

### Bug fixed along the way

`find()` records a lookup, so using it as an existence predicate corrupts the
counters. `find_notify_fiber.cpp` asserted `find(...) == false` immediately
before inserting, and `MONAD_ASSERT` has no `NDEBUG` guard, so every completed
disk read booked a phantom miss. Added a non-counting `contains()` and used it
there. Worth remembering whenever counters are added to a container.

`NodeCache` also stopped re-exporting `Base::clear`, which left `used_bytes_`
stale while `size()` read zero — harmless while unreachable, but the new
accessor would have published the inconsistency.

### Testing

`LruCache` had no test file at all; this adds one.

- `lru_cache_test` — hit/miss/eviction counting; that reads do **not** reset
  counters (the deliberate reversal of the old behaviour); that `clear()`
  leaves counters intact, which is the invariant keeping the window
  subtraction from underflowing; that a backwards delta saturates instead of
  wrapping to ~1.8e19.
- `static_lru_test` — same counting, plus that `contains()` does not count as a
  lookup (pins the bug above).
- `node_lru_cache_test` — counters across `NodeCache`'s own byte-budget
  eviction loop, and `used_bytes()` tracking.
- `db_cache_test` — that proposal-overlay hits are **not** charged to the LRU
  counters, that cumulative and per-block views diverge correctly, and that a
  second block with no reads reports zero rather than the first block's totals.
- `lru_weight_cache_tests` — counting, non-destructive reads, and that `clear()`
  preserves counters for the storage tier.

Not covered: the counters are not exercised concurrently, and the FFI
marshalling has no test — matching `triedb_storage_stats_read`, which the crate
also leaves untested for want of a db-handle fixture.

A monad-bft PR consuming the new FFI follows, and needs this merged and the
submodule bumped first.

### DIFF
--- a/category/core/CMakeLists.txt
+++ b/category/core/CMakeLists.txt
@@ -121,6 +121,7 @@ add_library(
   "tl_tid.c"
   "tl_tid.h"
   "detail/start_lifetime_as_polyfill.hpp"
+  "lru/cache_stats.hpp"
   "lru/lru_cache.hpp"
   "lru/static_lru_cache.hpp"
   "mem/batch_mem_pool.hpp"
--- a/category/core/lru/cache_stats.hpp
+++ b/category/core/lru/cache_stats.hpp
@@ -1,0 +1,111 @@
+// Copyright (C) 2025-26 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <category/core/config.hpp>
+
+#include <atomic>
+#include <cstdint>
+
+MONAD_NAMESPACE_BEGIN
+
+struct CacheStatsSnapshot
+{
+    uint64_t hits{0};
+    uint64_t misses{0};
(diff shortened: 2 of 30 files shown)


maxkozlovsky:
@claude  review fixes

PR #2496, opened 2026-08-20
offset trie improvments

goodlyrottenapple:
offset trie fixes/improvements:
- constructor validates the data forms a tree, i.e. each child has
  exactly one parent and no nodes are unreachable
- Bound path length at MAX_PATH_NIBBLES as the decoder assumes the
  encoding is always a short string
- Rework checked_end onto offsets, fixing an out-of-bounds read
- hashes_ insert is now guarded on the RLP length, enforcing the
  invariant that only hash-referenced nodes get cached in the map

### DIFF
--- a/category/execution/ethereum/db/offset_trie.cpp
+++ b/category/execution/ethereum/db/offset_trie.cpp
@@ -35,7 +35,6 @@
 #include <bit>
 #include <cstdint>
 #include <cstring>
-#include <limits>
 #include <span>
 #include <utility>
 #include <vector>
@@ -61,6 +60,22 @@ NodeId OffsetTrie::read_root(byte_string_view const blob)
     return root;
 }
 
+namespace
+{
+    // Set (val = true) or clear (val = false) the bit for byte offset `off`
+    template <bool val>
+    void set_offset_value(std::span<uint64_t> const offsets, uint64_t const off)
+    {
+        uint64_t const bit = uint64_t{1} << (off & 63);
+        if constexpr (val) {
+            offsets[off >> 6] |= bit;
+        }
+        else {
+            offsets[off >> 6] &= ~bit;
+        }
+    }
+}
+
 OffsetTrie::OffsetTrie(byte_string_view const blob)
     : blob_(blob)
     , root{read_root(blob_)}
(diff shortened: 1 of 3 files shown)


mkolosick (dismissed):

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -79,12 +96,19 @@ OffsetTrie::OffsetTrie(byte_string_view const blob)
...
+        if (c == NULL_ID) {
+            return;
+        }
         uint64_t child_offset = static_cast<uint64_t>(c);
         MONAD_ASSERT(
-            c == NULL_ID ||
-            (child_offset < node_offset && child_offset < blob_.size() &&
-             ((node_offsets[child_offset >> 6] >> (child_offset & 63)) & 1)));
+            child_offset < node_offset && child_offset < blob_.size() &&
+            ((node_offsets[child_offset >> 6] >> (child_offset & 63)) & 1));
```
dhil: I think the right-most &&-operand expression is equivalent to
```
node_offsets[child_offset >> 6] & (uint64_t{1} << (child_offset & 63))
```
which I find a bit easier to read, because it makes the bitwise &-test stand out. No blocker though.

Is it the case that `node_offset < blob_.size()`? If so, then we can drop the `child_offset < node_offset` test.

category/execution/ethereum/db/offset_trie.hpp
```diff
@@ -574,23 +617,32 @@ class OffsetTrie
...
-                        return dest.shrink(33);
+                        return encode_rlp(hash, dest);
                     }
                     return child_ref_compute<priming_pass>(id, node, dest);
                 }});
     }
 
+    node_rlp_span encode_rlp(bytes32_t const &hash32, node_rlp_span dest)
+    {
+        static_assert(HASH_RLP_LEN == KECCAK256_SIZE + 1);
```
dhil, line 662: I suppose here we should assert on the dynamic size of `dest`.

category/execution/ethereum/db/offset_trie.hpp
```diff
@@ -574,23 +617,32 @@ class OffsetTrie
...
+        auto const hash_span = dest.last(HASH_RLP_LEN);
+        hash_span[0] = 0xa0;
+        std::memcpy(hash_span.data() + 1, hash32.bytes, KECCAK256_SIZE);
+        return dest.shrink(HASH_RLP_LEN);
+    }
+
+    node_rlp_span encode_rlp(hash_rlp_view const hash, node_rlp_span dest)
+    {
+        std::memcpy(dest.last(hash.size()).data(), hash.data(), hash.size());
```
dhil, line 671: Same here with a size assertion, I think.

dhil (dismissed):

category/execution/ethereum/db/offset_trie.hpp
goodlyrottenapple: would a debug assert be ok. We don't assert the size of dest elsewhere, since the top level assumption is that every `node_rlp_span` is a slice into a buffer of size `MAX_NODE_RLP`, which **should** always be sufficient for encoding any single rlp trie node.

dhil: I have no problem with omitting them entirely. I just mention it here because this is technically a public function. Maybe worth documenting this invariant at some point.

category/execution/ethereum/db/offset_trie.cpp
goodlyrottenapple: ~~`child_offset < node_offset` condition is to rule out a self reference cycle.~~ ah i see, you are right, we don't need a separate test for this

category/execution/ethereum/db/offset_trie.hpp
goodlyrottenapple: I added asserts on all the methods of `node_rlp_span` that access the `data()` handle, making sure we throw an assert if the buffer runs out.

mkolosick (approved):

dhil (approved):

Baltoli (approved):

PR #2503, opened 2026-08-20
[core] add a shared cache statistics counter

maxkozlovsky:
Introduce CacheStats, a small always-on counter for hits, misses and evictions, plus the plain snapshot type callers read it through.


maxkozlovsky:
Squashed into #2504 — a struct with no production caller did not warrant its own review cycle. The CacheStats type and its tests now land as part of that PR.

PR #2504, opened 2026-08-20
[mpt] count hits, misses and evictions in the trie node cache

maxkozlovsky:
Wire CacheStats into static_lru_cache and NodeCache. a follow-up exports them over the triedb FFI.

Two related fixes in NodeCache while making its occupancy observable. insert() charged the full size of the incoming node before evicting, so an overwrite -- which replaces an entry already accounted for -- could evict entries to make room it did not need; it now applies the net change. And NodeCache no longer re-exports Base::clear, which left used_bytes_ stale while size() read zero.

### DIFF
--- a/category/core/lru/static_lru_cache.hpp
+++ b/category/core/lru/static_lru_cache.hpp
@@ -17,6 +17,7 @@
 
 #include <category/core/assert.h>
 #include <category/core/config.hpp>
+#include <category/core/lru/cache_stats.hpp>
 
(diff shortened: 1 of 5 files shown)


maxkozlovsky:
Recreated as #2575. A PR inside a GitHub stack cannot have its base changed, so once #2503 was squashed into this one there was no way to retarget it off the now-dead `max/cache-stats` base — the diff rendered as 330 files. #2575 is the same commit against `main`.

PR #2506, opened 2026-08-20
triedb: expose node-cache counters via FFI

maxkozlovsky:
Add triedb_node_cache_stats_read -> TriedbHandle::node_cache_stats(), mirroring triedb_storage_stats_read, so monad-node can export the trie node cache as Prometheus metrics.

Each triedb handle owns an independent NodeCache, so these are per-handle and not comparable across handles. used_bytes and entries accompany the hit/miss/eviction counts because the cache is bounded twice -- by node_lru_max_mem and by the slot count derived from it -- and only the pair shows which bound is binding.


maxkozlovsky:
Recreated as #2576, stacked on #2575. Closed along with #2504 so the pair could be reopened with correct bases.

PR #2507, opened 2026-08-21
crypto: replace ethash with our own hash256 type and keccak

goodlyrottenapple:
**Motivation:**
We currently have two versions of keccak in the code base, the ethash version, used for the the ethash_hash256 union aliased as monad::hash256, and ethash::keccak256 behind the SHA3 opcode and the OneSSL version used for all other keccak computation. Since the main codebase moved away from the ethash version earlier and the merge of the interpreter/compiler repo re-introduced it in the SHA3 codepath, it would seem prudent to finish the refactor and remove ethash as a direct dependency (still used by evmone and silkpre, both of which are also being phased out).

**Main changes:**
- The monad_crypto library collects keccak together with the vendored silkpre precompiles under `<category/crypto/...>`. 
- the C `keccak256` is now `monad_keccak256`, matching `monad_*` silkpre_vendored methods. 
- `ethash` stays configured for silkpre and evmone, which link it privately.
- fixes `SHA3_squeeze` declared arity: `keccak.c` declared four parameters, matching the asm, while OpenSSL's portable `keccak1600.c` defines a fifth (int next). Non-x86_64 builds hashed with an uninitialized register and produced a wrong digest for every input.


category/vm/runtime/CMakeLists.txt
```diff
@@ -63,7 +63,6 @@ target_include_directories(monad-vm-runtime
 
 target_link_libraries(monad-vm-runtime
     PUBLIC evmc::evmc
```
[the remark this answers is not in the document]
goodlyrottenapple: this will be fixed in a followup PR

category/core/hash256.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include <category/core/config.hpp>
+
+#include <category/crypto/hash256.h>
+
+MONAD_NAMESPACE_BEGIN
+
+// The 32-byte digest produced by keccak256() and blake3().
+using hash256 = ::monad_hash256;
```
Baltoli: Is there any particular reason to keep this typedef around - it's a legacy name from ethash, and a `hash256` has no meaningful difference from just a raw `bytes32_t`? Or does keeping this make the diff nicer?

category/crypto/hash256.h
```diff
@@ -0,0 +1,43 @@
...
+//
+// The alignas reproduces the alignment of the ethash_hash256 union this type
+// replaced. That union declared uint64_t/uint32_t members alongside the byte
+// array, which gave it 8-byte alignment; only the byte array was ever used
+// here, so the members are gone but the alignment is pinned explicitly rather
+// than silently dropped to 1.
+struct monad_hash256
+{
+    alignas(8) uint8_t bytes[32];
+};
```
Baltoli: Ah - re. my comment above (will teach me to read before posting), is the 8-byte alignment actually important?

Baltoli: (it might legitimately be, I'm really not sure)

category/crypto/keccak.c
```diff
@@ -23,10 +23,10 @@
 extern size_t
 SHA3_absorb(uint64_t A[5][5], unsigned char const *inp, size_t len, size_t r);
 
-extern void
-SHA3_squeeze(uint64_t A[5][5], unsigned char *out, size_t len, size_t r);
+extern void SHA3_squeeze(
+    uint64_t A[5][5], unsigned char *out, size_t len, size_t r, int next);
```
Baltoli: Concerning UB - where did this originate from? Was silkpre vendoring this bad code?

CMakeLists.txt
```diff
@@ -100,6 +100,11 @@ add_subdirectory("third_party/concurrentqueue")
 add_subdirectory("third_party/cthash")
 
 # ethash
+# No monad code uses ethash: hashing goes through category/core/keccak.hpp and
+# the digest type is monad::hash256 (category/core/hash256.hpp). It is still
+# configured because two third-party libraries link ethash::keccak privately —
+# silkpre, and evmone under MONAD_COMPILER_TESTING/BENCHMARKS. Retiring the
+# submodule means finishing the silkpre -> silkpre_vendor migration.
```
Baltoli: Makes sense - we can hopefully retire both of these libraries soon

category/crypto/hash256.h
goodlyrottenapple: I'm not sure either, but it seemed safer to assume it is.

category/crypto/keccak.c
goodlyrottenapple: This code has been sitting here unused. It was committed with the intent of enabling builds for apple silicon, but I don't know of anyone who actually builds and runs the client on non x86 hardware (outside the riscv zkvm build that is). We could potentially just drop this code, seeing as no one noticed it not working.

Baltoli: Let's drop it in that case. The context for the Apple Silicon stuff was Ken working on some core code on the train ages ago, but I believe nobody is using it now.

category/core/hash256.hpp
goodlyrottenapple: yeah somewhat smaller diff but we can rip this bandaid off now, seeing as we don't imbue the C++ version with any operators/etc.

category/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+  "keccak.h")
+
+target_include_directories(monad_crypto PUBLIC "${CATEGORY_MAIN_DIR}")
+
+# keccak256() is a thin wrapper over OpenSSL's SHA3 core: hand-written AVX2 /
+# AVX-512 assembly, so x86-64 is the only supported host architecture.
+if(NOT ${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64")
+  message(FATAL_ERROR "monad_crypto requires x86-64: keccak256() has no "
+                      "implementation for ${CMAKE_SYSTEM_PROCESSOR}")
+endif()
```
Baltoli, line 36: I presume the reason for pulling out `monad_crypto` is to isolate this platform-specific code?

category/crypto/hash256.h
```diff
@@ -0,0 +1,43 @@
...
+
+// The 32-byte digest produced by the hash primitives in this directory, and by
+// the keccak256() / blake3() wrappers in category/core.
+//
+// The alignas reproduces the alignment of the ethash_hash256 union this type
+// replaced. That union declared uint64_t/uint32_t members alongside the byte
+// array, which gave it 8-byte alignment; only the byte array was ever used
+// here, so the members are gone but the alignment is pinned explicitly rather
+// than silently dropped to 1.
+struct monad_hash256
```
Baltoli, line 33: Might be nice (in a followup?) to add a C++ layer on top of this base type:
```cpp
namespace monad {
  struct hash256 : public monad_hash256 {};
}
```

Benefits aren't massive:
- Namespacing, no explicit C names lying around in C++ code
- `operator==` and other behaviour if we ever need / want it?
- Consistency with how the rest of the client deals with fundamental vocabulary types (address, bytes32, etc)

category/crypto/keccak.h
```diff
@@ -25,7 +25,7 @@ extern "C"
 
 constexpr size_t KECCAK256_SIZE = 32;
 
-void keccak256(void const *in, size_t len, uint8_t out[KECCAK256_SIZE]);
+void monad_keccak256(void const *in, size_t len, uint8_t out[KECCAK256_SIZE]);
```
Baltoli, line 28: Similarly for a followup - we could do the same as the SHA code and have a C++ wrapper of this that takes spans / iterators? Might not be worth the effort if there are limited call sites

Baltoli (approved):

category/crypto/keccak.h
Baltoli: nvm I see you have this

Baltoli (dismissed):

Baltoli:
Looks sensible to me

category/crypto/CMakeLists.txt
goodlyrottenapple: yes, plus i thought it would make sense to bundle the silkpre vendor plus this code into it own library. In a followup PR, I also plan to add an ethash_vendor of keccak specifically for the zkvm target to this lib.

dhil (dismissed):
Looks like a clean refactoring to me.

dhil:
@claude please provide a critical review of this PR. Focus particularly on whether this refactoring has introduced new soundness issues and regressions.

category/vm/runtime/keccak.hpp
```diff
@@ -42,7 +41,7 @@ namespace monad::vm::runtime
             ctx->deduct_gas(word_size * bin<6>);
         }
 
-        auto const hash = ethash::keccak256(ctx->memory.data + *offset, *size);
+        auto const hash = keccak256({ctx->memory.data + *offset, *size});
```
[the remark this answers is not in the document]
goodlyrottenapple: zkvm change will be addressed in a separate PR

Baltoli (approved):

dhil (approved):

PR #2511, opened 2026-08-21
Eip 8246 remove selfdestruct burn

brett-monad:


brett-monad:
Duplicate of #2512, which has the correct base (`eip-8024-dupn-swapn-exchange`) and the description. This one targeted `main`, so it showed the whole stack rather than just the EIP-8246 commit.

PR #2512, opened 2026-08-21
EIP-8246 (remove `SELFDESTRUCT` burn)

brett-monad:
Removes the two paths by which `SELFDESTRUCT` still destroys ETH. This is a prerequisite of EIP-7708 rather than an adjacent EIP — 7708's current text lists `requires: 8246` and no longer specifies a burn — which is what lets the 7708 implementation land with no `Burn` log at all. That is a sequencing constraint rather than a code one: 8246 must be live before 7708 is, or 7708 would claim complete native-value logging while two burn paths went unlogged.

**Rule 1:** a same-transaction account destructing to itself keeps its balance. `State::selfdestruct` moves the balance whenever the beneficiary differs; for the destruct-to-self case one predicate, `selfdestruct_to_self_burns`, holds the revision rule: always before Cancun, only for a same-transaction account under EIP-6780, never under EIP-8246.

**Rule 2:** at finalization a destructed account holding a balance is preserved — nonce reset, code and storage cleared, balance untouched — rather than deleted. Nonce and code are cleared first and EIP-161 is applied to the result, so an empty account is still deleted. Rule 2 also covers the EIP's second burn, value sent to an account already marked for selfdestruction, because it keys on the balance rather than on who the beneficiary was.

Gated on a new `eip_8246_active()`, which is `>= MONAD_ETH_AMSTERDAM` in both trait families; pre-Amsterdam revisions must keep burning.

### Reserve balance is deliberately unchanged

A preserved account is a funded, nonce-0, codeless address, and no predicate distinguishes it from an ordinary pre-funded address — three `Account` fields match by construction, and the fourth, `incarnation`, is drawn from the same source for both and is absent from the merkle account. So it takes the existing MIP-4 floor exactly as a pre-funded address does today. Exempting it would mean making the storage-generation key carry account provenance, and that equivalence is the reason not to.

### Spec tests

No change to the Amsterdam exclusion list. The Amsterdam suite is disabled at
this commit and opens once every Amsterdam EIP is in the tree. Nothing needs
excluding: the pinned bundle is generated with 8246 in effect, and its nine
selfdestruct fixtures all pass against this tree. The unit tests are the gate
on this PR.

### Testing

Full `MONAD_COMPILER_TESTING=ON` build clean. `ctest` clean apart from the known `BacktraceTest.works`. All spec-test jobs pass. `clang-format` and `license.sh` clean.

Known limitation: the exec-event schema cannot express a code-cleared surviving account.

https://claude.ai/code/session_01T2YcdwU1mj3Lqunp9Y7pqa


category/execution/ethereum/state3/state.cpp
```diff
@@ -479,24 +488,84 @@ EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
...
+    // Three things the EIP-8246 preservation below rests on, none of them
+    // stated where they are decided. destruct_touched_dead, which runs straight
+    // after this, is a second and untemplated deletion pass, so it cannot be
+    // made revision-aware; it spares a preserved account only because the
+    // balance test below and is_dead are the same EIP-161 emptiness test.
+    // BlockState::can_merge runs before execute_final, which is what stops
+    // relaxed merge zeroing a preserved balance and committing an empty
+    // account. And the reserve-balance check runs at depth 0 in
+    // execute_message, before finalization, so the preservation is invisible to
+    // it.
```
Baltoli: This PR could do with a pass going through and trimming down the comments substantially: "Three things the EIP-8246 preservation below rests on, none of them stated where they are decided" is very unfortunate Claudeish

category/execution/ethereum/state3/state.cpp
```diff
@@ -479,24 +488,84 @@ EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
...
+        // set_to_state_incarnation does, for the RPC full-state-override path
+        // -- so iterating it is what keeps eth_call correct. Zero rather than
+        // erase, so anything reading through the block delta sees the zero.
+        auto const orig = original_.find(it.first);
+        MONAD_ASSERT(orig != original_.end());
+        AccountState::StorageMap zeroed{};
+        for (auto const &kv : orig->second.storage_) {
+            zeroed = zeroed.insert({kv.first, bytes32_t{}});
         }
+        account_state.storage_ = std::move(zeroed);
```
Baltoli: The point of this logic - to check my understanding - is to ensure that every storage key that might have seen a non-zero value by falling back to the original map will in fact see a zero value in the _current_ map?

category/execution/ethereum/state2/test/test_state.cpp
```diff
@@ -542,16 +542,637 @@ TYPED_TEST(InMemoryStateTraitsTest, selfdestruct_self_same_tx)
 
     State s{bs, Incarnation{1, 1}};
 
-    // Behavior doesn't change in cancun if in same txn
+    // The return value is the balance read before the guard, at every
+    // revision: it is what the call tracer records as the SELFDESTRUCT frame's
+    // value, and it does not track whether value actually moved. Pinned
+    // independently by simulate_v1_trace_multiple_selfdestructs_recursive,
+    // whose frame 2 is revision-independent for this reason.
```
Baltoli: Can we confirm the expected behaviour from Geth in call traces here? Do post-8246 `SELFDESTRUCT` frames still include the value?

category/execution/ethereum/state3/state.cpp
Baltoli: Similarly:
> `destruct_touched_dead`, which runs straight after this, is a second and untemplated deletion pass

why can't it be templated? would that simplify the code?

category/execution/ethereum/execute_transaction.cpp
```diff
@@ -410,6 +410,11 @@ Receipt ExecuteTransaction<traits>::execute_final(
     }
 
     // finalize state, Eqn. 77-79
+    //
+    // Runs after the coinbase award above, which under EIP-8246 decides the
+    // outcome for a beneficiary destructed in this transaction: the award is
+    // what leaves it holding a balance and preserved, rather than empty and
+    // deleted by EIP-161.
```
Baltoli: Not sure this comment is adding much; none of the logic here has actually changed

category/execution/ethereum/state2/test/test_state.cpp
```diff
@@ -542,16 +542,637 @@ TYPED_TEST(InMemoryStateTraitsTest, selfdestruct_self_same_tx)
...
+    }
+}
+
+TYPED_TEST(InMemoryStateTraitsTest, selfdestruct_self_same_tx_zero_balance)
+{
+    BlockState bs{this->tdb, this->vm};
+    State s{bs, Incarnation{1, 1}};
+
+    // EIP-8246 preserves only accounts that hold a balance. An empty one is
+    // still deleted by EIP-161, which is load-bearing rather than a spec
```
goodlyrottenapple: nooooo, please not load-bearing

category/execution/ethereum/execute_transaction.cpp
brett-monad: Removed.

category/execution/ethereum/state2/test/test_state.cpp
brett-monad: Yes. geth master has 8246 behind `IsAmsterdam`, and `opSelfdestruct6780` passes the pre-transfer `balance` to the SELFDESTRUCT frame unconditionally, including destruct-to-self where nothing moves. Ours records the same value.

category/execution/ethereum/state2/test/test_state.cpp
brett-monad: Reworded.

category/execution/ethereum/state3/state.cpp
brett-monad: Trimmed throughout; two of the blocks became code (selfdestruct_to_self_burns, zero_storage).

brett-monad: Removed the misleading comment. It could be templated, but there is nothing to select on: the semantics are unchanged from EIP-161 onward, and the execution layer's static asserts already exclude revisions before that.

category/execution/ethereum/state3/state.cpp
brett-monad: Yes. Restructured as `zero_storage` for clarity.

category/execution/ethereum/state2/test/test_state.cpp
```diff
@@ -93,6 +94,31 @@ namespace
...
+        State &s, Address const &address, uint256_t const &balance,
+        std::source_location const loc = std::source_location::current())
+    {
+        testing::ScopedTrace const trace(
+            loc.file_name(), static_cast<int>(loc.line()), "expect_preserved");
+        ASSERT_TRUE(s.account_exists(address));
+        EXPECT_EQ(s.get_balance(address), balance);
+        EXPECT_EQ(s.get_nonce(address), 0);
+        EXPECT_EQ(s.get_code_hash(address), NULL_HASH);
+    }
```
Baltoli, line 120: I think that _generally_ in test code I would favour explicit repetition of postcondition checks like this. My thoughts are:
- This introduces some awkward machinery - we don't use `ASSERT_NO_FATAL_FAILURE` anywhere else, and it requires some detailed working knowledge of GTest to understand (`ASSERT_*` vs. `EXPECT_*`). I had to look this up.
- `expect_preserved` isn't a totally clear name
- Generally closer to the style of the existing test code - we don't have helpers like this elsewhere

category/execution/ethereum/state3/state.cpp
```diff
@@ -475,6 +484,23 @@ State::selfdestruct(Address const &address, Address const &beneficiary)
...
+    AccountState::StorageMap zeroed{};
+    // Reads land only in the original map, writes in both.
+    for (auto const &kv : orig->second.storage_) {
+        zeroed = zeroed.insert({kv.first, bytes32_t{}});
+    }
+    for (auto const &kv : account_state.storage_) {
+        zeroed = zeroed.insert({kv.first, bytes32_t{}});
+    }
+    account_state.storage_ = std::move(zeroed);
+}
```
Baltoli, line 502: Codex suggested - and I didn't know this existed until now - that because these maps are `immer` under the hood, that the batch update can be done more efficiently with transients: https://sinusoid.es/immer/transients.html#

category/execution/ethereum/state3/state.cpp
```diff
@@ -445,26 +445,35 @@ evmc_page_storage_status State::update_page(
     return account_state.page_tracker_.update_page(key, status);
 }
 
+// Pre-Cancun always; from EIP-6780 only for an account created in this
+// transaction; never from EIP-8246.
```
Baltoli, line 449: Maybe add a more descriptive comment to the actual declaration of this function in the header?

category/execution/ethereum/state3/state.cpp
```diff
@@ -475,6 +484,23 @@ State::selfdestruct(Address const &address, Address const &beneficiary)
 
 EXPLICIT_TRAITS_MEMBER(State::selfdestruct);
 
+// Zeros rather than erasures: a missing key falls back to the original map or
+// the block delta and could still read a pre-destruct value.
```
Baltoli, line 488: Wording here could be clearer; also "erasures" typo

category/execution/ethereum/state3/state.cpp
```diff
@@ -475,6 +484,23 @@ State::selfdestruct(Address const &address, Address const &beneficiary)
...
+void State::zero_storage(Address const &address, AccountState &account_state)
+{
+    auto const orig = original_.find(address);
+    MONAD_ASSERT(orig != original_.end());
+    AccountState::StorageMap zeroed{};
+    // Reads land only in the original map, writes in both.
```
Baltoli, line 494: What does this mean?

category/execution/ethereum/state3/state.cpp
```diff
@@ -486,17 +512,35 @@ void State::destruct_suicides()
...
+            continue;
+        }
+
+        if constexpr (!traits::eip_8246_active()) {
+            account.reset();
+            continue;
+        }
+
+        // EIP-8246: keep the account as balance-only. Mutate in place: the
+        // incarnation is the storage-generation key the commit builders read.
```
Baltoli, line 536: What does this mean?

category/execution/ethereum/state2/test/test_state.cpp
brett-monad: Reverted to explicit checks.

category/execution/ethereum/state3/state.cpp
brett-monad: Done.

category/execution/ethereum/state3/state.cpp
brett-monad: Done, at the declaration.

category/execution/ethereum/state3/state.cpp
brett-monad: Reworded.

category/execution/ethereum/state3/state.cpp
brett-monad: Folded into the function comment.

category/execution/ethereum/state3/state.cpp
brett-monad: Reworded.

category/execution/ethereum/state3/state.cpp
```diff
@@ -486,17 +511,35 @@ void State::destruct_suicides()
...
+        if (account->incarnation != incarnation_) {
+            continue;
+        }
+
+        if constexpr (!traits::eip_8246_active()) {
+            account.reset();
+            continue;
+        }
+
+        // EIP-8246: nonce reset, code and storage cleared, balance unchanged.
```
0xSqualo, line 532: Should `debug_traceCall` follow Geth’s finalized-transaction path before producing a `prestateTracer` diff, with the EIP-8246 survivor representation covered explicitly?

[Geth implementation](https://github.com/ethereum/go-ethereum/blob/89def4a53b19dbb796d7ff81b438f50d270aaca0/core/state_processor.go#L219-L244)

PR #2523, opened 2026-08-27
rpc: add eth_call execution timeout via VM deadline polling

Chen-Yifan:
CAVEAT: timeout does not apply to precompiles.

eth_call previously enforced its timeout only at queue pickup; once execution started, a slow call could pin a PriorityPool fiber indefinitely. Fibers are cooperatively scheduled, so cancellation must be cooperative: runtime::Context now carries an absolute deadline that the VM polls at its gas checks (per opcode in the interpreter, and in deduct_gas for the shared runtime functions), amortizing the clock read over 256 checks. When the deadline passes, execution exits through the existing stack unwind with the new StatusCode::Cancelled. Paths that arm no deadline pay one load and one never-taken branch per check.

A cancelled child frame cancels every ancestor via an unconditional deadline check after each call/create return, so the abort cannot be mistaken for an ordinary failed subcall and produce a nondeterministic result. The VM records the cancellation on vm::Host, and the executor reports timeouts from that flag rather than inferring them from the status code (`EVMC_REJECTED` has other producers, e.g. precompiles) or the wall clock (a call that completes just past the deadline is a valid result and is returned).

The RPC pools are reclassified from gas to time: the low/high gas pools become short_tx/long_tx timeout pools, each with a queue timeout enforced at pickup, and an execution VM timeout. The gas clamp with its out-of-gas retry (`MONAD_ETH_CALL_LOW_GAS_LIMIT`) is removed since client gas limits are not a usable routing signal. Every call starts in the short pool; a call cancelled with a timeout there escalates to the long pool with a fresh budget, and a timeout in the long pool is final.

Expect almost zero effect for vm performance for execution, which doesn't enforce any timeout.


category/rpc/monad_executor.cpp
```diff
@@ -1371,34 +1389,47 @@ struct monad_executor
...
                 };
                 try {
-                    // check for timeout
-                    if (std::chrono::steady_clock::now() - call_begin >
-                        active_pool->timeout) {
+                    // Total budget (queuing + execution), measured from
+                    // call_begin; the VM polls this at its gas checks and
+                    // aborts with StatusCode::Cancelled once it has passed.
+                    auto const deadline = call_begin + active_pool->timeout;
+                    bool execution_cancelled = false;
```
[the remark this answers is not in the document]
Chen-Yifan: added tests

category/rpc/monad_executor.cpp
```diff
@@ -1243,9 +1260,6 @@ struct monad_executor
 
     mpt::RODb db_;
 
-    // The VM for executing eth calls needs to unconditionally use the
-    // interpreter rather than the compiler. If it uses the compiler, then
-    // out-of-gas errors can be misreported as generic failures.
     vm::VM vm_{vm::VM::InterpreterOnly};
```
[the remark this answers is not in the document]
Chen-Yifan: added back

category/rpc/monad_executor.h
```diff
@@ -85,8 +90,8 @@ struct monad_executor_pool_state
 
 struct monad_executor_state
 {
-    struct monad_executor_pool_state low_gas_pool_state;
-    struct monad_executor_pool_state high_gas_pool_state;
+    struct monad_executor_pool_state short_tx_pool_state;
+    struct monad_executor_pool_state long_tx_pool_state;
```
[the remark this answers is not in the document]
Chen-Yifan: thanks, renamed.

category/execution/ethereum/execute_transaction.cpp
```diff
@@ -364,7 +364,12 @@ Result<evmc::Result> ExecuteTransaction<traits>::execute_impl2(State &state)
         chain_ctx_,
         trace_transfers_};
 
-    return ExecuteTransactionNoValidation<traits>::operator()(state, host);
+    auto result =
+        ExecuteTransactionNoValidation<traits>::operator()(state, host);
+
+    MONAD_ASSERT(!host.execution_cancelled());
```
dhil: I think we need this to be a throw, as I reckon you can forge a simulation request that will cause a timeout.
```suggestion
    MONAD_ASSERT_THROW(!host.execution_cancelled(), "transaction timeout");
```

category/vm/runtime/call.cpp
```diff
@@ -181,6 +181,14 @@ namespace monad::vm::runtime
         // is deallocated by the `Environment` destructor.
         ctx->propagate_stack_unwind();
 
+        // A cancelled child frame (execution deadline passed) must cancel
+        // this frame too: otherwise the caller would observe the abort as an
+        // ordinary failed call and keep executing, producing a
+        // nondeterministic result. The child only aborts once the deadline
+        // has passed, so an unconditional clock check here is sufficient to
+        // propagate the cancellation.
```
dhil: I think we can be more terse here, e.g.
```
// Propagate the child frame's deadline cancellation to this frame to ensure deterministic execution.

File: category/vm/runtime/create.cpp [hunk 6]
```diff
@@ -121,6 +121,14 @@ namespace monad::vm::runtime
         // is deallocated by the `Environment` destructor.
         ctx->propagate_stack_unwind();
 
+        // A cancelled child frame (execution deadline passed) must cancel
+        // this frame too: otherwise the caller would observe the abort as an
+        // ordinary failed create and keep executing, producing a
+        // nondeterministic result. The child only aborts once the deadline
+        // has passed, so an unconditional clock check here is sufficient to
+        // propagate the cancellation.
+        ctx->check_deadline_now();
```
dhil: I think a single sentence suffices here.
```suggestion
        // Propagate the child frame's cancellation, if relevant.
        ctx->check_deadline_now();
```

category/vm/runtime/types.hpp
```diff
@@ -248,13 +251,66 @@ namespace monad::vm::runtime
         exit_stack_ptr_t exit_stack_ptr = nullptr;
         bool is_stack_unwinding_active = false;
 
+        /// Sentinel meaning no execution deadline is armed.
```
dhil: I'd say "pending" rather than armed here.

category/vm/host.hpp
```diff
@@ -40,6 +42,39 @@ namespace monad::vm
             evmc::address const &, evmc::bytes32 const &,
             evmc_storage_status) noexcept = 0;
 
+        /// Install an absolute execution deadline; every frame's
+        /// `runtime::Context` inherits it and exits with
+        /// `StatusCode::Cancelled` once it has passed. `time_point::max()`
+        /// disarms it; default: no deadline.
```
dhil: I'd say "effectively stops it" instead of "disarms it".

test/vm/unit/monad_vm_interface_tests.cpp
```diff
@@ -852,3 +853,88 @@ TEST(MonadVmInterface, message_memory)
         mem1.get()[0] = 0;
     }
 }
+
+namespace
+{
+    // JUMPDEST; PUSH1 0; POP; PUSH1 0; JUMP -- an infinite loop, bounded
```
dhil, line 861: I think that we should add a few tests that exercise cancellation deep in the call stack too.

dhil (commented):
Initial review. Mostly just some comments about comments. Though, I think the `MONAD_ASSERT` must be changed.

category/execution/ethereum/execute_transaction.cpp
Chen-Yifan: I use assert here because this is an execution only path which monad_executor rpc module doesn't go through. Execution should never enforce any cancellation.

category/vm/host.hpp
Chen-Yifan: fixed, sorry about claude wording

category/execution/ethereum/execute_transaction.cpp
dhil: Yes, but we probably want to attach this cancellation logic to `eth_simulateV1` too. The problem with using `MONAD_ASSERT` in the execution client is that it renders the client non-modular. `eth_simulateV1` is built as a modest wrapper around the execution client, so modularity is a key property we want.

Chen-Yifan: makes sense, and execution will catch the exception anyways. Updated.

PR #2499, opened 2026-08-20
EIP-7997 (deterministic factory contract)

Baltoli:

PR #2524, opened 2026-08-28
test: run the Amsterdam spec suite for the implemented EIP set

brett-monad:
**Draft, and deliberately not mergeable as-is.** It also does not yet include EIP-7997 (#2499), the fifth Amsterdam EIP, so this is not the complete fork set. 

**Draft note:** The four EIP commits below it are cherry-picks of #2439, #2407, #2512 and #2469, so this branch will conflict with them on merge. It exists to prove the end state and to hold the spec-test work while those four land; once they have, it gets rebuilt by rebasing onto `main` and reduces to its own top commit.

Every branch implementing an Amsterdam EIP currently has the `MONAD_NEXT` Amsterdam spec job switched off or narrowed, because each one moves ahead of the pinned fixture bundle in a different way. #2439 and #2469 blanket it via the `WILL_FAIL` guard; #2407 excludes one fixture; #2512 excludes nine. With all four present the suite can actually run.

**Result: 641 fixtures run, 633 pass, 8 skip, none fail.**

### What it takes

**Retire the `MonadEip7981SpecTestFixtures` bundle.** Pinned at `tests-monad_eip7981@v0.1.0`, the only version ever published, and it predates both EIP-7708 and the SLOTNUM header field — so all 555 of its fixtures fail, including its own 17 EIP-7981 ones: they expect no `Transfer` log in the bloom, and their genesis headers carry no `slotNumber`, which the loader hashes. The whole `ExternalProject` goes rather than just its path in the job, since a declared bundle still downloads 195 MB on every configure whether anything reads it or not. EIP-7981 coverage is expected to return through the Amsterdam bundle; if upstream keeps it separate, the `ExternalProject` comes back with a regenerated pin.

**Exclude 17 fixtures from `tests-monad_amsterdam@v0.5.0`**, in two groups with different causes:

- **8 gas.** v0.5.0 was generated without EIP-7981, which `main` implements, so its intrinsic-gas budgets predate the access-list cost increase and these fail with `intrinsic gas greater than limit`.
- **9 selfdestruct.** v0.5.0 implements EIP-7708 but not EIP-8246, so it expects a `Transfer` log where rule 1 means nothing moves, an account deleted that rule 2 preserves, and a balance burned that rule 2 keeps. Each fixture reports the bloom mismatch first and the post-state size and balance mismatches underneath — reading only the first line makes this look like a 7708 emission bug, which it is not.

### When the exclusions come out

Both groups clear when upstream publishes a bundle generated against a spec with EIP-7981, EIP-8246 and EIP-7708 all active. No such release exists today: v0.5.0 has 7708, SLOTNUM and 8024 but not 7981, and `eip7981@v0.1.0` has 7981 but predates the others. At that point the change here is two deletions from the exclusion list — no bundle to re-add, no path to re-wire.

One thing worth not assuming away: **EIP-8246 has never been exercised by any spec fixture.** It has no upstream coverage at all — the `EIP8246` fork class is a stub and its `selfdestruct_no_burn` test directory generates nothing. So those nine are the first real test of that implementation rather than a confirmation of it, and a residual failure there should be read as information about our code before it is read as a stale fixture.

### Fixtures that pass without an implementation

`amsterdam/eip7954_increase_max_contract_size` runs green even though there is no
`eip_7954_active` predicate and no implementation anywhere in the tree. That is
correct, not an accident: Monad's existing max code size already subsumes the
limit EIP-7954 raises Ethereum's to, so there is nothing to gate.

### Not covered

The 8 skips are the `for_monad_tentomonad_nextattime15k` fork-transition fixtures, whose network has no `revision_map.hpp` entry — including EIP-7708's own two, which pin when emission starts. That harness gap predates this change and a new bundle will not close it.


Baltoli:
Superseded: https://github.com/category-labs/monad/pull/2537

PR #2532, opened 2026-09-01
Add monad-mpt --rescan-devices

maxkozlovsky:
Growing a node's database meant --archive then --restore into a freshly created larger pool: scratch space, plus a full dump and restore. Extending a device the pool already owned was worse than unsupported -- the metadata saying a device belongs to a pool lives at its very end, so lvextend left it unopenable, with no resize2fs step to follow.

--rescan-devices reconciles a database with whatever storage it is now given: --storage lists every device it should contain, and one offline run joins any blank suffix and takes up the space of an extended last device, under one confirmation and one metadata growth.

Taking up an extend needs the previous size of the device, which the extend itself strands mid-device, so db_metadata now records every device's size on each writable open, in padding it already reserves -- no magic bump, no migration. That recording is the only source, and is acted on only where the stranded footer's config_hash confirms it, so an extend must be preceded by a writable open under a release carrying this code.

Refusing two sources that name one device is needed here, but the hole is not new: creating a pool with a device listed twice was silently accepted, and no later open could detect it. That check goes in the constructor, for every mode.

### DIFF
--- a/category/async/storage_pool.cpp
+++ b/category/async/storage_pool.cpp
@@ -25,6 +25,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <bit>
 #include <cassert>
 #include <cerrno>
 #include <cstddef>
@@ -35,6 +36,7 @@
 #include <filesystem>
 #include <limits>
 #include <mutex>
+#include <optional>
 #include <span>
 #include <utility>
 #include <variant>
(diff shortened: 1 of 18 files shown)


maxkozlovsky:
Superseded by #2563, which is the same feature rewritten against the single-device storage pool landed in #2562.

PR #2535, opened 2026-09-03
Add disabled Amsterdam spec tests

ryankeleti:
To be added in the mega Amsterdam PR stack

PR #2537, opened 2026-09-04
Enable Amsterdam spec tests

Baltoli:
The inverse of #2535. That PR pinned the combined Amsterdam fixture bundle and switched the `MONAD_NEXT` Amsterdam spec job off behind a blanket `BlockchainTests.*` exclusion, with a `WILL_FAIL` guard so ctest stayed green while nothing ran. Now that every EIP in that bundle is implemented on this stack (#2439 → #2499 → #2512 → #2469 → #2407), the suite runs for real, so this drops the scaffolding.

### Change

- `exclude/MONAD_NEXT_amsterdam.cmake`: replace the blanket `BlockchainTests.*` entry with the single EIP-4844 blob fixture Monad cannot run.
- `CMakeLists.txt`: remove the `WILL_FAIL` guard and its comment, since the guard can never fire again.

The feature PRs underneath leave the exclusion file alone. The fixtures are generated with the whole fork active, so opening or filtering the suite part-way up the stack either runs zero tests with the guard disarmed (which is what broke CI on #2512) or fails on the EIPs not yet in the tree (#2469). Enabling belongs in one place, here.

### Measured

Against `tests-monad_eip7708+7843+7981+7997+8024+8246@v0.2.0` on this branch, unfiltered:

| | |
|---|---|
| fixtures | 711 |
| pass | 697 |
| skip | 13 |
| fail | 1 |

The single failure is the blob fixture already excluded, so the ctest job passes as configured. The skips are the `for_monad_tentomonad_nextattime15k` fork-transition fixtures, whose network has no `revision_map.hpp` entry — a harness gap that predates this stack.

Sits on top of #2407.

https://claude.ai/code/session_019VKRBQPBtf5k1JbBH28hVe


ryankeleti (dismissed):

test/ethereum_test/exclude/MONAD_NEXT_amsterdam.cmake
```diff
@@ -13,8 +13,5 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-# Amsterdam spec tests are mutually dependent, so enable them together
-# when all EIPs land.
 set(MONAD_NEXT_amsterdam_excluded_tests
```
dhil: What's our practice here? In some of the other cmake files we set the corresponding variable to the empty string `""`, whilst here we just leave it undefined. Moreover, do we need these morally empty cmake files at all?

Baltoli: Good catch - doesn't matter in practice because of how these variables get used (list concat), but better to be consistent. We keep the empty-exclusion files to make the CMake code more regular.

dhil (dismissed):

ryankeleti (dismissed):

dhil (dismissed):

cmcl (dismissed):

ryankeleti (dismissed):

dhil (dismissed):

dhil (dismissed):

dhil (dismissed):

PR #2196, opened 2026-04-13
Introduce monad::Address, replacing evmc::address dependency

Baltoli:
## Summary

Second step in removing evmc C++ wrapper types from internal code (first: `bytes32_t` in daa26346c).

- Introduces `monad::Address` as a standalone struct in `category/core/address.hpp`, inheriting from `evmc_address` (the C struct) — mirrors the `bytes32_t` pattern
- Deletes `category/execution/ethereum/core/address.hpp` (the old `using Address = evmc::address` alias) and updates all 134 consumer files
- EVMC host virtual overrides keep `evmc::address const&` to match the base class; all internal APIs use `Address const&`
- Implicit conversion from `evmc_address` ensures seamless interop at the FFI boundary
- `std::hash<Address>` uses wyhash (via ankerl), matching `bytes32_t`; `static_assert` for `standard_layout` and `trivially_copyable` codifies FFI safety
- `_address` literal and `address_from_hex()` reuse existing `from_hex<FixedBytes>` template
- Migrates `evmc::address` → `Address` in test code (evm-as tests, fuzzer, micro benchmarks)
- Coalesces `using evmc::literals::operator""_address` / `_bytes32` into `using namespace monad::literals;`
- Adds `monad_core` as a dependency of `monad-vm-evm` (for transitive `unordered_dense` include path)
- Adds 30 unit tests in `category/core/address_test.cpp`

## Test plan

- [x] Full ctest suite passes (3515 tests, 4 pre-existing timeouts on long ethereum compliance suites)
- [x] `address_test`: 30/30 tests pass (constructors, operators, literals, hashing, conversions, is_zero, sizeof/alignof, type traits, constexpr comparisons, evmc_address round-trip)
- [x] clang-format clean
- [x] clang-tidy clean
- [x] Trait instantiation check clean
- [ ] CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2165, opened 2026-03-30
Introduce monad::bytes32_t and uint256_be_t, replacing evmc_bytes32 dependency

Baltoli:
Replace the `using bytes32_t = evmc::bytes32` alias with a standalone struct inheriting from `evmc_bytes32`. This gives us control over construction, comparison, and hashing while maintaining ABI compatibility with the EVMC C layer.

`bytes32_t` features:
- Zero-initializing default constructor
- Implicit converting constructor from `evmc_bytes32`
- Explicit `uint64_t` constructor (big-endian storage)
- Variadic byte constructor accepting 1–32 `uint8_t` arguments (no narrowing)
- Implicit conversion to `byte_string_view`
- Explicit conversion to `bool` (non-zero check via `std::any_of`)
- `operator==` and `operator<=>` (lexicographic)
- `std::hash` and `ankerl::unordered_dense::hash` specializations (wyhash)

Other changes:
- Add `uint256_be_t` as a type alias for `bytes32_t`, replacing `evmc_uint256be` in internal code to convey big-endian uint256 semantics
- Add `monad::literals::operator""_bytes32` and `bytes32_from_hex()`, replacing the blanket `using namespace evmc::literals` import
- Migrate all internal `evmc_bytes32`/`evmc_uint256be` usage to `bytes32_t`/`uint256_be_t` across staking, reserve balance, WAL reader, event recording, and test/fuzzer code
- Remove `using namespace monad` from `instrumentable_compiler.hpp` header to fix ambiguous `operator""_bytes32` resolution with evmone headers
- Add explicit `bytes32_t()` casts in `blockchain_test.cpp` and `fuzzer.cpp` where `evmc::bytes32` (from `monad_c_bytes32` ctypes) is compared with `bytes32_t`

Conversion to `evmc::bytes32` at the EVMC boundary works implicitly through the base class path (`bytes32_t` → `evmc_bytes32` → `evmc::bytes32` constructor). Remaining `evmc::bytes32` uses are at the EVMC interface boundary (host overrides, C API callbacks, arrays passed to EVMC functions) and must stay as evmc types.

PR #2538, opened 2026-09-04
[execution/vm] Port evmc_status_code to an in-tree Monad type

khordadi:
Closes EXE-160. Unblocks EXE-173 (`evmc::Result` + the `vm::Host` vtable).

Last of the four evmc leaf types, after `bytes32_t` (#2165), `Address` (#2196) and `monad_eth_revision` (#2304). Follows the last of those closely.

## How to review

- **Commit 1** adds the type. Nothing references it yet. Builds and tests green standalone.
- **Commit 2** migrates. Two conversions; every other line is `EVMC_X` → `MONAD_STATUS_X`.

## The issue's three open questions

**The fork.** `d51cc94` on top of the 12.1.0 tag adds `EVMC_MONAD_RESERVE_BALANCE_VIOLATION = 18`; upstream stops at 17. So 22 enumerators, and the one with no upstream reference gets its own test. It also can't go through the `MONAD_STATUS_##name == EVMC_##name` macro, because evmc spells it with a vendor prefix — that assert is longhand.

**The boundary.** A status code changes type exactly where `evmc_result` does. That's forced, not chosen: `vm::Host` derives from `evmc::Host` (`vm/host.hpp:28`), so `call()`'s signature isn't ours until EXE-173.

**Proving the mirror.** Per-code `static_assert` on `std::to_underlying`, as in `revision.cpp`. A wrong value names both numbers: `note: expression evaluates to '19 == 18'`. That's what lets `to_evmc_status_code` be a bare cast.

## Scope

| | |
|---|---|
| `PrecompileResult::status_code`, `CallFrame::status` | **moved** — Monad-owned |
| `evmc::Result::status_code`, all sites | **kept** — layout fixed by `evmc_result`, behind the vtable. EXE-173 |
| variant from `Context::copy_result_data` | **kept** — private carrier for `copy_to_evmc_result`, not storage |
| `Receipt::status` | **untouched** — `uint64_t` R_z, not a status code |
| transitive `<evmc/...>` includes | **left** — separate sweep, per the issue |

Wire and ABI encodings are byte-identical, and the lines that carry them aren't in the diff: `call_frame_rlp.cpp:65` and `record_txn_events.cpp:419` already used width-agnostic conversions. `exec_event_ctypes.h` is untouched, so the bindgen Rust side is unaffected.

Kept flat-C rather than `enum class` because the safety that would buy is already there — mixing the two enums is a hard error under `-Werror` on both compilers, which is also what caught a `CallFrame` initialiser I'd missed.

## Verification

`ctest` 3892/3892 · clang-19 ASAN+UBSAN clean · clang-tidy clean · `check-trait-instantiations` · clang-format · g++-15 and clang++-19 × Debug/RelWithDebInfo · commit 1 green standalone.

TSAN not run: `clang-tsan.cmake` has a hardcoded absolute blacklist path, and this diff adds no concurrency.

## Found in passing — please file

Three, all pre-existing, none fixed here.

**1. Trace RLP can't represent the status domain.** `call_frame_rlp.cpp:65` encodes the status as one byte. Negatives are lossy, and the decoded value is then outside the enum's `[-32,31]` range, so every later read is UB — UBSAN flags it inside `std::to_underlying`, which is what `record_txn_events.cpp:419` calls in production:

```
INTERNAL_ERROR (-1)   in -1 -> out 255  LOSSY
runtime error: load of value 255, which is not a valid value for type 'monad_status_code'
```

Identical for `evmc_status_code` before this PR, so not a regression. I found no live path that persists a negative code, so this is latent. Fixing it needs a compatibility decision for stored traces.

**2. `Receipt{.status = EVMC_SUCCESS}` is 0 — a *failed* receipt.** Six sites in `monad_executor_test.cpp`. R_z is 1 for success (`record_txn_events.cpp:374`). `test_db.cpp` uses plain `.status = 1`.

**3. `precompiles_test.cpp:528` signed overflow** (UBSAN). MONAD_SEVEN's blake2F price-doubling multiplies the `INT64_MAX` default gas by 2, giving `-2`. Line is `d88013808c`, June; not in this diff. Tests still pass, so that gas value isn't checked where the comment above it implies.

---

Review rationale is on the diff lines it applies to, not in the source.

### DIFF
--- a/category/execution/ethereum/precompiles.cpp
+++ b/category/execution/ethereum/precompiles.cpp
@@ -24,6 +24,7 @@
 #include <category/execution/ethereum/precompiles.hpp>
 #include <category/execution/ethereum/precompiles_impl.hpp>
 #include <category/vm/evm/explicit_traits.hpp>
+#include <category/vm/evm/status_code.h>
 
 #include <evmc/evmc.h>
 #include <evmc/evmc.hpp>
@@ -166,8 +167,8 @@ std::optional<evmc::Result> check_call_eth_precompile(evmc_message const &msg)
 
     auto const [status_code, output_buffer, output_size] = execute_func(input);
     return evmc::Result{evmc_result{
-        .status_code = status_code,
-        .gas_left = (status_code == EVMC_SUCCESS)
+        .status_code = to_evmc_status_code(status_code),
+        .gas_left = (status_code == MONAD_STATUS_SUCCESS)
                         ? msg.gas - static_cast<int64_t>(cost.value())
                         : 0,
         .gas_refund = 0,
@@ -201,9 +202,9 @@ from_impl_result(PrecompileImplResult result, uint8_t *out)
     }
     if (size == 0) {
         std::free(out);
-        return {EVMC_SUCCESS, nullptr, size};
+        return {MONAD_STATUS_SUCCESS, nullptr, size};
     }
-    return {EVMC_SUCCESS, data, size};
+    return {MONAD_STATUS_SUCCESS, data, size};
 }
 
 PrecompileResult ecrecover_execute(byte_string_view const input)
@@ -218,11 +219,11 @@ PrecompileResult ecrecover_execute(byte_string_view const input)
     auto const s{load_be_unsafe<uint256_t>(&d[96])};
 
     if (!Secp256k1Signature{r, s}.has_valid_range()) {
-        return {EVMC_SUCCESS, nullptr, 0};
+        return {MONAD_STATUS_SUCCESS, nullptr, 0};
     }
 
     if (v != 27 && v != 28) {
-        return {EVMC_SUCCESS, nullptr, 0};
+        return {MONAD_STATUS_SUCCESS, nullptr, 0};
     }
 
     uint8_t *out{static_cast<uint8_t *>(std::aligned_alloc(8, 32))};
@@ -235,10 +236,10 @@ PrecompileResult ecrecover_execute(byte_string_view const input)
             std::span<uint8_t const, 64>{&d[64], 64},
             v != 27)) {
         std::free(out);
-        return {EVMC_SUCCESS, nullptr, 0};
+        return {MONAD_STATUS_SUCCESS, nullptr, 0};
     }
     else {
-        return {EVMC_SUCCESS, out, 32};
+        return {MONAD_STATUS_SUCCESS, out, 32};
     }
 }
 
@@ -277,7 +278,7 @@ PrecompileResult expmod_execute(byte_string_view const input)
     uint64_t const mod_len = load_be_unsafe<uint64_t>(&header[88]);
 
     if (mod_len == 0) {
-        return {EVMC_SUCCESS, nullptr, 0};
+        return {MONAD_STATUS_SUCCESS, nullptr, 0};
     }
 
     uint64_t const base_len = load_be_unsafe<uint64_t>(&header[24]);
@@ -420,21 +421,21 @@ PrecompileResult p256_verify_execute(byte_string_view const input)
         p256_verify_impl(input, std::span<uint8_t, 32>{out, 32});
     if (result.data == nullptr) {
         std::free(out);
-        return {EVMC_SUCCESS, nullptr, 0};
+        return {MONAD_STATUS_SUCCESS, nullptr, 0};
     }
-    return {EVMC_SUCCESS, result.data, result.size};
+    return {MONAD_STATUS_SUCCESS, result.data, result.size};
 }
 
 PrecompileResult identity_execute(byte_string_view const input)
 {
     if (input.empty()) {
-        return {EVMC_SUCCESS, nullptr, 0};
+        return {MONAD_STATUS_SUCCESS, nullptr, 0};
     }
     auto *const out = static_cast<uint8_t *>(std::malloc(input.size()));
     MONAD_ASSERT(out != nullptr);
     auto const result =
         identity_impl(input, std::span<uint8_t>{out, input.size()});
-    return {EVMC_SUCCESS, result.data, result.size};
+    return {MONAD_STATUS_SUCCESS, result.data, result.size};
 }
 
 MONAD_NAMESPACE_END
--- a/category/execution/ethereum/precompiles.hpp
+++ b/category/execution/ethereum/precompiles.hpp
@@ -19,6 +19,7 @@
 #include <category/core/byte_string.hpp>
 #include <category/core/config.hpp>
 #include <category/core/int.hpp>
+#include <category/vm/evm/status_code.h>
 #include <category/vm/evm/traits.hpp>
 
 #include <evmc/evmc.h>
@@ -123,14 +124,14 @@ uint64_t p256_verify_gas_cost(byte_string_view);
 
 struct PrecompileResult
 {
-    evmc_status_code status_code;
+    monad_status_code status_code;
     uint8_t *obuf;
     size_t output_size;
 
     static constexpr PrecompileResult failure() noexcept
     {
         return {
-            .status_code = EVMC_PRECOMPILE_FAILURE,
+            .status_code = MONAD_STATUS_PRECOMPILE_FAILURE,
             .obuf = nullptr,
             .output_size = 0,
         };
--- a/category/execution/ethereum/trace/call_frame.hpp
+++ b/category/execution/ethereum/trace/call_frame.hpp
@@ -20,6 +20,7 @@
 #include <category/core/config.hpp>
 #include <category/core/int.hpp>
 #include <category/execution/ethereum/core/receipt.hpp>
+#include <category/vm/evm/status_code.h>
 
 #include <evmc/evmc.hpp>
 #include <nlohmann/json_fwd.hpp>
@@ -81,7 +82,7 @@ struct CallFrame
     uint64_t gas_used{};
     byte_string input{};
     byte_string output{};
-    evmc_status_code status{};
+    monad_status_code status{};
     uint64_t depth{};
     std::optional<std::vector<Log>> logs{};
 
--- a/category/execution/ethereum/trace/call_tracer.cpp
+++ b/category/execution/ethereum/trace/call_tracer.cpp
@@ -25,6 +25,7 @@
 #include <category/execution/ethereum/core/transaction.hpp>
 #include <category/execution/ethereum/trace/call_frame.hpp>
 #include <category/execution/ethereum/trace/call_tracer.hpp>
+#include <category/vm/evm/status_code.h>
 
 #include <evmc/evmc.h>
 #include <evmc/evmc.hpp>
@@ -149,7 +150,7 @@ void CallTracer::on_enter(evmc_message const &msg)
                      ? byte_string{}
                      : byte_string{msg.input_data, msg.input_size},
         .output = {},
-        .status = EVMC_FAILURE,
+        .status = MONAD_STATUS_FAILURE,
         .depth = depth,
         .logs = std::vector<CallFrame::Log>{},
     });
@@ -173,7 +174,7 @@ void CallTracer::on_exit(evmc::Result const &res)
                            ? byte_string{}
                            : byte_string{res.output_data, res.output_size};
     }
-    frame.status = res.status_code;
+    frame.status = from_evmc_status_code(res.status_code);
 
     if (frame.type == CallType::CREATE || frame.type == CallType::CREATE2) {
         frame.to = is_zero(res.create_address)
@@ -217,7 +218,7 @@ void CallTracer::on_self_destruct(
         .gas_used = 0,
         .input = {},
         .output = {},
-        .status = EVMC_SUCCESS, // TODO
+        .status = MONAD_STATUS_SUCCESS, // TODO
         .depth = parent.depth + 1,
         .logs = std::vector<CallFrame::Log>{},
     });
--- a/category/execution/ethereum/trace/rlp/call_frame_rlp.cpp
+++ b/category/execution/ethereum/trace/rlp/call_frame_rlp.cpp
@@ -24,6 +24,7 @@
 #include <category/execution/ethereum/rlp/decode.hpp>
 #include <category/execution/ethereum/rlp/encode2.hpp>
 #include <category/execution/ethereum/trace/call_frame.hpp>
+#include <category/vm/evm/status_code.h>
 
 #include <boost/outcome/try.hpp>
 
@@ -125,7 +126,7 @@ Result<CallFrame> decode_call_frame(byte_string_view &enc)
     BOOST_OUTCOME_TRY(call_frame.output, decode_string(payload));
     BOOST_OUTCOME_TRY(
         auto const status, decode_unsigned<unsigned char>(payload));
-    call_frame.status = static_cast<enum evmc_status_code>(status);
+    call_frame.status = static_cast<enum monad_status_code>(status);
     BOOST_OUTCOME_TRY(call_frame.depth, decode_unsigned<uint64_t>(payload));
 
     if (MONAD_UNLIKELY(payload.empty())) {
--- a/category/rpc/monad_executor_test.cpp
+++ b/category/rpc/monad_executor_test.cpp
@@ -59,6 +59,7 @@
 #include <category/vm/code.hpp>
 #include <category/vm/evm/delegation.hpp>
 #include <category/vm/evm/monad/revision.h>
+#include <category/vm/evm/status_code.h>
 #include <category/vm/evm/traits.hpp>
 #include <category/vm/interpreter/intercode.hpp>
 #include <category/vm/utils/evm-as.hpp>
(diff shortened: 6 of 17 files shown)


category/vm/evm/status_code.h
```diff
@@ -0,0 +1,66 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <evmc/evmc.h>
```
khordadi, line 18: The only tie to evmc; goes with EXE-173.

Consequence: this header can't be included from the C ABI headers, which stay evmc-free unless `MONAD_C_CTYPES_USE_EVMC` opts in. So their status fields stay plain ints — correct either way, given the mirroring.

category/vm/evm/status_code.h
```diff
@@ -0,0 +1,66 @@
...
+    MONAD_STATUS_CALL_DEPTH_EXCEEDED = 10,
+    MONAD_STATUS_STATIC_MODE_VIOLATION = 11,
+    MONAD_STATUS_PRECOMPILE_FAILURE = 12,
+    MONAD_STATUS_CONTRACT_VALIDATION_FAILURE = 13,
+    MONAD_STATUS_ARGUMENT_OUT_OF_RANGE = 14,
+    MONAD_STATUS_WASM_UNREACHABLE_INSTRUCTION = 15,
+    MONAD_STATUS_WASM_TRAP = 16,
+    MONAD_STATUS_INSUFFICIENT_BALANCE = 17,
+    // Fork-only: no upstream evmc counterpart.
+    MONAD_STATUS_RESERVE_BALANCE_VIOLATION = 18,
```
khordadi, line 52: Fork-only: `d51cc94`, on top of the 12.1.0 tag. Upstream stops at 17.

No upstream reference means nothing would catch this drifting on a fork rebase, hence its own test.

category/vm/evm/status_code.h
```diff
@@ -0,0 +1,66 @@
...
+
+    MONAD_STATUS_INTERNAL_ERROR = -1,
+    MONAD_STATUS_REJECTED = -2,
+    MONAD_STATUS_OUT_OF_MEMORY = -3
+};
+
+char const *monad_status_code_to_string(enum monad_status_code code);
```
khordadi, line 59: No caller yet — I checked, `to_json` emits `"REVERT"`/`"ERROR"`, not code names.

Kept to match `monad_eth_revision_to_string`, and because `-Wswitch` on it is the trip-wire below. Drop it if you'd rather not carry an uncalled function.

category/vm/evm/status_code.cpp
```diff
@@ -0,0 +1,166 @@
...
+#undef MONAD_ASSERT_STATUS_EQ
+
+// Fork-only, and evmc's spelling carries a vendor prefix the macro can't reach.
+static_assert(
+    std::to_underlying(MONAD_STATUS_RESERVE_BALANCE_VIOLATION) ==
+    std::to_underlying(EVMC_MONAD_RESERVE_BALANCE_VIOLATION));
+
+// Safe while every code has an evmc counterpart. If you add a Monad-only one,
+// give this a range guard: -Wswitch on to_string below is the only trip-wire.
+evmc_status_code to_evmc_status_code(monad_status_code const code)
```
khordadi, line 60: Cast is value-preserving via the asserts above, and needs no range guard since there's no divergent tail (unlike `to_evmc_revision`).

But the asserts only cover the codes they name: a Monad-only enumerator would cast into nothing. Only `-Wswitch` on `to_string` catches that — hence the comment.

khordadi (commented):
Author's notes on the diff — decisions a reviewer would otherwise have to ask about.

category/vm/evm/status_code.cpp
```diff
@@ -0,0 +1,166 @@
...
+    case EVMC_MONAD_RESERVE_BALANCE_VIOLATION:
+        return MONAD_STATUS_RESERVE_BALANCE_VIOLATION;
+    case EVMC_INTERNAL_ERROR:
+        return MONAD_STATUS_INTERNAL_ERROR;
+    case EVMC_REJECTED:
+        return MONAD_STATUS_REJECTED;
+    case EVMC_OUT_OF_MEMORY:
+        return MONAD_STATUS_OUT_OF_MEMORY;
+    }
+    MONAD_ABORT("unhandled evmc_status_code");
```
khordadi, line 114: Unreachable in-tree: only Monad's VM and evmone sit behind here, both returning defined codes. `evmc.h:282` does permit a VM to return undefined negatives, so if a third-party VM ever appears the answer is `MONAD_STATUS_INTERNAL_ERROR`, not an abort.

test/vm/unit/status_code_tests.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace
+{
+    // evmc_status_code is two contiguous ranges. Sweeping them rather than a
+    // list of names keeps these tests total over the enum; if evmc gains a code
+    // outside them, -Wswitch in from_evmc_status_code fails first.
+    constexpr int status_min = -3;
```
khordadi, line 32: Sweeping the range beats a name list, which can silently fall behind the header.

Literals rather than derived bounds: deriving would turn a future Monad-only addition into a crash here (`from_evmc_status_code`'s abort) instead of a pass.

File is in `test/vm/unit` because `category/vm/` has no test wiring — `monad_add_test_folder` isn't called there, so a test beside the header would run nowhere.

category/execution/ethereum/precompiles.hpp
```diff
@@ -123,14 +124,14 @@ uint64_t p256_verify_gas_cost(byte_string_view);
 
 struct PrecompileResult
 {
-    evmc_status_code status_code;
+    monad_status_code status_code;
```
khordadi, line 127: ~20 producers move, one consumer converts. That ratio is why this field is worth migrating and `evmc::Result::status_code` isn't.

category/execution/ethereum/precompiles.cpp
```diff
@@ -165,8 +166,8 @@ std::optional<evmc::Result> check_call_eth_precompile(evmc_message const &msg)
 
     auto const [status_code, output_buffer, output_size] = execute_func(input);
     return evmc::Result{evmc_result{
-        .status_code = status_code,
-        .gas_left = (status_code == EVMC_SUCCESS)
+        .status_code = to_evmc_status_code(status_code),
```
khordadi, line 169: **Seam 1 of 2.** The boundary is wherever `evmc_result` is — `vm::Host` derives from `evmc::Host` (`vm/host.hpp:28`), so `call()` is fixed until EXE-173.

The comparison below had to move too: cross-enum is a hard error under `-Werror`.

category/execution/ethereum/trace/call_frame.hpp
```diff
@@ -81,7 +82,7 @@ struct CallFrame
     uint64_t gas_used{};
     byte_string input{};
     byte_string output{};
-    evmc_status_code status{};
+    monad_status_code status{};
```
khordadi, line 85: `sizeof(CallFrame) == 216` still holds — `int` underlying type, same as before. Though it wouldn't have caught a width change either way; the field sits in padding.

category/execution/ethereum/trace/call_tracer.cpp
```diff
@@ -173,7 +174,7 @@ void CallTracer::on_exit(evmc::Result const &res)
                            ? byte_string{}
                            : byte_string{res.output_data, res.output_size};
     }
-    frame.status = res.status_code;
+    frame.status = from_evmc_status_code(res.status_code);
```
khordadi, line 177: **Seam 2 of 2**, reverse direction. A switch, not a cast, so `-Wswitch` breaks the build if evmc gains a code.

The `res.status_code` comparison above stays evmc — that's an `evmc::Result` field.

category/execution/ethereum/trace/rlp/call_frame_rlp.cpp
```diff
@@ -125,7 +126,7 @@ Result<CallFrame> decode_call_frame(byte_string_view &enc)
     BOOST_OUTCOME_TRY(call_frame.output, decode_string(payload));
     BOOST_OUTCOME_TRY(
         auto const status, decode_unsigned<unsigned char>(payload));
-    call_frame.status = static_cast<enum evmc_status_code>(status);
+    call_frame.status = static_cast<enum monad_status_code>(status);
```
khordadi, line 129: Type only; value identical. The encode side (line 65) isn't in this diff, so the RLP is byte-identical.

That encoding has two pre-existing problems, preserved here and filed: negatives are lossy (`-1` → `255`), and since the enum has no fixed underlying type its range is `[-32,31]`, so out-of-range reads are UB. Both were equally true of `evmc_status_code`. Fixing needs a compatibility decision for stored traces.

category/rpc/monad_executor_test.cpp
```diff
@@ -1617,7 +1618,7 @@ TEST_F(EthCallFixture, static_precompile_OOG_with_call_trace)
         .gas = 22000,
         .gas_used = 22000,
         .input = byte_string(data),
-        .status = EVMC_OUT_OF_GAS,
+        .status = MONAD_STATUS_OUT_OF_GAS,
```
khordadi, line 1621: Careful: `Receipt` also has `.status` — a `uint64_t` R_z, not a status code. I targeted `CallFrame`'s by its neighbouring `.depth` and left all 6 `Receipt` sites alone.

Those say `.status = EVMC_SUCCESS`, which is 0 — a *failed* receipt. Filed.

category/execution/ethereum/trace/rlp/call_frame_rlp.cpp
```diff
@@ -125,7 +126,7 @@ Result<CallFrame> decode_call_frame(byte_string_view &enc)
...
     BOOST_OUTCOME_TRY(call_frame.depth, decode_unsigned<uint64_t>(payload));
```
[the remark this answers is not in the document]
khordadi: It's not introduced by this PR.

test/vm/unit/status_code_tests.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    std::vector<monad_status_code> all_status_codes()
+    {
+        std::vector<monad_status_code> codes;
+        for (int i = status_min; i <= status_max; ++i) {
+            codes.push_back(
+                from_evmc_status_code(static_cast<evmc_status_code>(i)));
+        }
+        return codes;
+    }
```
goodlyrottenapple: Could use magic enum instead, e.g. 
```suggestion
    auto const all_status_codes = [] {
        std::array<
            monad_status_code,
            magic_enum::enum_count<evmc_status_code>()>
            codes;
        std::ranges::transform(
            magic_enum::enum_values<evmc_status_code>(),
            codes.begin(),
            from_evmc_status_code);
        return codes;
    }();
```

khordadi: Doesn't compile under GCC with our flags — `magic_enum` can't enumerate `evmc_status_code`. Self-contained repro, no monad headers needed:

```cpp
#include <evmc/evmc.h>
#include <magic_enum/magic_enum.hpp>
#include <algorithm>
#include <array>

// stand-ins for the two monad names, so this needs no monad headers
enum monad_status_code { MONAD_STATUS_SUCCESS = 0 };
monad_status_code from_evmc_status_code(evmc_status_code);

auto const all_status_codes = [] {
    std::array<
        monad_status_code,
        magic_enum::enum_count<evmc_status_code>()>
        codes;
    std::ranges::transform(
        magic_enum::enum_values<evmc_status_code>(),
        codes.begin(),
        from_evmc_status_code);
    return codes;
}();
```

```bash
g++-15 -std=c++23 -Wall -Wextra -Wconversion -Werror \
  -Ithird_party/evmc/include -Ithird_party/magic_enum/include \
  -fsyntax-only t.cpp
```

```
magic_enum.hpp:630:19: error: the result of the conversion is unspecified
  because '-128' is outside the range of type 'evmc_status_code'
  [-Werror=conversion]
  630 |   constexpr E v = static_cast<E>(V);
      |                   ^~~~~~~~~~~~~~~~~
```

`magic_enum` probes `[-128, 127]` with `static_cast<E>(V)`. `evmc_status_code` has no fixed underlying type, so its range is only `[-32, 31]` and the `-128` cast is unspecified — `-Wconversion` reports it, `-Werror` makes it fatal. Clang accepts the same file, which is why it looks fine at a glance.

Reduces to three lines:

```cpp
#include <evmc/evmc.h>
#include <magic_enum/magic_enum.hpp>
auto n = magic_enum::enum_count<evmc_status_code>();
```

It works on `enum class Transition` in the staking fuzzer because that has a fixed underlying type.

Baltoli: > magic_enum can't enumerate evmc_status_code

In the period where we're working on replacing evmc, we're still free to make changes to our own fork of evmc: in this case if we can get some simplification from magic_enum it might be worth doing so. It should be harmless / well-defined to just give evmc's status code an underlying type?

category/vm/evm/status_code.h
Baltoli: What do you mean by a trip-wire if the function is uncalled?

test/vm/unit/status_code_tests.cpp
goodlyrottenapple: or just put a pragma around the magic enum header in the test file


```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wconversion"
#include <magic_enum/magic_enum.hpp>
#pragma GCC diagnostic pop
```

category/vm/evm/status_code.h
khordadi: Compile-time trip-wire: `-Wswitch` still checks the switch even if the function is never called, so adding an unhandled enum value can break the build.

category/vm/evm/status_code.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+char const *monad_status_code_to_string(monad_status_code const code)
```
goodlyrottenapple: Perhaps another case for magic_enum?

test/vm/unit/status_code_tests.cpp
khordadi: Works; verified on gcc-15 and clang-19, whole tree builds, tests pass. I tried both:
- **`enum evmc_status_code : int`:** no pragma needed, but we need to change both `third_party/evmc` and `third_party/evmone/evmc`. This also fix the RLP UB.
- `#pragma`: 4 lines in this file, no cross-repo work. I'm going with this in this PR.

category/vm/evm/status_code.cpp
khordadi: I decided to remove the function and its tests altogether; if we need it later we can implement it using magic_enum.

category/vm/evm/status_code.h
```diff
@@ -0,0 +1,64 @@
...
+extern "C"
+{
+#endif
+
+// Mirrors evmc_status_code 1:1; values asserted equal in status_code.cpp, so
+// every wire and ABI encoding of a status is unchanged. Coarser VM exit codes
+// live in vm::runtime::StatusCode.
+//
+// The <evmc/evmc.h> include serves only the conversions below; both go with
+// EXE-173. Do not include this from the evmc-free C ABI headers.
```
Baltoli: This comment seems a bit superfluous - I don't think it's adding any documentary value, would probably just delete it

category/vm/evm/status_code.cpp
Baltoli, line 1: I would likewise take out the comments in this file; they're stating the obvious or something that will get flagged by the compiler anyway

category/vm/evm/status_code.h
```diff
@@ -0,0 +1,64 @@
...
+enum monad_status_code
```
Baltoli:
```suggestion
enum monad_status_code : int
```

test/vm/unit/status_code_tests.cpp
```diff
@@ -0,0 +1,110 @@
...
+          MONAD_STATUS_OUT_OF_MEMORY}) {
+        EXPECT_LT(std::to_underlying(to_evmc_status_code(code)), 0);
+        EXPECT_EQ(
+            std::to_underlying(
+                from_evmc_status_code(to_evmc_status_code(code))),
+            std::to_underlying(code));
+    }
+}
+
+// Fork-only, so no upstream reference; most likely value to move on a rebase.
```
Baltoli: We're in control of this code; it's not something that's going to silently change under us - I think this test is redundant

test/vm/unit/status_code_tests.cpp
```diff
@@ -0,0 +1,110 @@
...
+         {MONAD_STATUS_INTERNAL_ERROR,
+          MONAD_STATUS_REJECTED,
+          MONAD_STATUS_OUT_OF_MEMORY}) {
+        EXPECT_LT(std::to_underlying(to_evmc_status_code(code)), 0);
+        EXPECT_EQ(
+            std::to_underlying(
+                from_evmc_status_code(to_evmc_status_code(code))),
+            std::to_underlying(code));
+    }
+}
```
Baltoli: I think these tests are also redundant - my reading is that `MirrorsEvmcValues` and `RoundTripsThroughEvmc` are complete, the remaining tests are just restatements

Baltoli (approved):

goodlyrottenapple (approved):

PR #2539, opened 2026-09-05
[execution/vm] Add in-tree `monad_tx_context` type

khordadi:
Part of EXE-171 (EXE-162), reduced to the type. `monad_tx_context` mirrors `evmc_tx_context` field-for-field with `bytes32_t`/`Address` members, including the fork's `block_round`; `tx_context.cpp` asserts layout, field offsets and integer field types. Purely additive, nothing references the type yet.

The producer migration moves into the `vm::Host` port, EXE-173. With `get_tx_context` returning the in-tree type the host holds no `evmc_tx_context` copy, which this PR's earlier shape needed (256 bytes per host instance).

**Verification.** gcc-15 RelWithDebInfo build clean; `clang-format-19` and `clang-tidy-19` on the new files clean. `tx_context.cpp` includes `<evmc/evmc.h>` exactly as `revision.cpp` and `status_code.cpp` already do, so the zkVM guest sees nothing new.

Conflicts trivially with #2540 on `category/vm/evm/CMakeLists.txt`; whichever lands first, the other rebases.


category/vm/evm/tx_context.hpp
```diff
@@ -0,0 +1,51 @@
...
+    int64_t block_timestamp;
+    int64_t block_gas_limit;
+    monad::bytes32_t block_prev_randao;
+    monad::bytes32_t chain_id;
+    monad::bytes32_t block_base_fee;
+    monad::bytes32_t blob_base_fee;
+    monad::bytes32_t const *blob_hashes;
+    size_t blob_hashes_count;
+    // EOF TXCREATE initcodes: never populated; kept for evmc layout parity.
+    void const *initcodes;
```
khordadi, line 41: Kept only so offsets match; nothing in-tree populates it (TXCREATE is EOF). Once evmc is cut the mirror can drop both fields and re-layout.

category/vm/evm/tx_context.hpp
```diff
@@ -0,0 +1,51 @@
...
+    size_t initcodes_count;
+    // Monad consensus round (EIP-7843 SLOTNUM); 0 for EVM.
+    uint64_t block_round;
+};
+
+struct evmc_tx_context;
```
khordadi, line 47: Forward declaration keeps this header evmc-free; the only evmc include is in tx_context.cpp.

category/vm/evm/tx_context.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <type_traits>
+
+// Layout equality with evmc_tx_context; the bit_cast below relies on it.
+static_assert(std::is_standard_layout_v<monad_tx_context>);
+static_assert(std::is_trivially_copyable_v<monad_tx_context>);
+static_assert(sizeof(monad_tx_context) == 264);
+static_assert(sizeof(monad_tx_context) == sizeof(evmc_tx_context));
+static_assert(alignof(monad_tx_context) == alignof(evmc_tx_context));
+
+#define MONAD_ASSERT_TX_CONTEXT_FIELD_EQ(field)                                \
```
khordadi, line 31: Offsets only: per-field `sizeof` on the two pointer members trips `bugprone-sizeof-expression`. The type asserts below cover the narrowed-integer case that offsets alone would miss.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -52,7 +53,8 @@ class EvmcHostBase : public vm::Host
     BlockHashBuffer const &block_hash_buffer_;
 
 protected:
-    evmc_tx_context const &tx_context_;
+    // evmc copy for the get_tx_context vtable slot; goes with EXE-173.
+    evmc_tx_context const tx_context_;
```
khordadi, line 57: By value because the vtable slot returns a pointer, so the evmc copy needs stable storage. `reinterpret_cast` of the caller's `monad_tx_context` would be an aliasing violation for the interpreter's reads. 264 bytes, once per transaction.

khordadi (commented):
Notes for the reviewer on the non-obvious lines.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -107,7 +109,7 @@ class EvmcHostBase : public vm::Host
         evmc_storage_status) noexcept override;
 };
 
-static_assert(sizeof(EvmcHostBase) == 72);
+static_assert(sizeof(EvmcHostBase) == 328);
```
khordadi, line 112: Growth is the 264-byte copy replacing an 8-byte reference (+256).

category/execution/ethereum/evmc_host_test.cpp
```diff
@@ -40,39 +41,22 @@
...
 
 #include <cstdint>
-#include <cstring>
 
 using namespace monad;
 
 using db_t = TrieDb;
 
-bool operator==(evmc_tx_context const &lhs, evmc_tx_context const &rhs)
+bool operator==(monad_tx_context const &lhs, monad_tx_context const &rhs)
```
khordadi, line 49: Members have their own `==`, so the memcmp form goes. Compared field set unchanged; `blob_base_fee` still excluded as before.

category/execution/ethereum/evmc_host.hpp
```diff
@@ -52,7 +53,8 @@ class EvmcHostBase : public vm::Host
     BlockHashBuffer const &block_hash_buffer_;
 
 protected:
-    evmc_tx_context const &tx_context_;
+    // evmc copy for the get_tx_context vtable slot; goes with EXE-173.
```
goodlyrottenapple:
```suggestion

```

category/vm/evm/tx_context.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+// Bridge for the evmc::HostInterface::get_tx_context vtable slot; goes with
+// EXE-173.
+evmc_tx_context to_evmc_tx_context(monad_tx_context const &);
```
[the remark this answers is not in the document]
khordadi: Moot: `to_evmc_tx_context` was removed when the PR was cut back to the type only (no caller); the bridge returns with the call-site migration under EXE-173, where it will be `noexcept`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

category/vm/evm/tx_context.hpp
```diff
@@ -0,0 +1,45 @@
...
+    monad::bytes32_t block_prev_randao;
+    monad::bytes32_t chain_id;
+    monad::bytes32_t block_base_fee;
+    monad::bytes32_t blob_base_fee;
+    monad::bytes32_t const *blob_hashes;
+    size_t blob_hashes_count;
+    // EOF TXCREATE initcodes: never populated; kept for evmc layout parity.
+    void const *initcodes;
+    size_t initcodes_count;
+    // Monad consensus round (EIP-7843 SLOTNUM); 0 for EVM.
```
goodlyrottenapple:
```suggestion

```

goodlyrottenapple (dismissed):

category/vm/evm/tx_context.hpp
Baltoli, line 1: Is this intended to be C or C++? The name is spelled like a C type (`monad_tx_context`), but it uses C++ namespaces (`monad::bytes32_t tx_gas_price`) and is in a C++ header. If the intention is for this to be a C type, we should commit to that (and vice-versa for C++)

Baltoli (changes requested):

category/vm/evm/tx_context.hpp
khordadi: C++. The members have constructors, so a C compiler cannot parse either header. Making it real C would require a C bytes32 and a C address type in core.

goodlyrottenapple: The original `evmc_tx_context` is a C type and part of a C interface. I would lean towards making the monad version C as well

khordadi: I had another look, and as [the OCaml oracle build derives layouts from a C header](https://github.com/category-labs/monad-execution-specs/blob/e5ff4efb592d6aeda764eed86776ed24c4c2f7cf/bindings/c_evmc/dune#L17-L32), I think we must use C (at least eventually)! However, we need a C `bytes32` and an `address` type first to compile it as C.

goodlyrottenapple: there are https://github.com/category-labs/monad/blob/2e1c9fa9ba7ca7c71aca818151831e61e0bb6939/category/execution/ethereum/core/base_ctypes.h

khordadi: This can't be used in the vm because it's under `category/execution`, and is not linked to `monad-vm`. :(

goodlyrottenapple: actually based on the slack conversation, its fine to keep these as cpp i think

category/vm/evm/tx_context.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <category/core/address.hpp>
+#include <category/core/bytes.hpp>
+
+#include <cstddef>
+#include <cstdint>
+
+// Mirrors evmc_tx_context field-for-field, including the category-labs fork's
+// block_round; layout equality is asserted in tx_context.cpp.
+struct monad_tx_context
```
goodlyrottenapple:
```suggestion
struct TxContext
```

also wrap in the `monad` namespace?

khordadi: Used `monad::vm`.

goodlyrottenapple (approved):

Baltoli (approved):

PR #2541, opened 2026-09-07
zkVM guest optimizations #1: misc soundness

anne-laure-s:
This PR hardens soundness for the guest:
- more bindings inside the guest
- encoding/decoding consistency
- sound unreachable 

This is the first PR of the optimization series r10

### DIFF
--- a/category/execution/ethereum/db/offset_trie.cpp
+++ b/category/execution/ethereum/db/offset_trie.cpp
@@ -215,7 +215,7 @@ NodeViewBase OffsetTrie::find_original(NodeId id, NibblesView key) const
                     // There is only a single NullView node, namely the NULL_ID
                     // magic header. However, since id is strictly greater than
                     // 0 inside the loop, get_original can never reach it.
-                    std::unreachable();
+                    MONAD_ABORT("malformed trie: node not found");
                 },
             });
     }


category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -212,10 +212,15 @@ NodeViewBase OffsetTrie::find_original(NodeId id, NibblesView key) const
...
-                    // 0 inside the loop, get_original can never reach it.
-                    std::unreachable();
+                    // Unreachable as written: the only EMPTY-tagged byte is
+                    // the magic header at NULL_ID, id is non-null inside the
+                    // loop, and the constructor checks every child offset is a
+                    // recorded node start. That last premise is the fragile
+                    // one — it lives in a different function, it is the most
+                    // expensive thing the constructor does, and an optimisation
+                    // that drops it turns this arm from diagnosed into
+                    // undefined. The abort is one cold switch arm; keep it.
```
goodlyrottenapple: lets drop this comment. too verbose

anne-laure-s: I changed it to keep the original comment.

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -383,18 +390,37 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
...
     // TODO: Reserve txn vector size for better perf
     while (!ls.empty()) {
         if (ls[0] >= 0xc0) {
+            auto const before = ls;
             BOOST_OUTCOME_TRY(auto tx, decode_transaction_legacy(ls));
+            if (raw != nullptr) {
+                // The list including its header -- what the trie holds for a
+                // legacy transaction.
+                raw->push_back(before.substr(0, before.size() - ls.size()));
+            }
```
dhil: Since this touches the core execution client, I am a little skeptical of the potential performance penalty of the newly added branch here (same for the one in the `else`-branch). I am wondering whether this function should carry a template parameter to statically control whether this innermost branch is taken. Thoughts?

zkvm/guest/body_roots.hpp
```diff
@@ -26,6 +27,9 @@ MONAD_NAMESPACE_BEGIN
 //! the shape of the transactions, receipts and withdrawals tries, and it is
 //! what binds the block BODY the guest executed to the header whose hash is a
 //! public value. The empty trie returns NULL_ROOT.
+bytes32_t ordered_trie_root(std::span<byte_string_view const> items);
+
+//! Convenience overload for callers holding owned buffers.
```
dhil: Is this overload ever called?

category/execution/ethereum/core/rlp/transaction_rlp.cpp
anne-laure-s: I changed it for a template, let me know if that's good for you

zkvm/guest/body_roots.hpp
anne-laure-s: It's called [here in this diff](https://github.com/category-labs/monad/pull/2541/changes#diff-230c3fb26f9dd2ba59f9e8bd840d9e308299326e66c8675499091f125c7b96e7R171) and also outside of the diff: execute_block_zkvm.cpp:190 and :216

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -382,26 +382,58 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
...
-        if (ls[0] >= 0xc0) {
-            BOOST_OUTCOME_TRY(auto tx, decode_transaction_legacy(ls));
-            transactions.emplace_back(std::move(tx));
-        }
-        else {
-            BOOST_OUTCOME_TRY(auto str, parse_string_metadata(ls));
-            BOOST_OUTCOME_TRY(auto tx, decode_transaction_eip2718(str));
-            transactions.emplace_back(std::move(tx));
+    // `remember_raw` is a template parameter and not a runtime flag, so the
+    // node pays nothing at all for a capture only the guest asks for.
```
dhil: I think we can reword this to be more direct, e.g.

```suggestion
    // `remember_raw` controls whether the transaction slices are written to `raw`.
```
This probably also suggests that `raw` ought to be named something like `tx_byte_slices`.

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -382,26 +382,58 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
...
+    template <bool remember_raw>
+    Result<std::vector<Transaction>> decode_transaction_list_impl(
+        byte_string_view &enc, std::vector<byte_string_view> *const raw)
```
dhil: I suspect we'll need to slap a maybe unused attribute on this one now, because when `remember_raw = false` then `raw` is unused.
```suggestion
        byte_string_view &enc, [[maybe_unused]] std::vector<byte_string_view> *const raw)
```

anne-laure-s: Not needed, this builds

category/execution/ethereum/core/rlp/transaction_rlp.cpp
anne-laure-s: Rewrote the comment & changed the name into `raw_transactions`

category/execution/ethereum/core/rlp/transaction_rlp.cpp
dhil: I suppose that implies the variable reference check runs before template instantiation.

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -382,26 +382,62 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
...
-        else {
-            BOOST_OUTCOME_TRY(auto str, parse_string_metadata(ls));
-            BOOST_OUTCOME_TRY(auto tx, decode_transaction_eip2718(str));
-            transactions.emplace_back(std::move(tx));
+    // `remember_raw_transactions` controls whether the transaction slices are
+    // written to `raw_transactions`.
+    template <bool remember_raw_transactions>
+    Result<std::vector<Transaction>> decode_transaction_list_impl(
+        byte_string_view &enc,
+        std::vector<byte_string_view> *const raw_transactions)
```
goodlyrottenapple, line 392: Perhaps a bit cleaner with

```suggestion
    template <class Out>
        requires std::same_as<Out, std::nullptr_t> ||
                 std::same_as<Out, std::vector<byte_string_view> *>
    Result<std::vector<Transaction>> decode_transaction_list_impl(
        byte_string_view &enc,
        Out const raw_transactions)
```


then do

`if constexpr (std::same_as<Out, std::vector<byte_string_view> *>) {...}` in the body

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -382,26 +382,62 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
...
+}
+
+Result<std::vector<Transaction>> decode_transaction_list(
+    byte_string_view &enc,
+    std::vector<byte_string_view> *const raw_transactions)
+{
+    return raw_transactions != nullptr
+               ? decode_transaction_list_impl<true>(enc, raw_transactions)
+               : decode_transaction_list(enc);
 }
```
goodlyrottenapple, line 441:
```suggestion
Result<std::vector<Transaction>> decode_transaction_list(
    byte_string_view &enc,
    std::vector<byte_string_view> &raw_transactions)
{
    return decode_transaction_list_impl<true>(enc, &raw_transactions);
}
```

category/async/config.hpp
dhil, line 205: Why is the below assertion being removed? (the GitHub UI wouldn't let me highlight it). Lines 206 - 213.

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -341,7 +341,7 @@ Result<Transaction> decode_transaction_eip2718(byte_string_view &enc)
         BOOST_OUTCOME_TRY(
             txn.max_fee_per_blob_gas, decode_unsigned<uint256_t>(payload));
         BOOST_OUTCOME_TRY(auto hashes_payload, parse_list_metadata(payload));
-        while (hashes_payload.size() >= sizeof(bytes32_t)) {
+        while (!hashes_payload.empty()) {
```
dhil, line 344: This is a slight change of semantics. I don't know whether it matters. But to me it reads like before we'd allow trailing bytes in the payload, whereas now we don't. Is there any existing use-case for allowing the trailing bytes?

category/mpt/db_metadata_context.hpp
dhil: I see the assertion has been moved here. I presume because we are including the async config somewhere in zkVM.

category/vm/interpreter/intercode.cpp
dhil: Drive by change? I think we should make sure to keep orthogonal changes like these out of this PR.

category/async/config.hpp
dhil: Further down, I saw that it has been moved to another file.

category/vm/interpreter/intercode.cpp
goodlyrottenapple: you are seeing these because commits from #2288 have leaked in due to recent refactors of that branch.

category/async/config.hpp
goodlyrottenapple: another commit from #2288 which probably should have a separate PR anyway

category/vm/interpreter/intercode.cpp
dhil: I am not following why I see these changes when this PR is targetted towards that PR. Is it the case that these changes will disappear from the diff if this PR is rebased?

goodlyrottenapple: yes these will disappear. I suspect the issue is i changed the commits in https://github.com/category-labs/monad/pull/2288 and github can't rebase the current PR cleanly so it's still pointing onto the old commits from https://github.com/category-labs/monad/pull/2288

dhil: OK, I will hold off looking further at this PR until it has been rebased then. CC @anne-laure-s.

goodlyrottenapple: I've force-pushed and this branch is in sync with https://github.com/category-labs/monad/pull/2288 now

category/execution/ethereum/db/offset_trie.cpp
```diff
@@ -215,7 +215,7 @@ NodeViewBase OffsetTrie::find_original(NodeId id, NibblesView key) const
                     // There is only a single NullView node, namely the NULL_ID
                     // magic header. However, since id is strictly greater than
                     // 0 inside the loop, get_original can never reach it.
-                    std::unreachable();
```
goodlyrottenapple: This can go into a separate PR i think; unrelated to the other soundness commits

dhil: Is this change even strictly needed for the zkVM?

goodlyrottenapple: no, dont believe so. just claude being over-zealous.

goodlyrottenapple:
Closed in favour of #2554

PR #2542, opened 2026-09-07
[cmd/vm/mce] Drop evmone dependency in favour of the monad test host

Baltoli:
In #2251 we switched the VM fuzzer over to using our own state (rather than evmone's version). This PR applies the same moral change to the `mce` instrumentation state; in `mce` we don't really need any actual state management, just a way to respond to host callbacks without an error: the programs being run are just microbenchmarks.


dhil (approved):

andreaslyn (approved):

khordadi (approved):

Baltoli:
Rust CI job is unrelated; upstream 500 in Docker and no Rust changes

PR #2544, opened 2026-09-07
[test/vm] Use the OCaml spec VM as the unit-test differential oracle

Baltoli:
This PR switches a handful of regression tests that previously compared against evmone to instead use the OCaml spec. There are some changes to what revisions get tested via `execute_and_compare` as a result (previously: evmone didn't test Monad revisions, now: OCaml spec only tests `MONAD_EIGHT`), but we don't meaningfully change coverage as a result. There is also an LSAN exclude added to avoid trying to sanitise the OCaml runtime.

### DIFF
--- a/scripts/vm/test.sh
+++ b/scripts/vm/test.sh
@@ -3,6 +3,7 @@
 set -euxo pipefail
 
 export UBSAN_OPTIONS="halt_on_error=1"
+export LSAN_OPTIONS="suppressions=$(dirname "$0")/lsan.supp"
 ulimit -s 131072
 
 ./build/test/vm/unit/vm-unit-tests
--- a/test/vm/unit/evm_fixture.hpp
+++ b/test/vm/unit/evm_fixture.hpp
@@ -18,21 +18,20 @@
 #include <category/core/address.hpp>
 #include <category/core/bytes.hpp>
 #include <category/core/int.hpp>
+#include <category/vm/evm/monad/revision.h>
 #include <category/vm/evm/revision.h>
 #include <category/vm/evm/switch_traits.hpp>
+#include <category/vm/evm/traits.hpp>
 #include <category/vm/runtime/allocator.hpp>
 #include <category/vm/runtime/types.hpp>
 #include <category/vm/vm.hpp>
 #include <monad/test/traits_test.hpp>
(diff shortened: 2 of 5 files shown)


scripts/vm/test.sh
```diff
@@ -3,6 +3,7 @@
 set -euxo pipefail
 
 export UBSAN_OPTIONS="halt_on_error=1"
+export LSAN_OPTIONS="suppressions=$(dirname "$0")/lsan.supp"
```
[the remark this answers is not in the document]
Baltoli: Only runs in CI in this specific configuration; keeping the change simple for now as I don't think anyone uses this locally.

dhil (dismissed):

andreaslyn (dismissed):

m-alvarez (dismissed):

test/vm/unit/evm_fixture.hpp
```diff
@@ -146,16 +150,16 @@ namespace monad::vm::compiler::test
                         rt_ctx, icode);
             }
             else {
-                MONAD_ASSERT(impl == Evmone);
-                evmc::VM const evmone_vm{evmc_create_evmone()};
+                MONAD_ASSERT(impl == Spec);
+                evmc::VM spec_vm{evmc_create_monadml_evm()};
```
goodlyrottenapple, line 154:
```suggestion
                evmc::VM const spec_vm{evmc_create_monadml_evm()};
```

Baltoli: _Claude-generated:_ Applied, squashed in.

Baltoli: _Claude-generated:_ Correction: reverted. `evmc::VM::execute` is not const-qualified (evmc.hpp:716), so a `const` VM can't call it. The old code got away with `const` because it only called `get_raw_pointer()`, which is.

goodlyrottenapple (approved):

test/vm/unit/evm_fixture.hpp
goodlyrottenapple: whoops sorry for the noise

m-alvarez (approved):

PR #2547, opened 2026-09-09
Unify recover_address and ecrecover precompile calls to `monad_recover_address`

goodlyrottenapple:


goodlyrottenapple:
@claude re-review please

category/execution/ethereum/precompiles.cpp
```diff
@@ -227,13 +228,18 @@ PrecompileResult ecrecover_execute(byte_string_view const input)
     uint8_t *out{static_cast<uint8_t *>(std::aligned_alloc(8, 32))};
     MONAD_ASSERT(out != nullptr);
 
-    return from_impl_result(
-        ecrecover_impl(
+    std::memset(out, 0, 12);
+    if (!recover_address(
```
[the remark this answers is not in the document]
goodlyrottenapple: moved the corresponding commit from #2543 to this PR

goodlyrottenapple:
@claude re-review please

category/execution/ethereum/core/ecrecover/impl.hpp
Baltoli, line 1: Seems a bit odd for this to remain as `impl.hpp` when it no longer actually contains the implementation - might be worth renaming?

zkvm/category/execution/ethereum/core/ecrecover/impl.hpp
Baltoli, line 1: I guess it is legitimately an implementation here in the shadow, though - which is presumably why the non-ZK side has to call it `impl.hpp` still?

category/execution/ethereum/core/ecrecover/impl.hpp
goodlyrottenapple: open to suggestions, i couldn't really think of what would be best

zkvm/category/execution/ethereum/core/ecrecover/impl.hpp
```diff
@@ -0,0 +1,55 @@
...
+MONAD_NAMESPACE_BEGIN
+
+[[gnu::always_inline]] inline bool recover_address(
+    std::span<uint8_t, 20> const out, std::span<uint8_t const, 32> const msg,
+    std::span<uint8_t const, 64> const sig, uint8_t const recid)
+{
+    auto const *msg_hash =
+        reinterpret_cast<zkvm_secp256k1_hash const *>(msg.data());
+
+    auto const *signature =
```
Baltoli, line 35:
```cpp
    // TODO(dhil): Check `sig` is well-formed; the patch
    // before had subscript 64 on a `uint8_t[64]`, which I think was copy-pasted
    // from the previous implementation, but it used an array of length 128.
```
was in the prestate - resolved?

category/execution/ethereum/core/ecrecover/impl.hpp
Baltoli: I think it's probably fine, on a second pass through the PR

category/execution/CMakeLists.txt
```diff
@@ -54,8 +54,9 @@ add_library(
   "ethereum/core/contract/events.hpp"
   "ethereum/core/contract/storage_array.hpp"
   "ethereum/core/contract/storage_variable.hpp"
-  "ethereum/core/ecrecover.cpp"
   "ethereum/core/ecrecover.hpp"
+  "ethereum/core/ecrecover/impl.cpp"
+  "ethereum/core/ecrecover/impl.hpp"
```
mkolosick, line 59: duplicate

mkolosick (changes requested):

zkvm/category/execution/ethereum/core/ecrecover/impl.hpp
goodlyrottenapple: @dhil do you remember the context? I couldn't quite see what this meant, but given we've vendored the ecrecover and re-shuffled the precompiles quite a bit recently, just assumed it stale at this point.

category/execution/CMakeLists.txt
goodlyrottenapple: not sure i see what you mean?

mkolosick: I'm sorry, my brain just kept transposing the `.hpp` from the line above onto the first one.

mkolosick (dismissed):

zkvm/category/execution/ethereum/core/ecrecover/impl.hpp
goodlyrottenapple: yeah the intention was to not have to shadow `ecrecover.hpp` just the call to `monad_recover_address` but since `monad_recover_address` needs the thread local context on host, I've pushed it into one more wrapper under `ecrecover/impl.*`

category/execution/ethereum/core/signature.hpp
```diff
@@ -48,14 +49,27 @@ struct Secp256k1Signature
...
+    {
+        return Secp256k1Signature::has_upper_s(s);
+    }
+
+    [[gnu::always_inline]] constexpr static inline bool
+    is_valid(uint256_t const &r, uint256_t const &s)
+    {
+        return Secp256k1Signature::has_valid_range(r, s) &&
+               !Secp256k1Signature::has_upper_s(s);
+    }
```
Baltoli: This has only one caller (the member function below) - it shares a name with that member function but does a subtly different check, so I'd probably just inline this static method into the one below

category/execution/ethereum/precompiles.cpp
```diff
@@ -216,7 +217,7 @@ PrecompileResult ecrecover_execute(byte_string_view const input)
     auto const r{load_be_unsafe<uint256_t>(&d[64])};
     auto const s{load_be_unsafe<uint256_t>(&d[96])};
 
-    if (!Secp256k1Signature{r, s}.has_valid_range()) {
+    if (!Secp256k1Signature::has_valid_range(r, s)) {
```
Baltoli: What motivates the static method vs. the method call in the pre state?

goodlyrottenapple: just to avoid copying the two uint256 values, mainly thinking of the zkvm here, tho claude tells me this would be optimised away anyway to a no-copy code....

zkvm/category/execution/ethereum/core/ecrecover/impl.hpp
dhil: I think it is obsolete now. This came up during the precompile refactoring. It looked to me like the stub had been copied from x86_64 client, but changed the length of the array, and thus committed an OOB. The initial unification of the zkVM and x86_64 precompile interfaces resolved this issued as it removed all these fixed width arrays from the signatures such that the caller was responsible for passing in the pointer + length instead. I left the TODO there to follow up on whether the zkVM was doing the right thing now (it clearly wasn't before).

category/execution/ethereum/precompiles.cpp
Baltoli: My gut instinct is to keep the code simple and that the difference here will be tiny in practice - I'd _guess_ that a compiler would be able to optimise the object-constructed version away, but haven't verified.

goodlyrottenapple: ok, will revert and can revisit if ever an issue

Baltoli (approved):

zkvm/category/execution/ethereum/precompiles_impl.hpp
```diff
@@ -156,35 +156,6 @@ inline bool init_trusted_setup()
     return true;
 }
 
-[[gnu::always_inline]] inline PrecompileImplResult ecrecover_impl(
-    std::span<uint8_t const, 32> const msg,
-    std::span<uint8_t const, 64> const sig, uint8_t recid,
-    std::span<uint8_t, 32> const out)
-{
-    auto const *msg_hash = reinterpret_cast<zkvm_bytes_32 const *>(msg.data());
-    // TODO(dhil): Check `sig` is well-formed; the patch
```
dhil, line 165: We should remember to verify that this precompile is doing the right thing.

dhil (approved):

PR #2554, opened 2026-09-11
zkvm: execute block

goodlyrottenapple:
First substantial piece of the zkvm guest. Includes:
- new execute block which removes all concurrency and optimistic execution, executing all txns in a block sequentially
- witness block hash buffer which aborts if execution attempts to access a block hash that wasn't included in the witness
- `ordered_trie` MPT optimised for calculating the root of the receipts, etc tries which are of the form `[[rlp(0), receipts[0]], ..., [rlp(n), receipts[n]]]`
- x86 test runner harness for testing/debugging the witness format

The followup PR adds the ZisK/SP1 framework deps for building actual guest ELFs from the above code.


zkvm/guest/execute_witness.cpp
anne-laure-s, line 1: I feel like a more descriptive name could be better, like execute_witness.cpp for instance

zkvm/guest/ffi.cpp
```diff
@@ -0,0 +1,154 @@
...
+    monad::mpt::OffsetTrie trie{witness.value().encoded_nodes};
+    // A witness must carry a materialised pre-state trie; an overlay-id root
+    // signifies an empty trie.
+    MONAD_ASSERT(!monad::mpt::is_overlay_id(trie.root));
+    monad::PartialTrieDb pdb{std::move(trie), std::move(code_index)};
+
+    monad::byte_string_view block_view = witness.value().block_rlp;
+    // The byte slice each transaction was decoded from, kept so the
+    // transactions-root check can be made against those bytes rather than
+    // against a re-encoding of what was decoded from them.
```
anne-laure-s:
```c++
// Keep each transaction's original bytes to verify transactions_root
// without re-encoding the decoded transaction.
```

anne-laure-s:
```c++
// Keep each transaction's original bytes to verify transactions_root
// without re-encoding the decoded transaction.
```

zkvm/guest/execute_block.cpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+    }
+
+    // YP eq. 22 — cumulative gas fixup.
+    uint64_t cumulative_gas_used = 0;
+    for (auto &r : receipts) {
+        cumulative_gas_used += r.gas_used;
+        r.gas_used = cumulative_gas_used;
+    }
+    if (MONAD_UNLIKELY(cumulative_gas_used != block.header.gas_used)) {
```
anne-laure-s, line 193: This gas check, along with the `receipts_root`, `logs_bloom`, and `requests_hash` checks, appears redundant if the verifier already authenticates the block hash as **canonical** and the guest only proves the corresponding state transition. This assumes the executed body and initial state are correctly bound to that certified hash.

If we instead want the guest to establish execution-layer block validity **independently**, these checks belong, but additional validation is needed. See [this branch](https://github.com/category-labs/monad/commits/al/zkvm-validate-with-parent/); feel free to cherry-pick it or included it later if wanted (additional cost is negligible).

goodlyrottenapple: I assume we indeed want the latter, i.e. tie the gas used/receipts/bloom/etc. to the header?

PR #2555, opened 2026-09-11
rlp: block and txn decoding hardening + decode block raw_transaction return

goodlyrottenapple:
This PR includes some RLP parser hardening which is not strictly necessary in the usual x86 host case because consensus already does validation but could be a potential corner-case issue in the zkvm witness path, where a malicious block with added garbage could cause divergence in our client compared to a different implementation which would reject the witness outright.


category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -341,7 +341,7 @@ Result<Transaction> decode_transaction_eip2718(byte_string_view &enc)
         BOOST_OUTCOME_TRY(
             txn.max_fee_per_blob_gas, decode_unsigned<uint256_t>(payload));
         BOOST_OUTCOME_TRY(auto hashes_payload, parse_list_metadata(payload));
-        while (hashes_payload.size() >= sizeof(bytes32_t)) {
+        while (!hashes_payload.empty()) {
```
dhil, line 344: As I commented on the original PR, we need to figure out whether this slight change is semantics is acceptable. Before we allow potentially trailing bytes in the hash payload, now we mandate that there are none.

dhil (commented):
Thanks for hoisting this change out of the larger PR. It is nicely self-contained now. The only thing I see we need to figure out is whether the hash payload decoding change is OK.

category/execution/ethereum/core/rlp/block_rlp.cpp
```diff
@@ -209,13 +209,19 @@ decode_block_header_vector(byte_string_view &enc)
     return ommers;
 }
 
-Result<Block> decode_block(byte_string_view &enc)
+// `Out` is std::nullptr_t or std::vector<byte_string_view> *, matching
+// decode_transaction_list, so the raw-transaction capture is compiled out of
+// the plain decode_block overload rather than guarded at runtime.
```
goodlyrottenapple:
```suggestion
```

goodlyrottenapple:
@claude re-review please

category/execution/ethereum/core/rlp/transaction_rlp.cpp
Baltoli: This is fine I think - there are no blob hashes for Monad in production, so this isn't actually live code. It would only affect replay or ZK, where we have pre-validated blocks anyway (i.e. blocks that don't have the bad trailing-bytes form).

category/execution/ethereum/core/rlp/block_rlp.cpp
```diff
@@ -230,4 +233,15 @@ Result<Block> decode_block(byte_string_view &enc)
...
+Result<Block> decode_block(byte_string_view &enc)
+{
+    return decode_block_impl(enc, nullptr);
+}
+
+Result<Block> decode_block(
+    byte_string_view &enc, std::vector<byte_string_view> &raw_transactions)
+{
+    return decode_block_impl(enc, &raw_transactions);
+}
```
Baltoli, line 249: I might add a comment here or on the declaration explaining _why_ we need this overload - it took me a moment to figure out what was going on (I gather this PR is the bottom of a stack and will have other changes layered on top that actually use the zero-copy slice recovery?)

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -382,19 +382,33 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
...
 {
     std::vector<Transaction> transactions;
     BOOST_OUTCOME_TRY(auto ls, parse_list_metadata(enc));
 
     // TODO: Reserve txn vector size for better perf
     while (!ls.empty()) {
         if (ls[0] >= 0xc0) {
+            auto const before = ls;
             BOOST_OUTCOME_TRY(auto tx, decode_transaction_legacy(ls));
+            if constexpr (std::same_as<Out, std::vector<byte_string_view> *>) {
```
Baltoli: Maybe pull the same_as result into a variable at the top of the function? `recover_raw_transactions`?

category/execution/ethereum/core/rlp/transaction_rlp.cpp
```diff
@@ -382,19 +382,33 @@ Result<Transaction> decode_transaction(byte_string_view &enc)
     }
 }
 
-Result<std::vector<Transaction>> decode_transaction_list(byte_string_view &enc)
+template <class Out>
```
Baltoli:
```suggestion
template <typename Out>
```
Nitpicking

category/execution/ethereum/core/rlp/transaction_rlp.hpp
```diff
@@ -40,6 +42,20 @@ Result<AuthorizationList> decode_authorization_list(byte_string_view &);
 Result<Transaction> decode_transaction_legacy(byte_string_view &);
 Result<Transaction> decode_transaction_eip2718(byte_string_view &);
 Result<Transaction> decode_transaction(byte_string_view &);
-Result<std::vector<Transaction>> decode_transaction_list(byte_string_view &enc);
+
+// If raw_transactions is non-null, appends a view of transaction. Each view
+// aliases enc's underlying buffer, so it is valid only while that buffer lives,
+// and is independently decodable with decode_transaction.
```
Baltoli, line 49: Oops, just saw this comment - disregard my request elsewhere to add one

category/execution/ethereum/core/rlp/transaction_rlp.hpp
```diff
@@ -40,6 +42,20 @@ Result<AuthorizationList> decode_authorization_list(byte_string_view &);
...
+template <class Out>
```
Baltoli:
```suggestion
template <typename Out>
```

Baltoli (commented):
General metaprogramming refactor that I thought up:

```cpp
  enum class RawTransactions : bool
  {
      Discard,
      Collect
  };

  template <RawTransactions collect>
  Result<std::vector<Transaction>> decode_transaction_list(
      byte_string_view &enc, std::vector<byte_string_view> *raw_transactions);

  // Body

  if constexpr (collect == RawTransactions::Collect) {
      MONAD_ASSERT(raw_transactions != nullptr);
  }
  ...
  if constexpr (collect == RawTransactions::Collect) {
      raw_transactions->push_back(...);
  }

  // Public overloads:

  inline Result<std::vector<Transaction>>
  decode_transaction_list(byte_string_view &enc)
  {
      return decode_transaction_list<RawTransactions::Discard>(enc, nullptr);
  }

  inline Result<std::vector<Transaction>> decode_transaction_list(
      byte_string_view &enc, std::vector<byte_string_view> &raw_transactions)
  {
      return decode_transaction_list<RawTransactions::Collect>(
          enc, &raw_transactions);
  }
```

This avoids the `std::is_same` stuff and possibly reads a bit cleaner? Up to you though

Baltoli (dismissed):

category/execution/ethereum/core/rlp/block_rlp.cpp
```diff
@@ -209,13 +210,19 @@ decode_block_header_vector(byte_string_view &enc)
     return ommers;
 }
 
-Result<Block> decode_block(byte_string_view &enc)
+template <typename Out>
+    requires std::same_as<Out, std::nullptr_t> ||
+             std::same_as<
+                 Out, std::reference_wrapper<std::vector<byte_string_view>>>
```
dhil, line 216: I'd hoist these constraints to a concept, since they are used multiple times, e.g.
```c++
template <typename Out>
concept OptionalOutput = // Maybe a better name, but you get the idea.
    std::same_as<Out, std::nullptr_t> ||
    std::same_as<
        Out,
        std::reference_wrapper<std::vector<byte_string_view>>
    >;

goodlyrottenapple: perhaps `NullptrTyOrRef`?

dhil: Or simply NullableRef? Either way I won't get hung up on the exact name. I just think it is worth abstracting as it is used multiple times.

dhil (approved):

Baltoli (approved):

PR #2556, opened 2026-09-11
[rpc] Bespoke simulation error type

dhil:
This patch adds a small simulation error type, meant to be returned by `eth_simulateV1` and friends rather than throwing. This patch is only the starting point -- no attempt has been made to replace the candidate throw expressions yet, save for the gas limit exceeded one.

The point of this change is to work towards a modular and reusable set of primitive for simulations.


dhil:
@kjcamann this is a small prototype illustrating what the `SimulationError`-type, that we discussed previously, might look like.

category/rpc/simulation_error.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace
+{
+    evmc_status_code simulation_status(SimulationError const error)
+    {
+        switch (error) {
+        case SimulationError::Success:
+            return EVMC_SUCCESS;
+        case SimulationError::GasLimitExceeded:
+            return EVMC_INTERNAL_ERROR;
```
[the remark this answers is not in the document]
dhil: It is deliberate for now to maintain status quo.

category/rpc/simulation_error.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <boost/outcome/config.hpp>
+// TODO unstable paths between versions
+#if __has_include(<boost/outcome/experimental/status-code/status-code/config.hpp>)
+    #include <boost/outcome/experimental/status-code/status-code/config.hpp>
+    #include <boost/outcome/experimental/status-code/status-code/quick_status_code_from_enum.hpp>
+#else
+    #include <boost/outcome/experimental/status-code/config.hpp>
+    #include <boost/outcome/experimental/status-code/quick_status_code_from_enum.hpp>
+#endif
+#include <boost/outcome/success_failure.hpp>
```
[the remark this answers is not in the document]
dhil: I've replicated what `reserve_balance_error.cpp` does now.

category/rpc/simulation_error_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+TEST(SimulationErrorTest, BlockDomain)
+{
+    Result<void>::error_type const error = BlockError::GasAboveLimit;
+
+    auto const info = simulation_error_info(error);
+
+    EXPECT_EQ(info.status_code, EVMC_REJECTED);
+    EXPECT_EQ(info.message, "gas above limit");
+}
```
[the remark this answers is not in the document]
dhil: Yes, I will consider this in some foreseeable future.

dhil:
@claude review this PR please, taking into account the previous comments.

category/rpc/monad_executor.cpp
```diff
@@ -1041,16 +1042,20 @@ namespace
...
-                !(gas_used > 0 &&
-                  gas_consumed_so_far >
-                      std::numeric_limits<uint64_t>::max() - gas_used),
-                "gas limit exceeded");
+            if (MONAD_UNLIKELY(
+                    gas_used > 0 &&
+                    gas_consumed_so_far >
+                        std::numeric_limits<uint64_t>::max() - gas_used)) {
+                return Result<nlohmann::json>::error_type{
+                    SimulationError::GasLimitExceeded};
```
Baltoli, line 1050: Kind of wordy, especially if we're going to do more call sites in a future change. Maybe wrap this up into a helper? (it's a shame there's one too many levels of conversion and you can't return `GasLimitExceeded` directly)

dhil: There are at least three different notions of gas limit exceeded at play here:
1. Block gas limit exceeded
2. Transaction intrinsic gas exceeds the limit
3. Simulation gas limit exceeded.

The follow-up patch converts remaining throws into returning error values as well https://github.com/category-labs/monad/pull/2557. We can abstract the construction, but I don't think it will buy much, we still have to write the explicit `return`.

Baltoli: Yeah not a strong opinion, feel free to disregard

category/rpc/simulation_error.cpp
```diff
@@ -0,0 +1,120 @@
...
+    evmc_status_code simulation_status(SimulationError const error)
+    {
+        switch (error) {
+        case SimulationError::Success:
+            return EVMC_SUCCESS;
+        case SimulationError::GasLimitExceeded:
+            return EVMC_INTERNAL_ERROR;
+        }
+        return EVMC_INTERNAL_ERROR;
+    }
```
Baltoli, line 48: Is this going to grow in the future? Why not just:
```cpp
if (error == SimulationError::Success) {
  return EVMC_SUCCESS;
}

return EVMC_INTERNAL_ERROR;
```

category/rpc/simulation_error.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    evmc_status_code block_status(BlockError const error)
```
Baltoli, line 50: Is the above just for consistency with this one?

category/rpc/simulation_error.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+    else if (error.domain() == block_error.domain()) {
+        status_code = block_status(static_cast<BlockError>(error.value()));
+    }
+
+    return SimulationErrorInfo{
+        .status_code = status_code,
+        .message = std::string{error.message().c_str()},
+    };
+}
```
Baltoli, line 101:
```suggestion
SimulationErrorInfo simulation_error_info(Result<void>::error_type const &error)
{
    evmc_status_code status_code = EVMC_REJECTED;
    if (error.domain() == quick_status_code_from_enum_domain<SimulationError>) {
        status_code =
            simulation_status(static_cast<SimulationError>(error.value()));
    }
    else if (error.domain() == quick_status_code_from_enum_domain<BlockError>) {
        status_code = block_status(static_cast<BlockError>(error.value()));
    }

    return SimulationErrorInfo{
        .status_code = status_code,
        .message = std::string{error.message().c_str()},
    };
}
```

Baltoli (approved):
One suggestion; otherwise looks sensible

category/rpc/simulation_error.cpp
dhil: Yes! We can do something like that, though, I think we might want to map groups of errors to `EVMC_INTERNAL_ERROR` and `EVMC_REJECTED` respectively. I need to survey what Geth and other clients are doing regarding status codes. The legacy behaviour of our client is to just return `EVMC_INTERNAL_ERROR` because that's how exceptions were mapped.

category/rpc/simulation_error.cpp
dhil: It was to do an exhaustive match such that the semantics of future additions would have to be considered too.

category/rpc/simulation_error.cpp
dhil: Yes good point!

category/rpc/simulation_error.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+MONAD_NAMESPACE_BEGIN
+
+enum class SimulationError
+{
+    Success = 0,
+    GasLimitExceeded,
+};
+
+struct SimulationErrorInfo
```
kjcamann, line 43: It was a little confusing to me at first why this type existed, and it seemed superfluous. But I now think I understand that:

- A `SimulationError` is a new error domain for the things that can go wrong only in a simulation
- A `SimulationErrorInfo` is a description of _any_ error that occur during a simulation

The first is "pertains only to simulations" and the latter is "happens during simulations" and massages the available data into a nicer form.

I am not sure if it's possible to name this better (I couldn't think of a better way at least) but maybe comment it?

This is the only feedback I had.

kjcamann (dismissed):

category/rpc/simulation_error.hpp
dhil: I've renamed the info-thingie to `SimulationErrorResponse`. I think it reflects the idea better: its the data used to construct the rpc response in the event of an error.

category/rpc/simulation_error.cpp
dhil: Implemented now.

PR #2561, opened 2026-09-15
[test/vm] Replace evmc::MockedHost with an in-tree mock host

khordadi:
Closes EXE-161, part of EXE-162. The VM unit tests no longer depend on `evmc::MockedHost`, so the `vm::Host` port (EXE-173) has one test host to move instead of two, and `evmc::mocked_host` leaves the test link line.

**What changes.** `test/vm/utils/mocked_host.hpp` replaces the `test/vm/unit/runtime` wrapper, which layered MIP-8 page tracking over evmc's mock. It is derived from evmc's `mocked_host.hpp` (Apache-2.0; notice in the file header). It implements `vm::Host` directly, including `update_page`, carries the page tracking over on the production `compute_page_key`, and keys its state on `Address`/`bytes32_t`; evmc types remain only where the `vm::Host` virtuals require them. It sits in `monad-vm-test-utils` so the fuzzer and benchmark targets can pick it up later without a new target.

**Storage status.** The EIP-2200 status is the one piece of logic rewritten rather than moved. `mocked_host_tests.cpp` checks the fifteen `original -> current -> new` transitions listed on `evmc_storage_status`; they partition the three values by zero-ness and pairwise equality, so the table is exhaustive. The runtime storage tests reach only the six clean-slot cases, and the spec comparison drives both VMs through the same mock, so neither would notice a wrong dirty-slot status.

**Behaviour.** No test-observable change. `evm_fixture.hpp` and `monad_vm_interface_tests.cpp` used `evmc::MockedHost` directly and now get the wrapper's semantics: `access_storage` warmth per page rather than per slot, and `update_page` from the page set rather than from the storage status. The only SSTORE either reaches runs under `execute_and_compare`, pinned to MONAD_EIGHT where MIP-8 is inactive.

**Dropped, nothing uses them:**
- the 100/200 caps on recorded calls and account accesses
- the copy of call input bytes behind `recorded_calls` (its pointer fields are nulled on record)
- `MockedAccount::nonce`
- the per-slot access flag the wrapper never returned
- `access_storage`'s side effect of creating the account and slot (`account_exists` after SLOAD on an absent account is now false)
- the second `recorded_account_accesses` entry per SSTORE from the wrapper's extra `get_storage`
- evmc's always-warm rule for `0x01..0x09` in `access_account` (production decides that per fork in `EvmcHost`; tests that need a warm address pre-warm it)

**Verification.**

| `vm-unit-tests` | main | this PR |
|---|---|---|
| gcc-15 RelWithDebInfo | 1291 passed / 34 skipped | 1292 / 34 |
| clang-19 ASAN+UBSAN, spec comparison on | 1785 / 196 | 1786 / 196 |

The difference is the new test. `monad-compiler-fuzzer`, `execution-benchmarks`, `vm-micro-benchmarks` and `mce` build. `clang-format-19` and `clang-tidy-19` clean on the new files.

**Review map:** `test/vm/utils/mocked_host.hpp` (`storage_status`, then the virtuals) → `test/vm/unit/mocked_host_tests.cpp` → the five include/type swaps → the deleted wrapper.


test/vm/utils/mocked_host.hpp
```diff
@@ -0,0 +1,331 @@
...
+            auto const a = Address{addr};
+            auto const already_accessed =
+                std::ranges::find(recorded_account_accesses, a) !=
+                recorded_account_accesses.end();
+            record_account_access(a);
+            // EIP-2929: precompiles are always warm.
+            if (a >= 0x0000000000000000000000000000000000000001_address &&
+                a <= 0x0000000000000000000000000000000000000009_address) {
+                return EVMC_ACCESS_WARM;
+            }
```
goodlyrottenapple: I wonder if it might be better to just delete this? AFAIK nothing exercises this behaviour anyway and it's just prone to go stale.

khordadi: Deleted!

test/vm/utils/mocked_host.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+    public:
+        std::unordered_map<Address, MockedAccount> accounts;
+        evmc_tx_context tx_context{};
+        bytes32_t block_hash;
+        evmc_result call_result{};
+        mutable std::vector<int64_t> recorded_blockhashes;
+        mutable std::vector<Address> recorded_account_accesses;
+        // input_data is not copied; only the scalar fields outlive the call.
+        std::vector<evmc_message> recorded_calls;
```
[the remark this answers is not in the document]
khordadi: Done: `call()` now nulls `input_data`/`memory_handle`/`memory` and zeroes their sizes on the recorded copy, so a deref fails fast; tests only read `.flags` (head 3fc0faaf0).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

test/vm/unit/runtime/fixture.cpp
```diff
@@ -172,9 +172,8 @@ namespace monad::vm::compiler::test
         auto const codehash = keccak256({code.data(), code.size()});
         bytes32_t codehash_bytes;
         std::copy(codehash.bytes, codehash.bytes + 32, codehash_bytes.bytes);
-        auto const account = evmc::MockedAccount{
-            .nonce = 0,
-            .code = evmc::bytes(code.data(), code.size()),
+        auto const account = vm::test::MockedAccount{
+            .code = byte_string(code.data(), code.size()),
```
Baltoli, line 176: Explicitly out of scope here for this PR, but it's probably worth cleaning these namespaces up somewhat. I don't think there's a good semantic distinction between `vm::compiler::test` and `vm::test`.

khordadi: Filed as [EXE-212](https://linear.app/category-labs/issue/EXE-212/collapse-vmcompilertest-into-vmtest-in-the-vm-unit-tests) under EXE-162.

goodlyrottenapple (approved):

Baltoli (approved):

PR #2562, opened 2026-09-15
Support exactly one storage device

maxkozlovsky:
A storage pool has always been a list of devices, but nothing has used more than one: the node, the packaged systemd units and every operator deployment name a single /dev/triedb, and each unit's DeviceAllow cgroup would block a second device even if one were configured. Kernel RAID or LVM aggregates the real disks and the pool sees one block device.

Carrying the list cost more than it returned. Chunk ids are pool-wide and capped at the 20 bits of chunk_offset_t, so extra devices buy no addressable space, only dead conventional chunks reserved on each one. Worse, interleave_chunks_evenly decided a creation-time chunk layout that was never recorded and never folded into config_hash, so opening a multi-device pool under a different value of that flag silently remapped every chunk id.

So the device is a single member rather than a vector, and OnDiskDbConfig, the snapshot and statesync C entry points, the Rust shims over both, monad, monad-cli and monad-mpt each take one path in place of an array and its length. A two-device pool is unrepresentable rather than merely unused.

One device also makes a chunk a function of its id, so the pool stops materialising a vector<chunk_t> per chunk type and AsyncIO stops keeping an array of references beside it: chunk() constructs one per i/o, which retires chunk_t's descriptors and virtual destructor, AsyncIO's fd deduplication and the pool's lock. Opening a 700 GiB, 358,400-chunk pool, monad-mpt's peak RSS falls from 72.0 MB to 51.3 MB.

The config hash folds the same quantities in the same order, so pools created before this still open.


maxkozlovsky:
@claude re-review

category/rpc/monad_executor.cpp
```diff
@@ -1258,18 +1259,39 @@ struct monad_executor
...
-            for (auto const &file :
-                 std::filesystem::directory_iterator(triedb_path)) {
-                paths.emplace_back(file.path());
-            }
-        }
-        else {
-            paths.emplace_back(triedb_path);
+        std::filesystem::path path{triedb_path};
+        std::error_code ec;
+        if (std::filesystem::is_directory(path, ec)) {
```
Chen-Yifan: can just assert !is_directory. You would need to change monad-bft repo to be compatible. For example, the `--triedb-path` in `compose.yaml` to `/monad-triedb/test.db` instead of the parent directory, and monad-rpc/README.md

category/async/storage_pool.hpp
```diff
@@ -192,45 +190,35 @@ class storage_pool
         std::pair<file_offset_t, file_offset_t> capacity() const;
     };
 
-    /*! \brief A zone chunk from storage, which is always managed by a shared
-    ptr. When the shared ptr count reaches zero, any file descriptors or other
-    resources associated with the chunk are released.
+    /*! \brief A chunk of storage, derived from its id and cheap to copy. It
+    borrows the pool's device -- including the descriptor every chunk reads
```
Chen-Yifan: nit: I would rewrite this to remove the llm favored dashes

category/mpt/update_aux.cpp
Chen-Yifan:
```suggestion
        auto const chunk =
```

auto const& enforces the same lifetime but would recommend use by value. There are a few other places that still uses reference, can let your claude do a pass.

Chen-Yifan (commented):
Looks nice! Removing the vectors of chunks in storage pool and call chunk() on demand makes it much cleaner. And the static functions make the constructor finally readable.

Chen-Yifan (dismissed):

Chen-Yifan (approved):

Akatsukis (approved):

PR #2563, opened 2026-09-15
Add monad-mpt --rescan-devices

maxkozlovsky:
Growing a node's database meant --archive then --restore into a freshly created larger pool: scratch space, plus a full dump and restore. Extending the device the pool owns was worse than unsupported -- the metadata saying a device belongs to a pool lives at its very end, so lvextend left it unopenable, with no resize2fs step to follow.

--rescan-devices is that step: one offline run takes up the space an extended device now offers, under one confirmation and one metadata growth.

Taking up an extend needs the previous size of the device, which the extend itself strands mid-device, so db_metadata now records that size on each writable open, in padding it already reserves -- no magic bump, no migration. That recording is the only source, and is acted on only where the stranded footer's config_hash confirms it, so an extend must be preceded by a writable open under a release carrying this code.

With one device the relocated footer is the whole commit record: there is no sibling whose config_hash has to be durable first, so a crash at any point either leaves the old footer stranded and the run repeatable, or the new one in place and the run complete.


maxkozlovsky:
@claude re-review

category/mpt/cli_tool_impl.cpp
```diff
@@ -1569,6 +1622,19 @@ the same type, size and device id, otherwise the database cannot be opened.
...
+                "reconcile an existing database with the storage it is now "
+                "given: take up the space its device gained from being "
+                "extended in place. --storage must name the database's own "
+                "device, which keeps its contents. Taking up an extended "
+                "device needs the size the database recorded for it, which "
+                "every writable open writes, so only extend a device of a "
+                "database this release has already opened. Run with the daemon "
+                "stopped. An interrupted run is finished by re-running the "
+                "identical command, which resumes it from wherever it "
+                "stopped.");
```
Chen-Yifan, line 1637:
```suggestion
            cli_ops_group->add_flag(
                "--rescan-device",
                impl.rescan_devices,
                "reconcile an existing database with the storage it is now "
                "given: take up the space an existing database's device gained  from being extended in place. Must only extend a device of a database this release has already opened. An interrupted run is finished by re-running the identical command, which resumes it from wherever it stopped.");
```

`--rescan-device` or maybe `--grow-device` since we are explicit about supporting single device now. And help message can be more concise, can omit the detailed reason why it must open writable before running `--rescan-device` since the assertion message in `validate_device_to_rescan_()` covers it.

category/async/storage_pool.cpp
```diff
@@ -55,6 +57,423 @@ MONAD_ASYNC_NAMESPACE_BEGIN
...
+    return candidate;
+}
+
+auto storage_pool::validate_device_to_rescan_(
+    std::filesystem::path const &source, device_info_ const &info,
+    std::optional<file_offset_t> const recorded_size,
+    std::optional<db_metadata_budget> const &budget)
+    -> std::optional<grown_device_>
+{
+    std::optional<grown_device_> grown;
```
Chen-Yifan, line 271: nit: suggest renaming to `device` or `grown_device`

category/async/storage_pool.cpp
```diff
@@ -55,6 +57,423 @@ MONAD_ASYNC_NAMESPACE_BEGIN
...
+        return std::nullopt;
+    }
+    return candidate;
+}
+
+auto storage_pool::validate_device_to_rescan_(
+    std::filesystem::path const &source, device_info_ const &info,
+    std::optional<file_offset_t> const recorded_size,
+    std::optional<db_metadata_budget> const &budget)
+    -> std::optional<grown_device_>
```
Chen-Yifan, line 269: nit: there are a mixed usage of auto and explicit nested return type in this file (this function uses auto f() -> return_type, `read_device_info_()` and `device_info_at_previous_size_()`), uses the leading style) suggest unifying them.

category/async/storage_pool.cpp
```diff
@@ -55,6 +57,423 @@ MONAD_ASYNC_NAMESPACE_BEGIN
...
+    // The hash the device must have produced before it grew is the one the
+    // stranded footer itself stores. unique_hash folds the device's size, so
+    // the recomputed hash depends on the recorded size -- through the chunk
+    // count it is folded into, since the hash itself folds only its low 32
+    // bits -- so this pins
+    // the previous size rather than merely narrowing it, and four bytes
+    // spelling MND0 in trie data cannot pass it. The footer predates the grow
+    // and this operation never rewrites it, so it stays a fixed point to check
+    // against even across an interrupted run.
+    if (compute_config_hash_(device_info_at_previous_size_(
```
Chen-Yifan, line 258: `compute_config_hash_()` really reads are only `unique_hash`, `chunks` and `chunk_capacity`, but caller has to synthesize a `device_info_` object. Suggests simplifying it to take scalars directly, which can help simplify and remove some helper function like `device_info_at_previous_size_()` and `device_info_of_()`.

category/async/storage_pool.cpp
```diff
@@ -55,6 +57,423 @@ MONAD_ASYNC_NAMESPACE_BEGIN
...
+    rescan_preview ret{};
+    if (auto const grown =
+            validate_device_to_rescan_(source, info, recorded_size, budget)) {
+        ret.grown_previous_size = grown->previous_size;
+        ret.grown_previous_chunks = grown->previous_chunks;
+    }
+    return ret;
+}
+
+uint32_t storage_pool::compute_config_hash_(device_info_ const &device)
```
Chen-Yifan, line 165:
```suggestion
uint32_t storage_pool::compute_config_hash_(
    uint64_t const unique_hash, size_t const chunks,
    uint32_t const chunk_capacity)
```

category/async/storage_pool.cpp
```diff
@@ -55,6 +57,423 @@ MONAD_ASYNC_NAMESPACE_BEGIN
...
+        aligned_alloc(DISK_PAGE_SIZE, DISK_PAGE_SIZE * 2));
+    MONAD_ASSERT(buffer != nullptr);
+    auto const unbuffer = make_scope_exit([&]() noexcept { ::free(buffer); });
+    auto const offset = round_down_align<DISK_PAGE_BITS>(
+        ret.size - sizeof(device_t::metadata_t));
+    auto const bytesread = ::pread(
+        fd,
+        buffer,
+        static_cast<size_t>(ret.size - offset),
+        static_cast<off_t>(offset));
```
Chen-Yifan, line 124: no need for alignment since it is not open with O_DIRECT. Can this be replaced with `read_footer_for_size_()` helper?

Chen-Yifan (commented):
Looks good overall. Added nits and suggestions for cleanup.

category/async/storage_pool.hpp
```diff
@@ -144,21 +152,46 @@ class storage_pool
             }
         } *const metadata_;
 
+        // True if this open wrote the device's footer: the device was blank,
+        // or it was opened with mode::truncate.
+        bool const is_freshly_initialised_;
```
Chen-Yifan, line 157: is this field necessary? only saw it used in a unit test.

